PR #124, opened 2020-03-19
Review and update server foundation service docs - prep

skeeey:
see https://github.com/open-cluster-management/backlog/issues/1003


cadawson (changes requested):
I cannot find the path to Workloads -> Helm releases.  Is this the correct path in the UI?  Thanks!

cadawson (approved):
Thank you!

PR #334, opened 2020-04-08
open-cluster-management/backlog#1498 Replace cmd

cadawson:


cadawson:
Hi @dhaiducek . I am replacing the command in the installation content that you noticed during the testathon.  If it looks ok, we will move forward with pushing it live. Thanks!

PR #396, opened 2020-04-17
open-cluster-management/backlog#1449--known issue and app doc update

bcpswope:


bcpswope:
Yup--will take care of that in another issue! Thank you!

PR #430, opened 2020-04-22
adding a enforce and inform table

dockerymick:
https://github.com/open-cluster-management/backlog/issues/459

### DIFF
--- a/governance/config_policy_ctrl.md
+++ b/governance/config_policy_ctrl.md
@@ -17,7 +17,7 @@ Each _policy_ within the policy document contains the following elements.
 
   - A `namespace` selector that specifies which namespaces within the cluster that the policy is applied to.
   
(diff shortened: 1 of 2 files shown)


governance/config_policy_ctrl.md
```diff
@@ -27,6 +27,20 @@ Each _policy_ within the policy document contains the following elements.
     
     - A `policy-template` is used to create one or more policies for third party or external security controls. For example, you can create a certificate expiration policy with the certificate policy controller. For more information about other policy and their  controllers, see [Red Hat Advanced Cluster Management for Kubernetes policy controllers](../governance/policy_controllers.md).
 
+### Inform and enforce summary
+
+When you create a configuration policy, set the `remediationAction` to define how policy violations are reported. View the following summary table to know which policies support the `enforce` feature: 
+
+| Sample policy| Enforce support | Enforce Action |
```
ycao56, line 34: all yes, enforce would create/patch the object on managed cluster with the definition in objectTemplate

dockerymick: @gparvin since all configuration policies support the enforce feature, I think we should reconsider what information is in the table and the location of it. According to https://github.com/open-cluster-management/backlog/issues/1875, we might want to create the Inform and enforce summary onto the policy controller page.

PR #431, opened 2020-04-22
Bare metal host adoption is not available, removing mentions of it

dockerymick:
https://github.com/open-cluster-management/backlog/issues/1720


dockerymick:
@cadawson thanks for the clarity and update to the file!

PR #433, opened 2020-04-23
changes to support portal

vikram-redhat:
Hi @cadawson @bcpswope @dockerymick - 

A couple of changes in this PR:

1. Moved the release notes topics in to their own top level folder to help support the migration to portal easier. I have also fixed the paths in summary.md as well and any place else that this move would have caused issues with.

2. Two files were named overview.md. In this initial push, I am keeping all topics in a single folder and therefore, no two files could be named the same. I have renamed them to install_overview.md and app_management_overview.md respectively. I have also changed the paths so that there are no broken links anywhere.

If this all seems ok, could you please merge into both doc_stage/prod?


bcpswope:
@vikram-redhat thanks, agree with these changes and will note the file naming convention for future! Thanks.

PR #437, opened 2020-04-23
adding note about the replicated policies

dockerymick:
https://github.com/open-cluster-management/backlog/issues/1436

### DIFF
--- a/governance/compliance_intro.md
+++ b/governance/compliance_intro.md
@@ -2,7 +2,9 @@
 
 Use the Governance and risk page to view and manage the number of policy violations in your clusters and applications. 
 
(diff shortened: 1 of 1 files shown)


drkho:
Thanks. Can we add more info like the followings?

**Note**: When you search for a policy in the hub cluster, you might also receive the name of the replicated policy on your managed cluster. For example, if you search for `policy-dhaz-cert`, the following policy name might appear: `default-policy-dhaz-cert`.  _**The name of the replicated policy is constructed as _namespace-in-hub_._policyname_. So, in "default.policy-dhaz-cert", default is the namespace in hub while policy-dhaz-cert is the policy name.**_

 (Btw, it is `default.policy-dhaz-cert` rather than `default-policy-dhaz-cert`) 🙂

PR #441, opened 2020-04-24
check in of master and docinfo for books

vikram-redhat:


bcpswope:
Hi. I need the access console topic, very important topic....  looks like it was removed here but I cannot see a reason in my nav file bc it is working.

Please hold off on this merge until I get clarity about this file.

manage_cluster/docinfo.xml
```diff
@@ -0,0 +1,10 @@
+<productname>{product-title}</productname>
+<productnumber>{product-version}</productnumber>
+<subtitle>Manage Cluster</subtitle>
+<abstract>
```
bcpswope, line 4: I don't want to mess you up by changing myself, but can we please lower-case the "cluster" in line 3? Thanks!

vikram-redhat: Done.

security/docinfo.xml
```diff
@@ -0,0 +1,10 @@
+<productname>{product-title}</productname>
+<productnumber>{product-version}</productnumber>
+<subtitle>Security</subtitle>
+<abstract>
+    <para>Security and Governance in {product-title}</para>
+</abstract>
```
bcpswope, line 6: same here, can we go with a lowercase 'g'

Thanks!

manage_cluster/docinfo.xml
bcpswope: oh wow, that was fast!

vikram-redhat:
> Hi. I need the access console topic, very important topic.... looks like it was removed here but I cannot see a reason in my nav file bc it is working.

It's removing the one with the extra space in the name. I thought I had removed it in an earlier commit, but it seems to have snuck back.

bcpswope:
Ok, I see. Thanks @vikram-redhat are you ready for a merge!?

PR #452, opened 2020-04-26
remove duplicates

vikram-redhat:
Hey folks,

A few edits to make things easier on the transformation side. This is for the installation book, and I will follow up with others shortly.

(sorry about the whitespace edits, please ignore them if you can)

* The install section seems to include duplicate supported browsers and OSs content, first in the requirements module and then in independent modules. I am taking them out the individual modules from the adoc files as this causes our build to complain.

* Asciidoc (and our build system) is stricter on using attributes in the headers. There are ways around it, but hoping that we can not use them. So I have taken out start.sh from the header and also fixed the links to it.

* Our build system barfs on <your_username> style tags, as the content is parsed finally into DocBook XML. My edits here to take these out are just so it can pass the build. Hopefully, it doesn't change the meaning of the content.


bcpswope:
@cadawson

bcpswope:
We need to merge this. @vikram-redhat I am going to merge if that is ok with you.

PR #487, opened 2020-04-30
open-cluster-management/backlog#2012 cis def false

cadawson:
Changing default cis setting to false


cadawson:
Hi @gparvin . PR 1 of 2 for changing the default setting for CIS.  We should be set to go, but let me know if you see any problems.  Thanks! 
FYI:  @dockerymick and @swopebrandi1

PR #488, opened 2020-04-30
open-cluster-management/backlog#2012 cis def false

cadawson:
Setting the default value of cisController to false.


cadawson:
Hi @gparvin . PR 2 of 2 for changing the default setting for CIS. We should be set to go, but let me know if you see any problems. Thanks!
FYI: @dockerymick and @swopebrandi1

dockerymick:
Closing this PR; Following here instead: https://github.com/open-cluster-management/rhacm-docs/pull/487

PR #522, opened 2020-05-04
removing to keep stage/production source same--file not needed

bcpswope:


bcpswope:
open-cluster-management/backlog#1580

PR #524, opened 2020-05-04
 open-cluster-management/backlog#1774--terms needed in arch topic

bcpswope:
Adding definitions for review by architects before I add them to the glossary

### DIFF
--- a/about/architecture.md
+++ b/about/architecture.md
@@ -14,30 +14,24 @@ Learn more about the following components for Red Hat Advanced Cluster Managemen
 
 ## Hub cluster
 
-The hub cluster is the common term that is used to define the central controller that runs in a Red Hat Advanced Cluster Management for Kubernetes cluster. From the hub cluster, you can access the console and components found on that console, as well as APIs such as the `rcm-api`, which handles API requests related to cluster lifecycle management.
+The _hub_ cluster is the common term that is used to define the central controller that runs in a Red Hat Advanced Cluster Management for Kubernetes cluster. From the hub cluster, you can access the console and product components, as well as APIs such as the `rcm-api`, which handles API requests related to cluster lifecycle management, which is defined later in this topic. 
 
-The hub cluster aggregates information from multiple clusters by using an asynchronous work request model. With a graph database, the hub cluster maintains the state of clusters and applications that run on it. The hub cluster also uses `etcd`, a distributed key value store, to store the state of work requests and results from multiple clusters and provides a set of REST APIs for the various functions that it supports.
+The hub cluster aggregates information from multiple clusters by using an asynchronous work request model. With a graph database, the hub cluster maintains the state of clusters and applications that run on it. The hub cluster also uses `etcd`, a distributed key value store, to store the state of work requests and results from multiple clusters, and provides a set of REST APIs for the various functions that it supports.
 
 ## Managed cluster
 
-The managed cluster is the term used to define the Red Hat Advanced Cluster Management for Kubernetes `multicluster-endpoint`, which is the agent that is responsible for a single Kubernetes cluster. The managed cluster initiates a connection to the hub cluster, receives and applies requests, then returns the results. 
-
-See [Managing your clusters](../manage_cluster/intro.md) to learn how to create and import clusters, which is part of _Cluster lifecycle_.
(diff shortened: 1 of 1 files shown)


about/architecture.md
```diff
@@ -18,22 +18,22 @@ The hub cluster aggregates information from multiple clusters by using an asynch
 
 ## Managed cluster
 
-The managed cluster is used to define the Red Hat Advanced Cluster Management for Kubernetes `multicluster-endpoint`, which is the agent that is responsible for a single Kubernetes cluster. The managed cluster initiates a connection to the hub cluster, receives work requests, applies work requests, then returns the results. The managed cluster connects to various services within the cluster for operations, including the Kubernetes API service, and Weave for topology.
+The managed cluster is the term used to define the Red Hat Advanced Cluster Management for Kubernetes `multicluster-endpoint`, which is the agent that is responsible for a single Kubernetes cluster. The managed cluster initiates a connection to the hub cluster, receives work requests, applies work requests, then returns the results. The managed cluster connects to various services within the cluster for operations, including the Kubernetes API service, and Weave for topology.
```
jnpacker, line 21: 1. We dropped the weavescope pods as part of multicluster-endpoint. 
2. Possibly just remove the last sentence.
3. When you say `work` are you referring to `work` as an action, or the actual `work` resource?  We don't expose the concept of the `work` resource to end users, so I'm asking for clarification.

about/architecture.md
```diff
@@ -18,22 +18,22 @@ The hub cluster aggregates information from multiple clusters by using an asynch
...
 
 See [Managing your clusters](../manage_cluster/intro.md) for to learn how to create and import clusters, which is part of _Cluster lifecycle_.
 
 ## Cluster lifecycle
 
-Create, import, and manage clusters across public and private clouds. From the console, you can view cluster health status, for instance, then you can move toward creating applications on those clusters.
+Red Hat Advanced Cluster Management for Kubernetes cluster lifecycle defines creating, importing, and managing clusters across public and private clouds. From the console, you can view cluster health status, for instance, then you can move toward creating applications on those clusters.
```
jnpacker, line 27: 1. Instead of  then you can move toward creating applications.  you can view an aggregation of all cluster health statuses or view individual health metrics.  You can upgrade managed 4.x OpenShift clusters individually or in bulk, as well as destroy any OpenShift clusters you provisioned with Advanced Cluster Management for Kubernetes

about/architecture.md
gparvin, line 21: To me it sounds like `multicluster-endpoint` is an agent running on the hub that is used to manage other clusters.  I think the first sentence needs to be re-worded to make it very clear that a managed cluster is a separate cluster from the hub cluster where RHACM has installed a small set of services to address that cluster's lifecycle, the application lifecycle and the governance and risk of that cluster.  Those lifecycle pieces should be the focus of the second sentence instead of talking about making a connection and doing "work".  The last sentence needs to have the last bit removed: `, and Weave for topology.` because there is no more weave.

about/architecture.md
```diff
@@ -18,22 +18,22 @@ The hub cluster aggregates information from multiple clusters by using an asynch
...
 
 ## Application lifecycle
 
-After you configure a Red Hat Advanced Cluster Management for Kubernetes hub cluster and a managed cluster, you can view and deploy applications with application resources. Your _Application_ is used to only _view_ your resource, while other application resource examples are for deployment. A multi-cluster application uses a Kubernetes specification, but with additional automation of the deployment and lifecycle management of resources to individual clusters.
+Red Hat Advanced Cluster Management for Kubernetes application lifecycle defines the processes used to managed application resources. After you configure a Red Hat Advanced Cluster Management for Kubernetes hub cluster, create and import a managed cluster, you can view and deploy applications with application resources. Your _Application_ is used to only _view_ your resource, while other application resource examples are for deployment. A multi-cluster application uses a Kubernetes specification, but with additional automation of the deployment and lifecycle management of resources to individual clusters.
```
jnpacker, line 31: - `manage` instead of `managed`
- I don't think we need the sentence about create/import and just talk about working with `managed clusters`. Since you sometimes just import a cluster, like EKS or AKS.
- Leave out the sentence about Your Application is used to only view your.....
- `A multi-cluster application uses a Kubernetes specification, but with additional automation of the deployment and lifecycle management of resources to individual clusters.` Becomes
A multi-cluster application allows can deploy resources on multiple clusters, while maintaining easy to reconcile service routes, as well as full control of Kubernetes resource updates for all aspects of the application.

about/architecture.md
bcpswope: yes, there was once debate about whether or not to discuss `work` resources, I think originally here we decided not to and used it to really just say `requests`. If `requests` is enough, I will remove work. @jnpacker

PR #533, opened 2020-05-05
Explaining remediationAction

dockerymick:
https://github.com/open-cluster-management/backlog/issues/2077

### DIFF
--- a/governance/create_policy.md
+++ b/governance/create_policy.md
@@ -57,14 +57,6 @@ Complete the following steps to create a policy:
                verbs: ["get", "list", "watch", "create", "delete","patch"]
    ```
 
-  - The `disabled` parameter provides the ability to enable and disable your policies.
-  - The value _enforce_ provides automatic remediation based on the policies.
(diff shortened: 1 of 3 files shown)


drkho:
@dockerymick Hi, I don't see references to `**Important**: Some policy controllers may not support the enforce feature. If the enforce feature is not supported by the controller, you must remediate it manually. See, `remediationAction` in the [Policy elements section](../policy_overview.md#policy_elements_2).` in other updates besides policy_controllers.md. 

Is that missed by accident? Can we add the same to remind users in other files?

dockerymick:
@drkho I have updated the explanation for the enforce value.  Based on our slack convo, it is concluded that it's best to continue to have the note only on the policy controller page. The enforcement limitation affected the policy controllers most. The controllers have an enforcement section that explains if the feature is supported. If the controllers (parent policy) does not support enforcement, then its respective policies (child/replicated policies) also do not support enforcement. 

Do you agree that we are ready to merge this PR? Do you have other thoughts?

drkho:
Understood. /approve

PR #547, opened 2020-05-07
minor fix cis policy doc

vikram-redhat:
@dockerymick I think this is a fix in one your files that got merged yesterday.

### DIFF
--- a/governance/cis_policy_ctrl.md
+++ b/governance/cis_policy_ctrl.md
@@ -41,9 +41,9 @@ Complete the following steps to enable the CIS policy controller feature flag. T
        value: "yes"
    ```
 
-4. Save your deployment and close the Visual Web Terminal. 
+4. Save your deployment and close the Visual Web Terminal.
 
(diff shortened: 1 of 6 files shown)


dockerymick:
Yes, as a personal note the link will be broken in Markdown. Locally, I am working to restructure the GRC area similar to the Application section. Thanks @vikram-redhat

vikram-redhat:
Thanks @dockerymick! We can use anchored links, just to remember to make sure they are unique across the book and if you use it then make sure it exists. In this case the referenced link didn't exist: https://github.com/open-cluster-management/rhacm-docs/pull/547/files#diff-a73a73af8ef5c52cb6fc427d8638411eR46.

vikram-redhat:
@bcpswope I added some more fixes for manage_applications as well to this commit. Sorry!

These fixes were required as one of your commits from yesterday made modifications to headers/titles, but there were references (or anchored links) to those sections through some of the docs.

PR #549, opened 2020-05-07
open-cluster-management/backlog#2133 remove tip

cadawson:


cadawson:
Hello @thuyn-581 .  This PR should resolve both issues open-cluster-management/backlog#2132 and open-cluster-management/backlog#2133.  Can you make sure those are covered, so we can move close those issues?  Thanks!

PR #560, opened 2020-05-10
open-cluster-management/backlog#1169 Rev and restruc

cadawson:


cadawson:
This content was restructured while the PR was being reviewed. Created this PR to avoid conflicts with PR 512.

PR #571, opened 2020-05-13
Fix a few link issues in governance

cadawson:


cadawson:
Soo, I didn't notice that your PR was from stage to prod, and not just a normal peer review.  Seems not merging it resulted in some broken links, so I merged it this morning.  After merging it, most of the links were fixed, but these three were still broken.  I fixed them in this PR.  We can try to merge early, if you want to.  Or we can wait until this afternoon. Either way, please take a look at this and make sure that it is what you had in mind.  Thanks @dockerymick !

PR #591, opened 2020-05-15
open-cluster-management/backlog#1956 Add rel img

cadawson:


manage_cluster/sync_rel_img.md
```diff
@@ -0,0 +1,23 @@
+# Synchronizing available release images
+
+The release images are updated frequently, so you might want to synchronize the list of release images to ensure that you can select the latest available versions. Complete the following steps to refresh the list:
```
jnpacker, line 3: Should we put the URL for the git repo.
There are also make commands that can be run.
`make subscribe-stable`
`make subscribe-fast` # includes stable
`make subscribe-candidate` # includes stable & fast

manage_cluster/master.adoc
```diff
@@ -21,3 +25,5 @@ include::modules/import_gui.md.adoc[leveloffset=+2]
 include::modules/import_cli.md.adoc[leveloffset=+2]
 include::modules/modify_endpoint.md.adoc[leveloffset=+2]
 include::modules/upgrade.md.adoc[leveloffset=+1]
+
```
bcpswope, line 28: looks like some extra spaces here I would remove just incase

manage_cluster/sync_rel_img.md
cadawson: Added the URL. Would like to push this and add the make commands to a refresh, as I could not find how they are used.

PR #608, opened 2020-05-18
Trouble app3

bcpswope:


troubleshoot_acm/trouble_app_deploy.md
```diff
@@ -0,0 +1,24 @@
+# Troubleshooting application Kubernetes deployment version
```
bcpswope, line 1: @dhaiducek please help with this troubleshooting? I gathered what I could from the opened issues but it is not clear how user should get around this.

jnpacker: It's not a deprecated Kubernetes version, but actually a deprecated API version.
**A managed cluster will show a similar error when one or more of your application's Kubernetes resource have an `apiversion` that has been deprecated. See the ...

bcpswope: thanks for that reminder @jnpacker 


Also, line 20 and on in resolving just has placeholder for now.

troubleshoot_acm/trouble_app_deploy.md
```diff
@@ -0,0 +1,24 @@
...
+```
+
+Or with new Kubernetes API version in your YAML file named `old.yaml` for instance, you might receive:
+
+```
+error: unable to recognize "old.yaml": no matches for kind "Deployment" in version "deployment/v1beta1"
+```
+
+## Resolving the problem: Application deployment version
+
```
bcpswope, line 21: place holder content until I get workaround steps--thank you!

troubleshoot_acm/trouble_app_deploy.md
```diff
@@ -0,0 +1,24 @@
...
+1. If the current version of Kubernetes API fails, enter the previous version. For example, see the following YAML entry:
```
dhaiducek, line 22: This is a case of using an old API version in a resource when it's no longer supported by a Kubernetes instance on a managed cluster. The resolution is to update the `apiVersion` in the resource.

For example, if the error is with a Deployment YAML, then `apiVersion: extensions/v1beta1` or `apiVerision: apps/v1beta1` should be updated to `apiVersion: apps/v1`.

You can verify the versions available to a resource on a cluster by running `kubectl explain <resource>` on the managed cluster and checking for `VERSION`.

dhaiducek: Looking back at the Subscription YAML after making the change will show whether the change was successful and the Subscription was successfully propagated.

troubleshoot_acm/trouble_app_deploy.md
```diff
@@ -1,24 +1,36 @@
...
-1. If the current version of Kubernetes API fails, enter the previous version. For example, see the following YAML entry:
+1. Update the `apiVersion` in the resource. For example, if the error displays for _Deployment_ YAML, you need to update the `apiVersion` from `extensions/v1beta1` to `apps/v1`.
 
-2. Validate?
+  See the following example:
+  
+  ```
+  apiVersion: apps.open-cluster-management.io/v1
+  kind: Deployable
+  ```
```
dhaiducek, line 29: Sorry I hadn't provided this as requested 🙂 
For this example, we should use:
```
apiVersion: apps/v1
kind: Deployment
```

bcpswope: will switch--I just took one from the docs/dev repo

bcpswope:
@dhaiducek and I worked it out, changes are in after chat and he explained when user gets error.

PR #612, opened 2020-05-18
adding known issue for certificate policy

dockerymick:
https://github.com/open-cluster-management/backlog/issues/2302

The Certificate policy file itself is still a work in progress. I will come back and address other comments after this PR is closed for the known issue

### DIFF
--- a/release_notes/known_issues.md
+++ b/release_notes/known_issues.md
@@ -164,9 +164,11 @@ The CIS policy controller is disabled by default when you install Red Hat Advanc
 
 You must enable the CIS policy controller. For more information, see _Enable the CIS controller_ on the [CIS policy controller page](../security/create_cis_pol.md) to update the policy.
 
+### Certificate policies fail to report status
 
+When you apply multiple certificate policies onto your managed cluster and the policies share the same value for `namespaceSelector`, only one of the policies work as expected. You are able to create and apply multiple certificate policies on your managed cluster. However, you must have a different parameter value for the `namespaceSelector`. 
 
-
+For more information, see [Certificate policy controller](../security/cert_policy_ctrl.md).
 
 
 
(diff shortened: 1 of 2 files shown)


release_notes/known_issues.md
```diff
@@ -164,9 +164,11 @@ The CIS policy controller is disabled by default when you install Red Hat Advanc
 
 You must enable the CIS policy controller. For more information, see _Enable the CIS controller_ on the [CIS policy controller page](../security/create_cis_pol.md) to update the policy.
 
+### Certificate policies fail to report status
 
+When you apply multiple certificate policies onto the same managed cluster and the policies share the same value for `namespaceSelector`, only one of the policies work as expected. You are able to create and apply multiple certificate policies on your managed cluster. However, you must have a different parameter value for the `namespaceSelector`.
```
cadawson, line 169: You can create and apply multiple certificate policies on a single managed cluster, but each policy must have a different parameter value for the `namespaceSelector`. When two or more policies on the same managed cluster use the same `namespaceSelector` value, then only the first policy listed is applied. 
Suggestion for simplification. If you use this, please make sure that it is the first one in the list, rather than the last one.  See what you think.

security/cert_policy_ctrl.md
```diff
@@ -48,7 +48,7 @@ View the following example of a certificate policy and review the element in the
 | Spec | Required. Specifications of which certificates to monitor and refresh.|
 | Spec.Conditions |  Required. <!--add description--> |
 | Spec.Ownership | Required. <!--Add description--> |
-| Spec.NamespaceSelector| Required. <!--add description--> parameter values for `Include` and `Exclude`. |
+| Spec.NamespaceSelector| Required. Namespace where you want the policy to be applied to. Enter parameter values for `Include` and `Exclude`. **Note**: When you create multiple certificate policie and apply them to the same managed cluster, each policy `NamespaceSelector` must be assigned a different value.|
```
cadawson: Namespace to which you want to apply the policy. Enter...

PR #622, opened 2020-05-19
Peer review was needed

bcpswope:


install/perform_scale.md
```diff
@@ -36,6 +37,7 @@ Search keeps the data in-memory to achieve fast response times. The memory requi
 
 Table: Physical memory data
 By default, we deploy the datastore with a memory limit of 1Gi.  If you are managing larger clusters, you may need to increase this limit by editing the deployment named `search-prod-xxxxx-redisgraph` in the open-cluster-management namespace.
+
 ### Write throughput (cache recovery time)
 
 Most clusters in steady state generate a small number of resource updates. The highest rate of updates happen when the data in RedisGraph is cleared, which causes the remote collectors to synchronize their full state around the same time.
```
bcpswope, line 43: Please try to avoid anthropomorphism--this is IBM Style guidance. Not always possible. Like 43.

install/perform_scale.md
```diff
@@ -36,6 +37,7 @@ Search keeps the data in-memory to achieve fast response times. The memory requi
 
 Table: Physical memory data
 By default, we deploy the datastore with a memory limit of 1Gi.  If you are managing larger clusters, you may need to increase this limit by editing the deployment named `search-prod-xxxxx-redisgraph` in the open-cluster-management namespace.
```
bcpswope, line 39: We is not allowed per IBM style....devs will give it to you, please rewrite if seen.

bcpswope: may/might

open-cluster-management usually offset by code

bcpswope: Please see line 28-- etc not used correctly but also should not be used in technical documentation-- See style manual for reference. This can be a for example list or something similar.

install/perform_scale.md
cadawson: Couldn't figure out a good way to reword this one.

bcpswope: I am not sure what `steady` and `full` state means the way it is written. I usually ask if this is something the user gets in an output like `ready` state and then write it according to the output. 

A lot of anthropomorphisms in this topic so it is something we just need to be mindful of... knowing the purpose of the paragraph is beneficial. 

Compared to the voice and table errors, I think just come back to it.

If full state is a state, I would say:

which causes the remote collectors to synchronize  to `full` state around the same time.

maybe come back to this after tech preview?

install/perform_scale.md
```diff
@@ -24,7 +25,7 @@ The scalability of the Search component depends on the performance of the data s
 
 ### Physical memory
 
-Search keeps the data in-memory to achieve fast response times. The memory required is proportional to the number of resources (ConfigMaps, Deployments, Pods, etc) and their relationships in the cluster.
+Search keeps the data in-memory to achieve fast response times. The memory required is proportional to the number of resources and their relationships in the cluster. Resources can include ConfigMaps, deployments, pods, and more.
```
jlpadilla, line 28: @bcpswope We can simplify it like this.
```suggestion
Search keeps the data in-memory to achieve fast response times. The memory required is proportional to the number of Kubernetes resources and their relationships in the cluster.
```

install/perform_scale.md
jlpadilla: `steady state` is a general term which means that nothing is being changed in the system.  In this particular case it means that after the initial synchronization, we only need to process a very small amount of data from the clusters to stay in sync.

@swopebrandi1  we opened the door for anthropomorphisms back when we decided to use `Search` as a name!

How about this:
```
When a cluster joins, it synchronizes its entire state, but after that, only a small amount of data needs to be processed to keep the search data in sync with the cluster. Under normal operation, a cluster synching its full state doesn't cause a problem because clusters join at different times. The highest rate of updates is observed when the search data is cleared, which causes the remote collectors to synchronize their full state around the same time.
```

PR #628, opened 2020-05-19
adding task file for namespace policy

dockerymick:


security/create_ns_policy.md
```diff
@@ -0,0 +1,104 @@
...
+   kubectl get namespacepolicy <policy-name> -n <namespace> -o yaml
+   ```
+
+2. View a description of your namespace policy by running the following command:
+
+   ```
+   kubectl describe namespacepolicy <name> -n <namespace>
+   ```
+
+## Creating a namespace policy from the console
```
bcpswope, line 49: Should this be at the same level as the CLI header?

security/create_ns_policy.md
```diff
@@ -0,0 +1,104 @@
...
+
+2. From the navigation menu, click **Governance and risk** to view a table list of your policies.
+
+  **Note**: You can filter the table list of your policies by selecting the _All policies_ tab or _Cluster violations_ tab.
+
+3. Select one of your policies to view more details.
+
+4. View the policy violations by selecting the _Violations_ tab.
+
+## Updating
```
bcpswope, line 73: Since the general heading "Updating" can be used in other spots and headers across books must be unique, I would be more specific about what use is updating on line 73.

security/create_ns_policy.md
dockerymick: Yup it should thank you

security/create_ns_policy.md
dockerymick: Thank you for the feedback and reminder !

PR #632, opened 2020-05-19
updating policy overview

dockerymick:
https://github.com/open-cluster-management/backlog/issues/2153

### DIFF
--- a/security/policy_example.md
+++ b/security/policy_example.md
@@ -10,7 +10,7 @@ The policy requires a _PlacementPolicy_ that defines the clusters that the polic
 
 * The `PlacementPolicy` and `PlacementBinding` can be included in the same `.yaml` file or separate `.yaml` files.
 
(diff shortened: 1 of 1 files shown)


security/policy_example.md
```diff
@@ -10,6 +10,8 @@ The policy requires a _PlacementPolicy_ that defines the clusters that the polic
 
 * The `PlacementPolicy` and `PlacementBinding` can be included in the same `.yaml` file or separate `.yaml` files.
 
+* When you create a policy, you can only apply it to your hub cluster namesapces.
```
ycao56, line 13: I think we are trying to explain
If you just create a policy yaml, it won’t do anything until you create a placementrule to apply it to expected managed cluster and bind the placementrule with policy through placementbinding

ycao56: Also you should not create policy in cluster namespace as cluster namespace is reserved by acm to operate. If you happen to create a policy in cluster namespace, it will be deleted by acm

PR #662, opened 2020-05-20
Update cert_policy_ctrl.md

dockerymick:
updating YAML example

Confirmed w/https://codebeautify.org/yaml-validator

https://github.com/open-cluster-management/backlog/issues/2237

### DIFF
--- a/security/cert_policy_ctrl.md
+++ b/security/cert_policy_ctrl.md
@@ -18,8 +18,6 @@ View the following example of a certificate policy and review the element in the
     Namespace:
     Labels: category=system-and-information-integrity 
   Spec:
-    Conditions:
-      Ownership:
     NamespaceSelector:
       Exclude:
       Include:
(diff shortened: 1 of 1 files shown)


bcpswope:
looking at the space between the label and the value, for instance, this is our dev sample:
https://github.com/open-cluster-management/multicloud-operators-subscription/blob/master/examples/e2e_example/github_channel_e2e.yaml

So I was wondering about that-- not sure where your dev samples are.

security/cert_policy_ctrl.md
```diff
@@ -26,35 +26,34 @@ View the following example of a certificate policy and review the element in the
...
+| Spec.RemediationAction | Required. Specifies the remediation of your policy. Set the parameter value to `inform`. Certificate policy controller only supports `inform` feature.|
 | Spec.Disabled | Required. Set the value to `true` or `false`. The `disabled` parameter provides the ability to enable and disable your policies.|
-| Spec.MinimumDuration | Required. parameter specifies the smallest duration before a certificate is considered non-compliant. When the certificate expiration is greater than the `minimumDuration`, then the certificate is considered compliant. <!--is there a default parameter value-->| 
-| Status | Required. Reports the status of the policy. <!--expand explanation if possible--> |
-| Status.CompliancyDetails | Required. <!--details needed--> |
-| Status.Events| Required. <!--add details-->
+| Spec.MinimumDuration | Required. parameter specifies the smallest duration before a certificate is considered non-compliant. When the certificate expiration is greater than the `minimumDuration`, then the certificate is considered compliant. The default parameter value is `100h`. |
```
gparvin, line 53: I feel like we should say this field uses Golang's time duration format.  That format is described here: https://golang.org/pkg/time/#ParseDuration
The sentence that best describes the format of the field is: A duration string is a possibly signed sequence of decimal numbers, each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".

It's unlikely anyone will use a format other than the hour format similar to our example `100h`.  We could just detail that the format is:

Required. This parameter specifies the smallest duration (in hours) before a certificate is considered non-compliant. When the certificate expiration is greater than the `minimumDuration`, then the certificate is considered compliant. The default parameter value is `100h` and must include a sequence of decimal  numbers followed by an `h` suffix.

security/cert_policy_ctrl.md
```diff
@@ -26,35 +26,34 @@ View the following example of a certificate policy and review the element in the
...
+| Spec.RemediationAction | Required. Specifies the remediation of your policy. Set the parameter value to `inform`. Certificate policy controller only supports `inform` feature.|
 | Spec.Disabled | Required. Set the value to `true` or `false`. The `disabled` parameter provides the ability to enable and disable your policies.|
-| Spec.MinimumDuration | Required. parameter specifies the smallest duration before a certificate is considered non-compliant. When the certificate expiration is greater than the `minimumDuration`, then the certificate is considered compliant. <!--is there a default parameter value-->| 
-| Status | Required. Reports the status of the policy. <!--expand explanation if possible--> |
-| Status.CompliancyDetails | Required. <!--details needed--> |
-| Status.Events| Required. <!--add details-->
+| Spec.MinimumDuration | Required. parameter specifies the smallest duration before a certificate is considered non-compliant. When the certificate expiration is greater than the `minimumDuration`, then the certificate is considered compliant. The default parameter value is `100h`. | 
+| Status | Required. Reports the compliance status of the policy.|
```
gparvin, line 54: The indication that these fields are required doesn't make sense for Status fields.  The status is information provided by the controller that shows the status of the policy.  So think of these as output only fields.

security/cert_policy_ctrl.md
```diff
@@ -11,12 +11,12 @@ Certificate policy controller does not support the `enforce` feature.
...
-  APIVersion:  policies.ibm.com/v1alpha1
-  Kind:         CertificatePolicy
+  APIVersion: policies.ibm.com/v1alpha1
+  Kind: CertificatePolicy
   Metadata:
+    Name: certificate-policy-example
+    Namespace:
+    Labels: category=system-and-information-integrity 
   Spec:
     Conditions:
```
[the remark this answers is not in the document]
dockerymick: This was the policy example brought over from ICP. Currently making updates based on your suggestion

dockerymick:
@gparvin Thanks for the review Gus, i have made more updates

PR #667, opened 2020-05-20
Updates for CIS 

dockerymick:
https://github.com/open-cluster-management/backlog/issues/2166

### DIFF
--- a/security/create_cis_pol.md
+++ b/security/create_cis_pol.md
@@ -202,7 +202,7 @@ Complete the following steps to enable the CIS policy controller feature flag fr
    oc edit deployment $(oc get deployment -o custom-columns=:.metadata.name | grep 'grcui$')
    ```
 
-3. Add the following contents to the deployment. Your deployment might resemble the following content:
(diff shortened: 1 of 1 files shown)


drkho:
@dockerymick Sorry that I have misled you with my original post. I prefer not to expose the information of the followings. So, the original content is fine. We just need to use "Add" instead of "Update"

`Add the following contents to the deployment. Your deployment might resemble the following content:`
      ~~- name: grcUiApiUrl~~
      ~~value: https://grc-590ea-grcuiapi:4000/grcuiapi~~
      ~~- name: searchApiUrl~~
      ~~value: https://search-search-api:4010/searchapi~~
and
     ~~image: quay.io/open-cluster-management/grc-ui@sha256:6f3bd3490d61305e4ddad46fc21fee3ac8151dadd604d1240de3b2a0a3ee29fa
     imagePullPolicy: Always~~


@gparvin , what is your take?

dockerymick:
@drkho I will close this PR and create a different one with your suggestion for the moment.

PR #673, opened 2020-05-20
open-cluster-management/backlog#1774-- first round

bcpswope:


dockerymick:
@bcpswope should we bold the terms?

about/glossary_terms.md
```diff
@@ -0,0 +1,33 @@
...
+
+- Channel: A custom resource definition that points to repositories where Kubernetes resources are stored, such as a GitHub repository, Helm chart repository, object store, or hub cluster namespace. Channels support multiple subscriptions from multiple targets.
+
+- Cluster lifecycle: Defines the process of creating, importing, and managing clusters across public and private clouds. 
+
+- Console: The graphical user interface for Red Hat Advanced Cluster Management.
+
+- Governance and risk: The  Red Hat Advanced Cluster Management processes used to manage security and compliance.
+
+- Hub cluster:  The central controller that runs in a Red Hat Advanced Cluster Management for Kubernetes cluster. From the hub cluster, you can access the console and components found on that console, as well as APIs.
```
dockerymick, line 22: Suggesting: "You can access the console, its components, as well as APIs from the hub cluster."

bcpswope: gonna pass on just this one because in tech doc, we want to try as much as possible to avoid possession with "its" or "their" and anthropomorphism.

bcpswope:
@dockerymick not sure about bold, as a general rule, unless part of a header we would avoid bold because a screen reader cannot tell user why something is bold.

We could make these headers, but trying to keep from creating build errors today; we can certainly work that out when we have stage branch building in the portal.

PR #692, opened 2020-05-21
app was not building in QSG

bcpswope:


about/quick_start.md
```diff
@@ -27,15 +27,16 @@ You are now ready to create and import clusters. From your hub cluster, you can
 2. If you have a cluster that you want to import manually, you can view [Importing a cluster](../manage_cluster/import.md) to learn how to import a managed cluster.
 
 3. When you no longer need to manage a cluster, you can detach that cluster from the _Cluster page_.
-## Manage applications
```
dockerymick, line 30: As discussed in Slack, the only change required should be create a line break after the steps. move Manage applications to line 31, then start "You  can start ..." on line 33 and so on

PR #712, opened 2020-05-29
open-cluster-management/backlog#1908 -- app route, search

bcpswope:
Regarding issue, I gathered from notes in comments the draft here but need help with clarification. 

1. User does not see search on a GKE cluster because the default route is fixed? 
2. User needs to take an action? 
3. Will this be an issue for every GKE cluster?

Any other comments on this PR, please make them in line or commit your changes.

@quchangl-github and @birsanv


birsanv:
@bcpswope I want @jnpacker to review this doc issue

For reference this is the doc explaining what a Route is https://www.openshift.com/blog/kubernetes-ingress-vs-openshift-route

In summary, Route is an OpenShift K8s object used to make a service object available externally
So a Route object is created on the managed cluster only if that cluster is a ocp one. 
Otherwise, an Ingress object would be created instead

On the PR I would say ( and this should be changed on the 1.0.0 doc as well )

**Application Route resource shows in the search page only for OpenShift managed cluster**

- If none of the target managed clusters for a deployed application is an OpenShift cluster, the Route resource is not created, although the Application Topology displays the object. Since the object is not deployed, it will not show up in the Search view when doing a lookup. This is expected.
This is the case for all Kubernetes resources which are platform specific, such as the Route resource. The Application definition defines these resources but they are created on the managed clusters only if they are matching the resource platform.

I'll add this to PR as a review comment

release_notes/known_issues.md
```diff
@@ -205,6 +205,38 @@ For instance, namespace `charts-v1` is used by the installer as a Helm type chan
 
 For technical preview, all channels need an individual namespace, except GitHub channels, which can share a namespace with andother GitHub channel. See the process for [Managing channels](../manage_applications/managing_channels.md) for more information.
 
+### Application route does not list in the search page for GKE cluster
```
birsanv, line 208: Application Route resource shows in the search page only for OpenShift managed cluster

release_notes/known_issues.md
```diff
@@ -205,6 +205,38 @@ For instance, namespace `charts-v1` is used by the installer as a Helm type chan
...
+<!--1.0.0:1908-->
+
+If the target managed cluster is a Google Cloud cluster, the cluster might not have the `Route` resource. Without the route, a random port will be assigned to the front-end service.
```
birsanv, line 211: If none of the target managed clusters for a deployed application is an OpenShift cluster, the Route resource is not created, although the Application Topology displays the object. Since the object is not deployed, it will not show up in the Search view when doing a lookup. This is expected.
This is the case for all Kubernetes resources which are platform specific, such as the Route resource. The Application defines these resources but they are created on the managed clusters only if they are matching the resource platform.

release_notes/known_issues.md
```diff
@@ -205,6 +205,38 @@ For instance, namespace `charts-v1` is used by the installer as a Helm type chan
 
 For technical preview, all channels need an individual namespace, except GitHub channels, which can share a namespace with andother GitHub channel. See the process for [Managing channels](../manage_applications/managing_channels.md) for more information.
 
+### Application route does not list in the search page for GKE cluster
+<!--1.0.0:1908-->
+
+If the target managed cluster is a Google Cloud cluster, the cluster might not have the `Route` resource. Without the route, a random port will be assigned to the front-end service.  
+
+The application uses the following route to expose the internal service to the fixed, default external port `80`. The default port for HTTP web server that you can easily open the application without specifying the port number:
```
birsanv, line 213: I am not sure if we want to get into this low level details on how the Route works, I'll let @jnpacker decide

release_notes/known_issues.md
jnpacker: I think it should be clear from the Topo changes @birsanv made that the object did not deploy (is missing) and so we would expect the object icon to reflect that, the details to reflect that its not found (since search doesn't have it) and therefore the resource not being in search is explained.  Can we confirm if this route issue applies to other providers. Alternately `generalize the statement` for `all` unsupported resources and put it in our troubleshooting section.

birsanv: @jnpacker my Topology changes are for 2.0 ( the ones showing errors, etc ). The changes I am suggesting above , for this PR, are for the current published doc ( 1.0.0, and fixpack )
@bcpswope with the changes I am making to the Topo in 2.0, this Known issue is no longer required; sorry, I missed that

PR #739, opened 2020-06-02
open-cluster-management/backlog#2353 prov conn links

cadawson:


cadawson:
Hi @chrisahl . I added the links to the procedures.  We should be set on this one.  Thanks!

PR #779, opened 2020-06-11
Create SECURITY.md

jpweiser:
All repos, particularly those which are public, should have a security policy defined on Github. This takes care of https://github.com/open-cluster-management/backlog/issues/2496


jpweiser:
Since the markdown file will cause conflict, we will add this section the readme instead. Will do this in a different PR.

PR #790, opened 2020-06-15
open-cluster-management/backlog#2431 -- added new file, list approved…

bcpswope:
… and discussed in issue, remove a few known issues, some waiting on response, added to adoc file.

### DIFF
--- a/release_notes/fixpack.md
+++ b/release_notes/fixpack.md
@@ -4,7 +4,7 @@ View a summarized list of Red Hat Advanced Cluster Management for Kubernetes fix
 
 1. Upgrade from product version 1.0 is now supported. 
 2. Search sync performance was improved.
-3. Security vulnerabilities were fixed. 
+3. Known security vulnerabilities were fixed. 
 4. Bare metal hosts stuck in `power on` state were fixed to allow for local management.
 5. Added support for provisioning newer versions of Red Hat OpenShift Container Platform.
 6. Fixed issue with Visual Web Terminal command.
--- a/release_notes/fixpack.md
+++ b/release_notes/fixpack.md
@@ -10,5 +10,4 @@ View a summarized list of Red Hat Advanced Cluster Management for Kubernetes fix
 6. Fixed issue with Visual Web Terminal command.
 7. Fixed subscriptions that were not deploying all subscribed resources.
 8. Management ingress certificate deployment simplified and documented.
-9. Fixed internal error 500 during login to the console. 
 
--- a/release_notes/fixpack.md
+++ b/release_notes/fixpack.md
@@ -1,0 +1,14 @@
+# Fix pack updates
+
+View a summarized list of Red Hat Advanced Cluster Management for Kubernetes fix pack 1.0.1 updates.
+
+1. Upgrade from product version 1.0 is now supported. 
+2. Search sync performance was improved.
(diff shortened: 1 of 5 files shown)


release_notes/fixpack.md
```diff
@@ -0,0 +1,14 @@
+# Fix pack updates
+
+View a summarized list of Red Hat Advanced Cluster Management for Kubernetes fix pack 1.0.1 updates.
+
+1. Upgrade from product version 1.0 is now supported. 
+2. Search sync performance was improved.
+3. Security vulnerabilities were fixed.
```
cadawson, line 7: Should we make this "Known security vulnerabilities were fixed"?

release_notes/known_issues.md
```diff
@@ -232,155 +223,7 @@ For more information, see [Certificate policy controller](../security/cert_polic
 ### Any authenticated user can import clusters
 <!--1.0.0:2312-->
 
-Any authenticated user of OpenShift Container Platform can provision projects and have administrator privileges to the project and its associated namespace. As the administrator of a namespace, you can generate commands to import clusters into Red Hat Advanced Cluster Management for Kubernetes. To run the generated commands and import the cluster, you must have cluster administrator privileges on the managed cluster. For more information, view the Role-based access control table in the [Security](../security/security_intro.md) topic.
-
-### Internal error 500 during login to the console
```
dockerymick, line 237: This section be added back. From Gus and I's meeting, I  understood that this will not be fixed until 2.0 as well. @gparvin may you confirm?

release_notes/fixpack.md
```diff
@@ -0,0 +1,14 @@
...
+
+1. Upgrade from product version 1.0 is now supported. 
+2. Search sync performance was improved.
+3. Known security vulnerabilities were fixed. 
+4. Bare metal hosts stuck in `power on` state were fixed to allow for local management.
+5. Added support for provisioning newer versions of Red Hat OpenShift Container Platform.
+6. Fixed issue with Visual Web Terminal command.
+7. Fixed subscriptions that were not deploying all subscribed resources.
+8. Management ingress certificate deployment simplified and documented.
+9. Fixed internal error 500 during login to the console.
```
dockerymick, line 13: This should be removed bc it won't be fixed until 2.0. The issue still exists for 1.0.1

bcpswope: See the issue that I got from Josh about this: https://github.com/open-cluster-management/backlog/issues/2431#issuecomment-642655202. Maybe not the same but this is why I added line 9... it was low-hanging fruit that came up on our call.

dockerymick: I'm aware of 2431, it will still be a known issue for 1.0.1

dockerymick: I received correspondence and informed me of the following info: "I believe we need to keep this known issue going forward.  Even in 2.0 there will be ways to still see it.  Example:
1. Install OCP
2. Install RHACM
3. Update OCP certificate
Now RHACM will show that error until the customer does the fix. 

He also said that he will check with the install team to verify that the issue is handled, but he is assuming that it is not.

bcpswope: Ok @dockerymick -- since this will be build changes and we need to check it before the end of the day, I went ahead and removed #9. Better to not have it just in case. Please take a look and let me know if we can merge.

PR #862, opened 2020-07-08
Creating custom policy controller [WIP]

dockerymick:
https://github.com/open-cluster-management/backlog/issues/1224

### DIFF
--- a/security/create_policy_ctrl.adoc
+++ b/security/create_policy_ctrl.adoc
@@ -18,13 +18,14 @@ Complete the following steps to create a policy controller from the CLI:
 
 . Clone the `multicloud-operators-policy-controller` repository by running the following command:
 +
----
+----
 git clone git@github.com:open-cluster-management/multicloud-operators-policy-controller.git
----
+----
 
 . Customize the controller policy by updating the policy schema definition. Your policy might resemble the following content:
 +
----
+[source,yaml]
+----
 metadata: 
   name: samplepolicies.policies.open-cluster-management.io
 spec:
@@ -34,10 +35,9 @@ spec:
     listKind: SamplePolicyList
     plural: samplepolicies
     singular: samplepolicy
----
+----
 
 Your policy controller is created.
-
 
 [#viewing-your-policy-controller-yaml-file-from-the-console]
 ==== Viewing your policy controller YAML file from the console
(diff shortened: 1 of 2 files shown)


security/create_policy_ctrl.adoc
```diff
@@ -0,0 +1,248 @@
...
+.. Verify that the controller is functional by running the following command:
++
+----
+kubectl get pod -n <namespace>
+----
+
+Your policy controller is deployed.
+
+[#deleting-a-policy-controller]
+=== Deleting a policy controller
```
ycao56: the command in this section deletes policy not policy controller

security/policy_controllers.adoc
```diff
@@ -5,8 +5,9 @@ Use the Red Hat Advanced Cluster Management for Kubernetes governance and risk p
 The policy controllers are Kubernetes CustomResourceDefinition (CRD) instance.
 For more information about CRDs, see link:https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/[Extend the Kubernetes API with CustomResourceDefinitions].
 
-Create and customize policies with the out of the box policy templates.
-*Important*: Some policy controllers might not support the enforce feature.
+Create and customize poolicy controllers and policies with the out of the box policy templates. Policy controllers monitor and report whether your cluster is compliant with policy. It can also enforce the policy to bring the cluster status to compliant. See xref:../create_policy_ctrl.adoc#creating-a-custom-policy-controller[Creating a custom policy controller] for more information.
```
ycao56: shouldn't mix custom policy controllers with out of box policies. this should be a separate section. e.g.
Bring your own policies by creating custom policy and policy controller

security/create_policy_ctrl.adoc
```diff
@@ -0,0 +1,248 @@
+[#managing-custom-policy-controllers]
+= Managing custom policy controllers
```
ycao56: We probably need to reorganize the sections. We should explain
1. when and why to write a custom policy and policy controller
2. how to do it
3. how to deploy it and integrate with grc dashboard

security/create_policy_ctrl.adoc
dockerymick: @willkutler may you provide the correct command?

security/create_policy_ctrl.adoc
dockerymick: @ycao thanks for your suggestions. We have not specified when and why the policy controller should be created, however I suggest that the explanation is added to the security/policy_controllers.adoc file

willkutler (approved):

security/create_policy_ctrl.adoc
dockerymick: @ycao56 Will shared that he wasn't sure which command should be used to delete a policy controller. May you provide the info?

dockerymick:
Still need to update the PR. Per @willkutler suggest, we should add a section named `Integrating custom policy controller with the Governance and risk dashboard` and add steps

security/create_policy_ctrl.adoc
willkutler: here's how you'd use the new policy controller with the grc dashboard:
1. deploy the controller
2. go to the grc dashboard and click "create policy"
3. edit the yaml so that an instance of your custom policy is listed under `policy-templates`
4. now any violations generated by that custom policy will show up as violations on the parent policy

ycao56:
@dockerymick We need to talk about the purpose of this doc. I am still not comfortable with the content. We already have a blog talking about how to write a custom policy controller. Why do we want to duplicate the information? Also the content of the blog was not 100% percent port over to this doc. I am confused about the intention.

dockerymick:
@ycao56 The purpose of this doc is to document only task for the policy controller. The requirement for RH content team is to utilize modular documentation, where concepts and task are separate. Concepts for the policy controller are on the Policy controller page and the task are on Managing custom policy controllers. Even though there is a blog that explains creating a custom policy controller, it was requested that the info also be added into the customer portal documentation for 2.0.

willkutler (approved):

security/create_policy_ctrl.adoc
```diff
@@ -0,0 +1,202 @@
...
+metadata: 
+  name: samplepolicies.policies.open-cluster-management.io
+spec:
+  group: policy.open-cluster-management.io
+  names:
+    kind: SamplePolicy
+    listKind: SamplePolicyList
+    plural: samplepolicies
+    singular: samplepolicy
+----
```
ycao56, line 30: there are more steps required to update policy see
image

security/create_policy_ctrl.adoc
```diff
@@ -0,0 +1,202 @@
...
+
+. Check the `status` field for compliance details by running the following command: 
+
++
+----
+kubectl describe SamplePolicy example-samplepolicy -n default
```
ycao56, line 90: you need to run policy controller and install crd and cr before you can check the result. see
image

security/create_policy_ctrl.adoc
```diff
@@ -0,0 +1,202 @@
...
+        and 0 groups violations
+      kube-public:
+      - 1 violations detected in namespace `kube-public`, there are 0 users violations
+        and 1 groups violations
+  compliant: NonCompliant
+----
+
+. Update the policy rules and policy logic to introduce new rules for your policy controller. Complete the following steps:
+
+.. Add new fields in your YAML file by updating the `SamplePolicySpec`. Your specification might resemble the following content:
```
ycao56, line 119: steps are missing to update policy controller

image

security/create_policy_ctrl.adoc
```diff
@@ -0,0 +1,202 @@
...
+kubectl apply -f deploy/operator.yaml -n <namespace>
+----
+
+.. Verify that the controller is functional by running the following command:
++
+----
+kubectl get pod -n <namespace>
+----
+
+. You must integrate your policy controller by creating a `policy-template` for the controller to monitor. For more information, see xref:../security/create_policy.adoc#creating-a-cluster-security-policy-from-the-console[Creating a cluster security policy from the console].
```
ycao56, line 223: do we have instructions on how user can use its own policy template and create policy through UI?

dockerymick: @ycao We have instructions on how to create a policy from the console here: https://github.com/open-cluster-management/rhacm-docs/blob/2.0_stage/security/create_policy.adoc

willkutler (approved):

PR #866, opened 2020-07-08
remove search known issue on memory leak

dockerymick:
https://github.com/open-cluster-management/backlog/issues/3302


jlpadilla:
Looks good, but I'd like to get the issue verified before we update the doc.

@evelinec Can you verify the memory leak fix. https://github.com/open-cluster-management/backlog/issues/1918

jlpadilla (approved):
Approving pending verification of the fix from QA.
https://github.com/open-cluster-management/backlog/issues/1918

PR #868, opened 2020-07-09
removing CIS

dockerymick:
https://github.com/open-cluster-management/backlog/issues/3380#issuecomment-655579346


cadawson (changes requested):
Looks good. Please look at create_cis_pol. It looks like you deleted the file from the repo, but it looks like it still shows up in the master.adoc. If it is still listed in the master.adoc, I think it should be removed. Thanks!

dockerymick:
Received the ok via [Slack](https://coreos.slack.com/archives/DURFHUR16/p1594323726072700)

PR #879, opened 2020-07-10
open-cluster-mgmt/backlog#1657 clst offln

cadawson:


cadawson:
Hi @laisongls  . This is an older one that I am catching up with.  Please take a look and see if we cover the steps in the troubleshooting topic in this PR. If changes are required, let me know. These changes are in the 2.0 version, but if the content ins correct, I can add it to 1.0, as well. Thank you.

laisongls:
Corrected the comment.

cadawson:
Thank you.  Is this ready to be merged, @laisongls ?

laisongls:
@cadawson the doc is NOT correct to determine the root cause of offline cluster for 2.0. I've been add a single comments but not sure why it has gone. Let me comment it here.
The troubleshooting works for v1.0. But since 2.0 don't have the pods `endpoint-connmgr` and `endpoint-workmgr`, so we have to modify the section as below. Also please modify the same section in `trouble_cluster_create.adoc`

```
If it is not available, try restarting the managed cluster.

If the managed cluster status is still offline, try to run the command `oc get managedcluster <-cluster name-> -o yaml` on HUB to check message and reason of the specific `type: ManagedClusterConditionAvailable` from `status.conditions` section. 

```

cadawson:
Hi @laisongls . No problem.  I wasn't sure if it would apply to both versions, and started with 2.0 because there was no risk of it being released, since 2.0 has not yet shipped.  

As for the comment, I thought you had updated  the content when I saw your message that you updated the comment. A misunderstanding. Sorry about that. 

Given your comment, I split off the 1.0.1 and 2.0 versions. They are both in PRs for your review.  See if they look ok now:
2.0: https://github.com/open-cluster-management/rhacm-docs/pull/879
1.0.1: https://github.com/open-cluster-management/rhacm-docs/pull/921

I renamed the create file to make this one, so that file no longer exists.  Thanks!

laisongls (approved):

laisongls (approved):

laisongls (approved):

PR #887, opened 2020-07-14
Remove troubleshooting storage class docs

jamestalton:
Remove storage class docs as we are not using the storage class anymore for 2.0 install


dockerymick (approved):
Looks good to me, thanks for the update!

PR #893, opened 2020-07-14
Remove references to Unknown status

KevinFCormier:
open-cluster-management/backlog#3076
-Unknown status is no longer used
-Pending status is shown before provision starts
-Failed status will be shown if provision failed, and a link to logs is provided


KevinFCormier:
@cadawson I've updated most of the content that needs to change with this PR, but there is 1 more change I'd like to see with this page (which is why I've marked this PR as do not merge for now) which is to the following section:
image

There are a few issues:
1. It assumes the user has found logs and identified errors, but in Procedure 2, we are not expecting them to do that. So we need to be more generic. Something like "If you located logs, review the contents for error messages and resolve any identified issues before destroying the cluster and attempting to create a new one.
2. It gives one example of a type of error you might see ("No subnets provided for zones") but the text makes it sound like this is the only error we are expecting. Then the resolution for this is in a new section. I think these need to be kept in the same section and given as just one example of how you might resolve an issue, or as a specific case we think users might run into often.

I am not familiar with formatting in `.adoc` so I left this part to you.

KevinFCormier:
@cadawson There are also some links to this page which will need updating, I think.

troubleshoot_acm/trouble_console_status.adoc
```diff
@@ -54,8 +56,8 @@ See log error:
    No subnets provided for zones
 ----
 
-[#resolution-cluster-in-console-with-unknown-status]
-== Resolution: Cluster in console with unknown status
+[#resolution-cluster-in-console-with-pending-or-failed-status]
+== Resolution: Cluster in console with pending or failed status
 
 Select different zones within the region.
 One or more of the zones that you selected is not supported.
```
dockerymick: I suggest that we provide steps for users to update the regions, or provide a reference to a doc that describes the task

dockerymick (commented):
I made a couple of updates and a comment

troubleshoot_acm/trouble_console_status.adoc
KevinFCormier: @TheRealHaoLiu had initially suggested linking to Hive documentation, but the documentation he had found was not suitable. I guess these logs would contain output from Hive and the OpenShift installer. I'm not sure if there's a suitable entry point to the OpenShift installer docs that would make sense in this case.

cadawson:
HI @KevinFCormier . Can you take a look at this one now? I noticed that there is still an _unknown_ status mentioned in there.  Should that be removed?  Thanks!

dockerymick:
@KevinFCormier the troubleshoot topic, **Troubleshooting a default storage class error when creating a cluster on bare metal** was removed on this [PR](https://github.com/open-cluster-management/rhacm-docs/pull/887/files)

troubleshoot_acm/trouble_console_status.adoc
```diff
@@ -1,15 +1,17 @@
...
 
-After creating a new cluster by using the Red Hat Advanced Cluster Management for Kubernetes console, the cluster shows a status of _Unknown_ in the console.
+After creating a new cluster by using the Red Hat Advanced Cluster Management for Kubernetes console, the cluster does not progress beyond the status of _Pending_ or displays _Failed_ status.
 
-[#identifying-the-problem-cluster-in-console-with-unknown-status]
-== Identifying the problem: Cluster in console with unknown status
+[#identifying-the-problem-cluster-in-console-with-pending-or-failed-status]
+== Identifying the problem: Cluster in console with pending or failed status
+
+If the cluster displays _Failed_ status, navigate to the details page for the cluster and follow the link to the logs provided. If no logs are found or the cluster displays _Unknown_ status, continue with the following procedure to check for logs:
```
KevinFCormier:
```suggestion
If the cluster displays _Failed_ status, navigate to the details page for the cluster and follow the link to the logs provided. If no logs are found or the cluster displays _Pending_ status, continue with the following procedure to check for logs:
```

KevinFCormier:
@cadawson Yes, that was 1 reference to _Unknown_ that I missed. Looks good to me now.

dockerymick (approved):
Fixed the merge conflict @KevinFCormier. Thanks Chris for assisting

PR #921, opened 2020-07-16
Dawson1657101

cadawson:


cadawson:
Hi @laisongls . No problem. I wasn't sure if it would apply to both versions, and started with 2.0 because there was no risk of it being released, since 2.0 has not yet shipped.

As for the comment, I thought you had updated the content when I saw your message that you updated the comment. A misunderstanding. Sorry about that.

Given your comment, I split off the 1.0.1 and 2.0 versions. They are both in PRs for your review. See if they look ok now:
2.0: #879
1.0.1: #921

I renamed the create file to make this one, so that file no longer exists. Thanks!

laisongls (approved):

laisongls (approved):

laisongls (approved):

PR #933, opened 2020-07-21
open-cluster-management/backlog#1580 Conref start

cadawson:


cadawson:
Not necessarily a priority, unless you want to use conrefs in 2.0. You should be able to start using them. Here is a starter set. If you want to, add any that would be helpful to you. Thanks!

modules/common-attributes.adoc
```diff
@@ -1,4 +1,11 @@
+:addons: Klusterlet addons
+:klusterlet: Klusterlet
```
swopebe, line 2: Thanks!! 


Please let's not conref this term--no need. I used this term a long time without needed to conref, with VSCode, it is easy to find/replace. Can we stick to just product names where people can accidentally use various names and not worry about glossary terms.

modules/common-attributes.adoc
```diff
@@ -1,4 +1,11 @@
...
+:ocp: Red Hat OpenShift Container Platform
+:ocp-short: OpenShift Container Platform
 :product-title: Red Hat Advanced Cluster Management for Kubernetes
 :product-title-short: Red Hat Advanced Cluster Management
 :product-version: 2.0
+:product-version-prev: 1.0
+:quay: Red Hat Quay
+:quay-short: Quay
 :imagesdir: ../images
```
swopebe, line 11: Other than leaving out typical term k-let, looks good. Simple list and not so invasive that we spend more time that needed searching through conrefs! Great addition to the doc!

modules/common-attributes.adoc
cadawson: That's fine. Normally, I wouldn't. In this case, it went from Klusterlet -> not Klusterlet -> Klusterlet in the span of about 1 year. That was why I thought those would be useful. I am fine with it, though. We might actually stick with Klusterlet for a while.

modules/common-attributes.adoc
cadawson: Thanks for the feedback! It will be nice to have them back.

dockerymick (approved):
this looks good to me! What do you all think about creating a conref for hub cluster?

modules/common-attributes.adoc
swopebe: It's an easy change with find and replace, and used often enough that it would suck to write a conref each time LOL.

swopebe:
We didn't have conref for hub cluster or managed cluster or klusterlet. Most terms have not changed, if they do- easy find/replace, and used so often that it is a pain to write conref each time.

The conref idea is especially useful in older processes and tools but with Atom and VS Code, it eliminates the need to conref common terms because they may or may not change. We talked in content call about making it mostly for product names that can be misrepresented. Terms should not need conref, these have been around since May 2018. Should they change, easy fix.

PR #936, opened 2020-07-21
[WIP] starting to add policy API

dockerymick:
https://github.com/open-cluster-management/backlog/issues/3188


apis/policy.json.adoc
```diff
@@ -0,0 +1,415 @@
...
+[options="header", cols=".^3a,.^4a"]
+|===
+|Name|Schema
+|**complianceType** +
+__required__| string
+|<<_rhacm-docs_apis_policy_jsonpolicy_policyrule,**PolicyRule**>> +
+__required__| object
+|===
+
+[[_rhacm-docs_apis_policy_jsonpolicy_policyrule]]
```
willkutler: this section should be able to be removed since configuration policy doesn't use policyRules, it just lists the rules out

apis/policy.json.adoc
```diff
@@ -0,0 +1,415 @@
...
+    "remediationAction": "enforce",
+    "namespaces": {
+      "include": [
+        "default"
+      ],
+      "exclude": [
+        "kube*"
+      ]
+    },
+    "configurationPolicy": {
```
willkutler: should be nested inside policy-templates, so 
```
"policy-templates": [
     {
          "kind": "ConfigurationPolicy",
          "apiVersion": "policy.open-cluster-management.io/v1"
          ....etc
     }
]
```
note the apiVersion change too

PR #937, opened 2020-07-21
open-cluster-management/backlog#3435 -- adding changes to file and kn…

swopebe:
…own issue addition

### DIFF
--- a/manage_applications/channel_sample.adoc
+++ b/manage_applications/channel_sample.adoc
@@ -108,7 +108,21 @@ spec:
 [#kubernetes-namespace-namespace-channel]
 == Kubernetes namespace (Namespace) channel
 
-The following example channel definition abstracts a namespace as a channel that holds deployable resources.
+You need to manually create deployable resources within the channel namespace. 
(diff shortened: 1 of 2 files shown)


manage_applications/channel_sample.adoc
```diff
@@ -130,6 +144,34 @@ The channel controller maintains the resources at the actual namespace location
...
+      labels:
+        app: gbchn
+        release: gbchn
+      name: gbchn
+    spec:
+      ports:
+      - port: 80
+      selector:
+        app: gbchn
+----
```
swopebe, line 174: we have two samples now in this namespace section so let's chat about whether or not we need both?

xiangjingli (approved):
LGTM

manage_applications/channel_sample.adoc
xiangjingli: Yes, we need both. In namespace channel, it requires to define a namespace type channel and relative deployable resource under the channel namespace. The two samples address it

PR #949, opened 2020-07-22
adding what's new blurb about tech preview

dockerymick:
https://github.com/open-cluster-management/backlog/issues/3562
https://github.com/open-cluster-management/backlog/issues/460#issuecomment-662603824

### DIFF
--- a/release_notes/whats_new.adoc
+++ b/release_notes/whats_new.adoc
@@ -1,7 +1,9 @@
 [#whats-new-in-red-hat-advanced-cluster-management-for-kubernetes]
-= What's new in Red Hat Advanced Cluster Management for Kubernetes
+= What's new in Red Hat Advanced Cluster Management for Kubernetes 
 
-Red Hat Advanced Cluster Management for Kubernetes provides visibility of your entire Kubernetes domain with built-in governance, cluster lifecycle and application lifecycle management.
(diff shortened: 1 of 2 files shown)


release_notes/whats_new.adoc
```diff
@@ -1,5 +1,7 @@
 [#whats-new-in-red-hat-advanced-cluster-management-for-kubernetes]
-= What's new in Red Hat Advanced Cluster Management for Kubernetes
+= What's new in Red Hat Advanced Cluster Management for Kubernetes 
+
+Red Hat Advanced Cluster Managent for Kubernetes is no longer a technology preview. See what is available in Red Hat Advanced Cluster Management for Kubernetes version 2.0.0.
```
cadawson: Can I suggest tweaking the wording about it not being tech preview? I agree with the approach, but some of the functionality is still tech preview. My thought is something like: "Red Hat Advanced Cluster Management (check spelling of Management) for Kubernetes is now a generally available product!" Or something like that. To me, it twists the statement from it not being a tech preview at all to being a GA product. It might be a minor thing, but I think it is easier to understand why there is tech preview function in the second example.  Maybe it's just late this afternoon. Thanks for doing that!

cadawson (approved):
Thank you!

PR #971, opened 2020-07-23
No issue-rbac add to arch

swopebe:
@dockerymick was I was trying to say on slack is that we need to reference RBAC from arch topic, this was a change I made earlier for 1.0 refresh based on internal user request. It made sense to me given how much we highlight RBAC for our product. So your change to the link is also here now. Please review. Also, all of them are "link" from our previous work on fixing these. I would like to see if they are building, most are intro topics, as well.


dockerymick (approved):
looks good! Thanks for the change Brandi

PR #987, opened 2020-07-27
Changes to the certificates page 2.0

dockerymick:
https://github.com/open-cluster-management/backlog/issues/3440

### DIFF
--- a/security/certificates.adoc
+++ b/security/certificates.adoc
@@ -30,6 +30,7 @@ The certificate can be imported into your client truststores to access Red Hat A
 
 View the following table of the internal services that use `cert-manager` to create and manage their certificates:
 
+.Internal services for certificate manager
 |===
 | Certificate Name | Issuer Name | Secret Name | Usage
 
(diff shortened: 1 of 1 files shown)


security/certificates.adoc
```diff
@@ -84,6 +79,6 @@ View the following table of the internal services that use `cert-manager` to cre
 | Topology
 |===
 
-Use the certificate policy controller to create and manage certificate policies.
+A custom expiration term is not supported for OpenShift Container Platform. For more information see Openshift Contaainer Platform https://docs.openshift.com/container-platform/4.5/security/certificate-types-descriptions.html#service-ca-certificates_ocp-certificates[Service CA certificates]. Use the certificate policy controller to create and manage certificate policies on managed clusters.
```
gparvin: This is the wrong link.  The link should be from our known issues: https://github.com/open-cluster-management/rhacm-docs/blob/2.0_stage/release_notes/known_issues.adoc#internal-error-500-during-login-to-the-console
I was trying to say we should either include the information from that known issue here or we should link to it.

dockerymick: thanks for clarifying!

PR #989, opened 2020-07-27
pointing the user to the troubleshoot topic

dockerymick:
https://github.com/open-cluster-management/backlog/issues/3839#issuecomment-663731504

### DIFF
--- a/install/uninstall.adoc
+++ b/install/uninstall.adoc
@@ -13,8 +13,7 @@ This adds the removal of other required components and settings to the items tha
 When you continue with this step, it removes all of the components and subscriptions that were not removed with the custom resource removal.
 If you complete this level of uninstallation, you must reinstall the operator before reinstalling the custom resource.
 
-+
-IMPORTANT: Before you uninstall the Red Hat Advanced Cluster Management hub cluster, you must detach all of the clusters that are managed by that hub cluster. For more information about detaching clusters, see the *Removing a cluster from management* section by selecting the information for your provider in  link:../manage_cluster/create.adoc#creating-a-cluster-with-red-hat-advanced-cluster-management-for-kubernetes[Creating a cluster with Red Hat Advanced Cluster Management for Kubernetes].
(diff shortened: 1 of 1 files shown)


install/uninstall.adoc
```diff
@@ -13,8 +13,7 @@ This adds the removal of other required components and settings to the items tha
 When you continue with this step, it removes all of the components and subscriptions that were not removed with the custom resource removal.
 If you complete this level of uninstallation, you must reinstall the operator before reinstalling the custom resource.
 
-+
-IMPORTANT: Before you uninstall the Red Hat Advanced Cluster Management hub cluster, you must detach all of the clusters that are managed by that hub cluster. For more information about detaching clusters, see the *Removing a cluster from management* section by selecting the information for your provider in  link:../manage_cluster/create.adoc#creating-a-cluster-with-red-hat-advanced-cluster-management-for-kubernetes[Creating a cluster with Red Hat Advanced Cluster Management for Kubernetes].
+IMPORTANT: Before you uninstall the Red Hat Advanced Cluster Management hub cluster, you must detach all of the clusters that are managed by that hub cluster. See, link:../troubleshooting/trouble_uninstall_detach.adoc[Troubleshooting failed uninstallation because resources exist] for the work around.
```
swopebe, line 16: We mentioned not sending users from **tasks** to **troubleshooting** in a call with Joydeep and I gave the justification to the meeting attendees: _Generally we don't do this in doc in order to prevent dual maintenance and having task sprinkled with things that will be fixed. If it is an invasive problem, the usual best-practice is to write it in the doc instead._ Is this a one-time things we are doing just for this topic? We don't want this to be common practice.

dockerymick: @bcpswope As discussed  Understood, thanks for the info and clarification. I will keep our maintenance goal noted. As discussed in Slack, yes this is an exception right now. @cadawson do you have any feedback and opinion?

cadawson (approved):
This looks good, in this case. For this issue, linking to the troubleshooting topic makes sense, in my opinion. Thanks!

PR #990, opened 2020-07-27
open-cluster-management/backlog#3210 ts reimp fail

cadawson:


troubleshoot_acm/trouble_cleanup_offline.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+[#symptom-reimporting-cluster]
+== Symptom: Reimporting a detached cluster fails
+
+After you detach a cluster and try to import it again, an error occurs. The error message looks similar to the following example: 
+
+----
+"error":"failed to install release: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "cert-policy-controller-role" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata;
+...
+----
```
hanqiuzh: ok, this symptom written here is when user detached an offline cluster, but because the managed cluster is offline, so it can not clean up all resources. and the user tried to cleanup the managed cluster without following the correct order (so some resources will be left). 

I want to say that the script may not fix these leftovers as the script is just deleting things in the right order, so the operators on managed cluster can remove the related resources. instead of explicitly deleting all resources. If user manually deleting the operators, we will not be able to remove all resources with the script.

Shall we update the symptom to just say when detaching offline cluster from hub, the managed cluster will still have resources leftover (e.g. pods in namespace `open-cluster-management-agent` and `open-cluster-management-agent-addon`)?  I guess I will also look at how to update the script to handle more scenarios, but that may take a while

PR #995, opened 2020-07-27
updates, adding a known issue (detaching a cluster)

dockerymick:
https://github.com/open-cluster-management/backlog/issues/3847


cadawson (approved):
Looks good.  Is it worth saying that it is "not deleted automatically, and must be removed manually?" Seems a little redundant, but is there a reason that they have to remove it? If so, it might be good to add the extra phrase.  Thanks!

dockerymick:
@cadawson Good point Chris. After I read the dev issue, the namespace is expected to be deleted when the cluster is detached, so I think we should ~~reference the user to the Removing a cluster from management section~~ 

I added the oc command to manually delete the ns

PR #999, opened 2020-07-27
workaround for updated cloud provider keys

dockerymick:
https://github.com/open-cluster-management/backlog/issues/3706


cadawson (approved):
Looks good! Thanks! I would remove the "also" from the first sentence, but it isn't world-ending.

PR #1012, opened 2020-07-28
open-cluster-management/backlog#3581 known issue add app

swopebe:


mikeshng (approved):
I've made some changes in https://github.com/open-cluster-management/rhacm-docs/pull/1012/commits/eacfe581954553300c3830e38bbe37b4d70069bc

- changed the topic to better reflect the actual issue. (It should be namespace channel subscription not namespace channel AND subscription) 
- use the `oc` client like all the most of the other commands in the doc.

PR #1031, opened 2020-07-29
open-cluster-management/backlog#4038 upgrd

cadawson:


cdoan1:
@cadawson there is technically no **upgrade** or **update** path from 1.0 to 2.0. So the current wording is not accurate. The doc that I'm writing now would probably be a best practices/guidance for **migrating** from 1.0 to 2.0. Customer will have existing 1.0 env, and they will remove 1.0, and redeploy 2.0---or in a separate cluster, deploy 2.0 while the old 1.0 is running, then move bits over--specifically here, first focusing on just the spoke migrating from one 1.0 hub to a 2.0 hub.

cdoan1:
@cadawson touched up the passage here https://gist.github.com/cdoan1/a30fa1b86df3f4229bd4e5934ec8fdee

swopebe:
@cdoan1 do you mind checking this update?

cdoan1 (approved):
LGTM, this just covers our general update procedure when a new channel update becomes available

PR #1036, opened 2020-07-30
no issue  - list and example formatting

cadawson:
Update to fix the list in this topic. Wanted you to have a look.


troubleshooting/trouble_app_deploy.adoc
```diff
@@ -38,5 +38,5 @@ See the following example:
 ----
   kubectl explain <resource>
 ----
-
++
```
swopebe: Check in file preview, this is not building. Removing line 41 addition (+)

swopebe (approved):

PR #1051, opened 2020-08-03
Post meeting, klusterlet and klusterlet add-on

swopebe:
https://github.com/open-cluster-management/backlog/issues/3986


Randy424 (approved):

swopebe:
got the wrong randy--oops, but this works! Thanks @Randy424!! LOL

 @randymgeorge gonna move with these for arch call.

PR #1072, opened 2020-08-10
Update known_issues.md

dockerymick:
https://github.com/open-cluster-management/backlog/issues/2757

Recreated PR on correct branch. Change is only applicable to 1.0

### DIFF
--- a/release_notes/known_issues.md
+++ b/release_notes/known_issues.md
@@ -222,7 +222,7 @@ In some cases, a component operator crashes. When the component operator crashes
    kubectl api-resources -o name --namespaced=true); do echo "===$api==="; kubectl get $api -n multicluster-endpoint
    ```
 
-2. Get a list of your subscriptions and delete them. Run the following command:
+2. Get a list of your subscriptions and delete them. Run the following commands:
 
    ```
    kubectl get subscriptions.apps.open-cluster-management.io
@@ -230,7 +230,7 @@ In some cases, a component operator crashes. When the component operator crashes
    kubectl delete subscriptions.apps.open-cluster-management.io
    ```
 
-3. Get a list of your CustomResourceDefinitions (CRDs). Delete the application CRDs by running the following command:
+3. Get a list of your CustomResourceDefinitions (CRDs). Delete the application CRDs by running the following commands:
 
    ```
    kubectl get crd |grep apps.open-cluster-management.io
@@ -254,8 +254,11 @@ In some cases, a component operator crashes. When the component operator crashes
    kubectl delete clusterrolebinding endpoint-appmgr
    ```
 
-6. What command can the user run to verify that the target cluster is clean? 
+6. Verify that there are no pods that exist in the `multicluster-endpoint` namespace by running the following command:
 
+   ```
+   kubectl get pods -n multicluster-endpoint
+   ```
 
 Your target cluster is cleaned.
 
--- a/release_notes/known_issues.md
+++ b/release_notes/known_issues.md
@@ -211,23 +211,23 @@ When you enter an unsupported value into the `yaml` content before creating your
 
    **Notice:** If you have this problem in one of your imported clusters, you cannot import another cluster successfully until you fix it. Even if you import a different cluster, the import fails until the problem is fixed.   
 
-## Self-destruct script does not clean target clusters 
+## Managed cluster not clean after it is detached from the hub cluster 
 <!--1.0.0:2757-->
 
-In some cases, a component operator crashes. When the component operator crashes, the self-destruct script does not remove all resources from target clusters. Complete the following steps to clean your target clusters:
+In some cases, a component operator crashes. When the component operator crashes, the self-destruct script does not remove all resources from target clusters. See the self-destruct script in the [`endpoint-operator` folder](https://github.com/open-cluster-management/endpoint-operator/blob/release-1.0.0/hack/self-destruct.sh). Complete the following steps to clean your target clusters:
 
-1. Remove your API resources by running the following command:
+1. List the remaining Advanced Cluster Management for Kubernetes resources by running the following command:
 
    ```
-   kubectl api-resources -o name --namespaced=true); do echo "===$api==="; kubectl get $api -n multicluster-endpoint
+   for api in $(kubectl api-resources -o name --namespaced=true); do echo "===$api==="; kubectl get $api -n multicluster-endpoint; done
    ```
 
 2. Get a list of your subscriptions and delete them. Run the following commands:
 
    ```
-   kubectl get subscriptions.apps.open-cluster-management.io
+   kubectl get subscriptions.apps.open-cluster-management.io --all-namespaces
 
-   kubectl delete subscriptions.apps.open-cluster-management.io
+   kubectl delete subscriptions.apps.open-cluster-management.io -n <namespace> <subscription-name>
    ```
 
 3. Get a list of your CustomResourceDefinitions (CRDs). Delete the application CRDs by running the following commands:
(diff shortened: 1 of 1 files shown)


chenz4027 (approved):
Thanks!

release_notes/known_issues.md
```diff
@@ -210,7 +210,58 @@ When you enter an unsupported value into the `yaml` content before creating your
       ```
 
    **Notice:** If you have this problem in one of your imported clusters, you cannot import another cluster successfully until you fix it. Even if you import a different cluster, the import fails until the problem is fixed.   
+
+## Self-destruct script does not clean target clusters
```
dhaiducek: I wasn't even aware there was a self-destruct script available until this issue. Do users know about it? If not, I think this should be titled something similar to `Managed clusters not clean after detaching from Hub` (and/or provide a link to the self-destruct script?)

@chenz4027 What do you think--do these steps cover the steps in the script so that we don't have to link to it?

release_notes/known_issues.md
```diff
@@ -210,7 +210,58 @@ When you enter an unsupported value into the `yaml` content before creating your
...
+
+## Self-destruct script does not clean target clusters 
+<!--1.0.0:2757-->
+
+In some cases, a component operator crashes. When the component operator crashes, the self-destruct script does not remove all resources from target clusters. Complete the following steps to clean your target clusters:
+
+1. Remove your API resources by running the following command:
+
+   ```
+   kubectl api-resources -o name --namespaced=true); do echo "===$api==="; kubectl get $api -n multicluster-endpoint
```
dhaiducek: This command is missing some code to make it work. However, it looks like this dumps out a list of resources. Should we just remove this step? Either that or we can fix the code and maybe rename it to "List the remaining RHACM resources..." instead of "Remove your API resources..."

For reference, the code should be:
```
for api in $(kubectl api-resources -o name --namespaced=true); do echo "===$api==="; kubectl get $api -n multicluster-endpoint; done
```

@chenz4027 Was the purpose here to delete each of the remaining resources after listing them?

release_notes/known_issues.md
```diff
@@ -210,7 +210,58 @@ When you enter an unsupported value into the `yaml` content before creating your
...
+   ```
+
+2. Get a list of your subscriptions and delete them. Run the following commands:
+
+   ```
+   kubectl get subscriptions.apps.open-cluster-management.io
```
dhaiducek: This is namespaced, so we'll probably need:
```
kubectl get subscriptions.apps.open-cluster-management.io --all-namespaces
```

release_notes/known_issues.md
```diff
@@ -210,7 +210,58 @@ When you enter an unsupported value into the `yaml` content before creating your
...
+
+   kubectl delete subscriptions.apps.open-cluster-management.io
```
dhaiducek: I think we'll need to add a namespace and resource:
```
kubectl delete subscriptions.apps.open-cluster-management.io -n <namespace> <subscription-name>
```

release_notes/known_issues.md
```diff
@@ -210,7 +210,58 @@ When you enter an unsupported value into the `yaml` content before creating your
...
+   ```
+
+3. Get a list of your CustomResourceDefinitions (CRDs). Delete the application CRDs by running the following commands:
+
+   ```
+   kubectl get crd |grep apps.open-cluster-management.io
+
+   kubectl delete apps.open-cluster-management.io
```
dhaiducek: Since we're deleting CRDs, we'll need:
```
kubectl delete crd <apps-crd-name>
```

release_notes/known_issues.md
```diff
@@ -210,7 +210,58 @@ When you enter an unsupported value into the `yaml` content before creating your
...
+   ```
+
+4. Run the following commands to delete the cluster role for the `endpoint-appmgr`:
    
+   ```
+   kubectl get clusterrolebinding | grep ^endpoint | awk '{print $1}'
```
dhaiducek: I think this line should be:
```
kubectl delete clusterrole $(kubectl get clusterrole | grep ^endpoint | awk '{print $1}')
```

Or, @chenz4027, is this to query for it first? And these should both be `clusterrole`, correct? In the original commands provided, the subshell says `clusterrolebinding`, so I wanted to make sure.

release_notes/known_issues.md
```diff
@@ -210,7 +210,58 @@ When you enter an unsupported value into the `yaml` content before creating your
...
+
+   kubectl delete clusterrole endpoint-appmgr
+   ```
+
+5. Run the following commands to delete the cluster rolebinding for the `endpoint-appmgr`:
+
+   ```
+   kubectl get clusterrolebinding | grep ^endpoint | awk '{print $1}'
```
dhaiducek: As above, this might need to be changed to (with confirmation from @chenz4027 about both commands):
```
kubectl delete clusterrolebinding $(kubectl get clusterrolebinding | grep ^endpoint | awk '{print $1}')
```

dhaiducek (changes requested):
Looks good, especially since it's only for 1.0.0. I've left comments about some of the lines of code, many that @chenz4027 can verify since she's seen the issue directly. 🙂

release_notes/known_issues.md
dockerymick: @dhaiducek thanks for your comment, that's a good point. There is a link for the self-destruct script: https://github.com/open-cluster-management/endpoint-operator/blob/master/hack/self-destruct.sh

@chenz4027 thoughts?

dockerymick: So far I have updated the title of the known issue and added the link for the self-destruct script

chenz4027: yep that's the script I was referring. The master branch is for 2.0 and I think they have a branch for 1.x as https://github.com/open-cluster-management/endpoint-operator/blob/release-1.0.0/hack/self-destruct.sh

release_notes/known_issues.md
chenz4027: yeah I think so - to delete the remaining resources. Do we have a combined command for delete maybe?

dockerymick:
@dhaiducek Thanks for the review, I revisited the issue and i misunderstood. I have updated the commands for the clusterrole and clusterrolebinding. @chenz4027 thanks for your response and review as well. May I ask for one more review from you both please? :)

chenz4027 (approved):
Looks good to me!

release_notes/known_issues.md
dhaiducek: Shooooot. I just noticed that `endpoint-operator` is a private repo, so we can't link to it at all from the docs. 🙁 

Okay, in that case we need to make sure these steps cover what the script is doing (or leave this issue as technical debt since we may not have a ton of 1.0 users?).

For the former, we could probably update the command in Step 1 to delete namespaced resources like this (haven't tested it yet):
```
for api in $(kubectl api-resources -o name --namespaced=true | grep multicloud.ibm); do echo "===$api==="; for resource in $(kubectl get $api -o name -n multicluster-endpoint); do kubectl delete ${resource} -n multicluster-endpoint --timeout=15s; done; done
```

release_notes/known_issues.md
```diff
@@ -210,7 +210,54 @@ When you enter an unsupported value into the `yaml` content before creating your
...
+4. Run the following commands to delete the cluster role for the `endpoint-appmgr`:
    
+   ```
+   kubectl delete clusterrole $(kubectl get clusterrole | grep ^endpoint | awk '{print $1}')
+   ```
+
+5. Run the following commands to delete the cluster rolebinding for the `endpoint-appmgr`:
+
+   ```
+   kubectl delete clusterrolebinding $(kubectl get clusterrole | grep ^endpoint | awk '{print $1}')
```
dhaiducek: @chenz4027 Should this be `kubectl get clusterrolebinding` instead of `clusterrole` for the inner command?
```
kubectl delete clusterrolebinding $(kubectl get clusterrolebinding | grep ^endpoint | awk '{print $1}')
                                                ^^^here
```

chenz4027: yep @dhaiducek Nice catch 

```
Kubectl delete clusterrolebinding $(kubectl get clusterrolebinding | grep ^endpoint | awk '{print $1}')
```

dockerymick:
@dhaiducek I removed the link to the repo and updated step 1

release_notes/known_issues.md
```diff
@@ -210,7 +210,54 @@ When you enter an unsupported value into the `yaml` content before creating your
       ```
 
    **Notice:** If you have this problem in one of your imported clusters, you cannot import another cluster successfully until you fix it. Even if you import a different cluster, the import fails until the problem is fixed.   
+
+## Managed cluster not clean after it is detached from the hub cluster 
+<!--1.0.0:2757-->
+
+In some cases, a component operator crashes. When the component operator crashes, the target clusters are not clean after they are detached from the hub cluster.  Complete the following steps to clean your target clusters:
+
+1. Delete the namesapce resources for Advanced Cluster Management for Kubernetes resources by running the following command:
```
dhaiducek: - "namespace" is misspelled
- We should provide an additional delete command here to delete the resources, like:
```
kubectl delete -n multicluster-endpoint <resource>
```

dhaiducek (changes requested):
Couple minor things and then I think this will be fine.

PR #1075, opened 2020-08-10
open-cluster-management/backlog#1573 master.adoc

swopebe:
-- reorg application topics


cadawson (approved):
Assuming the levels are where you want them and that the files will be created in the folder, this looks good.

PR #1104, opened 2020-08-17
Starting on the Fix Pack

dockerymick:
https://github.com/open-cluster-management/backlog/issues/4198

### DIFF
--- a/release_notes/fixpack.adoc
+++ b/release_notes/fixpack.adoc
@@ -1,4 +1,44 @@
 [#fix-pack-updates]
 = Fix pack updates
 
-There is currently no fix pack for this version of the product. Fixes are listed when fix packs are released.
+View a summarized list of {product-title} fix pack 2.0.1 updates.
+
+. {product-title} is bundled with the most recent version of `oc` (v4.5) and `kubectl` (v1.18) CLIs in the Visual Web Terminal.
+
+// ^ Might say, "Upgraded the oc and kubectl CLIs for the Visual Web Terminal."
+
+. As a user with an administrator role, you can access pod logs of managed cluster.
+
+// are there steps documented on how to do this? If not, maybe it should not be in FP, but a new issue be created and labeled 2.0 doc refresh.
+
+// Administrator role access to the pod logs of managed clusters is fixed.
+
+. Search component CustomResourceDefinition (CRD) is removed during uninstallation. 
+
+// Since the related issue #3689^ has been fixed, should the troubleshoot topic Troubleshooting reinstallation failure be removed? 
+
+// Might say, "The uninstall script was improved"
+
+. Added a label for `Bare metal` to the Cloud field options list, on the _Importing a cluster_ page.
+
+. The default network type is updated from OpenShiftSDN to OVNKubernetes to support IPV6.
+
+. As a user with the cluster role administrator role, you can manage your bare metal assets. 
+
+// ^is bm tech preview for 2.0.1? Should we mention tech preview fixes as well? 
+
+// Cluster role administrator access to manage bare metal assets is fixed. 
+
+. Subscriptions support `kustomization.yaml` files that contains an inline patch where the patch content inside the file as a single string.
+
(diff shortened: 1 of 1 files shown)


release_notes/fixpack.adoc
```diff
@@ -1,4 +1,15 @@
...
+ 
+. Upgraded the `oc` and `kubectl` CLIs to the latest versions for the Visual Web Terminal.
+. Administrator (`admin`) role access to the pod logs of managed clusters is fixed.
+. The product uninstallation process was improved.
+. Added a label for `Bare metal` to the Cloud field options list, on the _Importing a cluster_ page.
+. The default `Network type` when you create a cluster is updated from OpenShiftSDN to OVNKubernetes.
+. Subscriptions support `kustomization.yaml` files that contains an inline patch where the patch content inside the file is a single string.
+. Improved how cloud providers manage sensitive data. 
+. Removed DNS virtual IP parameter from the create cluster flow. (https://bugzilla.redhat.com/show_bug.cgi?id=1866513[BZ 1862506])
+// ^ check with doc team
```
swopebe: Referenced on Slack. See previous issue and discussion. 

If BZ is reported and accessed by external user, this makes sense. But if not, it is for internal-use only and not required for the user. We don't want to insert them just to insert them. Last time **all of this was internally reported** so it did not make sense to add internal bugs that user would not need to see. Also, we keep our list pretty basic by design compared to OCP. (By design meaning architects and such wanted a simple list, like the one you created, which is awesome!)

(And I know you already do this, just getting it doc'ed for future @dockerymick :-)! We look at the **GH issues** mostly to eliminate the known issues, and those come from the BZ for the ACM team.)

release_notes/fixpack.adoc
```diff
@@ -1,4 +1,15 @@
 [#fix-pack-updates]
 = Fix pack updates
 
-There is currently no fix pack for this version of the product. Fixes are listed when fix packs are released.
+View a summarized list of {product-title} fix pack 2.0.1 updates.
```
swopebe: Looks great!!!

release_notes/fixpack.adoc
dockerymick: @bcpswope Thanks Brandi, I will remove the link to BZ bc i think the link that I provided is internal. I also checked Errata and did not see the Bug yet: https://access.redhat.com/errata/#/?q=&p=1&sort=portal_publication_date%20desc&rows=10&portal_product=Red%20Hat%20Advanced%20Cluster%20Management%20for%20Kubernetes

release_notes/fixpack.adoc
```diff
@@ -1,4 +1,14 @@
...
+ 
+. Upgraded the `oc` and `kubectl` CLIs to the latest versions for the Visual Web Terminal.
+. Administrator (`admin`) role access to the pod logs of managed clusters is fixed.
+. The product uninstallation process was improved.
+. Added a label for `Bare metal` to the Cloud field options list, on the _Importing a cluster_ page.
+. The default `Network type` when you create a cluster is updated from OpenShiftSDN to OVNKubernetes.
+. Subscriptions support `kustomization.yaml` files that contains an inline patch where the patch content inside the file is a single string.
+. Improved how cloud providers manage sensitive data. 
+. Removed DNS virtual IP parameter from the create cluster flow.
+. Overview page continues to display cluster details even after managed clusters are detached.
```
jnpacker: Lets change it to this: Overview page will not become blank when cluster are detached.

release_notes/fixpack.adoc
```diff
@@ -1,4 +1,14 @@
 [#fix-pack-updates]
 = Fix pack updates
 
-There is currently no fix pack for this version of the product. Fixes are listed when fix packs are released.
+View a summarized list of {product-title} fix pack 2.0.1 updates.
+
```
jnpacker, line 5: We should add a bullet. Improvements and fixes to the cluster import process

PR #1105, opened 2020-08-18
open-cluster-management/backlog#4524 Upd 311

cadawson:


hanqiuzh (approved):

hanqiuzh:
3.11.200 looks good as the release note shows the bug was fixed in that version. we cannot install old versions, but the latest version of okd does fix this issue 😂 

and this is also the same for acm 2.1, do we need another pr for 2.1 branch?

PR #1117, opened 2020-08-19
adding a blurb about search fields

dockerymick:
Doc issue: https://github.com/open-cluster-management/backlog/issues/4619

### DIFF
--- a/console/search.adoc
+++ b/console/search.adoc
@@ -17,7 +17,7 @@ Search is case-sensitive when you query labels.
 See the following examples: name, namespace, status, and other resource fields.
 Auto-complete provides suggestions to refine your search.
 See the following example:
-  *** Search for a single field, such as `kind:pod` to find all pod resources. You can add more than one value.
+  *** Search for a single field, such as `kind:pod` to find all pod resources.
   *** Search for multiple fields, such as `kind:pod namespace:default` to find the pods in the default namespace.
 
 +
@@ -27,6 +27,7 @@ See the following example:
 Additionally, if you save and share a Search query with another user, returned results depend on access level for that user.
 For more information on role access, see _Using RBAC Authorization_ in the link:https://kubernetes.io/docs/reference/access-authn-authz/rbac/[Kubernetes documentation].
  ** You can also search with conditions by using characters, such as `+>, >=, <, <=, !=+`.
+ ** When you search for more than one value, the search returns either of the values that were queried.
 +
 See the following example:
 
--- a/console/search.adoc
+++ b/console/search.adoc
@@ -3,7 +3,9 @@
 
 For {product-title}, Search provides visibility into your resources across all your clusters.
 
-You can type any text value in the _Search box_ and results include anything with that value from any property, such as a name or namespace. For example, if you search for `RedHat`, you can receive results such as `RedHat123`. For more specific search results, include the property in your search. For example, search for `name:RedHat` to search only in the `name` property. 
+You can type any text value in the _Search box_ and results include anything with that value from any property, such as a name or namespace. If you search for `RedHat` and you manage a large number of clusters, it might take a longer time to receive search results. 
+
+For more specific search results, include the property selector in your search. You can combine related values for the property, for a more precise scope of your search. For example, search for `cluster:dev red` to receive results that match the string "red" in the `dev` cluster.
 
 . Click *Search* in the navigation menu.
 . Type a word in the _Search box_, then Search finds your resources that contain that value.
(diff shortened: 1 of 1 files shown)


console/search.adoc
```diff
@@ -21,7 +17,7 @@ Search is case-sensitive when you query labels.
 See the following examples: name, namespace, status, and other resource fields.
 Auto-complete provides suggestions to refine your search.
 See the following example:
-  *** Search for a single field, such as `kind:pod` to find all pod resources.
+  *** Search for a single field, such as `kind:pod` to find all pod resources. You can add more than one value.
```
jlpadilla: When you search for more than one value, we interpret it as an **OR**.  For example, `name: nameA,nameB` will return results with either `nameA` **OR** `nameB`

This works well for most fields like `name` because resources can only have 1 name.  If we did `nameA` **AND** `nameB` it won't match anything because a resource can't have both names at the same time.

However, this becomes a limitation for fields that can have multiple values for example `label` or `nodes`.  Currently it isn't possible to search for resources with `labelA` **AND** `labelB`.

console/search.adoc
```diff
@@ -3,11 +3,7 @@
 
 For {product-title}, Search provides visibility into your resources across all your clusters.
 
-NOTE: You can type any text value in the _Search box_ and results include anything with that value from any property, such as a name or namespace.
-For example, if you search for `RedHat`, you can receive results such as `RedHat123`.
-
-For more specific search results, include the property in your search.
-For example, search for `name:RedHat` to search only in the `name` property.
+You can type any text value in the _Search box_ and results include anything with that value from any property, such as a name or namespace. For example, if you search for `RedHat`, you can receive results such as `RedHat123`. For more specific search results, include the property in your search. For example, search for `name:RedHat` to search only in the `name` property.
```
jlpadilla, line 9: Can we add this note here?

> Searching for an arbitrary keyword like `RedHat` is slower, especially when you are managing a large number of clusters. When possible use a property selector for faster response. You could also combine both to narrow down the scope of your search. For example `cluster:dev red` will return any resources matching the string "red" in the cluster with name dev.

console/search.adoc
```diff
@@ -31,6 +27,7 @@ See the following example:
 Additionally, if you save and share a Search query with another user, returned results depend on access level for that user.
 For more information on role access, see _Using RBAC Authorization_ in the link:https://kubernetes.io/docs/reference/access-authn-authz/rbac/[Kubernetes documentation].
  ** You can also search with conditions by using characters, such as `+>, >=, <, <=, !=+`.
+ ** When you search for more than one value, the search returns either of the values that were queried.
```
jlpadilla: This might be ambiguous.  Here's some specific examples:
- search for `name:a,b` 
  - returns any resource with name `a` **OR** `b`
- search for `kind:pod name:a` 
  - Returns any pod named `a`, so in this case it's an **AND** because we match both conditions.
- search for `kind:pod name:a,b` 
  - Returns resources where kind is Pod **AND** the name is `a` **OR** `b`.

dockerymick:
@jlpadilla I have made more updates. I think we should specify which properties support multiple values. Thoughts?

console/search.adoc
jlpadilla: verified. thanks for making my explanation much shorter!  :-)

console/search.adoc
```diff
@@ -31,6 +29,9 @@ See the following example:
 Additionally, if you save and share a Search query with another user, returned results depend on access level for that user.
 For more information on role access, see _Using RBAC Authorization_ in the link:https://kubernetes.io/docs/reference/access-authn-authz/rbac/[Kubernetes documentation].
  ** You can also search with conditions by using characters, such as `+>, >=, <, <=, !=+`.
+ ** When you search for more than property selector with multiple values, the search returns either of the values that were queried. View the following examples:
+ *** When you search for `kind:pod name:a`, any pod named `a`.
+ *** When you search for `kind:pod name:a,b`, any pod named `a` and the name `a` or `b` are returned.
```
jlpadilla: This query will return only pods. My interpretation of the original text is that other results with names a or b could be returned, so want to make it clear.
```suggestion
 *** When you search for `kind:pod name:a,b`, any pod named `a` or `b` are returned.
```

jlpadilla:
> I think we should specify which properties support multiple values. Thoughts?

`label` and `role` are the most commonly used. I wouldn't document the specific properties because this is dynamic, so new properties could get added and it's difficult to keep the doc in sync.

Reminder that this is the case where we have a limitation.  Users might expect to be able to search for any resource with both `labelA` AND `labelB` but we don't support that.

jlpadilla (approved):
Thanks for the updates!

PR #1141, opened 2020-08-24
Certificate for bare metal

cadawson:


cadawson:
Can you take a look at this @jeswanke and @chrisahl ? This is the PR for 2.0.1. Thanks!

install/install_disconnected.adoc
```diff
@@ -29,6 +28,21 @@ If you do not already have a mirror registry, create one by completing the proce
 If you already have a mirror registry, you can configure and use your existing one.
 +
 
+. Bare metal only: Provide the certificate information for the disconnected registry in your `install-config.yaml` file. To access the image in a protected disconnected registry, you must provide the certificate information so {product-title-short} can access the registry.
```
chrisahl, line 31: If the UI stuff is implemented in the release, and the doc in file manage_cluster/prov_conn_bare.adoc shown below gets added, I don't think we need this section for additionalTrustBundle here.  Unless the user has ACM 2.0.0?  

@chenz4027 @cdoan1 @jeswanke Thoughts?

chenz4027: I'm thinking we may want to include this until we have the manage_cluster/prov_conn_bare.adoc had the content

cadawson: @jeswanke  confirmed that the fix is in the UI code for the 2.0.1 release, when the feature flag is enabled. I guess the question might be: Will people be running 2.0.0? If yes, then they will need that information. If they will all be pushed to 2.0.1, they should be able to follow the instructions with the UI change. 
Does it cause a problem to leave the manual steps in there, @chrisahl and @chenz4027, or does it just provide another way of accomplishing it?

chrisahl: Based on prior comments, there is no harm in leaving the manual steps for now.  Better to have too much help than not enough.   In 2.1 or 2.2 we might revisit the manual info and remove it.

chrisahl (approved):

PR #1170, opened 2020-09-02
added a reference to known issue for cloud providers

dockerymick:
https://github.com/open-cluster-management/backlog/issues/5034


cadawson (approved):
Looks good. Thanks! At least if we missed something, it will be consistent. :-)

PR #1191, opened 2020-09-08
open-cluster-management/backlog#4330 Inst 10 min

cadawson:


install/install_connected.adoc
```diff
@@ -127,7 +127,17 @@ error: unable to recognize "./mch.yaml": no matches for kind "MultiClusterHub" i
 +
 Run the command again in a few minutes when the resources are created.
 
-. View the list of routes after about 10 minutes to find your route:
+. Wait for 10 minutes before you view the status.
```
nathanweatherly: Hey! 

So the user can check the status immediately, they just won't see a status of `Running` for 10 minutes.

Maybe this line could be `Expect to wait 10 mintues before you view a `Running` status.

And then the line below of "The status should be `Running`." could be dropped ¯\_(ツ)_/¯

cadawson: Check the updates

cadawson:
Hi @nathanweatherly . When you have a chance, can you check the updates on this one? Thanks!

nathanweatherly (approved):

nathanweatherly:
Looking great! Thanks Chris!

PR #1192, opened 2020-09-08
Adding new parameters for cert policy controller

dockerymick:
https://github.com/open-cluster-management/backlog/issues/4286

### DIFF
--- a/security/cert_policy_ctrl.adoc
+++ b/security/cert_policy_ctrl.adoc
@@ -1,8 +1,10 @@
 [#certificate-policy-controller]
 = Certificate policy controller
 
-Certificate policy controller can be used to detect certificates that are close to expiring. Create durations that are long, or contain DNS names that fail to match specified patterns to notify you about the policy expiration. Configure and customize the certificate policy controller by updating the following parameters in your controller policy: `minimumDuration`, `minimumCADuration`, `maximumDuration`, `maximumCADuration`, `allowedSANPattern`, `disallowedSANPattern`.
-When a certificate expires in less than the minimum duration amount of time, the policy becomes noncompliant.
+Certificate policy controller can be used to detect certificates that are close to expiring. Create time durations (hours) that are long, or contain DNS names that fail to match specified patterns to notify you about the policy expiration.
+
+Configure and customize the certificate policy controller by updating the following parameters in your controller policy: `minimumDuration`, `minimumCADuration`, `maximumDuration`, `maximumCADuration`, `allowedSANPattern`, `disallowedSANPattern`.
+When a certificate expires in less than the minimum duration amount of time, the policy becomes non-compliant.
 The certificate policy controller is created on your hub cluster.
 
 The certificate policy controller communicates with the local Kubernetes API server to get the list of secrets that contain certificates and determine all non-compliant certificates.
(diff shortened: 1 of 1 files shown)


security/cert_policy_ctrl.adoc
```diff
@@ -1,9 +1,10 @@
 [#certificate-policy-controller]
 = Certificate policy controller
 
-Certificate policy controller can be used to detect certificates that are close to expiring.
-Configure and customize the certificate policy controller by updating the minimum duration parameter in your controller policy.
-When a certificate expires in less than the minimum duration amount of time, the policy becomes noncompliant.
+Certificate policy controller can be used to detect certificates that are close to expiring. Create time durations (hours) that are long, or contain DNS names that fail to match specified patterns to notify you about the policy expiration.
```
gparvin: I think it should be:
... close to expiring, have durations that are too long or contains DNS names that do no match specified patterns.

My notes for id:
Note that it does more than just check for expiration now.  Checking expiration is the default behavior, but you can add the new options to check the other things too.  There are probably lots of other checks we can keep adding in the future too -- so it might not be bad to consider a list format in the introduction paragraphs.

security/cert_policy_ctrl.adoc
```diff
@@ -1,9 +1,10 @@
...
+
+Configure and customize the certificate policy controller by updating the following parameters in your controller policy: `minimumDuration`, `minimumCADuration`, `maximumDuration`, `maximumCADuration`, `allowedSANPattern`, `disallowedSANPattern`.
+When a certificate expires in less than the minimum duration amount of time, the policy becomes non-compliant.
```
gparvin: The `minimumDuration` and minimumCADuration` are the only 2 options that deal with expiration now.  The other 4 options add in the extra features:
- checking the lifetime of the certificate
- checking DNS names against patterns

security/cert_policy_ctrl.adoc
```diff
@@ -81,14 +88,25 @@ Set the parameter value to `inform`.
...
 
 | spec.severity
-| Required. Specifies the severity when the policy is non-compliant. Use the following parameter values: `low`, `medium`, or `high`.
+| Required. Informs the user of the severity when the policy is non-compliant. Use the following parameter values: `low`, `medium`, or `high`.
 
 | spec.minimumDuration
-| Required. Parameter specifies the smallest duration (in hours) before a certificate is considered non-compliant.
-When the certificate expiration is greater than the `minimumDuration`, then the certificate is considered compliant.
-Default value is `100h`.
-The parameter value uses the time duration format from Golang.
```
gparvin, line 90: We have to have the format and default value.  The format applies to several values so it should be specified in each or marked in some way to make it clear.

gparvin (changes requested):
See comments added.

security/cert_policy_ctrl.adoc
dockerymick: @gparvin thanks for clarifying, it wasn't clear that these were multiple task done by the controller. My understanding was that time durations needed to be created, so that the controller can monitor the policy.

dockerymick:
@gparvin Thanks for the review, I have been able to make a few more edits based on your suggestions

security/cert_policy_ctrl.adoc
```diff
@@ -1,12 +1,22 @@
...
+- `maximumCADuration`
+- `allowedSANPattern`
+- `disallowedSANPattern`
+
+Your policy might become non-compliant due to either of the following scenarios:
+
+* When a certificate expires in less than the minimum duration of time or exceeds the maximum time.
+* When DNS names fail to match the specified pattern.
+
+The certificate policy controller is created on your hub cluster. The controller communicates with the local Kubernetes API server to get the list of secrets that contain certificates and determine all non-compliant certificates.
```
gparvin: The certificate policy controller is created on your hub cluster.
change to
The certificate policy controller is created on your managed clusters.

gparvin (approved):
Note that I added one small comment.  Looks good though.

swopebe (approved):
looks great!

PR #1217, opened 2020-09-14
etcd encryption policy

ckandag:
This PR is for issue https://github.com/open-cluster-management/backlog/issues/4529


dockerymick (approved):
I made a few changes for consistency, otherwise this looks good. Thanks for starting the draft. I also verified that all of the links worked

PR #1228, opened 2020-09-16
open-cluster-management/backlog#5106 self mng

cadawson:


install/install_connected.adoc
```diff
@@ -114,19 +114,22 @@ metadata:
   namespace: <namespace>
 spec:
   imagePullSecret: <secret>
+  disableHubSelfManagement: false
```
nathanweatherly, line 117: So this doesn't actually have to be here-- this won't appear in the UI when a customer installs, so it might be confusing.

There could be a note saying "your hub will be managed by default. If you'd like to opt-out of this management, set `spec. disableHubSelfManagement` to `true`"

install/install_connected.adoc
```diff
@@ -114,19 +114,22 @@ metadata:
...
 ----
 +
-Replace _namespace_ with your project namespace.
+Replace _namespace_ with your project namespace. 
 Replace _secret_ with the name of the secret that you created.
 +
+*Important:* The `local-cluster` namespace is used for the imported self-managed hub cluster. You must create the `local-cluster` namespace, if necessary, and ensure that no clusters were previously imported into it.
```
nathanweatherly: So the MCH operator actually creates the local-cluster namespace. The customer _shouldn't_ have a `local-cluster` namespace, but if they do, they should remove it

cadawson: Ah, ok. I thought it had to be there, but empty.

install/install_connected.adoc
cadawson: Yeah, that makes sense. I was trying to find the easiest way for the customer to get to the point where they have something there. I thought it would be less confusing if we just have them add it as false at the same time that they are adding the other stuff anyway. They can change it, if they need to, at that time, or just leave it alone. It is one more line to copy while they copy the other ones. It won't hurt their configuration. 
That was the thought process. If that is crazy, let me know.

nathanweatherly: I think you've got a good point. I hope though that if they are using the UI and _don't_ see that in the pre-populated spec, they don't freak out

nathanweatherly: I mean to that point, IDK if the `imagePullSecret` is populated in the default pre-populated spec.

OS

cadawson: We do include: If the setting is not included in the `.yaml` file that defines the custom resource, add it as shown in the example of the previous step.

install/install_connected.adoc
cadawson: Reworded. Check it out.

nathanweatherly: That's still not exactly right (sorry)

>*Important:* The `local-cluster` namespace is used for the imported self-managed hub cluster. You must not have a `local-cluster` namespace on you cluster prior to installing

cadawson: Should be better now.

install/install_connected.adoc
nathanweatherly: ~@cadawson for:~

~> We do include: If the setting is not included in the .yaml file that defines the custom resource, add it as shown in the example of the previous step.~

~Does "the setting" mean the pull secret?~

Sorry, I kept reading :+1:

nathanweatherly (approved):

PR #1235, opened 2020-09-16
open-cluster-management/backlog#2667 add 2.1

cadawson:
I suggest to have this reviewed with the cluster and search squad; I assigned @xiangjingli for any comments on the channel connectivity questions


cadawson:
Hi @TheRealHaoLiu . We started this and got some content out for 1.x. I would like to also push it to 2.0, if it is still accurate. If not, can we work on updating it for 2.0.x and 2.1? Thanks!

install/requirements.adoc
```diff
@@ -37,3 +37,23 @@ See the following versions that are tested and supported:
...
+* Outbound and inbound connectivity to the `WorkManager` service route on the ManagedCluster on port 443
+* Inbound connectivity to the hub cluster's kube API server from the `ManagedCluster` on port 6443
+* Inbound connectivity for post-commit hook from GitHub to the Hub. This setting is only required when you use certain application management functions.
+
+
+Managed cluster:
+* Inbound connectivity to the Kubernetes API server from the hub cluster on port 6443
+* Inbound connectivity to `WorkManager` service endpoint from the hub cluster on port 443
+* Outbound connectivity to the hub cluster's Kubernetes API server on port 6443
+* Outbound connectivity from the Hub to channel source, including Github, Object Store, and Helm repository. This is only required when you are using application lifecycle to connect to these sources.
```
xiangjingli: The outbound connectivity is also required from the managed cluster to channel source.  So I would suggest to change it to 

* Outbound connectivity from the managed cluster to channel source, including Github, Object Store, and Helm repository. This is only required when you are using application lifecycle to connect to these sources.

cadawson: Thank you, @xiangjingli . I added that change to the PR.

cadawson:
Comments added, merging.

PR #1251, opened 2020-09-21
Swope fixpack3

swopebe:
@jnpacker preliminary list of fixes that I gathered:

https://github.com/open-cluster-management/backlog/issues?q=is%3Aissue+milestone%3A%22ACM+2.0.3%22+label%3Abug+-label%3Aimagescan+-label%3Asquad%3Acicd+is%3Aclosed

Using that ^^^^ query. We can look together but in the interim, you may find I am missing big ticket items. I had to go with 2.0.3 milestone as best I could.

### DIFF
--- a/release_notes/fixpack.adoc
+++ b/release_notes/fixpack.adoc
@@ -1,6 +1,29 @@
 [#fix-pack-updates]
 = Fix pack updates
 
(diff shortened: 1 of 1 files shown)


cadawson (approved):
Aaapproved! Thank you!

PR #1252, opened 2020-09-21
Updating the docinfo.xml file

dockerymick:
https://github.com/open-cluster-management/backlog/issues/5555


cadawson (approved):
I see what you did here, and I think it should be good. Please run a build as soon as you merge it to see if it will mess up the builds. Any time we touch one of those build files, I get nervous that it might not run for some strange reason. I don't think there is any reason to believe that here, but I felt the same about some in the past.  Thank you!

PR #1255, opened 2020-09-22
open-cluster-management/backlog#5567 Inst cons

cadawson:


install/install_connected.adoc
```diff
@@ -175,21 +175,21 @@ Your {ocp-short} pull secret is associated with your Red Hat Customer Portal ID,
...
 ----
 apiVersion: operator.open-cluster-management.io/v1
 kind: MultiClusterHub
 metadata:
-  name: multiclusterhub
   namespace: <namespace>
-spec:
-  imagePullSecret: <secret>
+  name: multiclusterhub
+  labels: {}
```
nathanweatherly: I technically don't think this `labels` line is necessary-- do we ever tell customers to add specific labels? If not, we could just remove the `labels` line entirely

cadawson: I removed the labels line. I don't think we tell them to do that.

nathanweatherly (approved):

PR #1257, opened 2020-09-22
open-cluster-management/backlog#4770 peer review whole topic

swopebe:


manage_applications/subscribe_git_resources.adoc
```diff
@@ -12,20 +12,20 @@ Learn how to grant subscription administrator access.
...
+. From the terminal, log in to your {product-title-short} cluster.
 
 . Add the following subjects into `open-cluster-management:subscription-admin` ClusterRoleBinding with the following command:
 +
 ----
 oc edit clusterrolebinding open-cluster-management:subscription-admin
 ----
 +
-*Note:* `open-cluster-management:subscription-admin` ClusterRoleBinding has no subject initially.
+*Note:* `open-cluster-management:subscription-admin` initially ClusterRoleBinding has no subject.
```
cadawson: This looked like a sentence before, but not now. Should "initially" be capitalized and have a comma following it?

manage_applications/subscribe_git_resources.adoc
```diff
@@ -12,20 +12,20 @@ Learn how to grant subscription administrator access.
 
 . From the console, log in to your {ocp-short} cluster.
```
cadawson: Use {ocp} on the first occurrence.

cadawson (commented):
Just a couple of quick comments.

manage_applications/subscribe_git_resources.adoc
dockerymick: I agree here, I would change this to:

Initially, `open-cluster-management:subscription-admin` ClusterRoleBinding has no subject.

dockerymick: The `+` sign for alignment is also visible when i switch to the preview tab

dockerymick (changes requested):
I think you can remove the plus signs in the `Resource overwrite example` section. The codeblocks are not aligning well. The alignment looks fine in the section before that tho

manage_applications/subscribe_git_resources.adoc
swopebe: Yes, I saw that in the portal a couple of times--I am thinking it doesn't like bold after it but I have to investigate that one.

manage_applications/subscribe_git_resources.adoc
swopebe: good point

swopebe:
didn't even see those, removed @dockerymick

swopebe:
gonna merge so Roke can close within sprint

PR #1325, opened 2020-10-07
updating parameter table for cert

dockerymick:
https://github.com/open-cluster-management/backlog/issues/5807

@drkho @gparvin I think we should continue to keep "Required" to maintain consistency. Let me know your thoughts


drkho (commented):
This looks good...thanks.

gparvin:
@dockerymick I agree, I think this is good!

PR #1334, opened 2020-10-08
open-cluster-management/backlog#4856 Add upgrade

cadawson:


install/upgrade_hub.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+.. Select *Automatic* to enable automatic updates. 
+
+.. Select *Save* to commit your change. 
+
+.. Wait for the automatic updates to be applied to the operator. The updates automatically add the required updates to the latest version in the selected channel. When all of the updated updates are complete, the _Upgrade Status_ field indicates *Up to date*.
+
+. Now that the _Upgrade Status_ is *Up to date*, click the value in the _Channel_ field to edit it.  
+
+. Select the next available channel. You cannot skip channels when upgrading.
```
joeg-pro: Sug:  Maybe "Select the channel for the next available feature release." (instead of the "next available channel").

install/upgrade_hub.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+. Now that the _Upgrade Status_ is *Up to date*, click the value in the _Channel_ field to edit it.  
+
+. Select the next available channel. You cannot skip channels when upgrading. 
++
+*Important:* You cannot revert back to an earlier channel after upgrading to a later channel. You must uninstall the operator and reinstall it at an earlier version to use an earlier version.
+
+. Select *Save* to save your changes.
+
+. Wait for the automatic upgrade to complete. After the upgrade to the next channel completes, the updates to the latest releases within the channel are deployed.
```
joeg-pro: Sug:  "After the upgrade to the next release feature release completes,  the updates to the latest patch release within the channel are deployed."

Rational for this and preivous sugs:
Its think we should focus on the feature release as what is being available, not the channel.  The channel is "just" the mechanism by which that feature release and its later patch (z-stream) releases are delivered.

The word "release" is used to refer to a release of a new set of features (that is, an x.y.0 release like 2.1.0) and sometimes to refer to patches for such a release (the so-called z-stream updates x.y.1, .2... like 2..03).  I do not know if Red Hat has standard terminology to refer to the x.y.0 release as on kind of release and x.y.z (for z > 0) as another, but I have seen the words "feature release" and "patch release" used in other context so I propose those here.

install/upgrade_hub.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+. If you have to upgrade to later channels, repeat steps 7-9 until your operator is at the latest level of the desired channel. Make sure that all of the updates are deployed for your final channel.
```
joeg-pro: Sug:  I would suggest saying "upgrade to later feature release" rather than "later channels".

joeg-pro (approved):
I've made a few suggestions to focus more on the releases than the channels.  But I am ok with it as is if you don't agree with the sugs.

install/upgrade_hub.adoc
cadawson: Made this change.

install/upgrade_hub.adoc
cadawson: Updated the wording here around the suggestion.

PR #1344, opened 2020-10-09
Update trouble_cluster_offline_cert.adoc

hanqiuzh:
issue: https://github.com/open-cluster-management/backlog/issues/4916
removing some not needed steps. 

user don't need to wait for 2 hours, once showing offline on hub, user can start reimport.


cadawson (approved):
Thanks!

PR #1346, opened 2020-10-09
no issue - Han PR

cadawson:


cadawson:
Hi @hanqiuzh . Thanks for your PR. It was in the wrong branch, and would have probably caused conflicts when we tried to merge. I think this is the same thing, but make sure it is. Thanks!

hanqiuzh (approved):

PR #1363, opened 2020-10-14
peer review

swopebe:
-- a plug for peer reviews and style/product naming standards.


cadawson (approved):
Thanks Brandi! I would have gotten to it, but I appreciate you knocking it out!

swopebe:
no prob @cadawson -- it was a peer review that was needed/beneficial. I just happened to be the one to discover it, but peer reviews should be a major part of our culture so I don't mind.

PR #1381, opened 2020-10-20
open-cluster-management/backlog#4558 PEER REVIEW, and last two notes.

swopebe:


manage_applications/ansible_config.adoc
```diff
@@ -10,7 +10,7 @@
 
 * You must have Ansible Tower version 3.7.3 or a later version installed. It is best practice to install the latest supported version of Ansible Tower. See link:https://docs.ansible.com/ansible-tower/[Red Hat AnsibleTower documentation] for more details.
 
-* Version of {ocp-short} (They have OCP to install operator so we should mention what is tested here?)*
+* NOTE FOR DEV: Version of {ocp-short} (They have OCP to install operator so we should mention what is tested here?)*
```
swopebe, line 14: @mikeshng feel free to comment here or commit -- 

doc team peer review please ignore this line

manage_applications/ansible_config.adoc
```diff
@@ -71,7 +74,9 @@ You must create an Ansible Tower secret CR in the same subscription namespace. T
 
 Create the secret from the console by filling in the `Ansible Tower secret name` section. To create the secret using terminal, edit and apply the following `yaml`:
 
-*Notes:* The `namespace` is the same namspace as the subscription namespace. The `string data` `token` and `host` are from the Ansible Tower.
+*Notes:* The `namespace` is the same namspace as the subscription namespace. The `string data` `token` and `host` are from the Ansible Tower. 
+
+See the following YAML sample: *NOTE FOR DEV: Command `kubectl apply f` or something here??*
```
swopebe, line 84: Same here

manage_applications/ansible_config.adoc
```diff
@@ -71,7 +72,9 @@ You must create an Ansible Tower secret CR in the same subscription namespace. T
```
cadawson, line 72: This looks redundant, but there might be a reason for it.

manage_applications/ansible_config.adoc
```diff
@@ -44,21 +42,24 @@ When you create a subscription CR, the Git-branch and Git-path points to a Git r
 [#prehook]
```
cadawson, line 42: Line 42, 40, and the next several lines: In the Git root location, the two subfolders `prehook` and `posthook` should contain at least one `AnsibleJob` `Kind` resource.
I don't know if it would make it clearer to a reader if you split up `AnsibleJob` and `Kind`. Because they are both monospace, people might view it as an "AnsibleJob Kind" resource. Would it be helpful to split them with something like: In the Git root location, the two subfolders `prehook` and `posthook` should contain at least one `Kind` resource with the type of `AnsibleJob`? 
Might be clear as it is. I am not sure my solution would fit.

cadawson (approved):
A couple of comments here, and some commits to the branch. Looks good!

manage_applications/ansible_config.adoc
swopebe: yes, that needs to stay. We mention it a few times due to history of users not sure when resources and namespaces need to match.

manage_applications/ansible_config.adoc
swopebe: I will look at this with mike. I was actually thinking more like `Kind: AnsibleJob`.

swopebe: @mikeshng ^^^ ?

manage_applications/ansible_config.adoc
mikeshng: Since the Ansible team owns that operator. I will have to verify with them first.

manage_applications/ansible_config.adoc
mikeshng: `oc apply -f` or `kubectl apply -f` I don't know which we use more throughout the doc but they both will work.

swopebe: we have been moving towards `oc` so I will add that.

manage_applications/ansible_config.adoc
mikeshng: If we don't get any concrete answer then it should probably be the OCP version(s) that ACM is allowed to be installed on.

manage_applications/ansible_config.adoc
mikeshng: `Kind: AnsibleJob` sounds good to me.

mikeshng (approved):
Just waiting for the OCP version(s) requirement for the ansible platform operator

swopebe:
@mikeshng I added our command, the Kind annotation, and OCP 4.5 for now, until we hear otherwise.

PR #1382, opened 2020-10-20
peer review

swopebe:
Came across in commit while troubleshooting

If this is menu cascade: Click *this > this* (No need for the, tab not bold)

If explaining tabs: Click the *This* tab, then click *This*. (tab not bold, and need "the)

Not sure what you wanted but seems more like a menu cascade?


dockerymick (approved):
Thank Brandi!

PR #1391, opened 2020-10-20
open-cluster-management/backlog#4856 rev

cadawson:


cadawson:
Hi @cdoan1 .  Here are the updates that we we discussed today. Let me know if you have questions. Thanks!

PR #1398, opened 2020-10-21
open-cluster-management/backlog#4907 starting architecture review

swopebe:
@jnpacker I am looking at this and it seems not too out of date for 2.1, but maybe missing a few things. 

In the About: Architecture section, should we talk about observability (we will have it in What's new)

and in the image, should we change "endpoint" to "Klusterlet" -- what else am I missing at a higher level?


jnpacker:
@bcpswope Yes to both. Change endpoint to klusterlet and we should mention Observability

swopebe:
Klusterlet in endpoint, Obs first in the OCP box, bottom of managed cluster-- reorder maybe.

Observability paragraph from Joydeep.

dockerymick (commented):
I made a minor commit based on the live PR review i had earlier. I also added a conref. Changes look good to me!

dockerymick (approved):
I believe your last update is the image, going to approve

PR #1404, opened 2020-10-21
No issue, quality live meeting with Josh

swopebe:


manage_applications/app_model.adoc
```diff
@@ -9,6 +9,8 @@ When you need to create or update an application component resource, you need to
 [#application-secrets]
 == Application secrets
 
+//Review of this paragraph and section
+
 Secrets are Kubernetes resources that you can use to store authorization and other sensitive information, such as passwords, OAuth tokens, and SSH keys. By storing this information as secrets, you can separate the information from the application components that require the information to improve your data security.
```
jnpacker: paragraph 1 & 2 should go into the namespace channel type if we have a section for that, or in the samples section if we have a section for namespace channels (which is where secret subscriptions are supported).  If we don't have any namespace channel info, then paragraph 1 & 2 can be removed.

manage_applications/app_model.adoc
```diff
@@ -9,6 +9,8 @@ When you need to create or update an application component resource, you need to
 [#application-secrets]
 == Application secrets
 
+//Review of this paragraph and section
+
 Secrets are Kubernetes resources that you can use to store authorization and other sensitive information, such as passwords, OAuth tokens, and SSH keys. By storing this information as secrets, you can separate the information from the application components that require the information to improve your data security. 
 
 If you have applications that require secret resources to operate, you can use subscriptions to deploy the secrets to the managed clusters where the resources are needed.
```
jnpacker: paragraph 1 & 2 should go into the namespace channel type if we have a section for that, or in the samples section if we have a section for namespace channels (which is where secret subscriptions are supported).  If we don't have any namespace channel info, then paragraph 1 & 2 can be removed.

jnpacker: Paragraph 3 either goes under the channel section at the top, before we talk about each channel type or a modified version of the paragraph goes in each channel type section (git, helm, object storage).

jnpacker: All remaining paragraphs should be dropped, or incorporated where applicable into the Application section.

Also Application is last and we should move it up to the top, so it aligns with how you interact with the console UI.

Create application, pick a channel, define a subscription, and choose a placement rule.  So:
Application
Channel
Subscription
PlacementRule...  Assuming the flow works.

jnpacker (approved):

PR #1407, opened 2020-10-22
removing statement for consistency

dockerymick:
@ckandag please confirm that you agree with this update. @drkho may you also review to ensure that I understood your suggestion? 
https://github.com/open-cluster-management/backlog/issues/6130

### DIFF
--- a/security/rbac.adoc
+++ b/security/rbac.adoc
@@ -59,7 +59,6 @@ View the following examples:
 * To view `managedcluster x` cluster information, the following roles are required:
 
   ** A cluster-wide binding to the role `open-cluster-management:view:X `.
-  ** A namespace-scoped binding to the `view` role for namespace "X".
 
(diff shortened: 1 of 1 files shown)


drkho:
@dockerymick Thanks for looking into this. The [comment](https://github.com/open-cluster-management/backlog/issues/6130#issuecomment-707193948) I made was trying to clarify whether both statements are describing the same "read-only" use case.

In my opinion, I think we eventually would like customers to use both managed cluster roles( statement 2) to view cluster resources. So, perhaps we can remove the statement 1? Or move it as a note under statement 2?

dockerymick:
@drkho thanks for clarifying, I moved the first statement after the example. However, I didn't label it as a Note bc it will be unnecessary. In my opinion,  the statement will be noticed since it is following the listed bullet points.

ckandag (approved):

PR #1417, opened 2020-10-23
adding troubleshot topic

dockerymick:
According to the comment here: https://github.com/open-cluster-management/backlog/issues/4029#issuecomment-690517831

The fix to reduce the search-collector memory consumption was merged for 2.1, so I am curious if we should remove the second fix for Errata 2.0.4


console/search.adoc
```diff
@@ -39,4 +39,6 @@ See the following example:
   *** Search for `kind:pod restarts:>1` to find all pods that restarted at least twice.
 . If you want to save your search, click the *Save disk* icon.
 
+You might need to increase the memory limit for the `search-collector`. For more information, see link:../troubleshooting/trouble_search.adoc#troubleshooting-the-search-collector-pod[Troubleshooting the search collector pod].
+
```
swopebe: Usually we don't link to troubleshooting often in the regular docs--only in extreme situations we do this to avoid interrupting user with things they may not need and avoid maintenance.

in this case, it looks like it is only for IBM MCM customers, and that means it does not apply to everyone. Usually if it is something that applies across the board, it changes in the task and is not a troubleshooting topic. I think we mentioned this with another one in the past but there was an exception.

troubleshooting/trouble_search.adoc
```diff
@@ -0,0 +1,34 @@
+[#troubleshooting-the-search-collector-pod]
```
swopebe: looks great.

swopebe (commented):
I am not sure we should point to this from the main doc, see comment. That is the only change.

(I would choose a more specific file name since every t-shooting requires a file. trouble_search_coll.adoc or something.)

you may already be doing this or have done it, the master.adoc will need this reference.

swopebe (changes requested):
I would choose a different, more specific file name since there can be many in search.

prob you are already on this, but don't forget master.adoc

I think we should not point to it from search, see comment.

thank you!

console/search.adoc
dockerymick: Thanks for the reminder! I will remove the reference and only add it to the troubleshoot intro topic here: https://github.com/open-cluster-management/rhacm-docs/blob/2.1_stage/troubleshooting/troubleshooting_intro.adoc

troubleshooting/trouble_search.adoc
```diff
@@ -0,0 +1,34 @@
...
+[#resolving-the-problem-reinstallation-failure]
+== Resolving the problem: Reinstallation failure
+
+You must increase the memory limit in the `search-collector` deployment. Complete the following steps to increase the memory limit:
+
+. Log in to your {ocp} hub cluster. 
+. Access the `search-collector` deployment by running the following command:
++
+----
+oc edit deploy search-prod-ee318-search-collector
```
robdolares:
```suggestion
oc edit deployment $(oc get deployment -l component=search-collector -o jsonpath='{.items[0].metadata.name}')
```

robdolares (approved):
LGTM

swopebe (approved):

PR #1426, opened 2020-10-27
component work, VWT not tech preview (dev review)

swopebe:


swopebe:
Spoke to dev, no changes needed.

PR #1441, opened 2020-10-29
open-cluster-management/backlog#6605 rem ns

cadawson:


cadawson:
Discussed with the squad, and this should not be a required step. Looking at how to address. Closing this PR.

PR #1461, opened 2020-11-03
peer review needed

swopebe:
Changes to increase conciseness, parallelism, and highlight the two areas that STSM and OM have both said are important.


cadawson (approved):
looks good!

PR #1469, opened 2020-11-04
open-cluster-management/backlog#6467 App work (potential for refresh)

swopebe:


manage_applications/channel_sample.adoc
```diff
@@ -52,54 +52,40 @@ spec:
...
-| Optional.
-Identifies the namespace that the channel controller monitors for new or updated deployables to retrieve and promote to the channel.
-For a `Namespace` channel, the namespace must be on the hub cluster.
+| Optional. Identifies the namespace that the channel controller monitors for new or updated deployables to retrieve and promote to the channel.
 
 | spec.type
-| Required.
-The channel type.
-The supported types are: `Helm`, `Git`, and `Object storage`
+| Required. The channel type. The supported types are: `Helm`, `Git`, and `Object storage`
```
rokej: The supported types are HelmRepo, Git and ObjectBucket.

manage_applications/scheduling_deployment.adoc
```diff
@@ -1,24 +1,25 @@
...
-When a time window is defined for a subscription, the time ranges when a subscription is active changes. As part of defining a time window, you can define the subscription to be _active_ or _blocked_ during that window. The deployment of new or changed resources begins only when the subscription is active. Regardless of whether a subscription is active or blocked, the subscription continues to monitor for any new or changed resource. The active and blocked setting affects only deployments.
+When a time window is defined for a subscription, the time ranges when a subscription is active changes. As part of defining a time window, you can define the subscription to be _active_ or _blocked_ during that window. 
+
+The deployment of new or changed resources begins only when the subscription is active. Regardless of whether a subscription is active or blocked, the subscription continues to monitor for any new or changed resource. The active and blocked setting affects only deployments.
 
 When a new or changed resource is detected, the time window definition determines the next action for the subscription.
 
 * For subscriptions to `HelmRepo`, `Object storage`, and `Git` type channels:
```
rokej: For subscriptions to `HelmRepo`, `ObjectBucket`, and `Git` type channels:

manage_applications/channel_sample.adoc
swopebe: we had a few meetings on it for both UI and Doc, because there were inconsistencies and were left with Object storage

swopebe: Will change these:

Brandi Swope:coffee:  10:07 AM
https://github.com/open-cluster-management/backlog/issues/5020
10:07
If it is, then we have to write it as-is
10:07
but maybe next release get on the same term, I will match what you have in back-end
New

Roke Jung  10:09 AM
so it looks like UI changed the terms but not the API.

Brandi Swope:coffee:  10:09 AM
ok, will match the API and then maybe we can catch next time?

PR #1476, opened 2020-11-04
Updating the asterisk element

dockerymick:
Fixing the Pantheon book by removing an invalid element. The asterisk is considered an invalid element. There is a specific [HTML entity](https://theasciicode.com.ar/ascii-printable-characters/asterisk-ascii-code-42.html) that must be entered. 

Reason for update:
Pantheon-invalid-element

Invalid element from VSCode:

VSC-invalid-ascii-syntax

Update in PR: 
VSC-html-entry-asterisk

### DIFF
--- a/troubleshooting/trouble_obs_cert.adoc
+++ b/troubleshooting/trouble_obs_cert.adoc
@@ -6,13 +6,13 @@ The metrics collector might fail to push metrics to the {product-short} hub clus
 [#symptom-observability-invalid-certification]
 == Symptom 1: Certification is invalid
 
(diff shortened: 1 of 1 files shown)


swopebe (changes requested):
We would typically not use this HTML tag or the astericks, sorry, I didn't see this was a peer review or I would have said something earlier. I would like to get the goal behind this code before we merge?

dockerymick:
Goal: avoid use HTML entries for elements due to translation issues 

Agreed to surround the pod name with < >

swopebe (approved):

PR #1501, opened 2020-11-05
adding note about disabling observability cr

dockerymick:
https://github.com/open-cluster-management/backlog/issues/6859


cadawson (approved):
Looks good. Do you need to add it to the console section, as well?

PR #1515, opened 2020-11-09
UX changes for errata for 2.1, default information added

swopebe:
Please note filename changed for UX, which was not done last release to avoid breaking links.


dockerymick (approved):
Looks good to me, verified that the links work as well. Thanks!

PR #1520, opened 2020-11-10
open-cluster-management/backlog#6932

swopebe:
--small edits to intiate clean up and review

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -31,7 +31,7 @@ Review the known issues for {product-title}. The following list contains known i
 == Installation known issues
 
 [#openshift-container-platform-cluster-upgrade-failed-status]
-=== {ocp-short} cluster upgrade failed status
+=== OpenShift Container Platform cluster upgrade failed status
 
 // 2.0.0:3442
 
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -53,7 +53,7 @@ kubectl get crd | grep certificates.certmanager
 ----
 
 [#upgrade-leaves-clusterserviceversion-pending]
-=== Upgrade from version 2.0.4 to version 2.1 leaves the ClusterServiceVersion in a pending state.
+=== Upgrade from version 2.0.4 to version 2.1 leaves the ClusterServiceVersion in a pending state
 
 // 2.1.0:6747
 
@@ -64,7 +64,7 @@ NAME                                 DISPLAY                                    
 advanced-cluster-management.v2.1.0   Advanced Cluster Management for Kubernetes   2.1.0     advanced-cluster-management.v2.0.4   Pending
 ----
 
-To workaround this issue, complete the following steps to find and create the missing `clusterRole` custom resource:
+To work around this issue, complete the following steps to find and create the missing `clusterRole` custom resource:
 
 . Find all `clusterrolebinding` resources that were deployed by the {product-title-short} 2.1 CSV by entering the following command: 
 +
@@ -188,7 +188,7 @@ As a result, the policy was not propagated to managed cluster, or the violation 
 When you import a cluster that was previously managed and detached by a {product-title-short} hub cluster, the import process might fail the first time. The cluster status is `pending import`. Run the command again, and the import should be successful. 
 
 [#klusterlet-runs-on-a-detached-cluster]
-=== {klusterlet} runs on a detached cluster
+=== Klusterlet runs on a detached cluster
 
 // 2.0.0:3460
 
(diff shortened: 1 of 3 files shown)


release_notes/known_issues.adoc
```diff
@@ -188,7 +188,7 @@ As a result, the policy was not propagated to managed cluster, or the violation
 When you import a cluster that was previously managed and detached by a {product-title-short} hub cluster, the import process might fail the first time. The cluster status is `pending import`. Run the command again, and the import should be successful. 
 
 [#klusterlet-runs-on-a-detached-cluster]
-=== {klusterlet} runs on a detached cluster
+=== Klusterlet runs on a detached cluster
```
dockerymick, line 193: Are we removing Klusterlet from the [attributes](https://github.com/open-cluster-management/rhacm-docs/blob/2.1_stage/modules/common-attributes.adoc#L4)?

release_notes/known_issues.adoc
```diff
@@ -332,7 +332,7 @@ Replace _namespace_ with the namespace of the managed cluster.
 [#observability-annotation-query-failed]
 === Metrics are unavailable in the Grafana console
 
-* *Annotation query failed in the Grafana console*: 
+* Annotation query failed in the Grafana console:
```
dockerymick, line 337: I bolded this, so that users could notice that there are two explanations. i see why it would be beneficial to remove the bold tags

dockerymick (commented):
Made a couple of comments, thanks for creating the PR

release_notes/known_issues.adoc
swopebe: @dockerymick I understand...bold for emphasis was always discouraged in IBM Style manual, largely because a screen reader just says it is bold, without an explanation. If screen reader says: Click (bold) button, the user understands the bold emphasis.

If headers are bold, the screen reader reads it as a header.

Using bold randomly for emphasis was mentioned here: https://github.com/open-cluster-management/rhacm-docs/blob/gh-pages/_pages/ui_text_writing.md

(pulled from the manual)

We don't do it across the doc for this reason, so we can chat if we want to open this up for change... but the reason is largely accessibility/screen reader/semantic tagging.

swopebe: headers are a good option here, as well

release_notes/known_issues.adoc
swopebe: I removed the conref here to get rid of characters in headers to try to fix whatever is happening, but did not remove it from attributes because someone is using it. It would break those cases.

In a meeting we mentioned not creating this conref (early after the move). I was told it was needed because it keeps changing, but it does _not_ keep changing. It was always Klusterlet, then they wanted to use Endpoint and drop Klusterlet. The term did not change.

I am not conreffing it when I write it because the _conref_ is exactly what the _word_ is. 

If the term changes, we would change the conref, anyway. I see the point in doing this legacy tools, but not in the tools we use today.  Used so commonly, it is also why we don't conref hub cluster, managed cluster, etc... 

For headers, I was trying to remove whatever could be throwing off the whole topic.

release_notes/known_issues.adoc
dockerymick: H4 headers?

cadawson (approved):
Looks fine to me. Thanks!

swopebe:
@cadawson I was hoping to use this PR to get all of us to review the sections, parts of the examples, for instance, are not correct but I don't know what they should be. So I was hoping for more commits on this in areas where formatting may be causing problems?

release_notes/known_issues.adoc
swopebe: Unless you want to use a list (like you have) you can create subsections with headers, yes.

swopebe: (I mean if you want bold, that is)

dockerymick: I will leave it as is

dockerymick (approved):

release_notes/known_issues.adoc
dockerymick: I think we should also remove the conref from the header on line 34. The header is not highlighted in VSC
Screen Shot 2020-11-10 at 4 46 18 PM

cadawson:
Removed the conref in the header. Still looks the same to me, but it is probably a good idea to remove it. I did a couple of other attempts at the formatting changes, but nothing worked. I see the same thing that I am looking at in the 2.0 version, so maybe I am not seeing the same thing. We can look at it.

dockerymick:
@cadawson Thank you, i took a look from VSC and your change works. In the following screenshot I underlined the headers again to show that it is registering as a header. i also circled the branch, to point out that I was on the correct branch. A snapshot of the previous file can be found in my [comment](https://github.com/open-cluster-management/rhacm-docs/pull/1520#discussion_r520894686). 
Screen Shot 2020-11-11 at 8 15 16 AM

swopebe:
Gonna merge the known issues PR we have been working with, in an effort to make sure we didn't upset the build and see if it improved. Thanks for your attention and trail/error on this.

PR #1529, opened 2020-11-12
open-cluster-management/backlog#7040 Add ns and oc

cadawson:
Hi Mikela. This should be ok with adding the namespace to the command and changing kubectl to oc, but I wanted to give you the right of refusal or check with dev. Thanks!


dockerymick (approved):
Thanks Chris!

PR #1554, opened 2020-11-23
open-cluster-management/backlog#6952 --- adding troubleshooting topic

swopebe:


troubleshooting/trouble_git_server.adoc
```diff
@@ -0,0 +1,105 @@
+[#troubleshooting-application-git-server]
+= Troubleshooting application Git server connection 
+
+If you cannot create an application with a private Git repository, patch hub clusters and managed clusters to connect to private Git server with a custom certificate.
+
+[#symptom-git-server]
+== Symptom: Git server connection 
+
+After you create a channel, placement rule, application and subscription, you receive an error: `no such file or directory` when you clone a repository.
+
```
swopebe, line 10: I was confused from the notes when they receive the error and what the exact error is.... is it after they create.... all resources (can we just say any resource), or is it after clone attempt---


Obviously I need more clarity here, but I wrote this line with what I could tell. Please commit or comment.

@rokej @xiangjingli

troubleshooting/trouble_git_server.adoc
```diff
@@ -0,0 +1,105 @@
...
+After you create a channel, placement rule, application and subscription, you receive an error: `no such file or directory` when you clone a repository.
+ 
+[#resolving-git-server]
+== Resolving the problem: Git server connection 
+
+*Important:* Before you apply these instructions, upgrade to the product version 2.1, if you have not already.
+
+. On the hub cluster, update the channel CRD with with the link:(channels_crd.yaml)[https://github.com/open-cluster-management/multicloud-operators-channel/blob/master/deploy/crds/apps.open-cluster-management.io_channels_crd.yaml] 
+
```
swopebe, line 17: They just copy or add the entire thing?

troubleshooting/trouble_git_server.adoc
```diff
@@ -0,0 +1,105 @@
...
+oc edit csv advanced-cluster-management.v2.1.0 -n open-cluster-management
+----
+
+. Find the following containers:
+
++
+- `multicluster-operators-standalone-subscription` 
+- `multicluster-operators-hub-subscription` 
+
+. Update the container images. Your update should resemble the following:
```
swopebe: I would rather say: append the image names with ______. See the following example:

troubleshooting/trouble_git_server.adoc
```diff
@@ -0,0 +1,105 @@
...
+quay.io/open-cluster-management/multicluster-operators-subscription:2.2-PR337-91af6cb37d427d22160b2c055589a4418dada4eb`
+---- 
+
++
+The update recreates the following images in the `open-cluster-management` namespace: 
+
+
+- `multicluster-operators-standalone-subscription-xxxxxxx`
+
+- `multicluster-operators-hub-subscription-xxxxxxx`
```
swopebe: xxxxx is what we appended with?

troubleshooting/trouble_git_server.adoc
```diff
@@ -0,0 +1,105 @@
...
+
++
+The update recreates the following images in the `open-cluster-management` namespace: 
+
+
+- `multicluster-operators-standalone-subscription-xxxxxxx`
+
+- `multicluster-operators-hub-subscription-xxxxxxx` 
+
+. Check that the new pods are running with the new docker image.
```
swopebe: How do they "check" (line 53)

troubleshooting/trouble_git_server.adoc
rokej: @bcpswope I would say...

The logs from the subscription controller pod (multicluster-operators-hub-subscription-xxx pod in open-cluster-management namespace) indicate that it fails to clone the Git repository due to `x509: certificate signed by unknown authority` error and `BadGateway` error.

troubleshooting/trouble_git_server.adoc
rokej: @bcpswope Can't we just use https://github.com/open-cluster-management/backlog/issues/6952#issuecomment-726870103 as is for patching instruction? This instruction was tested by two different customers.

troubleshooting/trouble_git_server.adoc
rokej: @bcpswope Can't we just use open-cluster-management/backlog#6952 (comment) as is for patching instruction? This instruction was tested by two different customers.

troubleshooting/trouble_git_server.adoc
rokej: @bcpswope Can't we just use open-cluster-management/backlog#6952 (comment) as is for patching instruction? This instruction was tested by two different customers.  

xxxx represents a random characters which is different on every RHACM cluster.

troubleshooting/trouble_git_server.adoc
swopebe: @rokej that is what I used but I cannot write it in the docs the way it is.... it needs to be rewritten and formatted for docs, and cleaned up a bit -- also needs to be steps, like any other task. So I used what I could here to build it but needs more clarity.

troubleshooting/trouble_git_server.adoc
swopebe: see above comment

troubleshooting/trouble_git_server.adoc
rokej: @bcpswope Can't we just use open-cluster-management/backlog#6952 (comment) as is for patching instruction? This instruction was tested by two different customers.

Check with this command 

oc get pod multicluster-operators-hub-subscription-xxxxxxx -n open-cluster-management -o yaml

and look for the container image.

troubleshooting/trouble_git_server.adoc
swopebe: ok--I think that needs to be <what-it-is> just like we create any variable in the docs, this is keeping with our conventional standards.

PR #1562, opened 2020-12-01
https://github.com/open-cluster-management/backlog/issues/5972 -- rem…

swopebe:
…ove pipe and edit priority line


KevinFCormier (approved):
Thanks, @bcpswope!

For `rhacm-docs/manage_applications/channel_sample.adoc`, you can safely change the `spec.labels` entry in the table to `metadata.labels`. But I leave it to @xiangjingli as to whether you should add something for `spec.gates.labelSelector` to that table.

swopebe:
Sure, I will handle that one @KevinFCormier before I merge. Thanks so much.

swopebe:
channel doc also changed, will post both in issue

PR #1573, opened 2020-12-04
open-cluster-management/backlog#7527 Add ki clusim

cadawson:


cadawson:
Merging to get some sort of change in ASAP. Touched base with Josh last night, and he agreed that this was the right direction. Josh is on PTO today.

jnpacker (commented):
You could also have them delete the exist ClusterImageSet based on the name in the error and retry the deploy. This is probably better, so they don't have two references appear.

PR #1583, opened 2020-12-07
open-cluster-management/backlog#5334 Mgd clus fix

cadawson:


cadawson:
Using this PR to cover https://github.com/open-cluster-management/backlog/issues/5334 and https://github.com/open-cluster-management/backlog/issues/6605. Does that work?

manage_cluster/remove_managed_cluster.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+If there are remaining resources on the managed cluster that you removed, there are additional steps that are required to ensure that you remove all of the remaining components. Situations when these extra steps are required include the following examples:
+
+* The managed cluster was detached before it was completely created, and components like the `klusterlet` remain on the managed cluster. 
+
+* The hub that was managing the cluster was lost or destroyed before detaching the managed cluster, and there is no way to detach the managed cluster from the hub. 
+
+* The managed cluster was not in an online state when it was detached.
+
+* The namespace of the managed cluster was not removed.
```
hanqiuzh: > The namespace of the managed cluster was not removed.

this one is ambiguous, and I guess we can add the detailed namespace (open-cluster-management-agent, open-cluster-management-agent-addon)? and to make use of the script, we have to make sure pods are running in these namespaces, and the namespace is not in terminating state. I guess that's too complex. Or we can remove this item?

manage_cluster/remove_managed_cluster.adoc
```diff
@@ -0,0 +1,83 @@
...
+*Tip:* You can detach or destroy multiple clusters by selecting the check boxes of the clusters that you want to detach or destroy.
+Then select *Detach* or *Destroy*.
+
+*Note:* If you attempt to detach the hub cluster, which is named `local-cluster`, be aware that the default setting of `disableHubSelfManagement` is `false`. This setting causes the hub cluster to reimport itself and manage itself when it is detached and it reconciles the `MultiClusterHub` controller. It might take hours for the hub cluster to complete the detachment process and reimport. If you want to reimport the hub cluster without waiting for the processes to finish, you can enter the following command to restart the `multiclusterhub-operator` pod and reimport faster:
+
+----
+oc delete po -n open-cluster-management `oc get pod -n open-cluster-management | grep multiclusterhub-operator| cut -d' ' -f1`
+----
+
+You can change the value of the hub cluster to not import automatically by changing the `disableHubSelfManagement` value to `true`, as described in link:../install/install_connected.adoc#installing-while-connected-online[Installing while connected online].
```
hanqiuzh, line 29: line 23-29 & line 40-46 seems repeated? but they are in different section. I think it's ok 👍

hanqiuzh:
@cadawson  I agree it's good to combine 👍 left some comments

manage_cluster/remove_managed_cluster.adoc
cadawson: Hmm, ok. This is where I was trying to merge a couple of issues, and it might not be possible. The issue was raised about providing instructions for removing the namespace when you remove a cluster. It sounds like the namespace should be removed automatically, so no instructions should be required. What I tried to do is to use this issue to cover the case when the namespace is not removed. Will that work? Is the procedure the same?

cadawson: I am going to remove line 59 for now. It sounds like it might be a little tougher than I thought.

dhaiducek (approved):
Looks good to me! 👍

hanqiuzh (approved):

PR #1598, opened 2020-12-10
Updates for IAM policy

dockerymick:
https://github.com/open-cluster-management/backlog/issues/7566

### DIFF
--- a/security/iam_policy_ctrl.adoc
+++ b/security/iam_policy_ctrl.adoc
@@ -18,17 +18,16 @@ View the following example of an IAM policy and review the parameters in the YAM
 [source,yaml]
 ----
 apiVersion: policy.open-cluster-management.io/v1
-kind: IamPolicy
+kind: IamPolicy 
 metadata:
-  name: iam-grc-policy
-  label:
-    category: "System-Integrity"
+  name: {{name}}-example
 spec:
+  severity:
   namespaceSelector:
-    include:
-    exclude:
-  remediationAction:
-  maxClusterRoleBindingUsers:
+    include: ["*"]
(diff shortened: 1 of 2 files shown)


security/iam_policy_ctrl.adoc
```diff
@@ -18,17 +18,16 @@ View the following example of an IAM policy and review the parameters in the YAM
...
+  severity:
   namespaceSelector:
-    include:
-    exclude:
-  remediationAction:
-  maxClusterRoleBindingUsers:
+    include: ["*"]
+    exclude: ["kube-*", "openshift-*"]
+  remediationAction: 
+  maxClusterRoleBindingUsers: 5
```
dhaiducek, line 30: Since this is the YAML structure, I don't think we want to fill in anything here (saving the example entries for the sample at the bottom):
```
apiVersion: policy.open-cluster-management.io/v1
kind: IamPolicy 
metadata:
  name:
spec:
  severity:
  namespaceSelector:
    include:
    exclude:
  remediationAction:
  maxClusterRoleBindingUsers:
```

And I don't think `spec.remediationAction` is required nor `spec.severity`.

@ycao56 would you confirm, please?

ycao56: correct

dockerymick: Thanks for confirming!

dockerymick: @ycao Are remediationAction and severity optional parameters for all policies?

ycao56: severity is optional for all. but not remediationAction.

PR #1599, opened 2020-12-10
imageManifestVuln note added

dockerymick:
https://github.com/open-cluster-management/backlog/issues/7456


gparvin (approved):
Thanks for the update!

PR #1619, opened 2020-12-14
open-cluster-management/backlog#1580 for team feedback/commit

swopebe:


dockerymick (approved):
Added my commit, thanks for leading this Brandi

PR #1628, opened 2020-12-15
open-cluster-management/backlog#7720 No Int 4 dsc

cadawson:


cadawson:
@cdoan1 As we Slacked about, here is the proposed update that I believe will add the clarity for this issue. If it looks ok, I will add it to 2.2, as well, so it is moved forward. Thanks!

cadawson:
Slacked with Chris Doan about this. Added the change that should make this clearer. Merging.

PR #1630, opened 2020-12-15
open-cluster-management/backlog#7770 GCP rule

cadawson:


chrisahl (approved):
I am fine with this for now.  We may look at adding some code to UI to not allow some of these illegal combinations but this will cover us for now.

PR #1638, opened 2020-12-17
https://github.com/open-cluster-management/backlog/issues/6789 adding…

swopebe:
… bundle

I went ahead and added this for review.

### DIFF
--- a/console/vwt_search.adoc
+++ b/console/vwt_search.adoc
@@ -5,11 +5,13 @@ The Visual Web Terminal is a *tech preview* feature. You can use the Visual Web 
 
 With Visual Web Terminal, commands that return cluster resource data are displayed visually in interactive tabular format, rather than plain text that is displayed in a standard terminal. You can see the data, click on a row, and then see more details about that resource. 
 
-Before you begin, check the versions of the CLI bundles. For instance, check the version of {ocp}, kubectl, and Helm CLIs. Run the following commands to ensure you have the latest supported versions:
+Visual Web Terminal by default includes the `oc`, `kubectl`, `helm` CLI binaries in versions that are equivalent to the supported {ocp-short} version. You do not need to install any CLI binaries to use Visual Web Terminal.
 
-- `oc version` returns client side version as 4.6.z
-- `kubectl version` returns client side version as 1.19.z
-- `helm version` returns 3.3.4 or greater
+You can check the versions by running the following commands:
+
+- `oc version`
+- `kubectl version`
+- `helm version`
 
 *Notes:*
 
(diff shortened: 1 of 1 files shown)


console/vwt_search.adoc
```diff
@@ -5,6 +5,12 @@ The Visual Web Terminal is a *tech preview* feature. You can use the Visual Web
 
 With Visual Web Terminal, commands that return cluster resource data are displayed visually in interactive tabular format, rather than plain text that is displayed in a standard terminal. You can see the data, click on a row, and then see more details about that resource.
```
mark-nc, line 7: I do not think this is how we want to communicate this information.  This makes it sound like you cannot use Visual Web Terminal until you run those commands and check the versions and we certainly do not want to imply that as that is not a requirement.

Also, specifying exactly what to look for is problematic because when we build the image downstream, we pull in whatever version the downstream OCP tools repo provides and could change for 2.2.0 and then 2.2.1, etc.

My idea had been did we need to say something that implies our intent of what we provide to customers, which is something like "....the versions of oc, kubectl and helm provided match the latest OCP version supported by the ACM hub cluster".  That implies what we intend to package without going into details.  But the real question of this issue was whether we need to say something about this at all.  I think we need input on that from our architects.

swopebe, line 7: @cadawson do you mind peer review of line 8-13? See if it makes sense to you, please? Thank you!

console/vwt_search.adoc
```diff
@@ -5,6 +5,14 @@ The Visual Web Terminal is a *tech preview* feature. You can use the Visual Web
 
 With Visual Web Terminal, commands that return cluster resource data are displayed visually in interactive tabular format, rather than plain text that is displayed in a standard terminal. You can see the data, click on a row, and then see more details about that resource. 
 
+Visual Web Terminal by default includes the `oc`, `kubectl`, and `helm` CLI binaries in versions that are equivalent to the supported {ocp-short} version. You do not need to install any CLI binaries to use Visual Web Terminal.
```
cadawson: I might change this to read: "By default, Visual Web Terminal includes the versions of the `oc`, `kubectl`, and `helm` CLI binaries that are equivalent to the versions that are supported by {ocp-short}. You do not need to install any additional CLI binaries to use Visual Web Terminal."

mark-nc: In fact, it is impossible to install any new CLI binaries in Visual Web Terminal.  We do not include the `curl` command in the session environment so there is no way for the user to "download" and install anything.  So...we might want to modify that last sentence that Chris suggests to "....to use Visual Web Terminal and, in fact, it is not possible to install any additional CLI binaries."

swopebe: good edits, done, thanks!

mark-nc (approved):
Looks good!

PR #1644, opened 2021-01-04
open-cluster-management/backlog#7300 Inst scr merge

cadawson:


nathanweatherly:
I ran the script t prove that it doesn't blow up 😂 

@JakobGray  could you give this a look-over to make sure I didn't miss anything?

nathanweatherly:
Jake said it's cool. @cadawson I think we're good to merge! 

Note: I think this should also be back=ported to 2.1 doc

PR #1650, opened 2021-01-07
Update for observability 

dockerymick:
https://github.com/open-cluster-management/backlog/issues/6428


observability/manage_observe.adoc
```diff
@@ -135,6 +96,48 @@ route:
...
+----
+oc get mco observability -o yaml
+----
+
+. Access the `observability-metrics-custom-whitelist.yaml` file in the 
+`open-cluster-management-observability` namespace by running the following command:
++
+----
+cat observability-metrics-custom-whitelist.yaml
+----
```
clyang82, line 118: create this file instead of access it in namespace. cannot access file in k8s environment and there is no such configmap in k8s environment. it requires the user to create such configmap with name `observability-metrics-custom-whitelist`

dockerymick: Thanks for the review Chun! @clyang82 may you re-review and approve?

PR #1660, opened 2021-01-08
open-cluster-management/backlog#8039 back-end

swopebe:
-- merge by default edits, more info needed

### DIFF
--- a/manage_applications/subscribe_git_resources.adoc
+++ b/manage_applications/subscribe_git_resources.adoc
@@ -123,6 +123,7 @@ data:
   age: 20
 ----
 
+
 You log in as a subscription administrator and create a subscription with `apps.open-cluster-management.io/reconcile-option: replace` annotation. See the following example:
 
 ----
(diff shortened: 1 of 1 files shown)


manage_applications/subscribe_git_resources.adoc
```diff
@@ -124,6 +123,7 @@ data:
   age: 20
 ----
 
+
 You log in as a subscription administrator and create a subscription with `apps.open-cluster-management.io/reconcile-option: replace` annotation. See the following example:
```
chenz4027: I would suggest to flip the sequence of `merge` as `replace` as merge is the default setting in both front-end and back-end.

manage_applications/subscribe_git_resources.adoc
```diff
@@ -154,38 +154,4 @@ data:
...
-data:
-  name: user1
-  age: 20
-----
-
-When the `merge` option is used, entries from subscribed resource are either created or updated in the existing resource. No entry is removed from the existing resource.
-
-*Important:* If the exising resource you want to overwrite with a subscription is automatically reconciled by another operator or controller, the resource configuration
-is updated by both subscription and the controller or operator. Do not use this method in this case.
+*Important:* If the existing resource you want to overwrite with a subscription is automatically reconciled by another operator or controller, the resource configuration is updated by both subscription and the controller or operator. Do not use this method in this case.
```
chenz4027, line 165: Should we still say `merge` somewhere in the section? Letting the user know what we describe is the merge reconcile, etc.

chenz4027 (commented):
Comment with some general ideas. Also looking forward to hearing from @rokej

manage_applications/subscribe_git_resources.adoc
swopebe: Yes @chenz4027 -- good point.

@rokej would you agree with me reordering `replace` and `merge` since now in the UI, the user will see merge by default?

manage_applications/subscribe_git_resources.adoc
rokej: @chenz4027 @bcpswope Yes, flipping the sequence is fine.

manage_applications/subscribe_git_resources.adoc
rokej: @chenz4027 @bcpswope Yes, flipping the sequence is fine.

chenz4027 (approved):
LGTM!

rokej (approved):

PR #1664, opened 2021-01-12
removing observability troubleshoot topic (cert)

dockerymick:
https://github.com/open-cluster-management/backlog/issues/8224#issuecomment-757652696


cadawson (approved):
Looks good from this. We will find out if there are other links that we don't know about when we run it, I think. Thanks!

PR #1684, opened 2021-01-20
open-cluster-management/backlog#6557 Add ts clust

cadawson:


skeeey (approved):

cadawson:
Thank you, @skeeey . Will this apply to 2.2, as well, or is it something that will not happen in 2.2? If it does, then I can add it.

troubleshooting/trouble_cluster_offline_avail.adoc
```diff
@@ -0,0 +1,28 @@
...
+When the network that connects the managed cluster to the hub cluster is unstable, the status of the managed cluster that is reported by the hub cluster cycles between `offline` and `available`. 
+
+[#symptom-cluster-status-changing-from-offline-to-available]
+== Symptom: Cluster status changing from offline to available
+
+After a cluster is deployed, the cluster status changes from `offline` to `available`. The status continues to cycle between the two statuses. 
+
+[#identifying-the-problem-cluster-status-changing-from-offline-to-available]
+== Identifying the problem: Cluster status changing from offline to available
+
```
dockerymick: I recommend that you move this explanation to the beginning of the file. The intro is repetitive to the symptom, so I would remove the _Identifying the problem_ section.  For example, I would update this to read: 


"Troubleshooting cluster status changing from offline to available

The status of a deployed cluster alternates from `offline` to `available` without any manual changes to the environment or cluster.

Symptom:

When the network that connects the managed cluster to the hub cluster is unstable, the status of the managed cluster that is reported by the hub cluster cycles between `offline` and `available`."

The rest of the file looks good to me!

dockerymick (changes requested):
I made a suggestion for you. Let me know what you think. Otherwise the PR looks good to me! Links are functional

troubleshooting/trouble_cluster_offline_avail.adoc
cadawson: Sounds good. I didn't like the repetition, but thought it might need to be there for structure. Glad you didn't think so, either. Thank you!

dockerymick (approved):
Made a small commit, changes look good!

skeeey:
> Thank you, @skeeey . Will this apply to 2.2, as well, or is it something that will not happen in 2.2? If it does, then I can add it.

@cadawson , yes, I think this also need to apply to 2.2, thank you

cadawson:
Sounds good, @skeeey . Added it to both releases.

PR #1686, opened 2021-01-20
Cert updates

dockerymick:
https://github.com/open-cluster-management/backlog/issues/8446#issuecomment-761190425

### DIFF
--- a/security/certificates.adoc
+++ b/security/certificates.adoc
@@ -183,10 +183,9 @@ Observability certificates are automatically refreshed upon expiration. View the
 
 CA certificates can be associated with Git channel that are a part of the {product-title-short} application management. See link:../manage_applications/configuring_git_channel.adoc[Configuring application channel and subscription for a secure Git connection] for more details.
 
-//i decided not to reference the specific section bc there are no task in the section. Will ask Brandi to update her file to point the user to the doc that explains how to create a secret manage_applications/secret_sample.adoc
-
 Helm channels allow you to disable certificate validation. Helm channels must be configured in development environments because disabling certificate validation introduces security risks.
 
+//i decided not to reference the specific section bc there are no task in the section. Will ask Brandi to update her file to point the user to the doc that explains how to create a secret manage_applications/secret_sample.adoc
 
 [#managed-cluster-certificates]
 === Managed cluster certificates
--- a/security/certificates.adoc
+++ b/security/certificates.adoc
@@ -20,10 +20,10 @@ See the following topics to replace certificates:
 * xref:../security/cert_root_ca.adoc#replacing-the-root-ca-certificate[Replacing the root CA certificate]
 * xref:../security/cert_mgmt_ingress.adoc#replacing-the-management-ingress-certificates[Replacing the management ingress certificates]
 
-[#list-certificates]
-== List certificates
+[#list-managed-certificates]
+== List managed certificates
 
-You can view a list of certificates that use `cert-manager` internally by running the following command:
+You can view a list of managed certificates that use `cert-manager` internally by running the following command:
 
 ----
 oc get certificates.certmanager.k8s.io -n open-cluster-management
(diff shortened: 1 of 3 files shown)


security/certificates.adoc
```diff
@@ -32,24 +32,167 @@ oc get certificates.certmanager.k8s.io -n open-cluster-management
...
+* The `metrics-collector` restarts to mount the renewed certificates.
++
+*Note:* `metrics-collector` can push metrics to the hub cluster before and after certificates are removed. For more information about refreshing certificates across your clusters, see the <<refresh-a-certificate,Refresh a certificate>> section. Be sure to specify the appropriate namespace when you refresh a certificate.
+
+[#channel-certificates]
+=== Channel certificates
+
+CA certificates can be associated with Git channel that are a part of the {product-title-short} application management. See link:../manage_applications/configuring_git_channel.adoc[Configuring application channel and subscription for a secure Git connection] for more details.
+
+Helm channels allow you to disable certificate validation. Helm channels must be configured in development environments because disabling certificate validation introduces security risks.
```
gparvin: The second sentence about helm channels is unclear.  Not all helm channels must be configured in development  environments -- only the ones where they chose to disable certificate validation.  Disabling certificate validation is an extra configuration option they have to chose -- only when they do that is the connection "risky".

gparvin (commented):
I added one comment about helm channels just to clarify it's the disabling of the certificate validation that makes it risky.

security/certificates.adoc
dockerymick: @gparvin Thanks for clarifying, I have updated the sentence. May you re-review the PR when you have an opportunity please?

gparvin (approved):
Thanks!

security/certificates.adoc
```diff
@@ -32,24 +32,164 @@ oc get certificates.certmanager.k8s.io -n open-cluster-management
...
+* {product-title-short} sends the refreshed certificates to managed clusters.
+
+* The `metrics-collector` restarts to mount the renewed certificates.
++
+*Note:* `metrics-collector` can push metrics to the hub cluster before and after certificates are removed. For more information about refreshing certificates across your clusters, see the <<refresh-a-certificate,Refresh a certificate>> section. Be sure to specify the appropriate namespace when you refresh a certificate.
+
+[#channel-certificates]
+=== Channel certificates
+
+CA certificates can be associated with Git channel that are a part of the {product-title-short} application management. See link:../manage_applications/configuring_git_channel.adoc[Configuring application channel and subscription for a secure Git connection] for more details.
```
dockerymick: @rokej May you review this section since we are referencing application. We would like to point the user specifically to [Updating certificates and SSH keys section](https://github.com/open-cluster-management/rhacm-docs/blob/2.2_stage/manage_applications/configuring_git_channel.adoc#updating-certificates-and-ssh-keys). I think it would be beneficial to the user if we pointed them to the [Secret sample](https://github.com/open-cluster-management/rhacm-docs/blob/2.2_stage/manage_applications/secret_sample.adoc). The link would be at the end of  Updating certificates and SSH keys section. 


Do you have any thoughts on my proposed change?

dockerymick (commented):
@rokej may your review the Channel section of the security/certificates.adoc file please?

dockerymick:
> Thanks!

Thank you ! :) @gparvin

security/certificates.adoc
rokej: @dockerymick Yes, I agree with you on pointing the user specifically to Updating certificates and SSH keys section. However, the Secret sample does not contain samples about CA certificates or SSH keys. These samples are contained in other sections of ../manage_applications/configuring_git_channel.adoc. 

So probably we can just add a sentence like "see above sections for reconfiguring certificates and SSH keys" in Updating certificates and SSH keys section.

dockerymick: @gparvin Based on @rokej suggestion, what do you think about pointing the user to the section, [Using custom CA certificates for a secure HTTPS connection](https://github.com/open-cluster-management/rhacm-docs/blob/8446/manage_applications/configuring_git_channel.adoc#using-custom-CA-certificates-for-secure-HTTPS-connection) instead of directing them to the Updating certificates and SSH keys section?

gparvin: Sounds great!  Thanks!

PR #1689, opened 2021-01-22
Seems we were missing the steps to obtain the Ansible Token.

jnpacker:


berenss:
I found that before I did the steps starting from Users, I needed to have an Organization defined first. Perhaps your env already had it?

jnpacker:
Mine was a fresh install of the latest Tower operator.

manage_applications/ansible_config.adoc
```diff
@@ -8,20 +8,42 @@
...
+The Ansible Tower URL is the same URL that is used to log in to Tower. This is required by the _Application console_ or the Tower access secret when configuring an application with Ansible prehooks and posthooks. 
+
+See the following example URL: `https://ansible-tower-web-svc-tower.apps.my-openshift-cluster.com`. 
+
+[#obtain-a-token]
+== Obtaining a token
+
+. Log in to your Ansible Tower console.
+. Click *Users* in the console navigation.
+. Search for the user that you want to use.
```
swopebe: Search for the users that you want to use?? Can we better say this maybe without "use" like 

Search for the correct user? Search for the correct user role that you need?

jnpacker: "Search for the correct user" works for me!

jnpacker (commented):
Looks good

PR #1700, opened 2021-01-27
review comments

cadawson:


security/rbac.adoc
```diff
@@ -52,50 +52,50 @@ Some product resources are cluster-wide and some are namespace-scoped. You must
...
 
 [#cluster-lifecycle-RBAC]
 === Cluster lifecycle RBAC
 
-To create and admin all managed clusters
+To create and administer all managed clusters:
 
-* Create a cluster role binding to the cluster role `open-cluster-management:cluster-manager-admin`. This role is a super user, who has access to all resources and actions. This allows creating of cluster-scoped managedclusters, the namespace for the managedcluster, and resources in the namespace. This also allows access to provider connections and bare metal assets used to create managed clusters.
+* Create a cluster role binding to the cluster role `open-cluster-management:cluster-manager-admin`. This role is a super user, which has access to all resources and actions. This role allows you to create cluster-scoped `managedcluster` resources, the namespace for the `managedcluster` resource, and resources in the namespace. This role also allows access to provider connections and to bare metal assets that are used to create managed clusters.
```
jamestalton: You are not creating a `namespace for the `managedcluster` resource` as the managedcluster is a cluster scoped resource.  You are creating a namespace for the namespace-scoped resources which are created for managing the managed cluster.

Also I think we should use bare-metal?

jamestalton: Maybe...
`the namespace for the resources to managed the managed cluster`
Here I am saying managed cluster as I'm not talking about the managedcluster resource

cadawson: Thanks. That explanation helps. My understanding there was a bit off. I like your suggestion. How about: "This role allows you to create cluster-scoped `managedcluster` resources. the namespace for the resources that manage the managed cluster, and the resources in the namespace."

For bare metal assets, we don't hyphenate bare metal when we refer to it anywhere else. It looks like it is most often not hyphenated. Bare metal servers, bare metal recovery. IBM did not hyphenate it, so we didn't.

jamestalton: sounds good

PR #1713, opened 2021-02-01
making change to fix TS book

dockerymick:
No issue, fixing the book


swopebe:
@dockerymick looks like you and I submitted the same change, I just merged mine since it was pretty cut and dry, thank you for jumping in on it.

PR #1716, opened 2021-02-02
Adding troubleshoot topic for mco cr

dockerymick:
[Dev issue](https://github.com/open-cluster-management/backlog/issues/8264)
[Doc issue](https://github.com/open-cluster-management/backlog/issues/8303)


clyang82 (approved):
lgtm

troubleshooting/trouble_obs_mco.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+After you update the `statefulSetSize`, you might receive an alert that reads, `KubePersistentVolumeFillingUp`. The `data-observability-observatorium-thanos-receive-default-x` PVCs might be affected.
+
+
+[#resolving-pvc-does-not-change-after-editing-statefulsetsize]
+== Resolving the problem: PVC does not change after editing statefulSetSize
+
+If you have this problem, update the `storage` parameter in each `data-observability-observatorium-thanos-receive-default-x` PVC. Complete the following steps:
+
+. Confirm that the disk space is full by running a query. Log in to the {ocp} console.
```
cadawson: I think the procedure that you are providing here might be clearer as:
. Confirm that the disk space is full by running a query. 
.. Log in to the {ocp} console.
.. From the navigation menu, click *Monitoring* > *Metrics*. 
.. Enter the following query in the _Expression_ window and sort the query by ascending by the _Volume_ column:
+
----
kubelet_volume_stats_available_bytes{namespace="open-cluster-management-observability"}/kubelet_volume_stats_capacity_bytes{namespace="open-cluster-management-observability"}
----
+
*Note:* If the value is 0, then the disk is full.

CD: If that is actually the steps for running the query that you mention. Typically each action is a separate step. Of course, there can be exceptions. but that that is the general rule of thumb.

troubleshooting/trouble_obs_mco.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+. From the navigation menu, click *Monitoring* > *Metrics*. Enter the following query in the _Expression_ window and sort the query by ascending by the _Volume_ column:
++
+----
+kubelet_volume_stats_available_bytes{namespace="open-cluster-management-observability"}/kubelet_volume_stats_capacity_bytes{namespace="open-cluster-management-observability"}
+----
++
+*Note:* If the value is 0, then the disk is full.
+
+. Expand the `data-observability-observatorium-thanos-receive-default-x` PVC to reflect the edited `statefulSetSize` in the mco CR. Run the following command:
```
cadawson: Is MCO here multicluster observability or multicluster operator? I assume the former, but might not be clear to everyone.

cadawson: Maybe put "Run the following command..." in its own step. Unless, now that I read it again, that is how to "Expand the...". In that case, consider something like: Expand the ...in the MCO CR by running the following command:" Would not make sense here to use a nested list for a single step.

cadawson (changes requested):
Looks good. A couple of comments. Thanks!

PR #1726, opened 2021-02-02
adding ts topic for mcobservability 2.1

dockerymick:
https://github.com/open-cluster-management/backlog/issues/8303


troubleshooting/trouble_obs_mco.adoc
```diff
@@ -0,0 +1,46 @@
+[#pvc-does-not-change-after-editing-statefulsetsize]
+= PVC does not change after editing statefulSetSize
+
+When you update the `statefulSetSize` for any cluster that has the observability service enabled, the persistent volume claim (PVC) is not updated.
+
+
+[#symptom-pvc-does-not-change-after-editing-statefulsetsize]
+== Symptom: PVC does not change after editing statefulSetSize
```
bjoydeep: Symptom: Both of what is mentioned below will happen. It is just depends on what the user notices first:

1. Current data from the grafana dashboards in ACM Console disappear - therefore the default ACM-Clusters Overview dashboard will be largely empty. Historical data will still be available.
2. On the managed clusters, logs for the `metrics-collector-deployment-* `pod running in namespace `open-cluster-management-addon-observability` : will show an HTTP 500 error failing due to disk space

To confirm
Run a query to check that indeed the disk space is full.
Log into OpenShift console, click on Monitoring, select Metrics and enter this query in the Expression window:

`kubelet_volume_stats_available_bytes{namespace="open-cluster-management-observability"}/kubelet_volume_stats_capacity_bytes{namespace="open-cluster-management-observability"}`
and sort ascending by the Value column. The closer the value to 0, the more full the disk is.

troubleshooting/trouble_obs_mco.adoc
```diff
@@ -0,0 +1,46 @@
...
+On managed clusters, you might receive the following error message in the logs for the `metrics-collector-deployment-x` pod:
+
+----
+HTTP 500, failing due to disk
+----
+// need error message
+
+[#resolving-pvc-does-not-change-after-editing-statefulsetsize]
+== Resolving the problem: PVC does not change after editing statefulSetSize
+
```
bjoydeep, line 23: Resolving the problem: We need to expand the PVC size.
Note: Changing stateful set size (spec.Observability.statefulSetSize) in MultiClusterObservability CR will not work.
Steps:
To rectify
Expand the PVC size: as an example :
kubectl edit pvc data-observability-observatorium-thanos-receive-default-0
and change the storage under

```
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 

```
This may take some time to work. At the end when you run this command:
`kubectl get pvc data-observability-observatorium-thanos-receive-default-0 -o yaml`
you will see that value of storage under spec and status match.

Verify the fix:

1. Current data will start appearing in the grafana dashboards and 
2. metric collector pod logs - which was showing HTTP 500 error earlier -  will show http 200 on data sends

troubleshooting/trouble_obs_mco.adoc
dockerymick: @bjoydeep Since the user is confirming that the the disk is not fill, I think It is best that the confirmation remain in the resolve section

troubleshooting/trouble_obs_mco.adoc
dockerymick: @bjoydeep Can you confirm what the user is updating the `storage`  parameter to read? Should the value match the value used for the `statefulSetSize` or the `status` parameter?

troubleshooting/trouble_obs_mco.adoc
bjoydeep: Here is the sequence:
User checks to see that disk is full by running: `kubelet_volume_stats_available_bytes{namespace="open-cluster-management-observability"}/kubelet_volume_stats_capacity_bytes{namespace="open-cluster-management-observability"}`
Then the user adds space by expanding PVC
Finally - to verify that everything is in order, like you are saying it is good to run this query again.

troubleshooting/trouble_obs_mco.adoc
bjoydeep: `storage` parameter should be set to a size greater than that `statefulSetSize` because what was set in `statefulSetSize` was not enough.

bjoydeep (approved):
lgtm

PR #1729, opened 2021-02-03
open-cluster-management/backlog#8610--object-store

swopebe:


troubleshooting/trouble_object_store.adoc
```diff
@@ -0,0 +1,55 @@
...
+  namespace: ch-obj
+  labels:
+    name: obj-sub
+data:
+  AccessKeyID: YWRtaW4=
+  SecretAccessKey: cGFzc3dvcmRhZG1pbg==
+----
+
+. Run the following to test:
+
```
swopebe, line 51: @ianzhang366 what are they looking for when they run this? 

Run the following to test ____?

ianzhang366: after running the `annotate` command, the user can go to application UI page to verify if the application resource is deployed to the managed cluster. Or the user can log into their managed cluster to see if the application resource is created at the given namespace.

ianzhang366: @swopeb please help out with above ^^^

ianzhang366 (approved):

PR #1737, opened 2021-02-08
updating 2.2 troubleshoot for mco

dockerymick:
Revision of 2.2. after review of 2.1 @clyang82 can you help confirm that my changes are accurate for 2.2?

https://github.com/open-cluster-management/backlog/issues/8303

### DIFF
--- a/troubleshooting/trouble_obs_mco.adoc
+++ b/troubleshooting/trouble_obs_mco.adoc
@@ -1,19 +1,19 @@
-[#pvc-does-not-change-after-editing-statefulsetsize]
-= PVC does not change after editing statefulSetSize
+[#current-data-on-the-grafana-dashboard-disappears]
+= Current data on the Grafana dashboard disappears
 
-When you update the `statefulSetSize` for any cluster that has the observability service enabled, the persistent volume claim (PVC) is not updated.
+Current data from the Grafana dashboards in the {product-title-short} console disappears.
 
 
 [#symptom-pvc-does-not-change-after-editing-statefulsetsize]
 == Symptom: PVC does not change after editing statefulSetSize
 
-After you update the `statefulSetSize`, you might receive an alert that reads, `KubePersistentVolumeFillingUp`. The `data-observability-observatorium-thanos-receive-default-x` PVCs might be affected.
+After you update the `statefulSetSize`, you might receive an alert that reads, `KubePersistentVolumeFillingUp`. The `data-observability-observatorium-thanos-receive-default-*` PVCs might be affected.
 
 
 [#resolving-pvc-does-not-change-after-editing-statefulsetsize]
 == Resolving the problem: PVC does not change after editing statefulSetSize
 
-If you have this problem, update the `storage` parameter in each `data-observability-observatorium-thanos-receive-default-x` PVC. Complete the following steps:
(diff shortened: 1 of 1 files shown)


troubleshooting/trouble_obs_mco.adoc
```diff
@@ -1,38 +1,60 @@
...
 
-When you update the `statefulSetSize` for any cluster that has the observability service enabled, the persistent volume claim (PVC) is not updated.
+Current data from the Grafana dashboards in the {product-title-short} console disappears.
 
 
 [#symptom-pvc-does-not-change-after-editing-statefulsetsize]
 == Symptom: PVC does not change after editing statefulSetSize
 
-After you update the `statefulSetSize`, you might receive an alert that reads, `KubePersistentVolumeFillingUp`. The `data-observability-observatorium-thanos-receive-default-x` PVCs might be affected.
+After you update the `statefulSetSize`, you might receive an alert that reads, `KubePersistentVolumeFillingUp`. The `data-observability-observatorium-thanos-receive-default-*` PVCs might be affected.
```
clyang82: `you might receive an alert that reads,` I think we do not need to have `comma` in the end.

troubleshooting/trouble_obs_mco.adoc
```diff
@@ -1,38 +1,60 @@
...
 ----
 kubelet_volume_stats_available_bytes{namespace="open-cluster-management-observability"}/kubelet_volume_stats_capacity_bytes{namespace="open-cluster-management-observability"}
 ----
 +
-*Note:* If the value is 0, then the disk is full.
+*Note:* If the value is 0, then the disk is full. If the disk is full continue with the following tasks.
 
-. Expand the `data-observability-observatorium-thanos-receive-default-x` PVC to reflect the edited `statefulSetSize` in the MultiClusterObservability (mco) CR. Run the following command:
+. Expand the `data-observability-observatorium-thanos-receive-default-*` PVC to update the `storage` parameter to a greater value than the `statefulSetSize` value  from the MultiClusterObservability (mco) CR. Run the following command for each `data-observability-observatorium-thanos-receive-default-*` PVC:
 +
```
clyang82, line 33: Here has a format issue so that the whole line is bold.

troubleshooting/trouble_obs_mco.adoc
```diff
@@ -1,38 +1,60 @@
...
 ----
-kubectl get pvc data-observability-observatorium-thanos-receive-default-0 -o yaml
+kubectl get pvc data-observability-observatorium-thanos-receive-default-* -o yaml
```
clyang82: cannot run the command with `*`. suggest to keep using `0` here.

troubleshooting/trouble_obs_mco.adoc
dockerymick: Thanks for pointing this out, that's interesting bc the previous example of the PV name has the same syntax. I will continue to use `x` instead of the asterisk.

troubleshooting/trouble_obs_mco.adoc
dockerymick: Thanks for clarifying

dockerymick:
@clyang82 I have made updates based on your feedback. May you re-review when you have an opportunity?

clyang82 (approved):
lgtm

PR #1743, opened 2021-02-09
remove controllers from the supported policies page

dockerymick:
https://github.com/open-cluster-management/backlog/issues/7479


dockerymick:
Verified the the links work

PR #1748, opened 2021-02-10
Updates for observability 2.1

dockerymick:
https://github.com/open-cluster-management/backlog/issues/7173

### DIFF
--- a/observability/observability_install.adoc
+++ b/observability/observability_install.adoc
@@ -152,6 +152,6 @@ All the pods in `open-cluster-management-observability` namespace for Thanos, Gr
 
 To disable the observability service, uninstall the `observability` resource. See step 1 of link:../install/#removing-a-multiclusterhub-instance-by-using-commands[Removing a MultiClusterHub instance by using commands] for the procedure.
 
-To learn more about how to customizing the observability service, see xref:../observability/manage_observe.adoc#customizing-observability[Customizing observability].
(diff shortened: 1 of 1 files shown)


dockerymick:
@randymgeorge I went ahead and updated the infrastructure node link with the infrastructure machine set link. When you have an opportunity may you review the PR?

randymgeorge (commented):
Some comments for this PR: 

1. I would insert the statement about deploying on infrastructure nodes between the "create the Multiclusterobservability yaml" and the "apply the observability yaml" steps.   This add of the nodeselector will be to this multiclusterobservability yaml.  Thus, you would want to talk about this optional update before the apply. 
2. The new statement about Infrastructure nodes does not read right, i.e. Next, update the nodeSelector the infrastructure YAML  There is a the before and after nodeselector.  Also, I am not sure what infrastructure yaml is.  The change needs to be in the MultiClusterObservability yaml. 
3. The above statement references a knowledge article.  That in turn references the OpenShift documentation.  Should we just reference the OpenShift documentation directly? More of a question than an issue. 
4. The last change, customize -> customizing, makes the grammar in that sentence incorrect.

dockerymick:
@randymgeorge Thanks for your feedback. I agree, we should link to OCP doc. added the link to the OCP doc on machine sets that you provided yesterday. I also did a searched the file for customize and it wasn't present in the file. Was there a different file that you were looking at that requires this change?

randymgeorge (approved):

PR #1765, opened 2021-02-15
updating the links, improving explanation for header icons

dockerymick:
https://github.com/open-cluster-management/backlog/issues/8224#issuecomment-777870001

### DIFF
--- a/console/console.adoc
+++ b/console/console.adoc
@@ -1,9 +1,9 @@
 [#observability-in-the-console]
 = Observability in the console
 
-Learn more about console components that you can use to view, manage, or customize your console. 
+Learn more about console components that you can use to view, manage, or customize your console.  
 
-See the following image of the _Navigation_ from the {product-title} console, which is decsribed in more detail later in each section. See that the navigation represents major production function.
+See the following image of the _Navigation_ from the {product-title} console, which is described in more detail later in each section. See that the navigation represents major production function.
 
 image:../images/nav-2.0.png[Navigation]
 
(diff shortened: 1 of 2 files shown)


console/console.adoc
```diff
@@ -39,7 +39,7 @@ oc get routes -n <namespace-from-install>
 == Welcome page
 
 From the {product-title} Welcome page, you get more information about the product and you can access header features, such as _Search_ and _Create resource_.
-When you select _Create resource_, you can create a YAML file or JSON file for a deployment. Also from the _Header_, you can click the *Help* (?) icon to view the _About_ page and the documentation. From the _User_ menu, you can access the _Configure Client_ page. Additionally, you can access the xref:../console/vwt_search.adoc#searching-with-visual-web-terminal[Visual Web Terminal] from the header.
+When you select _Create resource_, you can create a YAML file or JSON file for a deployment. Also from the _Header_, you can click the *Help* (?) icon to view the _About_ page and the documentation. From the _User_ menu, you can access the _Configure Client_ page. Additionally, search capabilities are also available in the xref:../console/vwt_search.adoc#visual-web-terminal[Visual Web Terminal] from the header.
```
mark-nc: Just changing the link from the search subsection of the VWT page to the top of the page, though, does not fully address the last comments I added to issue 8224.   Line 41 talks about accessing features in the header like Search and Create Resource.  But line 42 does not reference the Search icon in the header and now just points to Visual Web Terminal with no explanation of context for why you would want to go there for search.  

Maybe I am missing the the point of this section but since it is about Observability in the console, shouldn't it be referencing the Search icon with a link to the Search page documentation and then saying that search is also one of the commands available in the Visual Web Terminal and then have the link to the VWT documentation.  If I am a new user and do not know about any of these things, I want to understand what these have to do with the overall topic of Observability.

dockerymick: Thanks for explaining. I made a couple of more updates.

mark-nc (approved):
lgtm

PR #1796, opened 2021-02-24
improving security book

dockerymick:
https://github.com/open-cluster-management/backlog/issues/9489


security/master.adoc
```diff
@@ -7,13 +7,12 @@ include::rbac.adoc[leveloffset=+2]
...
-include::cert_policy_ctrl.adoc[leveloffset=+4]
-include::iam_policy_ctrl.adoc[leveloffset=+4]
+include::grc_intro.adoc[leveloffset=+1]
+include::policy_example.adoc[leveloffset=+2]
+include::policy_controllers.adoc[leveloffset=+2]
+include::config_policy_ctrl.adoc[leveloffset=+3]
+include::cert_policy_ctrl.adoc[leveloffset=+3]
+include::iam_policy_ctrl.adoc[leveloffset=+3]
 include::policy_sample_intro.adoc[leveloffset=+3]
 include::memory_policy.adoc[leveloffset=+4]
```
swopebe: consider if 16, 17, and any others can be level 4?

security/master.adoc
```diff
@@ -28,6 +27,7 @@ include::gatekeeper_policy.adoc[leveloffset=+4]
 include::compliance_operator_policy.adoc[leveloffset=+4]
 include::e8_scan_policy.adoc[leveloffset=+4]
 include::manage_policy_overview.adoc[leveloffset=+3]
+include::create_policy_ctrl.adoc[leveloffset=+3]
 include::create_policy.adoc[leveloffset=+4]
```
swopebe: Consider is level 4 can be level 3 to get into the navigation of possible

security/master.adoc
dockerymick: They can be i believe, my plan was to only change a few at first

swopebe: ok, do your thing, then. We now know level 4 will not appear in the nav but you have your own discernment about what should be there.

swopebe (approved):

PR #1799, opened 2021-02-24
update spelling for the remaining files in install

dockerymick:
https://github.com/open-cluster-management/backlog/issues/9442


dockerymick:
Verified that the same change was not needed for 2.1

PR #1800, opened 2021-02-24
open-cluster-management/backlog#2667 Add submar

cadawson:


cadawson:
Hi @qiujian16 . I added the networking requirements for Submariner to the existing entries. Can you verify that they are correct? If there are additional ones that we need, let me know. Thank you!

install/requirements.adoc
```diff
@@ -58,4 +58,9 @@ Managed cluster:
 * Outbound connectivity to the Kubernetes API server of the hub cluster on port 6443
 * Outbound connectivity from the managed cluster to the channel source, which includes GitHub, Object Store, and Helm repository. This is only required when you are using application lifecycle to connect to these sources.
 
+Clusters that are using Submariner must have the following connections available:
+* Inbound and outbound on each of the managed clusters on port 4800/UDP
+* Inbound on each of the managed cluster on port 8080/TCP
+* Inbound and outbound on each of the managed clusters on ports 4500/UDP, 500/UDP, and any other ports that are used for IPSec traffic on the gateway nodes
```
qiujian16, line 65: or any other port. The port number can be configured with different port number for 4800, 4500, 500. But three ports needs to be opened

PR #1842, opened 2021-03-02
open-cluster-management/backlog#8221 review entire file, please

swopebe:


jnpacker:
@cadawson can you add more detail about this bullet, I'm not sure what it means
`You can install the {product-title-short} hub cluster on infrastructure nodes.`

berenss:
> @cadawson can you add more detail about this bullet, I'm not sure what it means
> `You can install the {product-title-short} hub cluster on infrastructure nodes.`

This means that we have done the necessary work (code, QE checkoff) to ensure that when we deploy RHACM operator on a cluster, it will spin up appropriately on nodes tagged as `infrastructure`.  This is an OpenShift designation that results in not using workers, and hence meaning that it does not consume an OpenShift subscription cost.  For RHACM this ensures that our product components do not run on workers and hence the RHACM hub does not cost a subscription. Our pricing model is that we only charge for the managed clusters running under the hub management.

berenss:
btw @cadawson @jnpacker this is something that we have been touting to the field (and IBM Cloud Pak teams) since our RHACM 2.0 GA. so the fact that we're making a statement about this now, might seem a little strange.  eg, it should probably not be a new feature. instead it's a fix to what was supposedly done/ready for months now.

cadawson:
So, it was something that we want to silently add because it should have already been there. We can remove it.

berenss:
that's right @cadawson. I would have expected to find a note in an z-stream fix errata I guess.
If there's clarity needed on the `support hub on infrastructure nodes` statement and what was actually done for 2.2, perhaps @nathanweatherly could chime in?

berenss:
somewhere in here we need to include the support for Azure Red Hat OpenShift (ARO) as an import and manage.

berenss:
Move the `Search Persistence` topic to be under `Observability` section. eg,
### observing-environments
> You can now define the storage settings for search persistence. Persistence is enabled by default with the searchCustomization custom resource. For more information, see Search in the console.

berenss:
Maybe get rid of the parent topic `Web console` once https://github.com/open-cluster-management/rhacm-docs/pull/1842#issuecomment-789159915 gets done.

berenss:
This is a really great set of features and reads very well for our customers and the field! thx @swopebe and team! 🚀

PR #1849, opened 2021-03-02
adding anchors for the arch doc

dockerymick:
https://github.com/open-cluster-management/backlog/issues/8220


dockerymick:
Verified that all the links working, so moving forward and merging

PR #1852, opened 2021-03-02
Adding security arch diagram for 2.2

dockerymick:
https://github.com/open-cluster-management/backlog/issues/9999


cadawson (changes requested):
Link looks good. A few comments and concerns:
Biggest: Are the roadmap items available in some format now, or are we preannouncing? If we are preannouncing, we don't want to do that. I know you are aware of that, so I trust that those are all available in some format now. If not, maybe we delay the art replacement until a refresh, unless we know it is already public knowledge. 
Other suggestions (don't hold up for these):
- Spell out Kubernetes, rather than k8s.
- Replace etc with Other, or list all, if possible. I know we might be trying to leave room for changes in the future. Etc just is not great to use there imo.
- We should probably use sentence-style capitalizations, where possible. I know we have used all caps with the largest sized ones, so that is ok. The lower-level ones would be better as sentence-style, where not a proper name. 
- Should we spell out SIEMs? I don't know what that stands for, but they might.
Looks good! Thanks, Mikela!

dockerymick:
@cadawson Thanks for the review Chris. It was confirmed that this diagram should not be used for external use via slack. Closing this PR and closing my issue

PR #1853, opened 2021-03-03
2.0.8 Errata for Thursday 2.0 refresh

swopebe:


dockerymick (approved):
Updates llok good Brandi, thank you !

juliana-hsu (commented):
Some comments on the errata:
9390 is associated with Bugzilla 1925281 please add the Bugzilla number - I'd suggest the following change:  "Fixed an issue with the hub subscription crashing when a Helm subscription is created for subscribing resources from a private Helm repo channel, where only spec.SecretRef is defined.
The private Helm repo channel secret has to be defined in the same channel namespace. (edited) 
10:13
9913 - I'd suggest the following change "Fixed an issue with the cert-manager-webhook pod failing to start with a permissions issue. The image was updated so there is no longer a dependency on specific user permissions."
10:17
The rest looks good.

juliana-hsu (approved):
Some comments on the errata:
9390 is associated with Bugzilla 1925281 please add the Bugzilla number - I'd suggest the following change:  "Fixed an issue with the hub subscription crashing when a Helm subscription is created for subscribing resources from a private Helm repo channel, where only spec.SecretRef is defined.
The private Helm repo channel secret has to be defined in the same channel namespace. (edited) 
10:13
9913 - I'd suggest the following change "Fixed an issue with the cert-manager-webhook pod failing to start with a permissions issue. The image was updated so there is no longer a dependency on specific user permissions."
10:17
The rest looks good.

PR #1854, opened 2021-03-03
open-cluster-management/backlog/issues/8329 

swopebe:
Searching for answers for the credentials doc and did a quick peer review. 

1. Italics not needed for general terms, and for code in text, as we discussed a few times, changed those so please double check.
2. `Beside` is replaced since non-directional for accessibility.
3. `File` after example of a file.

Maybe other small changes but:

I am not sure what line 50 is an example of so I could not edit. Maybe what it is an example of and why a list follows it can help... when I read it, I was confused.

Still did not find what I need for rotation topic so I will have to comment out what we don't have for 2.2 in this issue.


swopebe:
@cadawson please review? Thanks

swopebe:
@cadawson older PR I was asking for review on here?

cadawson (approved):
Thanks. Sorry it took a while. Was looking into the field and wanted to nail down the variable, an then lost track of it.

PR #1889, opened 2021-03-10
open-cluster-management/backlog#9648 clusterrole21

cadawson:


cadawson:
Hi @skeeey. Here is the 2.1 version of the clusterroles. Can you take a look? Thank you!

manage_cluster/define_clusterrole.adoc
```diff
@@ -0,0 +1,79 @@
...
+  - watch
+  - update
+  - delete
+  - deletecollection
+  - patch
+- apiGroups:
+  - register.open-cluster-management.io
+  resources:
+  - managedclusters/accept
+  resourceNames:
```
skeeey: We should remove `resouceNames` field in 2.1

cadawson:
Hi @skeeey . Thank you. I think I removed the wrong one the first time. I now removed the last one and added the one that I removed earlier. I believe you are saying that only the one instance after managedclusters/accept should be removed. Let me know if all of them should be removed. There are 2 other instances of resourceNames in the content. Thank you.

skeeey (approved):

skeeey:
@cadawson yeah, it is right this time

PR #1912, opened 2021-03-16
open-cluster-management/backlog#10489 Add err 209

cadawson:


cadawson:
Hi @juliana-hsu . The 2.0.9 errata should only be a container update, I think. I've included a line here that states that. Any other updates that should be addressed? Thanks!

cadawson:
Saw the confirmation from Juliana in a slack message that there were no other changes.

PR #1914, opened 2021-03-16
no issue - update variable guidance

cadawson:


cadawson:
The angle brackets in the command are to set it apart, and is mainly consistent with what exists in the docs.
The italic style for the variable that is outside of the content is consistent with the IBM style, and further sets it apart from the rest of the text.

swopebe:
it is set apart with code, italics in the command is more important, I think. But style manual (see my issue comment) has both, so if we are going to just choose one, why outside of the command?

_pages/ui_text_writing.md
```diff
@@ -54,7 +54,8 @@ Consider a small list of differences and similarities if you are coming from DIT
 |Element|Similarities|Differences|
 --------|------------|-------------|
 |Bold|Use for UI, clickable components|Use .md bold tag where you would have used `ui-control`. Do not use bold for emphasis, see Style Guide
-|Variable|Still build as italicized just like DITA|Use .md italics tag. No `varname` tag, so use <these> to identify a variable
+|Variable inside of a command or script sample|Still want to identify the variable.|No `varname` tag, so use angle brackets (<these>) to identify a variable. Surround the variable with the angle brackets (<>) to identify it within the existing monospace style. Ex: `oc get <namespace>`. 
+|Variable ouside of a command or script sample (in normal text).|Still want to identify the variable.|No `varname` tag, so use monospace and italic style to identify a variable. Surround the variable with monospace (`) and italic (_) styles to identify it as a variable. Ex: Replace `_namespace_` with the name of your project namespace. 
 |Menu cascade|Bold, same as DITA|No `menu cascade` tag, build as **Multi-cluster** > **Pods**|
```
swopebe: This is a new, extra step, discussing if necessary in Slack.

_pages/ui_text_writing.md
```diff
@@ -54,7 +54,8 @@ Consider a small list of differences and similarities if you are coming from DIT
 |Element|Similarities|Differences|
 --------|------------|-------------|
 |Bold|Use for UI, clickable components|Use .md bold tag where you would have used `ui-control`. Do not use bold for emphasis, see Style Guide
-|Variable|Still build as italicized just like DITA|Use .md italics tag. No `varname` tag, so use <these> to identify a variable
+|Variable inside of a command or script sample|Still want to identify the variable.|No `varname` tag, so use angle brackets (<these>) to identify a variable. Surround the variable with the angle brackets (<>) to identify it within the existing monospace style. Ex: `oc get <namespace>`.
```
swopebe: We could get rid of the DITA stuff since this was for onboarding onto markdown, much of this is older content that is not relevant to our team now.

swopebe: the whole file seems to have older IBM links and stuff in it. Do you mind updating this one? Also, not really a UI topic anymore, we can rename it to "writing guidance" or something? (I have renamed and reorged other files in this branch to more align with our current process, but only when it is time to update it.)

dockerymick: I agree with removing the DITA info, otherwise this update is good to me

_pages/ui_text_writing.md
dockerymick: Not sure on what has been decided yet. Using monospace only is the original process we have done. During the meeting, I understood that we should also add italics to a monospaced variable when it is outside of the code block. I think this is helpful to do to bring emphasis. The common denominator is to have the variable monospaced, so that might be the safest option

dockerymick (commented):
Made a couple of comments

_pages/ui_text_writing.md
swopebe: If it is working and we don't need the extra step every time we write one, I would rather stick with codeblock and <>, like we mostly have been doing. I just am struggling with the justification for a new step if we are not fully following the manual guidance (which is a bit dated) and it is not done in OCP doc.

swopebe: so do we agree <this> in the command (like as is, mostly) -- the command is the important part that they copy, and `this` in text (no need for italics)? This has been working thus far and I am thinking we don't need the extra step. Lots of people just copy the command without reading, so highlighting it in the command is worth the extra step.

dockerymick: Yea, that is fine. I agree to removing the extra step to italicize variables in the text, only monospace.

swopebe (changes requested):
requesting changes based on the discussion.

swopebe (approved):

dockerymick (approved):
made a very small update. This looks good! Thanks for working through this and updating the file Chris!

PR #1916, opened 2021-03-16
https://github.com/open-cluster-management/backlog/issues/9953--known…

swopebe:
… issue

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -327,10 +327,14 @@ Replace `_namespace_` with the namespace of your cluster resource.
 == Application management known issues
 
 [#deployable-resource-empty-spec] 
-=== Deployable resource with empty specification crashes the pod
-// 2.2:9953 app team plans to fix in 2.3
+=== Deployable resource with empty specification does not work
+// 2.2:9953 app team plans to fix in 2.3 to block user from creating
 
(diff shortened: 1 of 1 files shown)


release_notes/known_issues.adoc
```diff
@@ -326,6 +326,24 @@ Replace `_namespace_` with the namespace of your cluster resource.
 [#application-management-known-issues]
 == Application management known issues
 
+[#deployable-resource-empty-spec] 
+=== Deployable resource with empty specification crashes the pod
+// 2.2:9953 app team plans to fix in 2.3
+
+Applying a deployable resource with empty specification crashes the pod `multicluster-operators-application` container `multicluster-operators-deployable`.
```
swopebe: help me understand this?

1. I don't see any clue that the user gets from the sample that was given in the issue. (line 336)
2.  Why would user do this? 
3. Line 333 is a confusing sentence, can you help revise or make a comment that will help?

xiangjingli: This is like a negative case test from user. Or it could be created by user accidentally. 

The empty spec deployable resource can't do anything. But it crashes the container `multicluster-operators-deployable` in the pod `multicluster-operators-application`.  That is what we want to avoid. So you can see the workaround is to simply delete the  unnecessary deployable CR

We will fix this issue in 2.3. The known issue is for <=2.2

xiangjingli (approved):

Issue #169, opened 2020-03-24
Broken Link in summary.md - Configmap watcher service 


gghezzo:
## Documentation Report
While the summary.md is a temporary solution, the link to Configmap watcher service is not resolving (404) 
## Product release(s)
-[] [ACM 4.4] 

Note: For Common services and IBM Cloud Private version 3.2.1 and earlier defects, open up an issue in the [roadmap repo](https://github.ibm.com/IBMPrivateCloud/roadmap/issues/new?template=documentation-report.md).
https://github.com/open-cluster-management/rhacm-docs/blob/doc_stage/cert_manager/refresh_cert.md

## Type of documentation change

- [ ] New topic
- [ ] Update to an existing topic
- [x] Temporary issue with summary.md 

## Link to the topic(s) that require an update
https://github.com/open-cluster-management/rhacm-docs/blob/doc_stage/summary.md
404 - https://github.com/open-cluster-management/rhacm-docs/blob/doc_stage/cert_manager/refresh_cert.md
 

## Draft content or detailed description to get ID started

PR #1921, opened 2021-03-17
updating the subscription description in RBAC

dockerymick:
https://github.com/open-cluster-management/backlog/issues/10059

### DIFF
--- a/security/rbac.adoc
+++ b/security/rbac.adoc
@@ -37,9 +37,7 @@ Some product resources are cluster-wide and some are namespace-scoped. You must 
 | A user with cluster-wide binding to the `managed-cluster-x` role has view access to `managedcluster “X”` resource.
 
 | *open-cluster-management:subscription-admin*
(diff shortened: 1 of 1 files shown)


ckandag (commented):
@dockerymick : Not necessarily terrible to add it here.  but this table is for roles and their role, to add this note seems out of place.

Note the same exact information is already documented under the Application LifeCycle section at line #169.  

May be you could link to the App LC section from under the subscription role

dockerymick:
@ckandag thanks for your feedback! Yes, I brought the same concern to Derek. I like your idea of linking to the section for the reader to view more details

drkho (changes requested):
I still like to add a reference here...a link to the section below is ok.

Please note that the problem is an reference of `admin` in the name of **_open-cluster-management:subscription-admin_** here. It implies "admin" capability and can mislead users about the role's permissions to associated resources. Therefore, we need to explicitly state the requirement of access to other resources in other namespaces.

dockerymick:
@drkho to avoid being repetitive, would it be accurate to state the following on Line 169:

To perform application lifecycle tasks, users with the `admin` role must have access to the `application` namespace where the application is created, and to the `managed cluster` namespace.

drkho (approved):

PR #1937, opened 2021-03-24
open-cluster-management/backlog#9639 GCP no destroy

cadawson:


release_notes/known_issues.adoc
```diff
@@ -218,6 +218,9 @@ oc patch secret {CLUSTER-NAME}-aws-creds -n {CLUSTER-NAME} --type json -p='[{"op
 
 * Google Cloud Platform (GCP)
 
++
+You can identify this issue by a repeating log error message that reads, `Invalid JWT Signature` when you attempt to destroy the cluster. If your log contains this message, obtain a new Google Cloud Provider service account JSON key and enter the following command:
+
 +
 ----
 oc set data secret/{CLUSTER-NAME}-gcp-creds -n {CLUSTER-NAME} --from-file=osServiceAccount.json=$HOME/.gcp/osServiceAccount.json
```
cadawson: Questions here for Chris A: Our convention is typically <variable>, rather than {variable}. So, is the command listed the same as:
oc set data secret/<cluster-name>-gcp creds -n <cluster-name> --from-file=osServiceAccount.json=$HOME/.gcp/osServiceAccount.json  

Question 2: Do you have to update the key in osServiceAccount.json, or is it somehow transferred from the GCP service account? 

Question 3: Extra credit: Can I replace the {} in the commands on the other platforms with <> for the variables? I want to make sure some of the braces are not required in the code, for some reason. Thanks!

chrisahl (approved):

release_notes/known_issues.adoc
chrisahl: 1) greater than and less than instead of curly brackets are fine with me. 
2) The `$HOME/.gcp/osServiceAccount.json` is one place where the new GCP key may be stored when the user uses gcloud to log into GCP.  But they could also save it somewhere else.  They just need the command to point to the new valid json file
3)  yes, you can replace

PR #1958, opened 2021-04-02
open-cluster-management/backlog#11063--release note 

swopebe:


release_notes/errata.adoc
```diff
@@ -5,6 +5,48 @@ By default, Errata updates are automatically applied when released. See link:../
 
 *Important:* For reference, https://access.redhat.com/errata/#/[Errata] links and GitHub numbers might be added to the content and used internally. Links that require access might not be available for the user. 
 
+== Errata 2.2.2
+
+View a summarized list of {product-title} Errata 2.2.2 updates:
+
+* This Errata addresses multiple security issues and container image updates.
```
cadawson: I think errata should be lowercase here. If not, let's make sure we are consistent.

release_notes/errata.adoc
```diff
@@ -5,6 +5,48 @@ By default, Errata updates are automatically applied when released. See link:../
...
+
+* Added the ability to create Azure clusters for all possible regions. (GitHub 9700)
+
+* Fixed an issue with custom certificate authority on hub cluster. Submariner agent can now connect. (GitHub 9894)
+
+* Fixed an issue with `packageOverrides` that were incorrectly specified in subscription CR. Errors occurred on a pod on the hub cluster, or the `klusterlet-addon-appmgr` pod on the managed cluster. The log now ignores the override. (GitHub 9700)
+
+* Updated the Visual Web Terminal CLI to support {ocp-short} version 4.7. (GitHub 9640)
+
+* Updated import cluster commands with double quotes to handle un-escaped characters. Be sure to use base64 with `-d` option. (GitHub 10748)
```
cadawson: Un-escaped or unescaped?

cadawson: You must use base64 if you use the `-d` option.?

release_notes/errata.adoc
```diff
@@ -1,13 +1,55 @@
...
+
+* Fixed an issue with the cluster YAML editor. (https://bugzilla.redhat.com/show_bug.cgi?id=1941778[Bugzilla 1941778])
+
+* Fixed an issue with bare metal not selected for a provisioning cluster. (GitHub 9850)
```
cadawson: Not sure I follow this one. Is it: Fixed an issue where the bare metal option is not available when provisioning a cluster?

release_notes/errata.adoc
```diff
@@ -1,13 +1,55 @@
...
+
+* Added support for NodeJS version 14 from version 12 to limit vulnerabilities in the base image. (GitHub 9540)
+
+* Updated the `ServiceExport` API version in the base image. (GitHub 9840)
+
+* Fixed bare metal reuse issue. Now when a `clusterdeployment` is deleted, the bare metal asset that was originally reference can be reused for another `clusterdeployment`. (GitHub 9272)
```
cadawson: How about: Fixed an issue that prevented a bare metal asset from being used with a new `clusterdeployment` after a `clusterdeployment` that previously referenced it was deleted. ?

cadawson (approved):
Thank you! It looks great. Some considerations added, but not show-stoppers.

release_notes/errata.adoc
swopebe: I can rework it, we tend shy away from "being" as "to be verbs" have always been discouraged in concise writing and tech doc. I change these in peer reviews because of that.

release_notes/errata.adoc
swopebe: I am not sure about unescaped, as I have not written that in tech doc before.

I am asking about Base 64, as it seems it is a -d (data) option upon my google search when I wrote it. This will change if that is correct.

release_notes/errata.adoc
swopebe: @cadawson if you can doc test or provide a screenshot of that area, would be great. It is one of the ones I need to check if it was not available as or for, feel free to look at the issue:

https://github.com/open-cluster-management/backlog/issues/9850

(I was not yet done with this one, since your area, maybe you can check the product?)

release_notes/errata.adoc
swopebe: @cadawson: We are/were consistent here. (I think we should be careful when saying we are "_not consistent_" and look at the doc first for proof. In many situations lately, we are consistent, but someone is just looking to change something (which is great).)

All previous doc has `E`, we decided to change when NOT juxtaposed with the product, so lowercase going forward. That change is in. Thanks!

release_notes/errata.adoc
swopebe: Change is in

release_notes/errata.adoc
swopebe: No change, yes, need to look at the -d first.

release_notes/errata.adoc
swopebe: No change, yet, need to look at product first.

release_notes/errata.adoc
swopebe: @cadawson I researched this on my own, I am writing it this way because `D` is a command option for decode. (Just searched for it.)

swopebe: This is done

release_notes/errata.adoc
swopebe: Ok--I have decided to write this and let STSM check it:


* Fixed an issue where bare metal assets were not selected during cluster provisioning. (GitHub 9850)

swopebe: Done.

release_notes/errata.adoc
swopebe: @cadawson ? ^^^

release_notes/errata.adoc
cadawson: Sorry for the terminology miss. I meant consistent across us. I would think to use the lowercase, given a different opportunity. Didn't mean it was not consistent in the doc.

release_notes/errata.adoc
```diff
@@ -1,13 +1,55 @@
...
+
+* Fixed an issue with the cluster YAML editor. (https://bugzilla.redhat.com/show_bug.cgi?id=1941778[Bugzilla 1941778])
+
+* Fixed an issue where bare metal assets were not selected during cluster provisioning. (GitHub 9850)
+
+* Added support for NodeJS version 14 from version 12 to limit vulnerabilities in the base image. (GitHub 9540)
+
+* Updated the `ServiceExport` API version in the base image. (GitHub 9840)
+
+* Fixed a reuse issue with bare metal assets in `clusterdeployment`. The bare metal asset that was originally referenced can now be reused for another `clusterdeployment`. (GitHub 9272)
```
cadawson: Better now. One additional option:
A bare metal asset that was originally referenced by a `clusterdeployment` that is now removed can be reused for another `clusterdeployment`.

swopebe: see, I think that is confusing. Here I feel that a longer sentence makes it worse, so I would rather keep it concise. I can work on it more.

swopebe: Fixed a bare metal asset issue where assets that were originally referenced in `clusterdeployment` could not be reused for another `clusterdeployment`?

cadawson: That works. Thank you!

release_notes/errata.adoc
swopebe: Right. It is E (consistent) in all the doc, in all versions, change is from this release forward.

anthonyfister (approved):
LGTM

release_notes/errata.adoc
swopebe: done

jnpacker (approved):
Looks good.

jnpacker:
Some of these references listed feel like they could use links to existing/new knowledge center doc.  But I'm not sure how we would track that down...

swopebe:
@jnpacker I can probably add the known issues to this at the bottom..

PR #1959, opened 2021-04-05
Updating credentials doc

dockerymick:
https://github.com/open-cluster-management/backlog/issues/8330#issuecomment-804145355

### DIFF
--- a/security/credentials.adoc
+++ b/security/credentials.adoc
@@ -1,7 +1,9 @@
 [#credentials]
 = Credentials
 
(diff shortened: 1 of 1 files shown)


swopebe:
Thanks for pulling me in, we have a meeting today (I hope) to talk about placement and meaning of "credentials" compared to the product. This may move as a result, but want to hear from lead devs.

Looks good, left a comment for change.

dockerymick:
@swopebe thanks for the review :)

PR #1976, opened 2021-04-09
open-cluster-management/backlog#11273 -- rename

swopebe:
…name


swopebe:
@cadawson @dockerymick just the folders/links here. I would rather merge when 2.3 fresh build is not needed, that way I can take my time and build/check for errors without messing you up... just take a look when you can? 

thanks

PR #1983, opened 2021-04-12
open-cluster-management/backlog#11063 --update last two stop-ship

swopebe:


anthonyfister:
@swopebe 
[Issue 9850](https://github.com/open-cluster-management/backlog/issues/9850) is not entirely fixed from the user's perspective. Needs UI work. It is moving to 2.2.3.

[Issue 9437](https://github.com/open-cluster-management/backlog/issues/9437) is also not working yet and should be removed.  But our dependencies (OCP and Hive) do not support dual stack IPv6 so it is not possible for us to do so yet.  We may once the 2.2.z stream begins supporting OCP 4.8 (IF they support it by then).

anthonyfister (approved):
LGTM  Ship it!

PR #1984, opened 2021-04-12
removing cert files

dockerymick:
https://github.com/open-cluster-management/backlog/issues/10456
I thought that I submitted my review, so removing cert officially. The path to the files were already removed from the master.adoc file 

https://github.com/open-cluster-management/rhacm-docs/pull/1978


cadawson (approved):
Looks straightforward to me.

PR #2009, opened 2021-04-20
Update observability_enable.adoc

dockerymick:
update for 2.3
https://github.com/open-cluster-management/backlog/issues/10553

@songleo I am unsure of what the comment should read for enableDownSampling

### DIFF
--- a/observability/observability_enable.adoc
+++ b/observability/observability_enable.adoc
@@ -122,6 +122,7 @@ metadata:
   name: observability #Your customized name of MulticlusterObservability CR
 spec:
   availabilityConfig: High # Available values are High or Basic
+  enableDownSampling: true # The default value is true. This is not recommended as querying long-time ranges without non-downsampled data is not efficient and useful.
   imagePullPolicy: Always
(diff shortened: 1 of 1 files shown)


observability/observability_enable.adoc
```diff
@@ -122,6 +122,7 @@ metadata:
   name: observability #Your customized name of MulticlusterObservability CR
 spec:
   availabilityConfig: High # Available values are High or Basic
+  enableDownSampling: true # The default value is true. This is not recommended as querying long-time ranges without non-downsampled data is not efficient and useful.
```
dockerymick: @songleo the comment here reads the same as the comment in 2.2. Is a `true` value also not recommended?

songleo: @dockerymick sorry for confusing, it should be `This is recommended as querying long-time ranges downsampled data is efficient and useful.`

dockerymick: No worries! Thanks for clarifying. Making the change now and will continue to merge the PR

PR #2029, opened 2021-04-27
Discovery doc2

swopebe:


swopebe:
Please create a PR for changes to this draft: https://github.com/open-cluster-management/rhacm-docs/blob/2.3_stage/clusters/add_disc_cred.adoc

PR #2064, opened 2021-05-03
Open-cluster-management/backlog#11961 summaries/code WIP

swopebe:


anthonyfister (approved):
GitHub 11705 resolves BZ issue 1949225.  Should we add the BZ issue since they are publicly searchable?

swopebe:
> GitHub 11705 resolves BZ issue 1949225. Should we add the BZ issue since they are publicly searchable?

got it, it was not in the description as BZ so I didn't catch it.

PR #2080, opened 2021-05-07
add deleted metrics doc

songleo:
if a metric name ending in `-`,  we will delete this metric from the default allowlist.

https://github.com/open-cluster-management/backlog/issues/12106

Signed-off-by: Song Song Li <ssli@redhat.com>

### DIFF
--- a/observability/customize_observability.adoc
+++ b/observability/customize_observability.adoc
@@ -150,7 +150,7 @@ Complete the following steps to delete default metrics:
 oc get mco observability -o yaml
 ----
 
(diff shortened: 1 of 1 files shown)


observability/customize_observability.adoc
```diff
@@ -142,6 +142,53 @@ oc apply -n open-cluster-management-observability -f observability-metrics-custo
...
+----
+oc get mco observability -o yaml
+----
+
+. Create a YAML file for the `observability-metrics-custom-allowlist` ConfigMap by running the following command:
++
+----
+cat observability-metrics-custom-allowlist.yaml
+----
+
```
clyang82: That is existing issue. I do think we should remove this step ^^. 
We can combine the above step and the below step together.
for example: Just say
Create a new file `observability-metrics-custom-allowlist.yaml` with the following content ...

songleo: got it, let me update it. thanks.

dockerymick (approved):
Made a  couple of small changes. This looks good to me ! This PR can be merged

PR #2085, opened 2021-05-10
Updating the verb remove vs delete

dockerymick:
https://github.com/open-cluster-management/backlog/issues/12106
https://github.com/open-cluster-management/backlog/issues/12336

After rereading a few times, I thought it would be best to change delete to remove. @songleo are metrics deleted (no longer existing) when it is not in the allowlist or is data no longer being collected for that metric, but still existing as a metric?


songleo:
if we remove a metric or the metric is not in the allowlist, the metric does not exist. @dockerymick

songleo (approved):

PR #2086, opened 2021-05-10
open-cluster-management/backlog#11774 credential

swopebe:


clusters/discovery_credential.adoc
```diff
@@ -13,14 +13,16 @@ You need access to a https://cloud.redhat.com/[cloud.redhat.com] account.
 [#add-credential]
 == Add your credential
 
-If you have no credentials, see the following process. *Best practice:* Create only one credential in each namespace.
+*Best practice:* Create only one credential in each namespace.
```
zkayyali812, line 16: Perhaps this should state only 1 RHOCM credential in each namespace?

nathanweatherly: No, there's an edge case where 2 creds in one namespace is acceptable.

swopebe: ?

PR #2129, opened 2021-05-17
open-cluster-management/backlog#3597 VWT moved (intro)

swopebe:
@cadawson can you let me know if manage cluster labels needs to move or be removed from this topic?

@dockerymick can you please let me know if the order of Obs. and Search needs to change?


dockerymick:
@swopebe i agree with your change for VWT, Search is still in the header so I would keep it where it is

PR #2138, opened 2021-05-20
Adding doc for custom template functions

dockerymick:
https://github.com/open-cluster-management/backlog/issues/12034


release_notes/whats_new.adoc
```diff
@@ -50,4 +50,6 @@ You can now install your hub cluster in the Red Hat OpenShift Dedicated environm
 
 //updated the name of the section based on issue 11273
 
+* You can now add custom template functions for configuration policies. See link:../risk_compliance/custom_template.adoc#customizing-templates-for-config-policies[Customizing templates for configuration policies].
```
ckandag: You can now add/or include templates in configuration policies.

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,242 @@
+[#customizing-templates-for-config-policies]
+= Customizing tempplates for configuration policies
+
+Configuration policies now support inclusion of Golang templates in the object definitions. These templates are resolved at runtime on the target managed cluster using configuration policies related to that cluster. This gives you the ability to define configuration policies with dynamic content, and inform or enforce Kubernetes resources that are customized to the target cluster.
```
ckandag: 1.  Golang **text** templates.  

2.   These templates are resolved at runtime on the target managed cluster using configuration  related to that cluster.

"using configuration related to that cluster". or NOT "using configuration policies related to that cluster"

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,242 @@
...
+
+//we should reference golang doc along with provide tasks to create a custom template function
+//would this be a good link to use: https://golang.org/pkg/text/template/#hdr-Pipelines
+//or this link: https://golang.org/pkg/text/template/#hdr-Functions
```
ckandag: yes, agree, I intended to add but missed it. Please use this link 
https://golang.org/pkg/text/template/

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,242 @@
...
+The template syntax must be conformed to the Golang template language specification, and the resource definition generated from the resolved template must be a valid YAML. Any errors in template validation are recognized as policy violations. When you use a custom template function, the values are replaced at runtime.
+
+*Required access:* Not sure
```
ckandag: nothing additional  for templates support,   access requirements would be on the configuration policies  on specifically templates included in the policies

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,242 @@
...
+
+//[#prerequisites-templatized]
+// are there any prereqs?
+
```
ckandag, line 10: Perhaps, the previous sentence about syntax  could be the made the Pre-req  or NOT  - i,e "The template syntax must be conformed ...."

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,242 @@
...
+[#template-functions]
+== Template functions
+
+//what are generic template functions? Would it be accurate to say default template functions>
+Custom template functions, both resource-specific and generic template functions, are available for referencing Kubernetes resources on the cluster. 
+
```
ckandag: no saying default template function is not correct.

resource specific include name of the resource in the function name like fromSecret , fromConfigmap etc 
generic is not tied a specific resource and can work with any kube resource.

You can remove this   part  if its confusing " , both resource-specific and generic template functions,"

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,242 @@
...
+Customize the template for your configuration policies.
```
ckandag: this statement is not correct, pls remove it

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,242 @@
...
+// are there any prereqs?
+
+[#template-functions]
+== Template functions
+
+//what are generic template functions? Would it be accurate to say default template functions>
+Custom template functions, both resource-specific and generic template functions, are available for referencing Kubernetes resources on the cluster. 
+
+Customize the template for your configuration policies. 
+//should it be "`base64` encode" OR "`base64encode`"; throughout the doc i've seen "`base64` encode"
```
ckandag: here it is the function-name so needs to be as is `base64encode`

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,242 @@
...
+
+[#template-functions]
+== Template functions
+
+//what are generic template functions? Would it be accurate to say default template functions>
+Custom template functions, both resource-specific and generic template functions, are available for referencing Kubernetes resources on the cluster. 
+
+Customize the template for your configuration policies. 
+//should it be "`base64` encode" OR "`base64encode`"; throughout the doc i've seen "`base64` encode" 
+The resource-specific functions offered for this release make the references to the resources more accessible. The generic function, `lookup`, require you to input the reference to the resource. In addition to these functions, utility functions like `base64encode`, `base64decode`, `indent`, etc. are also available.
```
ckandag: This stmt doesnot capture the meaning accurately 

Replace the following with the below 
"The resource-specific functions offered for this release make the references to the resources more accessible. The generic function, `lookup`, require you to input the reference to the resource."

Replace to : The resource specific functions are for convenience and make accessing content of resources easy, while the generic lookup function is for an advanced user  who is familiar with the yaml structure  of the resource being looked up , location and type of the content  being referenced .


Here the idea is you can use the generic "lookup" function to achieve the same result as "fromSecret" function, but  with fromSecret  user doesnt  to worry about knowing the yaml structure of the Secret resource etc , it is done internally , hence making convenient to the user.

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,242 @@
...
+Continue reading to view descriptions and examples for some of the template functions that are available.
+
+[#fromsecret-func]
+=== _fromSecret_ function
+
+The `fromSecret` function returns the value of the given data key in the secret. View the following syntax for the function:
+
+----
+func fromSecret (ns string, secretName string, datakey string) (dataValue string, err error)
+----
```
ckandag, line 25: could you pls list what each param in the  function syntax is 

 input the namespace, name and the datakey of a k8s Secret resource

Please do this for the other functions as well

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,242 @@
...
+//instead of template can we call PASSWORD a parameter?
```
ckandag: PASSWORD is a parameter but here  value for the  Password  data key is template

I think its fine the way it is now

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,242 @@
...
+        product: '{{ fromClusterClaim "product.open-cluster-management.io" }}'
+        version: '{{ fromClusterClaim "version.openshift.io" }}'
+  remediationAction: enforce
+  severity: low
+----
+
+[#lookup-func]
+=== _lookup_ function
+//can you please clarify what "Unstructured struct" is? Is Unstrunctured the name of a resource? Is struct short for
+//structure?
```
ckandag: could you replace Unstructured struct with "JSON compatible map"

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,242 @@
...
+=== _base64enc_ function
+
+The `base64enc` function returns a `base64` encoded value of the input `data string`. View the following syntax for the function:
+
+----
+func base64enc ( data string) (enc-data string)
+----
+
+View the following example of the configuration policy that uses the `base64enc` function:
+//should we show more of the policy example? Is there more content that we can add that benefit the user?
```
ckandag: given we have full policy examples , it seemed unncessary the bury a small template in a large file

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,242 @@
+[#customizing-templates-for-config-policies]
+= Customizing tempplates for configuration policies
```
ckandag: This title doesnt sound correct.

I would vote for something like 

Templates in Configuration policies 

or Support for Templates in Configuration Policies

ckandag (changes requested):

risk_compliance/custom_template.adoc
dockerymick: @ckandag would the user need to be cluster admin to add the templates?

risk_compliance/custom_template.adoc
dockerymick: I think it would be best to show part of the policy so that the user sees that it is a part of a ConfigMap

risk_compliance/custom_template.adoc
dockerymick: Thanks for your explanation Chaitanya and clarification!

risk_compliance/custom_template.adoc
dockerymick: I understand why it can seem unnecessary. Maybe we can just add the YAML section that it would be a part of, Maybe at the `objectDefinition`? Let me know your thoughts. I prefer to add at least a snippet of the YAML file for consistency throughout the doc. Alternatively, we can remove some parts out of the other YAML examples

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+*Required access:* Not sure
+
+[#prerequisites-templatized]
+== Prerequisites
+
+* The template syntax must be conformed to the Golang template language specification, and the resource definition generated from the resolved template must be a valid YAML. See the Golang documentation about https://golang.org/pkg/text/template/[Package templates] for more information. Any errors in template validation are recognized as policy violations. When you use a custom template function, the values are replaced at runtime.
+
+[#template-functions]
+== Text template functions
```
ckandag: instead of Text template , could we say Custom Template functions.

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+Text template functions are available for referencing Kubernetes resources on the cluster. The resource-specific functions are for covenience and make content of the resources more accessible. If you use the generic function, `lookup`, it is best to be familiar with the YAML structure of the resource that is being looked up. In addition to these functions, utility functions like `base64encode`, `base64decode`, `indent`, etc. are also available.
```
ckandag: Same here , instead of Text template , could we say Custom Template functions , at the beginning of the stmt

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+Continue reading to view descriptions and examples for some of the text template functions that are supported.
```
ckandag: please replace text template functions with either Custom template functions or just template functions.

While the golang pkg is called text/template.  I do not think the custom functions. need to be called text template functions.

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,233 @@
...
+[#fromsecret-func]
+=== _fromSecret_ function
+
+The `fromSecret` function returns the value of the given data key in the secret. View the following syntax for the function:
+
+----
+func fromSecret (ns string, secretName string, datakey string) (dataValue string, err error)
+----
+
+When you use this function, enter the namespace, name, and data key of a Kubernetes `Secret` resource. You receive a policy violation if the secret or the data key does not exist on the target cluster. View the following configuration policy that enforces a `Secret` resource on the target cluster. The value for the `PASSWORD` data key is a text template that references the secret on the target cluster:
```
ckandag: i think just template is enough instead of text template

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,233 @@
...
+[#fromConfigmap-func]
+=== _fromConfigmap_ function
+
+The `fromConfigmap` function returns the value of the given data key in the ConfigMap. View the following syntax for the function:
+
+----
+func fromConfigMap (ns string, configmapName string, datakey string) (dataValue string, err Error)
+----
+
+When you use this function, enter the namespace, name, and data key of a Kubernetes `ConfigMap` resource. You receive a policy violation if the ConfigMap or the data key does not exist on the target cluster. View the following configuration policy that enforces a Kubernetes `ConfigMap` resource on the target managed cluster. The value for the `log-file` data key is a template that retrieves the value of the `log-file` from the ConfigMap, `logs-config` from the `default` namespace, and the `log-level` from the `default` namespace. The default value for `log-level` is `ERROR`:
```
ckandag: and the `log-level` from the `default` namespace. The default value for `log-level` is `ERROR`:

this part of the sentence doesn't sound right. Perhaps, change it to something like below

and value  of  `log-level` is set to  the data key `log-level`,  If the data key doesnt exists in the ConfigMap or if the value is empty then `log-level` defaults to the value `ERROR`

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,233 @@
...
+          {{with (fromConfigMap "default" "logs-config" "log-level") }} {{ . }} {{else}} ERROR {{end}}
+  remediationAction: enforce
+  severity: low
+----
+
+
+[#fromclusterclaim-func]
+=== _fromClusterClaim_ function
+
+The `fromClusterClaim` function returns the value of the `SpecValue` in the `ClusterClaim` resource. View the following syntax for the function:
```
ckandag: its `Spec.Value` not `SpecValue`

note the "."  between Spec and Value

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+[#lookup-func]
+=== _lookup_ function
+
+The `lookup` function returns the configuration resource as a JSON compatible map. View the following syntax for the function:
+
+----
+func lookup (apiversion string, kind string, namespace string, name string) (value string, err Error)
+----
+//will the user receive a policy violation if a certain data key or resource is not existing?
```
ckandag: no, user will not receive specific violation if data key or resource doesnt exists. That is why we made a note in the front that the lookup function is for a advanced or a power use who is familiar with the resource and yaml

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,233 @@
...
+        product: '{{ fromClusterClaim "product.open-cluster-management.io" }}'
+        version: '{{ fromClusterClaim "version.openshift.io" }}'
+  remediationAction: enforce
+  severity: low
+----
+
+[#lookup-func]
+=== _lookup_ function
+
+The `lookup` function returns the configuration resource as a JSON compatible map. View the following syntax for the function:
```
ckandag: its not configuration resource. pls replace with  kubernetes resource

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+----
+func lookup (apiversion string, kind string, namespace string, name string) (value string, err Error)
+----
+//will the user receive a policy violation if a certain data key or resource is not existing?
+When you use the function, enter the API version, kind, namespace, and name of the Kubernetes `ConfigMap` resource. View the following example of the configuration policy that enforces a Kubernetes `ConfigMap` resource on the target managed cluster. The value for the `metrics-url` data key is a template that retrieves the `v1/Service` Kubernetes resource `metrics` from the `default` namespace, and is set to the value of the `Spec.ClusterIP` in the queried resource:
```
ckandag: When you use the function, enter the API version, kind, namespace, and name of the Kubernetes **`ConfigMap`** resource.

its not just a ConfigMap resource , can be any resource. ( just for the example i used configmap but can be any kube resource)

So replace the above with 
When you use the function, enter the API version, kind, namespace, and name of the Kubernetes  resource.

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,233 @@
...
+----
+
+When you use the function, enter a data string with the specific number of spaces. View the following example of the configuration policy that uses the `indent` function:
+
+----
+Ca-cert:  |
+          {{ ( index ( lookup "v1" "Secret" "default" "mycert-tls"  ).data  "ca.pem"  ) |  base64dec | indent 4  }}
+----
+
+//are there other task like creating a text template possible for the user??
```
ckandag: no, we did not test or support Users creating there own templates

risk_compliance/custom_template.adoc
ckandag: ok

ckandag (changes requested):

dockerymick:
@ckandag Thank you so much for your feedback and for understanding the need for consistency. I believe the last thing for this PR is to add more YAML contents for `base64enc`, `base64dec`, and `indent`

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,279 @@
+[#support-custom-templates-in-config-policies]
+= Support for custom templates in configuration policies
```
ckandag: Please make this "Support for templates in configuration policies".  we have some custom templates but we support regular templates too.. So support for custom templates sounds incorrect.

release_notes/whats_new.adoc
```diff
@@ -58,6 +58,8 @@ You can now install your hub cluster in the Red Hat OpenShift Dedicated environm
 
 //updated the name of the section based on issue 11273
 
+* You can now add or include custom template functions in configuration policies. See link:../risk_compliance/custom_template.adoc#support-custom-templates-in-config-policies[Support for custom templates in configuration policies] for more information.
```
ckandag: Please make this "You can now add or include templates in configuration policies.". 
Just  "templates" NOT  "custom template functions" 
we have some custom templates functions but we support regular templates too.. So saying  custom templates sounds very incorrect.

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,279 @@
...
+
+Configuration policies now support inclusion of Golang text templates in the object definitions. These templates are resolved at runtime on the target managed cluster using configurations related to that cluster. This gives you the ability to define configuration policies with dynamic content, and inform or enforce Kubernetes resources that are customized to the target cluster.
+
+*Required access:* Cluster administrator
```
ckandag: this is incorrect, I dont believe i mentioned its cluster admin any where.    This infact is not even needed.   

whatever access is needed to create  configuration policies can do this too , there is no additional Access level just for templates.

risk_compliance/custom_template.adoc
```diff
@@ -0,0 +1,279 @@
...
+[#prerequisites-templatized]
+== Prerequisites
+
+* The template syntax must be conformed to the Golang template language specification, and the resource definition generated from the resolved template must be a valid YAML. See the Golang documentation about https://golang.org/pkg/text/template/[Package templates] for more information. Any errors in template validation are recognized as policy violations. When you use a custom template function, the values are replaced at runtime.
+
+[#custom-template-functions]
+== Custom template functions
+
+Custom template functions are available for referencing Kubernetes resources on the cluster. The resource-specific functions are for covenience and make content of the resources more accessible. If you use the generic function, `lookup`, it is best to be familiar with the YAML structure of the resource that is being looked up. In addition to these functions, utility functions like `base64encode`, `base64decode`, `indent`, etc. are also available.
+
```
ckandag, line 15: The above too is choppy and doesnt flow well. Like I had mentioned in my original draft documentation. 
I think you need say there are some like some resource-specific functions like   - ,- -  which are convenience. and a generic function , which is for advanced user who is already familiar with the yaml structure. 

Please read through and restructure this

ckandag (changes requested):

risk_compliance/custom_template.adoc
dockerymick: Thanks for clarifying. I confirmed with Yu Cao, that a cluster admin or admin is required to create config policies.

risk_compliance/custom_template.adoc
dockerymick: How about this: Custom template functions, such as resource-specific and generic template functions, are available for referencing Kubernetes resources on the cluster. The resource-specific functions are used for convenience and make content of the resources more accessible. If you use the generic function, `lookup`, which is more advance, it is best to be familiar with the YAML structure of the resource that is being looked up.

ckandag: @dockerymick : This looks better , one modification 
Custom template functions, such as resource-specific and generic   lookup functions, 

---while the generic function is named "lookup" the purpose of all of the functions is to lookup k8s resources. So I think saying "resource-specific and generic   lookup functions, "  makes more sense

release_notes/whats_new.adoc
ckandag: @dockerymick : Please also make this change

dockerymick: Done

PR #2141, opened 2021-05-22
add doc for BYO certificate support for alertmanager route.

morvencao:
ref:
- https://github.com/open-cluster-management/backlog/issues/12478
- https://github.com/open-cluster-management/backlog/issues/12728

Signed-off-by: morvencao <lcao@redhat.com>

### DIFF
--- a/risk_compliance/certificates.adoc
+++ b/risk_compliance/certificates.adoc
@@ -166,32 +166,60 @@ Observability certificates on managed clusters are automatically refreshed upon 
 
 * https://docs.openshift.com/container-platform/4.7/security/certificates/service-serving-certificate.html[OpenShift Service Serving Certificates] manages certificates for processes that run on your hub cluster.
 
-You can replace alertmanager certificates by updating the alertmanager route if you do not want to use the OpenShift default ingress certificate. The steps to use your own certificates for alertmanager route are similar to xref:../risk_compliance/cert_mgmt_ingress.adoc#replacing-the-management-ingress-certificates[Replacing the management ingress certificates] except the following three things:
-
-* Be sure to change the common name (CN) on the certificate to `alertmanager`.
-* You also need to change the SAN labeled, `DNS.1` in the `csr.cnf` configuration file with the correct hostname for your alertmanager route.
-* Create the following two secrets in the `open-cluster-management-observability` namespace, the observability service operator (`multicluster-observability-operator`) will take care of everything:
-+
-----
-oc -n open-cluster-management-observability create secret tls alertmanager-byo-ca --cert ./ca.crt --key ./ca.key
-oc -n open-cluster-management-observability create secret tls alertmanager-byo-cert --cert ./ingress.crt --key ./ingress.key
-----
-
-If you want to restore the default self-signed certificate for alertmanager route, please follow the similar steps as specified in ref:../risk_compliance/cert_mgmt_ingress.adoc#restore-the-default-self-signed-certificate-for-management-ingress[Restore the default self-signed certificate for management ingress] to delete two secrets in the `open-cluster-management-observability` namespace:
-
-+
-----
-oc -n open-cluster-management-observability delete secret alertmanager-byo-ca
-oc -n open-cluster-management-observability delete secret alertmanager-byo-cert
-----
-
-then the observability service operator will take care of everything.
-
 * {product-title-short} supports the management of certificates for managed clusters.
 
 * The `metrics-collector` restarts to mount the renewed certificates.
 +
 *Note:* `metrics-collector` can push metrics to the hub cluster before and after certificates are removed. For more information about refreshing certificates across your clusters, see the <<refresh-internal-certificates,Refresh internal certificates>> section. Be sure to specify the appropriate namespace when you refresh a certificate.
+
+[#replacing-cert-alertmanager]
+==== Replacing certificates for alertmanager route
+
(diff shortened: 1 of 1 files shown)


risk_compliance/certificates.adoc
```diff
@@ -166,6 +166,19 @@ Observability certificates on managed clusters are automatically refreshed upon
 
 * https://docs.openshift.com/container-platform/4.7/security/certificates/service-serving-certificate.html[OpenShift Service Serving Certificates] manages certificates for processes that run on your hub cluster.
 
+You can replace alertmanager certificates by updating the alertmanager route if you do not want to use the OpenShift default ingress certificate. The steps to use your own certificates for alertmanager route are similar to xref:../risk_compliance/cert_mgmt_ingress.adoc#replacing-the-management-ingress-certificates[Replacing the management ingress certificates] except the following three things:
```
dockerymick: @morvencao i like that you added a reference to the Replace management ingress certificates doc. I recommend that we keep this reference, but add the steps that they user would complete. It would be a cost to the user to jump to another page bc then we would have to think about how they get directed back to this certs page

risk_compliance/certificates.adoc
```diff
@@ -166,6 +166,19 @@ Observability certificates on managed clusters are automatically refreshed upon
...
+* Be sure to change the common name (CN) on the certificate to `alertmanager`.
+* You also need to change the SAN labeled, `DNS.1` in the `csr.cnf` configuration file with the correct hostname for your alertmanager route.
+* Create the following two secrets in the `open-cluster-management-observability` namespace, the observability service operator (`multicluster-observability-operator`) will take care of everything:
++
+----
+oc -n open-cluster-management-observability create secret tls alertmanager-byo-ca --cert ./ca.crt --key ./ca.key
+oc -n open-cluster-management-observability create secret tls alertmanager-byo-cert --cert ./ingress.crt --key ./ingress.key
+----
+
+*Note:* If you want to restore the default self-signed certificate for alertmanager route, you only need to delete two secrets in the `open-cluster-management-observability` namespace, the observability service operator will take care of everything.
```
dockerymick: Great note, I think we should also add a reference link to the section, Restore the default self-signed certificate for management ingress

dockerymick (changes requested):

risk_compliance/certificates.adoc
morvencao: Do we really want to do this?
I'm asking because there lots of duplicate stuff there, especially the `Example configuration file for generating a certificate`

risk_compliance/certificates.adoc
morvencao: thanks, added the reference link.

risk_compliance/certificates.adoc
morvencao: Actually I added these steps in my local before I created the PR, then I found the similar steps for management-ingress, that's why I reverted my work.

dockerymick: Understandable. I think we can assume that the user would already have the configuration file that is needed to generate a certificate. With that in mind, I think that it would be helpful to the user if we had a step to open or access the configuration. I made some edits in the file. 

Is there a command that that the user can run to change the CN and SAN?

risk_compliance/certificates.adoc
```diff
@@ -172,6 +172,55 @@ Observability certificates on managed clusters are automatically refreshed upon
...
+
+You can replace alertmanager certificates by updating the alertmanager route, if you do not want to use the OpenShift default ingress certificate. Complete the following steps:
+
+. Examine the observability certificate with the following command:
++
+----
+openssl x509  -noout -text -in ./observaility.crt
+----
+//is there a command for this step?
+. Change the common name (`CN`) on the certificate to `alertmanager`.
```
dockerymick, line 186: Is there a command for this? Or can the user edit the configuration after running the command on line 183?

morvencao: unfortunately no such command, we have to update the CN in the csr.conf and then generate the certificate.

risk_compliance/certificates.adoc
morvencao: unfortunately no such command, 
since we guide the end user to use csr.conf, so they have to change the CN and SAN in the csr.conf and then generate the certificate.

dockerymick:
@morvencao thanks for your review and feedback. I have removed the comments. When you have a chance, may you rereview the PR?

risk_compliance/certificates.adoc
```diff
@@ -172,6 +172,31 @@ Observability certificates on managed clusters are automatically refreshed upon
...
++
+----
+openssl x509  -noout -text -in ./observaility.crt
+----
+
+. Change the common name (`CN`) on the certificate to `alertmanager`.
+
+. Change the SAN in the `csr.cnf` configuration file with the hostname for your alertmanager route.
+
+. Create the two following secrets in the `open-cluster-management-observability` namespace. `multicluster-observability-operator` creates the secret. Run the following command:
```
morvencao: I found the sentence `'multicluster-observability-operator' creates the secret.` is confusing, I suggest to remove it.

morvencao:
Others looks good to me.

dockerymick:
I addressed @morvencao suggestions and since all looks good, I am going to merge this PR

morvencao:
thanks @dockerymick

PR #2156, opened 2021-05-27
Swope errata211

swopebe:
@juliana-hsu at first glance, this is all I saw.... i am waiting on feedback from the Search line (commented in the issue)

### DIFF
--- a/release_notes/fixpack.adoc
+++ b/release_notes/fixpack.adoc
@@ -7,13 +7,15 @@ By default, Errata updates are automatically applied. See link:../install/upgrad
 
 == Errata 2.0.11
 
-*Important:* You must run {ocp} version 4.5, or later, to upgrade to {product-title-short} version 2.0.11. If you cannot upgrade your {ocp} version 4.4 to a later version, you can continue to use {product-title-short} version 2.0.10.
-
 View a summarized list of {product-title} Errata 2.0.11 updates:
 
-Updated Search as a result of Kubernetes version deprecation, which caused errors in the console editor. (GitHub 11904)
+*Important:* You must run {ocp} version 4.5, or later, to upgrade to {product-title-short} version 2.0.11. If you cannot upgrade your {ocp} version 4.4 to a later version, you can continue to use {product-title-short} version 2.0.10.
 
-Fixed an issue that caused unexpected policies to display in the _Policy listing_ page. (GitHub 11853)
+. Updated Search as a result of Kubernetes version deprecation, which caused errors in the console editor. (GitHub 11904)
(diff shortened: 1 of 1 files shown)


dockerymick (approved):
Looks good Brandi! Thanks !

juliana-hsu (commented):
As you already mentioned we need Jorge / Sherin to comment on the wording for 11904.  Did you document that OCP 4.5 was no longer supported with ACM 2.0.11?  It would be similar wording that went into the 2.1 release.  The user has to upgrade OCP to 4.6 in order to upgrade to ACM 2.0.11.

juliana-hsu (commented):
Here's the OCP 4.5 deprecation info:  Starting with ACM 2.0.11, please update the release notes to let customers know that OCP 4.5 is no longer supported. Customers will still be able to use ACM 2.0.10 with OCP 4.5 without interruption, but they will need to upgrade to OCP 4.6 in order to upgrade ACM to 2.0.11.

swopebe:
Yes, @juliana-hsu you should see the OCP statement in there. I documented the ones labeled errata; for the known issue, we it was decided that this close to the Errata, we don't need a known issue bc the fix is coming so soon.

juliana-hsu:
The OCP upgrade statement in the errata isn't correct.  Please have a look at my comment above.

juliana-hsu:
Also, I'll wait on Jorge / Sherin's comments on 11904 before approving.

cadawson:
The versions need to be updated in the support note. I think it should be:

*Important:* You must run {ocp} version 4.6, to upgrade to {product-title-short} version 2.0.11. If you cannot upgrade your {ocp} version to version 4.6, you can continue to use {product-title-short} version 2.0.10.

Version 4.5 is no longer supported. In addition, 4.7 is not supported for 2.0.x. For those reasons, we have to specify 4.6. Other than that, it looks great to me. Thank you!

swopebe:
@cadawson I have made that change already, pinged it to Juliana. Additionally, the translation may not need to be mentioned, but with the label I added it, so also pinged about that. Will merge and refresh when I hear about those.

swopebe:
Restructured the template a bit with @cadawson: 

*Important:* {ocp} 4.5 is not supported with 2.0.11. You must run {ocp} version 4.6 to upgrade to {product-title-short} version 2.0.11. If you cannot upgrade your {ocp} version to 4.6, you can continue to use {product-title-short} version 2.0.10.

For future use, we will build this note as such.

juliana-hsu:
Changes look good.  Thanks!

PR #2189, opened 2021-06-04
starting on draft for ansible integration for governance

dockerymick:
https://github.com/open-cluster-management/backlog/issues/12646


risk_compliance/ansible_grc.adoc
```diff
@@ -0,0 +1,33 @@
...
+Complete the following steps to configure a policy violation automation:
+
+. Log in to your {product-title-short} hub cluster.
+. From the navigation menu, select *Governance*. 
+. Configure an automation for a specific policy by clicking *Configure* in the _Automation_ column. 
+. The _Create policy violation automation_ panel appears.
+. From the _Credential_ section, click the drop-down menu to select an Ansible credential. If you need to add a credential, see link:../credentials/credential_intro.adoc#managing-credentials-overview[Managing credentials overview].
+. Click the drop-down list to select a job template.
+. In the _Extra variables_ section, add the parameter values from the `extra_vars` section of the `PolicyAutomation`.
+. Select the frequency of the automation. You can select _Manual run_, _Run once mode_, or _Disable automation_.
```
gparvin, line 29: Are we going to have more documentation somewhere that describes these frequency fields?  I think we need some really good descriptions for these, especially for `Manual`, which isn't very intuitive to me.

gparvin (commented):
I think we need to have some descriptions that detail exactly when automation will run when the `Run once` mode is selected.  That should be the primary mode used since disabled and manual are not really automating anything.

I might need to test this some more to make sure, but I think Run Once only runs when the policy transitions from Compliant to NonCompliant.  It will not run when additional managed clusters add to the NonCompliance of an already NonCompliant policy.  It will also only run once, it will not run again if the policy stays NonCompliant.

risk_compliance/ansible_grc.adoc
dockerymick: Thanks for your feedback Gus! I think describing the mode values is a good idea. I can add it after the step.

risk_compliance/manage_policy_overview.adoc
```diff
@@ -5,7 +5,7 @@ Use the _Governance_ dashboard to create, view, and manage your security policie
 
 From the _Governance_ page, you can customize your Summary view by filtering the violations by categories or standards, collapse the summary to see less information, and you can search for policies. You can also filter the violation table view by policies or cluster violations.
 
-The table of policies list the following details of a policy: _Policy name_, _Namespace_, _Remediation_, _Cluster violations_, _Standards_, _Categories_, _Controls_, and _Created_. You can edit, disable, inform or remove a policy by selecting the *Actions* icon.
+The table of policies list the following details of a policy: _Policy name_, _Namespace_, _Remediation_, _Cluster violations_, _Controls_, _Automation_ and _Created_. You can edit, enable or disable, inform, or remove a policy by selecting the *Actions* icon. You can view the categories and standards of a specific policy by selecting the drop-down arrow to expand the row.
```
dockerymick, line 8: @gparvin Maybe we can add the value descriptions here or from the Configuring Ansible page. Thoughts?

dockerymick: I added description starting at line 10

gparvin (approved):
Thanks for adding these updates.  After testing ansible some I think everything here looks accurate.  
We are thinking about some other scenarios, but I don't have any documentation to provide you yet.  Certainly be aware of the following:
1. Today there's no "Delete" button to remove automation -- we're trying to get this in if we have time.
2. Once the automation is disabled, it will not switch back to Run Once mode if the policy is no longer in violation.  The user has to manually do that.  
3. We are going to test the behavior with gitops -- this will likely change how item 2 above behaves (we think)

PR #2193, opened 2021-06-07
Fix design_grafana typo

haoqing0110:
The dashboard JSON data should be copied to `data` not `metadata`


dockerymick (approved):

dockerymick:
@haoqing0110 Changes look good, great catch. I am not sure if there is an issue associated to this PR. If there is an issue for this change please add a reference link for the issue. Otherwise, great catch! Thank you for making the change and pulling me in for a review. Good to merge this PR

PR #2207, opened 2021-06-08
update invalid object storage secret example

songleo:
update invalid object storage secret example.

https://github.com/open-cluster-management/backlog/issues/13021

Signed-off-by: Song Song Li <ssli@redhat.com>


dockerymick (approved):
@songleo your changes look great here as well ! Thank you for making the updates.

PR #2229, opened 2021-06-10
Updating custom metrics doc

dockerymick:
https://github.com/open-cluster-management/backlog/issues/12852

### DIFF
--- a/observability/customize_observability.adoc
+++ b/observability/customize_observability.adoc
@@ -130,7 +130,7 @@ data:
 ----
 +
 ** In the `names` section, add the name of the custom metrics which is to be collected from the managed cluster.
-** In the `rules` section, enter a value for the `expr` parameter to define the query expression. The metrics are collected as the name that is defined in the `record` parameter from your managed cluster. The metric value returned are the results after you run the query expression. For more information, see the documentation for https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#recording-rules[recording rules].
+** In the `rules` section, enter a value for the `expr` parameter to define the query expression. The metrics are collected as the name that is defined in the `record` parameter from your managed cluster. The metric value returned are the results after you run the query expression.
 ** The `names` and `rules` sections are optional. You can use either one or both of the sections.
 
 . Create the `observability-metrics-custom-allowlist` ConfigMap in the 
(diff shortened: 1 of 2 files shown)


observability/customize_observability.adoc
```diff
@@ -130,7 +130,8 @@ data:
 ----
 +
 ** In the `names` section, add the name of the custom metrics which is to be collected from the managed cluster.
-** In the `rules` section, enter a value for the `expr` parameter to define the query expression. The metrics are collected as the name that is defined in the `record` parameter from your managed cluster. For more information, see the documentation for https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#recording-rules[recording rules]. 
+** In the `rules` section, enter a value for the `expr` parameter to define the query expression. The metrics are collected as the name that is defined in the `record` parameter from your managed cluster. The metric value returned are the results after you run the query expression. For more information, see the documentation for https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#recording-rules[recording rules].
```
marcolan018: 1. the "rules" is a list, and each entry in the list has "expr" and "record".
2. don't need the "for more information ..." part. We borrowed the term "recording" rule from prometheus, although the implementation is similar but they are different things indeed.

dockerymick: Thanks Marco, i am confused on what i need to update from your first suggestion. Should this statement:

In the `rules` section, enter a value for the `expr` parameter to define the query expression. The metrics are collected as the name that is defined in the `record` parameter from your managed cluster. The metric value returned are the results after you run the query expression.

be changed to this:

In the `rules` section, you can view the list of values for the `expr` and `record` parameter, which define the query expression. The metrics are collected as the name that is defined in the `record` parameter from your managed cluster. The metric value returned are the results after you run the query expression.

marcolan018: @dockerymick 
for the 1st item, I am worrying about whether the users will misunderstand that in rules section, they can only input one "expr"/"record" pair.

dockerymick: Gotchu! Ok, what do you think about this update:

In the `rules` section, enter only one value for the `expr` and `record` parameter pair to define the query expression. The metrics are collected as the name that is defined in the `record` parameter from your managed cluster. The metric value returned are the results after you run the query expression.

marcolan018: thanks, the update looks good

dockerymick:
Thanks @marcolan018 for your review and comments! Based on the review and agreed changes, we can merge this PR

PR #2232, opened 2021-06-10
Update observability certificates

dockerymick:
https://github.com/open-cluster-management/backlog/issues/8366

### DIFF
--- a/risk_compliance/certificates.adoc
+++ b/risk_compliance/certificates.adoc
@@ -27,13 +27,17 @@ Continue reading to learn more about certificate management:
 ** <<rotating-the-gatekeeper-webhook-certificate,Rotating the gatekeeper webhook certificate>>
 ** <<rotating-the-integrity-shield-webhook-certificate,Rotating the integrity shield webhook certificate (Technology preview)>>
 ** <<observability-certificates,Observability certificates>>
+*** <<observability-byo-certificates,Bring Your Own (BYO) observability certificate authority (CA) certificates>>
+*** <<openssl-commands-ca,OpenSSL commands to generate CA certificate>>
+*** <<create-secrets-byo-certs,Create the secrets associated with the BYO observability CA certificates>>
+*** <<replacing-cert-alertmanager,Replacing certificates for alertmanager route>>
 ** <<channel-certificates,Channel certificates>>
 ** <<managed-cluster-certificates,Managed cluster certificates>>
 
 [#list-managed-certificates]
 == List managed certificates
 
-You can view a list of managed certificates that use https://docs.openshift.com/container-platform/4.7/security/certificates/service-serving-certificate.html[OpenShift Service Serving Certificates] service internally.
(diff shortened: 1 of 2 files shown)


risk_compliance/certificates.adoc
```diff
@@ -160,17 +167,74 @@ oc delete po -n integrity-shield-operator-system integrity-shield-server-5fbdfbb
...
 
-Observability certificates on managed clusters are automatically refreshed upon expiration. View the following list to understand the effects when certificates are automatically renewed:
+View the following list to understand the effects when certificates are automatically renewed:
 
-* https://docs.openshift.com/container-platform/4.7/security/certificates/service-serving-certificate.html[OpenShift Service Serving Certificates] manages certificates for processes that run on your hub cluster.
+* Non-CA certificates are renewed automatically when the remaining valid time is no more than 73 days. After the certificate is renewed, the pods in the related deployments restart automatically to use the renewed certificates.
 
-* {product-title-short} supports the management of certificates for managed clusters.
+* CA certificates are renewed automatically wehn the remainging valid time is no more than one year. After the certificate is renewed, the old CA is not deleted bu co-exist with the renewed ones. Both old and renewed certificates are used by related deployments, and continue to work. The old CA certificates are deleted when they expire.
```
marcolan018: "bu" should be "but"

risk_compliance/certificates.adoc
```diff
@@ -106,6 +112,7 @@ View the following table for a summarized list of the pods that contain certific
 | work-webhook-serving-cert
 |===
 
+*Note*: Customization of the service certificates are managed by {ocp-short} and can be configured. For more information see https://docs.openshift.com/container-platform/4.7/security/certificates/replacing-default-ingress-certificate.html[Replacing the default ingress certificate].
```
dockerymick: @gparvin This is the note I was referring to. Earlier in the doc, we inform the user about the certificates being managed by the OpenShift Service Serving

gparvin: This note doesn't make sense here.  I think it should just be deleted or if we need additional clarification I think it needs to be prior to this note.  The complicated thing that makes this confusing is we have 4 "managers" of certificates:
1. The OpenShift ingress operator
2. The OpenShift Service Serving Certificates operator - these are internal certificates that are managed by OpenShift
3. These certificates -- which we are trying to call Internally managed certificates -- like number 2 above but we managed these ourselves
4. Others -- this is a category of certificates that doesn't fall into the list above.  Such as gatekeeper certs and the managed cluster certs

I agree that this can be pretty confusing, if you need to we can meet and maybe discuss ways to organize it better.

gparvin (commented):
Added a comment around a note that was unclear.

risk_compliance/certificates.adoc
dockerymick: @gparvin Thanks for your feedback! Yes, I would like to meet about this for sure!

dockerymick:
Going to merge this PR so that we have a clean doc to look at during the meeting between Gus and I.

PR #2246, opened 2021-06-16
open-cluster-management/backlog#3597--changes to KUI

swopebe:


mark-nc (commented):
These updates following our discussion look good.  Thanks!

PR #2295, opened 2021-06-23
open-cluster-management/backlog#13239 Swope errata1.8

swopebe:
…sion info

### DIFF
--- a/release_notes/errata.adoc
+++ b/release_notes/errata.adoc
@@ -5,11 +5,11 @@ By default, Errata updates are automatically applied when released. See link:../
 
 *Important:* For reference, link:https://access.redhat.com/errata/#/[Errata] links and GitHub numbers might be added to the content and used internally. Links that require access might not be available for the user.
 
-* *Important:* You must run {ocp} version 4.6, or later, to upgrade to {product-title-short} version 2.1.7 and 2.1.8. If you cannot upgrade your {ocp} version 4.5 to a later version, you can continue to use {product-title-short} version 2.1.6.
-
 == Errata 2.1.8
 
 View a summarized list of {product-title} Errata 2.1.8 updates:
+
+*Important:* You must run {ocp} version 4.6, or later, to upgrade to Errata 2.1.7 and later. If you cannot upgrade your {ocp} version 4.5 to a later version, you can continue to use {product-title-short} version 2.1.6.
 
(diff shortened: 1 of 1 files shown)


release_notes/errata.adoc
```diff
@@ -5,12 +5,12 @@ By default, Errata updates are automatically applied when released. See link:../
...
 *Important:* For reference, link:https://access.redhat.com/errata/#/[Errata] links and GitHub numbers might be added to the content and used internally. Links that require access might not be available for the user.
 
-* *Important:* You must run {ocp} version 4.6, or later, to upgrade to {product-title-short} version 2.1.7 and 2.1.8. If you cannot upgrade your {ocp} version 4.5 to a later version, you can continue to use {product-title-short} version 2.1.6.
-
 == Errata 2.1.8
 
 View a summarized list of {product-title} Errata 2.1.8 updates:
 
+*Important:* You must run {ocp} version 4.6, or later, to upgrade to Errata 2.1.7 and later. If you cannot upgrade your {ocp} version 4.5 to a later version, you can continue to use {product-title-short} version 2.1.6.
+
```
swopebe, line 13: This is why I don't like it here: We have a discussion about 2.1.6 in the 2.1.8 section. If we use 2.1.8, then we dismiss 2.1.7. It's in my opinion better to be inclusive and add it to the important notes above, where the user also reads about the links already. We would not put that in each section--that would not make sense.

cadawson: Yeah, not the best, I will agree. Let's come up with a plan for the next one. Thank you for working with me on it.

release_notes/errata.adoc
```diff
@@ -5,12 +5,12 @@ By default, Errata updates are automatically applied when released. See link:../
...
 * Resolves the issue with Observability `thanos-store-shard` pods in `crashloopback` state after upgrade to {ocp} 4.6.30. (GitHub 13081)
 
 * Fixes an issue with `placementrule` and `placementbinding` not deleted when user deletes a disabled policy in the console. (GitHub 12689)
```
cadawson, line 16: Fixes an issue with `placementrule` and `placementbinding` not deleting...?

cadawson (approved):
One minor suggestion to look at. If not, thank you and merge!

release_notes/errata.adoc
swopebe: If it includes most of the file by the time we are done with 2.1.X, it may be better in one spot. For now it is just two so will merge and check in another time.

PR #2299, opened 2021-06-24
updating the list of supported object stores for obs

dockerymick:
https://github.com/open-cluster-management/backlog/issues/13376

### DIFF
--- a/observability/observe_environments.adoc
+++ b/observability/observe_environments.adoc
@@ -18,16 +18,12 @@ image:../images/RHACM-ObservabilityArch.png[Multicluster observability architect
 
 By default, observability is included with the product installation, but not enabled. Due to the requirement for persistent storage, the observability service is not enabled by default. {product-title-short} supports the following S3 compatible, stable object stores:
 
-- Amazon S3 (or other S3 compatible object stores like Ceph)
+- Amazon S3 (or other S3 compatible object stores like Minio and Ceph)
 - Google Cloud Storage
 - Azure storage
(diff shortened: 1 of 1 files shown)


observability/observe_environments.adoc
```diff
@@ -22,6 +22,10 @@ By default, observability is included with the product installation, but not ena
 - Google Cloud Storage
 - Azure storage
 - {ocp-short}
+- OpenStack Swift
+- Tencent Cloud Objeect Storage (COS)
+- AliYun Object Storage Service (OSS)
+- File system
 +
```
bjoydeep: @dockerymick none of these are supported as STABLE and PRODUCTION READY.
Swift, Tencent, AliYun are all Beta for Production, not stable
Local filesystem is not for production Usage.
So I would not add any of these. Source: https://thanos.io/tip/thanos/storage.md/#supported-clients

IMHO - we state what we know now. And then link to the upstream doc for latest additions. They will not subtract. Else we will have to watch upstream doc and for every change there - we need to update ours. Is that a feasible model?

BTW OpenShift Container storage has been renamed. https://www.redhat.com/en/technologies/cloud-computing/openshift-data-foundation

dockerymick: @bjoydeep I agree with only documenting what we know is supported by ACM. Thank you for your suggestion in the issue as well! I will make an update soon

bjoydeep (approved):
lgtm

PR #2320, opened 2021-06-30
Update customize_observability.adoc

dockerymick:
https://github.com/open-cluster-management/backlog/issues/13758#issuecomment-870960667

### DIFF
--- a/observability/customize_observability.adoc
+++ b/observability/customize_observability.adoc
@@ -271,9 +271,9 @@ View the data from your managed clusters by accessing Grafana. Complete the foll
 . From the navigation menu, click *Infrastructure* > *Clusters*. 
 . Access your Grafana dashboards by clicking the *Grafana* link.
 . Access the Prometheus metric explorer by selecting the *Explore* icon from the Grafana navigation menu.
-. To query metrics that come from a single node cluster (SNO), add the following label in the query expression: `{clusterType="SNO"}`. For example, to _cluster_infrastructure_provider_ from a SNO cluster, use the following query expression: `cluster_infrastructure_provider{clusterType="SNO"}`
+. To query metrics that come from a single node cluster, add the following label in the query expression: `{clusterType="SNO"}`. For example, to _cluster_infrastructure_provider_ from a single node cluster, use the following query expression: `cluster_infrastructure_provider{clusterType="SNO"}`
 +
-*Note*: Do not set the `ObservabilitySpec.resources.CPU.limits` parameter if observability is enabled on SNO managed clusters. When you set the CPU limits, it causes the observability pod to be counted against the capacity for your managed cluster. See https://github.com/openshift/enhancements/blob/master/enhancements/workload-partitioning/management-workload-partitioning.md#management-workload-partitioning[Management Workload Partioning] for more information.
+*Note*: Do not set the `ObservabilitySpec.resources.CPU.limits` parameter if observability is enabled on single node managed clusters. When you set the CPU limits, it causes the observability pod to be counted against the capacity for your managed cluster. See https://github.com/openshift/enhancements/blob/master/enhancements/workload-partitioning/management-workload-partitioning.md#management-workload-partitioning[Management Workload Partioning] for more information.
 
 
 [#disable-metrics-collector]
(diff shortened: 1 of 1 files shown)


observability/customize_observability.adoc
```diff
@@ -272,6 +272,8 @@ View the data from your managed clusters by accessing Grafana. Complete the foll
 . Access your Grafana dashboards by clicking the *Grafana* link.
 . Access the Prometheus metric explorer by selecting the *Explore* icon from the Grafana navigation menu.
 . To query metrics that come from a SNO cluster, add the following label in the query expression: `{clusterType="SNO"}`. For example, to _cluster_infrastructure_provider_ from a SNO cluster, use the following query expression: `cluster_infrastructure_provider{clusterType="SNO"}`
++
+*Note*: Avoid setting `CPU.limits` in the `ObservabilitySpec.resources` YAML section if observability is enabled on OpenShift SNO clusters that are included as managed clusters.
```
dockerymick: @marcolan018 in the issue, I suggested the following update: 


Note: It is best practice to avoid setting limits on the MultiClusterObservability resource, so that the pod is not counted against your cluster capacity.

I think it would be helpful to explain to the user why they shouldn't set the CPU limits, so I recommend we update the statement to the following:

It is best practice to avoid setting `CPU.limits` in the `ObservabilitySpec.resources` YAML section if observability is enabled on OpenShift SNO clusters that are included as managed clusters. Your pod is not counted against your cluster capacity when the limit is not set.

marcolan018:
@CrystalChun can you help to review this doc change? Thanks

observability/customize_observability.adoc
CrystalChun: I agree with including an explanation. 

**Perhaps something like**:

> NOTE: Do not set `ObservabilitySpec.resources.CPU.limits` if observability is enabled on managed SNO clusters. Setting the CPU limits causes the observability pod to be counted against your managed cluster's capacity. See [workload partitioning](https://github.com/openshift/enhancements/blob/master/enhancements/workload-partitioning/management-workload-partitioning.md) for more information.

**Reasons**:

- We need to emphasize they definitely should not set it so using a direction action phrase such as "do not" instead of "best practice" so that the user definitely does NOT do it 
- (If I were reading this doc) Having the full YAML path together makes it easier to see exactly which part is being talked about
- SNO clusters implies openshift already
- Including a link to the feature will give context

---
Feel free to wordsmith @dockerymick!

dockerymick: Thanks for your feedback @CrystalChun. Based on your explanation, i recommend that we create a known issue and then have a simple line instructing the user to avoid setting the limit. I will make some changes soon

CrystalChun: I'm not sure if it's an issue perse, but correct me if I'm wrong! @marcolan018 What do you think? I'll ask the far edge tech leads as well.

Thanks Mikela!

dockerymick: @CrystalChun you're right, I had to reread your explanation haha. Making the changes now !

observability/customize_observability.adoc
```diff
@@ -272,6 +272,8 @@ View the data from your managed clusters by accessing Grafana. Complete the foll
 . Access your Grafana dashboards by clicking the *Grafana* link.
 . Access the Prometheus metric explorer by selecting the *Explore* icon from the Grafana navigation menu.
 . To query metrics that come from a SNO cluster, add the following label in the query expression: `{clusterType="SNO"}`. For example, to _cluster_infrastructure_provider_ from a SNO cluster, use the following query expression: `cluster_infrastructure_provider{clusterType="SNO"}`
++
+*Note*: Do not set the `ObservabilitySpec.resources.CPU.limits` parameter if observability is enabled on SNO managed clusters. When you set the CPU limits, it causes the observability pod to be counted against the capacity for your managed cluster. See https://github.com/openshift/enhancements/blob/master/enhancements/workload-partitioning/management-workload-partitioning.md#management-workload-partitioning[Management Workload Partioning] for more information.
```
CrystalChun: The Telco doc teams write `SNO` as `single node cluster` so let's change it to that?

CrystalChun:
One small comment, lgtm otherwise!

dockerymick:
Thanks for your reviews and suggestions Crystal! Moving forward and merging this PR

PR #2324, opened 2021-07-01
open-cluster-management/backlog#8597--sample repo

swopebe:
@xiangjingli we need careful examination of each of the samples; just the four files. I made some changes already but the samples themselves, please see if they are still correct? Thank you.


applications/channel_sample.adoc
```diff
@@ -216,7 +211,7 @@ The following example channel definition abstracts a Helm repository as a channe
 ----
 apiVersion: v1
 data:
-  insecureSkipVerify: "true" # deprecated
+  insecureSkipVerify: "true" # deprecated //NOTE FROM BRANDI, IS THIS REMOVED NOW???
 kind: ConfigMap
```
swopebe: See my comment here, it was deprecated last release... still? or removed?

xiangjingli: Yes, it has been deprecated. now the insecureSkipVerify setting should be defined in the channel spec.insecureSkipVerify instead of the channel referred configMap

- remove the channel configmap sample
```
apiVersion: v1
data:
  insecureSkipVerify: "true" # deprecated //NOTE FROM BRANDI, IS THIS REMOVED NOW???
kind: ConfigMap
metadata:
  name: insecure-skip-verify
  namespace: hub-repo
```

- use the new channel sample
```
apiVersion: v1
kind: Namespace
metadata:
  name: hub-repo
---
kind: Channel
metadata:
  name: Helm
  namespace: hub-repo
spec:
    pathname: [https://9.21.107.150:8443/helm-repo/charts] # URL points to a valid chart URL.
    insecureSkipVerify: true
    type: HelmRepo
```

- Add the new field in the Channel YAML table
```
spec.insecureSkipVerify | Optional. default value is false, if set to be true, the channel connection will be built by skipping the authentication
-- | --

xiangjingli (commented):
The rest lgtm

PR #2334, opened 2021-07-06
2.2.4 known issue for observability

dockerymick:
https://github.com/open-cluster-management/backlog/issues/13407

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -177,7 +177,7 @@ After the `_finalizers_` parameter is removed, the `ObervabilityAddon` resource 
 [#metrics-not-collected]
 === Metrics data no longer collected after upgrade
 
-After you upgrade {product-title-short} from 2.2.3 to 2.2.4, you might notice that the `metrics-collector` in the `open-cluster-management-addon-observability` namespace stops collecting data from your managed clusters. This is because the image manifests ConfigMap is not updated. 
+After you upgrade {product-title-short} from 2.2.3 to 2.2.4, you might notice that the `metrics-collector` in the `open-cluster-management-addon-observability` namespace stops collecting data from your managed clusters. This is because the image manifests ConfigMap was updated. 
 
 You might see the following Quay.io images being used in the `endpoint-observabililty-operator` YAML file: `quay.io/open-cluster-management/endpoint-monitoring-operator:2.2.0-6a5ea47fc39d51fb4fade6157843f2977442996e` and `quay.io/open-cluster-management/metrics-collector:2.2.0-ff79e6ec8783756b942a77f08b3ab763dfd2dc15`. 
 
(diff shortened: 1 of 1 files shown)


release_notes/known_issues.adoc
```diff
@@ -174,6 +174,15 @@ After the `_finalizers_` parameter is removed, the `ObervabilityAddon` resource
 
 {product-title-short} observability does not display data from an ROKS cluster on some panels within built-in dashboards. This is because ROKS does not expose any API Server metrics from servers they manage. The following Grafana dashboards contain panels that do not support ROKS clusters: `Kubernetes/API server`, `Kubernetes/Compute Resources/Workload`, `Kubernetes/Compute Resources/Namespace(Workload)`
 
+[#metrics-not-collected]
+=== Metrics data no longer collected after upgrade
+
+After you upgrade {product-title-short} from 2.2.3 to 2.2.4, you might notice that the `metrics-collector` in the `open-cluster-management-addon-observability` namespace stops collecting data from your managed clusters. This is because the image manifests configuration map is not updated.
```
marcolan018: “configration map" should be configmap.

I think the one below is better to describe the cause:
This is because the image manifests configmap is updated after multicluster observability operator.

swopebe: Please write as we should according to the manuals and kube doc: https://coreos.slack.com/archives/CUHCCV19P/p1609874948058400 @marcolan018 FYI this is how we should write it unless in code.

Known issues are for refreshes, so might not be best here but better in the Errata doc. Remember the only issues we have for Errata is the Errata release note. Errata release note for zStream.

dockerymick: @swopebe @marcolan018 updated configuration map to ConfigMap

release_notes/known_issues.adoc
```diff
@@ -174,6 +174,15 @@ After the `_finalizers_` parameter is removed, the `ObervabilityAddon` resource
 
 {product-title-short} observability does not display data from an ROKS cluster on some panels within built-in dashboards. This is because ROKS does not expose any API Server metrics from servers they manage. The following Grafana dashboards contain panels that do not support ROKS clusters: `Kubernetes/API server`, `Kubernetes/Compute Resources/Workload`, `Kubernetes/Compute Resources/Namespace(Workload)`
 
+[#metrics-not-collected]
+=== Metrics data no longer collected after upgrade
+
+After you upgrade {product-title-short} from 2.2.3 to 2.2.4, you might notice that the `metrics-collector` in the `open-cluster-management-addon-observability` namespace stops collecting data from your managed clusters. This is because the image manifests ConfigMap was updated.
```
marcolan018: Last sentense can be changed as below

This is because the image manifests ConfigMap was upgraded after multicluster observability operator upgraded.

swopebe: Meeting yesterday, all confirmed this should move to the Errata note for 2.2.4 and out of GA release known issues.

dockerymick: @swopebe Yes, however this issue is just a known issue. 2.2.4 does not have a fix for this issue  and 2.2.5 also does not address it, so this should still be a known issue in 2.2

PR #2344, opened 2021-07-07
Update install_connected.adoc

JakobGray:
Issue: https://github.com/open-cluster-management/backlog/issues/12957

### DIFF
--- a/install/install_connected.adoc
+++ b/install/install_connected.adoc
@@ -264,7 +264,7 @@ metadata:
 spec:
   sourceNamespace: openshift-marketplace
   source: redhat-operators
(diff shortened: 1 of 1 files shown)


install/install_connected.adoc
```diff
@@ -264,7 +264,7 @@ metadata:
 spec:
   sourceNamespace: openshift-marketplace
   source: redhat-operators
-  channel: release-2.2
+  channel: release-2.3
   installPlanApproval: Automatic
```
swopebe, line 268: yikes, missed that, good catch

swopebe (approved):

PR #2345, opened 2021-07-07
add upgrade instructions

dockerymick:
https://github.com/open-cluster-management/backlog/issues/12253

### DIFF
--- a/risk_compliance/create_gatekeeper.adoc
+++ b/risk_compliance/create_gatekeeper.adoc
@@ -7,6 +7,7 @@ Use the gatekeeper operator policy to install the gatekeeper operator and gateke
 
 * <<install-gatekeeper-operator-policy,Installing gatekeeper using a gatekeeper operator policy>>
 * <<creating-a-gatekeeper-policy-from-the-console,Creating a gatekeeper policy from the console>>
+* <<upgrading-gatekeeper-gatekeeper-operator,Upgrading gatekeeper and the gatekeeper operator>>
 * <<updating-gatekeeper-operator-policy,Updating gatekeeper operator policy>>
 * <<deleting-gatekeeper-operator-policy,Deleting gatekeeper operator policy>>
 * <<uninstalling-gatekeeper,Uninstalling gatekeeper policy, gatekeeper, and gatekeeper operator policy>>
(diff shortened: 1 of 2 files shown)


risk_compliance/create_gatekeeper.adoc
```diff
@@ -94,6 +95,22 @@ spec:
     other-annotation: "another test"
 ----
 
+[#upgrading-gatekeeper-gatekeeper-operator]
+== Upgrading gatekeeper and the gatekeeper operator
+
+You can upgrade the versions for gatekeeper and the gatekeeper operator. Complete the following steps:
+
+. When you install the gatekeeper operator with the gatekeeper operator policy, notice the value for `installPlanApproval`. The operator upgrades automatically when `installPlanApproval` is set to `Automatic`. You must approve the upgrade of the gatekeeper operator manually, for each cluster, when `installPlanApproval` is set to `Manual`.
+
```
cadawson, line 104: Do you have to do step 2 whether it is set to Automatic or manual, or is it only if it is manual?

cadawson (commented):
Here is the content in the advisory. It has been shipped, so if it is wrong, no one has said anything: 
Before applying this update, make sure all previously released errata
relevant to your system have been applied.

The requirements to apply the upgraded images are different whether or not you used the operator. Complete the following steps, depending on your installation:

- Upgrade gatekeeper operator:
The gatekeeper operator that is installed by the gatekeeper operator policy has `installPlanApproval` set to `Automatic`. This setting means the operator will be upgraded automatically when there is a new version of the operator. No further action is required for upgrade. If you changed the setting for `installPlanApproval` to `Manual`, then you must view each cluster to manually approve the upgrade to the operator.

- Upgrade gatekeeper without the operator:
The gatekeeper version is specified as part of the Gatekeeper CR in the gatekeeper operator policy. To upgrade the gatekeeper version:
  a) Determine the latest version of gatekeeper by visiting: https://catalog.redhat.com/software/containers/rhacm2/gatekeeper-rhel8/5fadb4a18d9a79d2f438a5d9.
  b) Click the tag dropdown, and find the latest static tag. An example tag is v3.3.0-1.
  c) Edit the gatekeeper operator policy and update the image tag to use the latest static tag. For example, you might change this line to image: 'registry.redhat.io/rhacm2/gatekeeper-rhel8:v3.3.0-1'.

Refer to https://open-policy-agent.github.io/gatekeeper/website/docs/howto/ for additional information.

Here is a link to it: https://errata.devel.redhat.com/docs/show/75236

As I recall, this was a quick thing, so you probably know more about it than I did.

risk_compliance/create_gatekeeper.adoc
dockerymick: I believe it is only if it is manual

cadawson: If they are really two different procedures (do one for automatic, and the other for manual), consider changing the top-level number list to a bulleted list, so it looks more like it is an option, rather than a procedure where you have to complete both steps. For example:
Complete the steps that apply to your situation:
* If you have automatic set up...
* If you have to update manually, complete the following procedure:
+
. Manual...
. Steps...
. Here...

Let me know if it doesn't make sense.

PR #2361, opened 2021-07-13
open-cluster-management/backlog#13526 Peer

cadawson:


cadawson:
Updated the link, as the main branch title changed from "master" to "main". The other link worked, but this is more accurate, and more in line with the direction of our docs.

PR #2371, opened 2021-07-14
Updating VWT

dockerymick:
https://github.com/open-cluster-management/backlog/issues/13188

### DIFF
--- a/console/vwt.adoc
+++ b/console/vwt.adoc
@@ -33,8 +33,15 @@ You can access the Visual Web Terminal by completing the following steps:
 . Log in to your console.
 . Access the terminal either from the console navigation where you can open in a new tab, or select *Visual Web Terminal* icon from the console _Header_. From the header, choose to open in a new tab or open in a current tab.
 . Start running commands in the input field. For example, you can start running any `oc` or `kubectl` commands and get a visual output.
-. *Optional*: Run or select `getting started` to pull up the _Getting Started_ page in the Visual Web Terminal overview.
+. *Optional*: Run or select the *getting started* link to pull up the _Getting started_ page in the Visual Web Terminal overview.
 . *Optional*: You can click the buttons to change the default context and namespace. By default, the context is the hub cluster and the namespace is the default namespace. After you change the default namespace, the `@default` namespace is changed until you manually change it again.
++
+*Notes*:
++
+** After you run a command in your session, the following actions are available: _Copy_, _Re-run_, and _End session_
+// (says re-execute, would like to change that to Reload or Rerun)
+** When you select a pod to see the details, the screen is split. The second screen that appears displays the command that is ran to display the details. The following details that are included in the split-screen tabs are _Summary_, _Logs_, _Terminal_, and _YAML_.
+** If you have multiple tabs open, you can selct the namespace and the end of the page to use the pop-up finder panel. You can search for the namespace that you want to use, and receive information independently if you have multiple tabs open. 
 . Use the <<supported-command-list,Supported command list>> to learn about commands that you can run using `kubectl` or `oc`.
 
 [#supported-command-list]
@@ -87,7 +94,7 @@ For example, `history help` displays the last 25 commands that contained the str
 
 | `oc`
 | Runs Red Hat OpenShift `oc` commands.
-*Tip:* See https://docs.openshift.com/enterprise/3.0/cli_reference/basic_cli_operations.html[Red Hat OpenShift CLI Operations] for more information about the `oc` commands.
+*Tip:* See https://docs.openshift.com/enterprise/3.0/cli_reference/basic_cli_operations.html[Red Hat OpenShift CLI Operations] for more information about the `oc` commands. For example, when you run `oc get pods`, a grid is displayed to show the pods on your cluster. When you hover over a square, the name of the pod and status are shown. Select _Click to show more detail_ for more information. If you want to view your pods in a table list view, select the *Table* icon.
 | `oc get svc`
 
 | `search`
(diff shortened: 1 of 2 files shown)


console/vwt.adoc
```diff
@@ -5,13 +5,14 @@ The {product-title} Visual Web Terminal is a Technology preview feature. You can
 
 With Visual Web Terminal, commands that return cluster resource data are displayed visually in interactive tabular format, rather than plain text that is displayed in a standard terminal. You can see the data, click on a row, and then see more details about that resource. 
 
-By default, Visual Web Terminal includes the versions of the `oc`, `kubectl`, and `helm` CLI binaries that are equivalent to the latest {ocp-short} version that is supported by {product-title-short} for the hub cluster.
+By default, Visual Web Terminal includes the versions of the `oc`, `kubectl`, `helm`, and link:../clusters/find_subctlfile.adoc[`subctl`] CLI binaries that are equivalent to the latest {ocp-short} version that is supported by {product-title-short} for the hub cluster.
```
swopebe: @cadawson can we re-think this filename again? "find_subctlefile" is a bit confusing but the file doesn't seem to exist, so I cannot read what it should or could be. From what I understand, we are offering new cli commands. Those files are usually: submariner_cli (or something like that).... But I would need to see the header and filename. 

This is all in the effort to have filenames tell user more what the topic is and match headers--as we mentioned in the call with James. It's a bit of a mess in some areas right now but we can start with new files now. (assuming you might have this in an issue somewhere?)

console/vwt.adoc
```diff
@@ -31,11 +33,23 @@ The Visual Web Terminal connects to the cloud that you are already logged into w
...
+** When you select a resource in the command results table to see the details, the screen is split. The second screen that appears displays the command that was run to display the details. The details that are included in the split-screen details tabs are _Summary_, _YAML_,  and others that are associated with the resource type, for example _Logs_ and _Terminal_. 
+. *Optional*: Run the command, `getting started` or select the *getting started* link to pull up the _Getting started_ page in a new tab on the Visual Web Terminal page. The new tab remains open until you close it.
+. *Optional*: You can click the buttons to change the default context and namespace using the pop-up finder panel. By default, the context is the hub cluster and the namespace is the default namespace. After you change the default namespace, the `@default` namespace is changed until you manually change it again.
++
+*Note*: If you have multiple tabs open, you can select the a diifferent namespace for every tab so that the commands that you run on that tab default to the namespace you selected. You can search for the namespace that you want to use.
```
swopebe: typo line 50.

release_notes/whats_new.adoc
```diff
@@ -30,10 +30,9 @@ Learn more about what is new this release:
 == Web console
 
 * New navigation --Brandi will add
-
-* **Technology Preview:** You can access Visual Web Terminal both the navigation and the header. 
-  - Visual Web Terminal: keep the updates here (new function only that he wants to highlight) status grid and scrolling
-
+* **Technology Preview:** You can access Visual Web Terminal from both the navigation and the header. 
+// - Visual Web Terminal: keep the updates here (new function only that he wants to highlight) status grid and scrolling
+  - Resource details are now availble in a split screen that supports scrolling through the history of the different resource details. See link:../console/vwt.adoc#starting-a-session[Starting a session] for more information.
```
swopebe, line 35: This looks good, given that it is tech preview and getting removed, we may not need to add anything else here.

swopebe (approved):
I approve but I am requesting we revisit file name for subctl file.

console/vwt.adoc
cadawson: @swopebe : I am not sure what this is linking to, right now. It looks like it might be referring to the commands from Submariner, because VWT can run them. Maybe this is a placeholder for https://submariner.io/operations/deployment/subctl/? I am not familiar with that file. I agree with you that a different name would be better, if it ends up being our file.

swopebe: @cadawson submariner (from foundation) CLI is integrated this release and if so, there should have been a doc issue opened if we are adding commands to the product. Can you ask the foundation team about this or maybe find the correct developer issues and open it to find yours? Foundation team scrum may be a good place to bring this up, since we need to make sure dev/doc issues are opened each release before closing the story.

mark-nc: What I told Mikela during the review is that it did not make sense to provide documentation for all the commands supported by the Submariner CLI `subctl` in the VWT documentation as there should already be in the 2.3 docs a whole section about Submariner support that would cover the CLI usage and we should just link to that as that is where users would be looking for Submariner information anyway.  If that documentation does not go into detail on use of `subctl` CLI, then I would hope that documentation has a link out to `subctl` CLI documentation elsewhere (perhaps a wiki in the Submariner repo)?

cadawson: Thanks @swopebe and @mark-nc . Let me check with Foundation. We can certainly link somewhere that will help the user with the commands.

swopebe: @dockerymick we may want to take out the link until Chris has a doc for it, from what I see, there is no issue for us.

I have one like this where I depend on a cluster file that does not exist, so I took out the link until it does.
Just a suggestion.

dockerymick: @swopebe @cadawson I named the file like that as a placeholder until i found the correct link. Apologies for the confusion

dockerymick: I agree that we should use that subctl link that @cadawson provided. at first I was thinking to point the user to [Deploying Submariner with the console](https://github.com/open-cluster-management/rhacm-docs/blob/2.3_stage/services/deploy_submariner.adoc)

dockerymick:
@mark-nc @swopebe @cadawson Thank you all for the reviews, comments, and approvals. I agreed to use the subctl link that Chris provided in the comments. Merging this PR

PR #2376, opened 2021-07-15
open-cluster-management/backlog#14098 KI no chg cred

cadawson:


release_notes/known_issues.adoc
```diff
@@ -117,6 +117,12 @@ For more information, see link:../observability/observability_enable.adoc#enabli
 [#cluster-management-issues]
 == Cluster management known issues
 
+[#no-change-upgrade-cred]
+== Cannot change credentials on clusters after upgrading to version 2.3
+//2.3:14098
+
+After you upgrade {product-title-short} to version 2.3, you cannot change the credential secret for any of the managed clusters that were managed by {product-title-short} before the upgrade.
```
leena-jawale: should it be `After you upgrade {product-title-short} to version 2.3, you cannot change the credential secret for any of the hive created managed clusters that were managed by {product-title-short} before the upgrade.` ?

cadawson: That would work for me. So, it is only those created by RHACM? @leena-jawale

cadawson: If so, I changed it a little, but I think it means the same thing.

leena-jawale: yep looks good

PR #2399, opened 2021-07-19
open-cluster-management/backlog#13523 Move ki

cadawson:


cadawson:
Moving to regular known issue.

PR #2413, opened 2021-07-20
open-cluster-management/backlog#14408 ki no ch lab

cadawson:


dhrpatel4 (approved):
Thanks @cadawson , this address the problem.

PR #2441, opened 2021-07-26
updating section title header size

dockerymick:
fixing format that i noticed


cadawson:
Thanks @dockerymick !

PR #2442, opened 2021-07-26
open-cluster-management/backlog#13207--cluster cred edit with link

swopebe:


credentials/credential_ansible.adoc
```diff
@@ -17,12 +17,11 @@ You need a credential to use {product-title} console to deploy and manage an {oc
 You must have the following prerequisites before creating a credential:
 
 * A deployed {product-title} hub cluster
-* Internet access for your {product-title} hub cluster.
-* Ansible login credentials, which include access key ID and secret access key. 
-* Account permissions that allow installing clusters on Ansible. 
-// link here
+* Internet access for your {product-title} hub cluster
+* Ansible login credentials, which include access key ID and secret access key; see https://docs.ansible.com/ansible-tower/latest/html/userguide/credentials.html[Credentials for Ansible Tower].
+* Account permissions that allow installing clusters on Ansible.
```
Randy424: "...installing clusters on Ansible."
Strictly speaking, we don't install clusters on Ansible. We just use Ansible to run jobs before or after install, or upgrade. That said, I would alter the sentence with this in mind: the account permissions needed for creating Ansible credentials are the same permissions needed to create any other credential on the cluster (so you could reuse text from another credential doc for explaining permissions if you wanted to), but if you wanted to keep things roughly as they are, and since "installing clusters on Ansible." doesn't work, you could use this: "* Account permissions that allow installing clusters on the hub". Those permissions would be sufficient for creating a secret, and the information itself is in line with the what's going on under the hood. 

Thanks!

Randy424: I just reviewed the [whole doc ](https://github.com/open-cluster-management/rhacm-docs/blob/e8a3d1685ab02f5d4e5f10c8130fad6dcc4361aa/credentials/credential_aws.adoc), and I think I understand the "clusters on Ansible" phrase a little better. This instruction is referring to permissions on the Ansible platform, not on RHACM. 

In this case, you just need an Ansible account with permissions that allow for the execution of Ansible jobs templates. We can also reference [this link](https://docs.ansible.com/ansible-tower/latest/html/userguide/users.html).

Ignore the latter half of my first comment! Thanks!

credentials/credential_ansible.adoc
```diff
@@ -42,13 +41,9 @@ Log in to Ansible to find the ID.
 . Add your _Ansible Secret Access Key_.
```
Randy424: `. You can optionally add a _Base DNS domain_ for your credential. If you add the base DNS domain to the credential, it is automatically populated in the correct field when you create a cluster with this credential.
. Add your _Ansible access key ID_ for your Ansible account.
Log in to Ansible to find the ID.
. Add your _Ansible Secret Access Key_.`

**We only need the following information:**
- The Ansible host URL
- An authentication token
- Name
- Namespace
You will want to rename _Ansible access key ID_ to _Ansible host URL_
You will want to rename _Ansible Secret Access Key_ to _Ansible authentication token_ (or something with equivalent meaning)

**You can remove the references/steps for adding:**
- Base DNS domain
- SSH keys
- Red Hat pull secret
Also, note that we don't create or install a cluster using Ansible credentials. We just use Ansible jobs to configure the environment or cluster before or after install/upgrade. Sentences that suggest using Ansible credentials to install a cluster should be corrected.

See this screenshot for reference to the earlier outline (it's an overview of all the fields in the create Ansible credential wizard page):

Screen Shot 2021-07-27 at 10 13 52 AM

Thank you!

credentials/credential_ansible.adoc
```diff
@@ -42,13 +41,9 @@ Log in to Ansible to find the ID.
...
 . Enter your _Red Hat OpenShift pull secret_.
 You can download your pull secret from https://cloud.redhat.com/openshift/install/pull-secret[Pull secret].
-. Add your _SSH private key_ and _SSH public key_, which allows you to connect to the cluster.
-You can use an existing key pair, or create a new one with key generation program.
-See ---[Generating an SSH private key and adding it to the agent] for more information about how to generate a key.
+. Add your _SSH private key_ and _SSH public key_, which allows you to connect to the cluster. You can use an existing key pair, or create a new one with key generation program.
```
Randy424: yeah-this should be removed.

credentials/credential_ansible.adoc
swopebe: I actually agree @Randy424 that install on ansible is confusing, I think this is in a draft somewhere and I took it over. I think your feedback can help make it better. I will also add that link.

PR #2445, opened 2021-07-26
open-cluster-management/backlog#12010 Add proxy

cadawson:


clusters/proxy.adoc
```diff
@@ -0,0 +1,101 @@
...
++ 
+Replace `provisioning-network/CIDR` with the IP address of the provisioning network and the number of assigned IP addresses, in CIDR notation.
++
+Replace `BMC-address-range/CIDR` with the BMC address and the number of addresses, in CIDR notation.
++
+After those settings are added, they are applied to the clusters that you create. 
+
+. Provision the cluster by completing the procedure for creating a cluster. See xref:../clusters/create.adoc#creating-a-cluster[Creating a cluster] to select your provider.
+
+[#enable-cluster-wide-proxy-existing]
```
jnpacker, line 45: @cadawson this is the part Philip provided?  I think we want to call out that this is for specifically make components proxy aware on the Managed Clusters.

For example to make the Application Manager (AppMgr) work behind a proxy, the following settings must be applied to its deployment.

jnpacker: The content seems fine, we just need to better describe that the second section is for enabling a component from the managed cluster to work behind a proxy. Such as AppMgr.

jnpacker:
@cdoan1 Since you have the most experience with this configuration, do you mind review the documentation and making any adjustments you feel we should have.

The goal was to cover hub side and managed cluster side configurations.

clusters/proxy.adoc
cadawson: Modified the text a bit, @jnpacker . Thanks.

clusters/proxy.adoc
```diff
@@ -0,0 +1,101 @@
...
+* The managed cluster is on a cloud provider, but the firewall ports enable communication from the managed cluster to the hub cluster.
+
+To create a cluster that is configured with a proxy, complete the following steps:
+
+. Configure the cluster-wide-proxy setting on the hub cluster by adding the following information to your `install-config.yaml` file:
++
+[source,yaml]
+----
+apiVersion: v1
+baseDomain: <domain>
```
cdoan1, line 20: This example needs the `Kind` to know what this is about?

cdoan1 (approved):

clusters/proxy.adoc
cadawson: I added a kind: Proxy, @cdoan1 . Does that resolve this issue? Thanks!

clusters/proxy.adoc
```diff
@@ -0,0 +1,102 @@
+[#creating-a-cluster-proxy]
+= Creating a cluster in a proxy environment
+
+You can create a {ocp} cluster when your hub cluster is connected through a proxy server.
+
+One of the following situations must be true for the cluster creation to succeed:
+
+* {product-title} has a private network connection with the managed cluster that you are creating, but they reach the Internet using a proxy.
+
```
swopebe, line 9: Please don't use "they" here.... They, we, etc.... is not good for tech doc.

swopebe: You can reword to use what they actually means here.

clusters/proxy.adoc
```diff
@@ -0,0 +1,102 @@
...
+* The managed cluster is on a cloud provider, but the firewall ports enable communication from the managed cluster to the hub cluster.
+
```
swopebe, line 11: Since we named these infrastructure provider in the UI, should "cloud provider" change?

clusters/proxy.adoc
```diff
@@ -0,0 +1,102 @@
...
+Replace `port` with the communciation port with the proxy server.
++
+Replace `wildcard-of-domain` with an entry for domains that should bypass the proxy.
++ 
+Replace `provisioning-network/CIDR` with the IP address of the provisioning network and the number of assigned IP addresses, in CIDR notation.
++
+Replace `BMC-address-range/CIDR` with the BMC address and the number of addresses, in CIDR notation.
++
+After those settings are added, they are applied to the clusters that you create. 
+
```
swopebe, line 42: After you add the previous values, the settings are applied to your clusters.

?
Previous orients a non-sighted user; sentence more specific and active....

clusters/proxy.adoc
```diff
@@ -0,0 +1,102 @@
...
+. Provision the cluster by completing the procedure for creating a cluster. See xref:../clusters/create.adoc#creating-a-cluster[Creating a cluster] to select your provider.
+
+[#enable-cluster-wide-proxy-existing]
+== Enabling cluster-wide proxy on an existing cluster
+
+When you enable the cluster-wide proxy feature, you have to make the components of the managed cluster aware so they can work behind the proxy. To enable the components like Application Manager, the operator pods are updated with the following environment variables: `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`. These environment variables specify where the pod routes the `HTTP` and `HTTPS` traffic to the proxy server. The operator pods are automatically updated with these environment variables, so the standalone subscriptions work after enabling the cluster-wide proxy.
```
swopebe: Application Manager is the name of the component? or do you mean something more specific here?

swopebe: Changing such as instead of "like"

swopebe (changes requested):
few small things here and otherwise looks great

clusters/proxy.adoc
dockerymick: I recommend writing, "{product-title} has a private network connection with the managed cluster that you are creating, which uses a proxy to reach the Internet." 

Maybe say access rather than reach? Both work in my opinion

clusters/proxy.adoc
```diff
@@ -0,0 +1,106 @@
...
+kubectl patch deployment <deployment-name> -p '{"spec":{"template":{"spec":{"containers":[{"env":[{"name":"NO_PROXY","value":"'<no-proxy>'"}],"name":"'<container-name>'"}]}}}}'
+----
++
+Replace `deployment-name` with the name of the deployment.
++
+Replace `no-proxy` with the comma delimited list of IPs, hostnames, and domains where traffic will not be routed through the proxy. For example, `.cluster.local,.svc,10.128.0.0/14, example.com`.
++
+Replace `container-name` with the name of the container.
+
+*Tip:* For several containers, you can create and run a file that resembles the following example on each of the containers:
```
dockerymick, line 92: Is the tip intended for Step 3 or the entire doc overall? If it is intended for step 3, I recommend adding a plus sign (+) on line 91

clusters/proxy.adoc
cadawson: Good point. Thanks!

clusters/proxy.adoc
cadawson: Component name

clusters/proxy.adoc
cadawson: It is actually for the overall topic. That was intentional.

clusters/proxy.adoc
cadawson: I see what you mean here. But it refers to both of them, and I think "which uses a proxy to reach the Internet" can be unclear whether it is both or just the managed cluster.

PR #2451, opened 2021-07-27
Update credential_ocm.adoc

JakobGray:
Replace `cloud.redhat.com` with `console.redhat.com` and remove redundant step.

Issue: https://github.com/open-cluster-management/backlog/issues/13208

### DIFF
--- a/credentials/credential_ocm.adoc
+++ b/credentials/credential_ocm.adoc
@@ -8,7 +8,7 @@ Add an {ocm} credential so that you can discover clusters.
 [#prerequisites-discovery]
 == Prerequisites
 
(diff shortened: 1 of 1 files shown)


swopebe:
@JakobGray if you do not right now have a link to this, it is risky just changing to URL. Have they officially moved the doc?

swopebe:
Please use the Portal link here @JakobGray

JakobGray:
@swopebe There isn't a Portal link that I am aware of, but based on the discussion in the issue [here](https://github.com/open-cluster-management/backlog/issues/13208#issuecomment-887861466) I think it is safe to make the cloud->console link change

PR #2464, opened 2021-07-28
open-cluster-management/backlog#13053 review

cadawson:


clusters/managedclustersets.adoc
```diff
@@ -50,24 +50,22 @@ Complete the following steps to create a managed cluster set by using the {produ
...
 kind: ManagedClusterSet
 metadata:
   name: <clusterset1>
 ----
-+
+
 Replace `clusterset1` with the name of your managed cluster set. 
 
 [#adding-clusters-to-a-managedclusterset]
-== Adding clusters to a ManagedClusterSet
```
swopebe, line 67: I thought we were doing `this` with these, per meeting discussion. @dockerymick is that right. 

Other than that, this looks great and the changes make a big difference. Thanks!

swopebe (approved):

clusters/managedclustersets.adoc
cadawson: Thanks for the review! In the regular text, we are putting them in monospace. I think when they are in titles is when @mickeyjack suggests italics because of how the monospace renders in titles.

PR #2466, opened 2021-07-28
open-cluster-management/backlog#12702--info about sync button

swopebe:


applications/manage_apps_git.adoc
```diff
@@ -31,6 +31,8 @@ If you enter a new Git repository path, you can optionally enter Git connection
 . Enter values for the optional fields, such as branch, path, and commit hash. Optional fields are defined in the fields or in the hover text.
 
 . Notice the reconcile option. The `merge` option is the default selection, which means that new fields are added and existing fields are updated in the resource. You can choose to `replace`. With the `replace` option, the existing resource is replaced with the Git source.
+
```
sahare, line 34: @swopebe Looks good, but couple of things that I can think of to change here: - If it's set to off, it will never reconcile. - The sync button is on the cards on single application view/page, not under action menu. Thanks.

sahare (approved):

applications/manage_apps_git.adoc
swopebe: completely valid and I didn't realize, thanks for the improvement!

PR #2467, opened 2021-07-28
open-cluster-management/backlog#14631 ki pwr multarch

cadawson:


dberg1 (approved):

release_notes/known_issues.adoc
```diff
@@ -123,6 +123,90 @@ For ROKS clusters, {product-title-short} observability does not display data in
...
++
+Replace `private-repo` with the path to your repository.
+
+. Create a release image on the hub cluster.
+
+.. Create a file that contains information similar to the following example:
++
+----
+$ kubectl login
+$ cat << EOF | kubectl apply -f -
```
jaywcarman: I like showing the straight yaml for the `ClusterImageSet` instead of mixing it in with shell commands. Lines 188 and 189 should be removed. I think those actions are covered by your statement on line 204. The `EOF` on line 199 should also be removed.

release_notes/known_issues.adoc
```diff
@@ -123,6 +123,90 @@ For ROKS clusters, {product-title-short} observability does not display data in
...
+
+.. Merge the list in your manifest list with the existing manifest:
++
+----
+$ podman manifest push mymanifest docker://<private-repo>/ocp-release:4.8.1
+----
++
+Replace `private-repo` with the path to your repository.
+
+. Create a release image on the hub cluster.
```
jaywcarman: Would it be better to say something like:
On the hub cluster, create a `ClusterImageSet` pointing to the manifest list in your repository.

jaywcarman (commented):
Looks great @cadawson! Just a couple of suggestions.

release_notes/known_issues.adoc
cadawson: So, this just gets added to the config.yaml file? Do we still need need to run the file as the next step, or just remove that step? @jaywcarman

jaywcarman: The general process we're trying to describe is creating a custom resource object - the general OCP documentation can be found here: https://docs.openshift.com/container-platform/4.8/operators/understanding/crds/crd-managing-resources-from-crds.html#crd-creating-custom-resources-from-file_crd-managing-resources-from-crds

So the yaml file should describe the custom resource, and can be named anything. To actually create the custom resource object the user will need to run `oc apply -f <file_name>.yaml`. (note: the OCP documentation uses `oc create` instead of `oc apply`, I think both would work).

cadawson: Thanks @jaywcarman !

cadawson:
One more time @jaywcarman , and I think we are there. Thanks!

jaywcarman (approved):
LGTM - Thanks @cadawson!

PR #2481, opened 2021-07-29
open-cluster-management/backlog#12767--peer review needed and final d…

swopebe:
…ev review.

### DIFF
--- a/clusters/placement_managed.adoc
+++ b/clusters/placement_managed.adoc
@@ -27,11 +27,10 @@ The `Placement' specification includes the following fields:
 
 - `ClusterSets` represents the `ManagedClusterSets` from which the `ManagedClusters` are selected. 
 
-+
-If not specified, `ManagedClusters` is selected from the `ManagedClusterSets` that are bound to the placement namespace. 
+  
+  ** If not specified, `ManagedClusters` is selected from the `ManagedClusterSets` that are bound to the placement namespace. 
 
-+
-If specified, `ManagedClusters` is selected from the intersection of this set and the `ManagedClusterSets` that are bound to the placement namespace.
+  ** If specified, `ManagedClusters` is selected from the intersection of this set and the `ManagedClusterSets` that are bound to the placement namespace.
 
 - `NumberOfClusters` represents the desired number of `ManagedClusters` to be selected, which meet the placement requirements. 
 +
--- a/clusters/placement_managed.adoc
+++ b/clusters/placement_managed.adoc
@@ -34,7 +34,7 @@ If specified, `ManagedClusters` is selected from the intersection of this set an
 
 - `NumberOfClusters` represents the desired number of `ManagedClusters` to be selected, which meet the placement requirements. 
 +
-If not specified, all `ManagedClusters` that meet the placement requirements are selected
+If not specified, all `ManagedClusters` that meet the placement requirements are selected.
 
 - `Predicates` represents a slice of predicates to select `ManagedClusters` with label and claim selector. The predicates are ORed.
 
@@ -83,7 +83,7 @@ spec:
                 - us-west-1
 ----
 
-- You can select `ManagedClusters` from particular `clusterSets`. See in the following sample where `claimSelector` only matches `clusterSets:` `clusterset1` `clusterset2`:
+- You can select `ManagedClusters` from particular `clusterSets`. See the following sample where `claimSelector` only matches `clusterSets:` `clusterset1` `clusterset2`:
 
 +
 [source,yaml]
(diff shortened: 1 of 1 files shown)


elgnay:
The doc looks fine except some minor issues.

Screen Shot 2021-07-30 at 11 16 47 AM

Screen Shot 2021-07-30 at 11 20 21 AM

Screen Shot 2021-07-30 at 11 28 00 AM

swopebe:
@elgnay the important special tag is actually part of the overview, the list is information about Placement (so I named it overview.) But it may be better if I move it up, rather than place it at the end?

I actually created a section for placement binding and had examples under it I think in the first draft, but that was when I was writing them as steps. So I can make that change and the others make sense.

clusters/placement_managed.adoc
```diff
@@ -34,7 +34,7 @@ If specified, `ManagedClusters` is selected from the intersection of this set an
 
 - `NumberOfClusters` represents the desired number of `ManagedClusters` to be selected, which meet the placement requirements. 
 +
-If not specified, all `ManagedClusters` that meet the placement requirements are selected
```
cadawson, line 37: Ln 23: I was wondering whether this should be a note under the previous bullet, rather than a bullet of its own? 
Ln 30: "is bound" or "are bound"?
Ln 33: LOVE the way you worded this bullet! Very clear, and not easy to do.  Is "this set" clear enough, or would "the specified set" be more clear? 
Ln 37: I added a period here.

clusters/placement_managed.adoc
```diff
@@ -43,7 +43,7 @@ If not specified, all `ManagedClusters` that meet the placement requirements are
 
 You need to bind a `ManagedClusterSet` to a namespace by creating a `ManagedClusterSetBinding` in that namespace. *Note:* You need role-based access to `CREATE` on the virtual sub-resource of `managedclustersets/bind`. See the following examples:
```
cadawson, line 44: Extra space at beginning. Will check this.

clusters/placement_managed.adoc
```diff
@@ -43,7 +43,7 @@ If not specified, all `ManagedClusters` that meet the placement requirements are
...
 
-- You can select `ManagedClusters` with the `labelSelector`. See in the following sample where the `labelSelector` only matches clusters with label `vendor: OpenShift`.
+- You can select `ManagedClusters` with the `labelSelector`. See the following sample where the `labelSelector` only matches clusters with label `vendor: OpenShift`:
```
cadawson, line 46: Removed "in" from "in the following example" from a couple of these to keep them consistent with the other ones. Add it back to all of them, if you prefer that way.

clusters/placement_managed.adoc
```diff
@@ -140,11 +136,10 @@ spec:
 
 One or multiple `PlacementDecisions` with label `cluster.open-cluster-management.io/placement={placement name}` are created to represent the `ManagedClusters` selected by a `Placement`.
 
-- If a `ManagedCluster` is selected and added to a `PlacementDecision`, components that consume this `Placement` may apply the workload on this `ManagedCluster`. After the `ManagedCluster` is no longer selected and it is removed from the `PlacementDecisions`, the workload that is applied on this `ManagedCluster` should be removed accordingly.
+- If a `ManagedCluster` is selected and added to a `PlacementDecision`, components that consume this `Placement` might apply the workload on this `ManagedCluster`. After the `ManagedCluster` is no longer selected and it is removed from the `PlacementDecisions`, the workload that is applied on this `ManagedCluster` should be removed accordingly.
```
cadawson: - Changed "may" to "might".
- "After the `ManagedCluster` is no longer selected and it is removed..." > "After the `ManagedCluster` is no longer selected and is removed..."
- When we say it "should be removed...", we are implying doubt, in my opinion. Can we confidently say "is removed"?

cadawson (changes requested):
Thanks so much for working this one! A few comments and suggestions, but go with what you think. It looks good!

clusters/placement_managed.adoc
swopebe: I am not sure about the last part, this section you can see in the draft in the issue.

clusters/placement_managed.adoc
swopebe: 33. If changes would use relative pronoun: the set that is specified.

Other changes I agree and will make, along with dev review. Thanks

clusters/placement_managed.adoc
swopebe: I don't, it was left over from the draft. I thought I removed them. Thanks for catching it!

PR #2492, opened 2021-07-31
open-cluster-management/backlog#12103 -- AWS S3 Dev review

swopebe:


mikeshng (changes requested):
- We need to replace all `xili` references with something like `sample`. I think a search and replace should be suffice. 
**DONE--bcs**
- Replace `9.28.236.243` with another sample IP address.
**DONE--bcs**
- Following the `You can continue to create other AWS subscription and placement rule objects.`, add the sample for placementrule and subscription. 
-**DONE--bcs**
```
kind: PlacementRule
metadata:
  name: towhichcluster
  namespace: obj-sub-ns
spec:
  clusterSelector: {}
---
kind: Subscription
metadata:
  name: obj-sub
  namespace: obj-sub-ns
spec:
  channel: ch-object-dev/object-dev
  placement:
    placementRef:
      kind: PlacementRule
      name: towhichcluster
```

I think the first 2 bullets should be done. For the 3rd bullet (sample yaml) lets wait for @xiangjingli feedback.

xiangjingli:
agree to add the sample yamls as @mikeshng suggested in the the 3rd item 

@swopebe   We also want to highlight the other 2 points for users:
- we support to subscribe all the objects under a specific subfolder in the object bucket.
To do it, add the subfolder annotation to the subscription that will force the object bucket subscription to only apply all the resources under the subfolder path.

Using the annotation below:
```
annotations:
  apps.open-cluster-management.io/bucket-path: <your bucket subfolder>
``` 

here is one sample yaml
```
kind: Subscription
metadata:
  annotations:
    apps.open-cluster-management.io/bucket-path: subfolder1
  name: obj-sub
  namespace: obj-sub-ns
  labels:
    name: obj-sub
spec:
  channel: ch-object-dev/object-dev
  placement:
    placementRef:
      kind: PlacementRule
      name: towhichcluster
```

- object bucket subscription Known issue: For now we only support One yaml manifest per object file. Don't support multiple yaml manifests in the same object file yet 
**DONE--bcs**

applications/manage_apps_object.adoc
```diff
@@ -154,5 +156,27 @@ stringData:
...
+metadata:
+  name: obj-sub
+  namespace: obj-sub-ns
+spec:
+  channel: ch-object-dev/object-dev
+  placement:
+    placementRef:
+      kind: PlacementRule
+      name: towhichcluster
+----
```
swopebe: @mikeshng see your recommended changes in the PR are added.

applications/manage_apps_object.adoc
```diff
@@ -1,17 +1,19 @@
 [#managing-apps-with-object-storage-repositories]
 = Managing apps with Object storage repositories
 
-When you deploy Kubernetes resources using an application, the resources are located in specific repositories. Learn how to deploy resources from Object storage repositories in the following procedure. Learn more about the application model at xref:../applications/app_model.adoc#application-model-and-definitions[Application model and definitions].
+When you deploy Kubernetes resources using an application, the resources are located in specific repositories. Learn more about the application model at xref:../applications/app_model.adoc#application-model-and-definitions[Application model and definitions]:
 
-*User required access:* A user role that can create applications. You can only perform actions that your role is assigned. Learn about access requirements from the link:../access_control/rbac.adoc#role-based-access-control[Role-based access control] documentation. 
+*User required access:* A user role that can create applications.
```
swopebe: @xiangjingli you think we can get a better role here, we normally can say admin or cluster admin.... do you know the lowest level needed?

applications/manage_apps_object.adoc
```diff
@@ -1,17 +1,19 @@
...
 
 * <<deploy-resources-from-object-storage-git-repo, Deploy resources from object storage>>
 * <<sample-yaml-object, Sample YAML>>
 * <<creating-aws-object-bucket, Creating your Amazon Web Services (AWS) S3 object bucket>>
 * <<subscribing-to-aws-bucket, Subscribing to the object in the AWS bucket>>
 * <<sample-aws-subscription, Sample AWS subscription>>
 
-[#deploy-resource-object-storage]
+[#deploy-resources-from-object-storage-git-repo]
 == Deploy resources from object storage
```
swopebe, line 17: @xiangjingli thank you for the additional annotations that you gave me in the issue. Where is the best spot to add this part to the procedure?

Meaning the sub folder additions...

applications/manage_apps_object.adoc
xiangjingli: admin or cluster admin is good. 

And  `A user role that can create applications` is the lowest level needed , user should at least have the permission to create/update app lifecycle related CRs (channel, subscription, placementrule, ) in the given Namespace.

applications/manage_apps_object.adoc
xiangjingli: I think we can add a new item after  `Sample AWS subscription`, sth like

### Sample AWS subscription
...
### Subscribe to objects under a specific subfolder in the Minio and AWS object buckets 
.... // here we inform how to specify the subfolder annotation with a sample yaml

xiangjingli (approved):
lgtm

PR #2501, opened 2021-08-02
open-cluster-management/backlog#11206

swopebe:
--collecting cluster TP and putting under one header to eliminate bold tags in the same section and easier UX, added discover, changes cluster management and application management to what it is called now with 2.3 changes


swopebe:
1st round of clean-up, finalization efforts

PR #2518, opened 2021-08-03
open-cluster-management/backlog#14867 skipMachinePools

cadawson:


dhrpatel4 (commented):
Hi this looks good, but Now we will have to modify line 73
so I believe that if we add this spec then that SNO cluster pool creating extra worker nodes will not exists any more
1:27
so if user create sno cluster pool without spec then they will get those extra machine pools but if they add spec from yaml then no extra nodes will be created

@cadawson Just adding this based on our discussion

cadawson:
I think we have it @dhrpatel4 . Let me know how it looks.

dhrpatel4 (approved):
Thanks @cadawson for the prompt changes.
looks good to me

PR #2522, opened 2021-08-03
Adding knownn issue for metrics-collector

dockerymick:
https://github.com/open-cluster-management/backlog/issues/14287

### DIFF
--- a/troubleshooting/trouble_metrics_collector.adoc
+++ b/troubleshooting/trouble_metrics_collector.adoc
@@ -17,11 +17,13 @@ error: response status code is 500 Internal Server Error, response body is x509:
 
 If you have this problem, complete the following steps:
 
-. Log in to your {product-title} hub clluster. 
+. Log in to your managed clluster. 
 
-. Navigate to the managed cluster that you want to update. 
-
-. Delete the secret named, `observability-controller-open-cluster-management.io-observability-signer-client-cert` that is in the `open-cluster-management-addon-observability` namespace.
+. Delete the secret named, `observability-controller-open-cluster-management.io-observability-signer-client-cert` that is in the `open-cluster-management-addon-observability` namespace. Run the following command:
++
+----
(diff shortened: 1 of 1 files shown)


troubleshooting/trouble_metrics_collector.adoc
```diff
@@ -0,0 +1,29 @@
...
+----
+error: response status code is 500 Internal Server Error, response body is x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "observability-client-ca-certificate")
+----
+
+[#resolving-observability-metrics-unavailable]
+== Resolving the problem: metrics-collector cannot verify observability-client-ca-certificate
+
+If you have this problem, complete the following steps:
+
+. Log in to your {product-title} hub clluster.
```
marcolan018: All the required opeations, to delete the secret to forcely trigger the recreation, should be performed on managed cluster side.
I am not sure user needs to login hub cluster to do that??

dockerymick: i am not sure either. It is the only way that i am familiar with. More specifically if the user logs in to the hub cluster, they would then select the applications icon and click the OCP link to access their managed clusters

marcolan018: if in our doc, this is the standard steps to perform operation on managed cluster, I am ok to keep consistent here.
If not, I think it's better to simply ask user to login managed cluster, then use kubectl/oc to update the k8s resource there.

dockerymick: I accept your suggestion of instructing the user to log in to their managed cluster. Is there a command that should be used to delete the secret?

marcolan018: oc delete observability-controller-open-cluster-management.io-observability-signer-client-cert -n open-cluster-management-addon-observability

PR #2537, opened 2021-08-05
open-cluster-management/backlog#14386

swopebe:
--adding table

### DIFF
--- a/install/plan_capacity.adoc
+++ b/install/plan_capacity.adoc
@@ -1,30 +1,41 @@
 [#sizing-your-cluster]
 = Sizing your cluster
 
-Each {product-title} cluster is unique. Guidelines provide sample deployment sizes. Recommendations are classified by size and purpose. {product-title-short} applies the following three dimensions for sizing and placement of supporting services:
+Each {product-title} cluster is unique and the following guidelines provide sample deployment sizes for you. Recommendations are classified by size and purpose. {product-title-short} applies the following three dimensions for sizing and placement of supporting services:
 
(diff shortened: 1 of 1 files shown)


install/plan_capacity.adoc
```diff
@@ -1,30 +1,41 @@
...
 
-|Worker|3|redisgraph/redis x 1|12Gi|6 CPU|
+|Worker|3|redisgraph/redis x 1|12 GB|6 CPU
 |===
 
-In addition to {product-title-short}, the {ocp} cluster runs additional services to support cluster features. The following node sizes (3 nodes of the types noted in the information that follows, distributed evenly across 3 availability zones) are recommended.
+In addition to {product-title-short}, the {ocp-short} cluster runs additional services to support cluster features. The following node sizes (3 nodes of the types noted in the information that follows, distributed evenly across 3 availability zones) are recommended.
+
+[#product-environment-sno]
+=== Product environment: Single Node {ocp-short}
```
CrystalChun: These minimums are for a specific use-case: provisioning and managing 1,000 single node openshift (SNO) clusters. I'm not sure if it'll be confusing to say this is a SNO product environment?

Perhaps the title could be 
```
=== Product environment: Managing 1K Single Node {ocp-short}
```

cadawson: Hi @CrystalChun . I talked with Brandi, and she suggested adding a table introduction that specifies the use case would resolve the concern. There might be other information that is added to that section in the future. Will that work? Thank you!

cadawson:
I moved the contents of the PR into the SNO document because it is dev preview. It is not in the Customer Portal docs. The new PR is: https://github.com/open-cluster-management/rhacm-docs/pull/2540. I think this PR can be closed.

PR #2540, opened 2021-08-05
open-cluster-management/backlog#14386 move sno 1k DUPLICATE

cadawson:


cadawson:
I don't want to lose this work. so I added it to a file for the SNO clusters that will not be in the Portal. Verizon can access this file and get the information, When we merge this content back in, we can use it in its original spot, if there is not a better one. Thanks!

swopebe:
@cadawson I have this info in the other PR; I don't think we need two, as I cleaned up the topic in the other PR and commented out the table I added for future reference. The one line could be added to the original PR. Closing as duplicate since the other one has changes to be merged.

PR #2554, opened 2021-08-09
open-cluster-management/backlog#15005 upd supp cld

cadawson:


berenss:
the list of hub support has another small issue that you can pickup here:
See Red Hat OpenShift Dedicated (OSD) **(ocp-short}** version 4.7, and later).

berenss (approved):

PR #2556, opened 2021-08-10
Swope prereq

swopebe:


install/install_disconnected.adoc
```diff
@@ -50,28 +52,59 @@ See xref:../install/plan_capacity.adoc#sizing-your-cluster[Sizing your cluster]
...
+----
 additionalTrustBundle: |
   -----BEGIN CERTIFICATE-----
   certificate_content
   -----END CERTIFICATE-----
 sshKey: >-
-....
-.. Save the `install-config.yaml` file.
+----
+
```
swopebe, line 80: This is the addition, but please see the whole file for placement. Add anyone else who needs input.

drkho:
@gparvin @swopebe I recommend to point, in the document, to the following [catalog](https://catalog.redhat.com/software/containers/search?p=1&product_listings_names=Red%20Hat%20Quay) for the users. What do you think? Or have we already done it?

Screen Shot 2021-08-10 at 5 07 33 PM

swopebe:
@drkho we can point them to the catalog, but where and why would be needed when I add that link... 

the install doc is here and we kinda need to make sure everyone sees that, then we change it from there:
https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.3/html/install/installing

And then the changes that I added, is that good?

Thanks for your help.

drkho:
👍🏼 Sounds like a good place to mention in. Thanks.

PR #2563, opened 2021-08-10
Update observability_enable.adoc

dockerymick:
https://github.com/open-cluster-management/backlog/issues/14806

### DIFF
--- a/observability/observability_enable.adoc
+++ b/observability/observability_enable.adoc
@@ -17,15 +17,15 @@ Monitor the health of your managed clusters with the observability service (`mul
  
 - You must install {product-title}. See link:../install/install_connected.adoc#installing-while-connected-online[Installing while connected online] for more information.  
 - You must configure an object store to create a storage solution. {product-title-short} supports the following cloud providers with stable object stores:
-* https://aws.amazon.com/getting-started/hands-on/lightsail-object-storage/[Amazon S3]
-* https://www.redhat.com/en/technologies/storage/ceph[Red Hat Ceph]
+* https://aws.amazon.com/getting-started/hands-on/lightsail-object-storage/[Amazon Web Services S3 (AWS S3)]
+* https://www.redhat.com/en/technologies/storage/ceph[Red Hat Ceph (S3 compatible API)]
 * https://cloud.google.com/storage[Google Cloud Storage]
 * https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction[Azure storage]
-* https://www.redhat.com/en/technologies/storage/ceph[Red Hat OpenShift Data Foundation]
+//* https://www.redhat.com/en/technologies/storage/ceph[Red Hat OpenShift Data Foundation (S3 compatible API)]
 * https://www.ibm.com/docs/en/baw/20.x?topic=storage-preparing-cloud-public-roks[Red Hat OpenShift on IBM (ROKS)]
-* https://www.ibm.com/docs/en/spectrum-scale/5.0.5?topic=overview-s3-api[IBM Spectrum Scale]
+//* https://www.ibm.com/docs/en/spectrum-scale/5.0.5?topic=overview-s3-api[IBM Spectrum Scale]
 +
-*Important*: When you configure your object store, ensure that you meet the encryption requirements necessary when sensitive data is persisted. For a complete list of the supported object stores, see https://thanos.io/tip/thanos/storage.md/#object-storage[Thanos documentation].
+*Important*: When you configure your object store, ensure that you meet the encryption requirements necessary when sensitive data is persisted. For more information on Thanos supported object stores, see https://thanos.io/tip/thanos/storage.md/#object-storage[Thanos documentation].
 
 [#enabling-observability]
 == Enabling observability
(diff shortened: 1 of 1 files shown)


observability/observability_enable.adoc
```diff
@@ -16,19 +16,17 @@ Monitor the health of your managed clusters with the observability service (`mul
...
-- You must configure an object store to create a storage solution. {product-title-short} only supports cloud providers with stable object stores, such as Amazon S3 (or other S3 compatible object stores like Ceph), Google Cloud Storage, Azure storage, Red Hat OpenShift Data Foundation, and Red Hat OpenShift on IBM (ROKS).
+- You must configure an object store to create a storage solution. {product-title-short} supports the following cloud providers with stable object stores:
+* https://aws.amazon.com/getting-started/hands-on/lightsail-object-storage/[Amazon S3]
+* https://www.redhat.com/en/technologies/storage/ceph[Red Hat Ceph]
+* https://cloud.google.com/storage[Google Cloud Storage]
+* https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction[Azure storage]
+* https://www.redhat.com/en/technologies/storage/ceph[Red Hat OpenShift Data Foundation]
+* https://www.ibm.com/docs/en/baw/20.x?topic=storage-preparing-cloud-public-roks[Red Hat OpenShift on IBM (ROKS)]
+* https://www.ibm.com/docs/en/spectrum-scale/5.0.5?topic=overview-s3-api[IBM Spectrum Scale]
 +
```
bjoydeep, line 28: A few comments :

I think @berenss wants to make sure that for all things we claim support formally, it should be tested by QE for start.

1. So question for QE team @Xiang-Yin-ACM @quchangl-github have you folks tested all of the above ? For things not yet tested, they will not make it YET.
2. For: Ceph and ODF, suggest we add word like `S3 Compatible API` - because both Ceph and ODF expose other APIs.
3. Have we tested ROKS?

Now, one of the key motivations for this is adding IBM Spectrum Scale. But this has to be tested by QE before we add this. @berenss do you agree?

observability/observability_enable.adoc
```diff
@@ -16,19 +16,17 @@ Monitor the health of your managed clusters with the observability service (`mul
...
 *Important*: When you configure your object store, ensure that you meet the encryption requirements necessary when sensitive data is persisted. For a complete list of the supported object stores, see https://thanos.io/tip/thanos/storage.md/#object-storage[Thanos documentation].
```
bjoydeep: @dockerymick can we change the wording of `For a complete list of the supported object stores` to something else. The list above is more complete than https://thanos.io/tip/thanos/storage.md/#object-storage. However that link has important information about configuration which is needed.

observability/observability_enable.adoc
clyang82: I did test on ROKS in the past. that is why propose to have it added in 2.3 document. Agree with @bjoydeep It should be tested by QE.

quchangl-github: QE has verified AWS S3, Azure, GCS, and ROKS in 2.3.
Ceph and OCS are verified in the previous version, 2.2.
ODF and Spectrum Scale are not verified.

dockerymick:
@bjoydeep @clyang82 @quchangl-github Thanks for your comments! I have hidden mentions of ODF and Spectrum Scale. May you all do another review? i have updated the issue milestone back to In Progress since QE needs to verify that we support those object stores.

clyang82:
I agree to hide Spectrum Scale since we have never tried. but for `ODF`, we support it since 2.1 or 2.2. we should not hide it to make the end user confusion. @quchangl-github Could you help to do quick verification?

quchangl-github:
@clyang82 I checked ACM doc 2.1 and 2.2, there's no information regarding support ODF. Only in 2.2, it mentions support OCS. But I'm not completely sure if ODF is totally the same as OCS, there's no ODF in OCP 4.8 operator hub, but there's still OCS. 

I'm trying to contact OCP to clear it out, and then will try to setup one ODF to verify. But ODF is new, QE needs time to verify it. Thanks.

berenss:
ODF is just a new branding. Underneath the code is still OCS

dockerymick:
Would it be best to maybe add that info in the doc, for example:

* Red Hat OpenShift Data Platform (formerly known as OpenShift Container Storage) 

Or do we want to avoid calling out the old name?

berenss:
@dockerymick - LGTM! Very many of the field teams and customers still refer to it as OCS

berenss:
correction: ODF = OpenShift Data Foundation
https://www.redhat.com/en/technologies/cloud-computing/openshift-data-foundation

quchangl-github:
I got an update from OCS/ODF team - rebranding from OCS to ODF is in 4.9 release, operator is still OCS but this will change in next release according to roadmap. 
I have verified ACM 2.3 with OCS, it's working. IMO, I prefer in 2.3 Doc like
Red Hat OpenShift Container Storage (known as Red Hat OpenShift Data Foundation), or not add ODF info here. And in 2.4 doc, we would add ODF instead of OCS in the support list.

For IBM Spectrum Scale, it's new for ACM. QE needs to take time to see up and verify it. I suggest we add it in 2.4 Doc.
@Xiang-Yin-ACM FYI.

dockerymick:
@quchangl-github Thanks for your comments and for verifying that information. I agree that we should not mention ODF to avoid pre-announcing development plans. I have a created an [issue for 2.4](https://github.com/open-cluster-management/backlog/issues/15313).

dockerymick:
@berenss thanks for your approval and @quchangl-github thanks again for confirming what are supported object stores. I will move forward with merging this PR. @quchangl-github may you review my [PR for 2.2](https://github.com/open-cluster-management/rhacm-docs/pull/2584/files)?

PR #2579, opened 2021-08-12
open-cluster-management/backlog#11111 rem note cred upd

cadawson:


jnpacker:
This statement is true, so why are we removing it?

credentials/credential_ansible.adoc
```diff
@@ -39,6 +39,8 @@ To create a credential from the {product-title} console, complete the following
 . Click *Create*.
 . Review the new credential information, then click *Add*. When you add the credential, it is added to the list of credentials.
 
+Starting with {product-title-short} version 2.3, the information that you provide when you update your Ansible secret in the Ansible credential is automatically updated in your managed clusters that use that credential. The update applies to all managed clusters that use the credential for cluster lifecycle, governance, and application management automations.
```
jnpacker: So it is an Ansible token and host (url) that are stored in the Ansible credential.  As well, when you update it, the change is made automatically to all instances of the secret in use with Application, cluster or policy automations.

---
Thought I would give an explanation of what is happening behind the scene, doesn't need to be used @cadawson .
When one of the lifecycles uses Ansible, it makes a copy of the parent credential (that you see in provider credentials). This means one credential could be copied a number of times for cluster provisioning, app automation and policies.  When you update the provider credential (which is a kube secret) it updates all the copies of the credential (kube secret) that were made to allow cluster, app and policy automation to work.

cadawson: Thank you for the explanation, @jnpacker . That helps. I've updated the text a little to provide a little more detail.

credentials/credential_ansible.adoc
```diff
@@ -39,6 +39,8 @@ To create a credential from the {product-title} console, complete the following
 . Click *Create*.
 . Review the new credential information, then click *Add*. When you add the credential, it is added to the list of credentials.
 
+Starting with {product-title-short} version 2.3, the Ansible Token and host URL that you provide when you create your Ansible credential are automatically updated in your managed clusters that use that credential when you update them in the credential. The updates are copied to any automations that use that Ansible credential, including those related to cluster lifecycle, governance, and application management automations. This ensures that the automations continue to run after the credential is updated.
```
jnpacker: I still think the part about managed clusters implies the credential is on the managed cluster, which it is not.  I think it should read ". . . Ansible credential are automatically updated in your automation that use that credential when you update them in the credential."  something to that affect.

jnpacker: FYI @cadawson

cadawson: @jnpacker Thanks, I included a slight change to address your comment. See if that works better. Thanks!

jnpacker (approved):

PR #2585, opened 2021-08-16
Swope errata7

swopebe:


anthonyfister (commented):
If we say that the errata delivers "a new set of container images" does it infer that all of the container images were updated?
Would it be better to say that the release delivers "updated container images"? (not inferring which or how many container images were updated.

swopebe:
We re-used that line from previous similar releases but it can change. I don't think anyone has disputed it, yet:

This errata release delivers updated container images.

1. Without the word "set", it can mean all container images were updated...that's not true and I would find that confusing.
2. I would rather we say "a set of updated container images" or "various container updates".
3. If only one: delivers a container image update.

Let me know. Thanks!

swopebe:
Quickly asking @cadawson to chime in

dockerymick (approved):
Looks good to me !

anthonyfister (commented):
I would prefer:   "various container updates"

cadawson:
Interesting points. 
I am good with what we use now, as I think a "set" might also be used when the group as a whole is new, not necessarily each of its parts is new. I am also ok with "various container updates". 

Another alternative could be:

This errata release delivers updates to one or more of the product container images.

anthonyfister:
It's probably more of an academic conversation at this point.
When I made the comment, I thought Brian had only updated one or two containers.
I wasn't suggesting removing the word "set".  But, as the reader, I interpreted "a new set of container images" to mean updating them all.
My suggestion was to offer something sufficiently vague, which is why I liked the suggestion of "updates various containers".
But in the end, we have updated more than one containers - in fact many of them.  So it's all academic.  I'm fine with it as-is.

anthonyfister (approved):

swopebe:
@anthonyfister how our content is interpreted is absolutely worth chatting about and we are word nerds, so we appreciate the perspective. I really like what @cadawson suggested:

This errata release delivers updates to one or more of the product container images. (It covers all the scenarios)

If you are OK with that, I will use it.

anthonyfister:
Looks good to me.  Thanks @swopebe .

PR #2594, opened 2021-08-17
https://github.com/open-cluster-management/backlog/issues/13046 add p…

swopebe:
…rereq

### DIFF
--- a/install/install_connected.adoc
+++ b/install/install_connected.adoc
@@ -14,7 +14,9 @@
 
 Before you install {product-title-short}, see the following requirements:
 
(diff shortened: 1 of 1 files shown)


install/install_connected.adoc
```diff
@@ -14,7 +14,9 @@
 
 Before you install {product-title-short}, see the following requirements:
 
-* Your {ocp} must have access to the {product-title-short} operator in the OperatorHub catalog from the console.
+* Your {ocp} must have access to the {product-title-short} operator in the OperatorHub catalog from the console.
```
swopebe, line 17: @drkho can you scan these prereqs, though I would like to understand more what the user does with the link in line 19.

PR #2597, opened 2021-08-18
Removing nginx, should only be pod policy

dockerymick:
https://github.com/open-cluster-management/backlog/issues/13684


swopebe (approved):
couple quick changes.

dockerymick:
Thanks for the review Brandi! @swopebe

PR #2644, opened 2021-09-02
Updating OCS to RH OpenShift Data Foundation

dockerymick:
https://github.com/open-cluster-management/backlog/issues/15313


dockerymick:
Thanks for the review @clyang82 !

PR #2658, opened 2021-09-09
Add Placement info in policy overview

dockerymick:
https://github.com/open-cluster-management/backlog/issues/15131

### DIFF
--- a/governance/policy_example.adoc
+++ b/governance/policy_example.adoc
@@ -8,11 +8,11 @@ For more information about CRDs, see https://kubernetes.io/docs/tasks/access-kub
 Each {product-title} policy can have at least one or more templates.
 For more details about the policy elements, view the following _Policy YAML table_ section on this page.
 
-The policy requires a _PlacementRule_ that defines the clusters that the policy document is applied to, and a _PlacementBinding_ that binds the {product-title} policy to the placement rule. For more on how to define a `PlacementRule`, see link:../applications/app_model.adoc#placement-rules[Placement rules] in the Application lifecycle documentation.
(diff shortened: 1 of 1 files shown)


governance/policy_example.adoc
```diff
@@ -53,30 +52,29 @@ spec:
   disabled:
 
 ---
-apiVersion: apps.open-cluster-management.io/v1
+apiVersion: policy.open-cluster-management.io/v1
```
gparvin: I don't think we should edit the existing yaml.  The `Placement` support is a new, additional way to use the policy so I was thinking we would have a new section that includes the sample in the issue.

For example, if you undo the yaml edits you currently have here, and then create a new section at the bottom of this file:

### Placement Sample

The sample `PlacementBinding` and `Placement` below could be combined with the policy above to deploy the policy using the Cluster `Placement` API instead of the `PlacementRule` API.

```
---
apiVersion: policy.open-cluster-management.io/v1
kind: PlacementBinding
metadata:
  name: binding-policy-role
placementRef:
  name: placement-policy-role
  kind: Placement
  apiGroup: cluster.open-cluster-management.io
subjects:
- name: policy-role
  kind: Policy
  apiGroup: policy.open-cluster-management.io
---
apiVersion: cluster.open-cluster-management.io/v1alpha1
kind: Placement
metadata:
  name: placement-policy-role
spec:
  predicates:
  - requiredClusterSelector:
      labelSelector:
        matchExpressions:
          - {key: environment, operator: In, values: ["dev"]}
```

gparvin (commented):
See comment about making the sample an extra sample -- not a change to the existing sample.

PR #2661, opened 2021-09-09
open-cluster-management/backlog#15628 Add 2.2.8 errata

cadawson:


cadawson:
Hi @anthonyfister. The four errata items that were marked are drafted in this PR. 
Question: Was the issue 14821 fixed, or is it being addressed as a known issue? It looked like a known issue, but there was no summary. If it is a known issue, we don't mention known issue additions in the errata. We focus on items where changes to the code fixed an issue. We can remove that one, if it is a known issue. Let me know if you have any updates. Thanks!

release_notes/errata.adoc
```diff
@@ -9,6 +9,18 @@ By default, errata updates are automatically applied when released. See link:../
 
 - For reference, link:https://access.redhat.com/errata/#/[Errata] links and GitHub numbers might be added to the content and used internally. Links that require access might not be available for the user.
 
+== Errata 2.2.8
+
+View a summarized list of {product-title} Errata 2.2.8 updates:
+
+* Fixes an issue that caused an application to be deleted when it was reconciled. The deletion was caused by a bad application manifest, such as incorrect kustomization, in a subscribed Git repository. With this fix, an error message is displayed in the application topology UI when there is a bad manifest, but the deployed application is not deleted. (https://bugzilla.redhat.com/show_bug.cgi?id=1972947[Bugzilla 1972947])
```
dockerymick: Are there other situations where the application is "bad" , maybe we can say misconfigured? I suggest the following change (should kustomization start with a c?): 

The deletion was caused by an incorrect kustomization for the application manifest, in a subscribed Git repository. With this fix, an error message is displayed in the application topology user interface when there is a misconfigured manifest, but the deployed application is not deleted.

release_notes/errata.adoc
```diff
@@ -9,6 +9,18 @@ By default, errata updates are automatically applied when released. See link:../
...
+
+* Decreases the number of unnecessary secrets that are created so the `cert manager` can start with less memory consumption. (GitHub 13127)
```
dockerymick: comma after created

release_notes/errata.adoc
```diff
@@ -9,6 +9,18 @@ By default, errata updates are automatically applied when released. See link:../
...
+
+* Fixes an issue that caused some {ocp} version 4.8 observability metrics to not display in the Grafana console. (GitHub 14821)
```
dockerymick: I have an active [PR](https://github.com/open-cluster-management/rhacm-docs/pull/2667) to document this specific known issue. If there is a fix, it makes sense for me to not doc the known issue, unless there are still some metrics that are not displayed.

dockerymick (changes requested):

release_notes/errata.adoc
dockerymick: @clyang82 thoughts? ^

dockerymick: Honestly, the statement is only a few sentences. I think that it is fine to keep the known issue as is, plus it is approved.

release_notes/errata.adoc
cadawson: Thanks. Made some updates to this from your comments, for clarity.

release_notes/errata.adoc
cadawson: I'm trying to determine whether it was fixed or still an issue.

anthonyfister (approved):
Looks good.  Thanks.

release_notes/errata.adoc
anthonyfister: It was not fixed in 2.2.z.   It was too complex and would have required mucho refactoring.
We decided to fix forward (in the next release) and document as a known issue for 2.2.z.  It's hard to keep up with changing metrics each time OCP comes out with a new version.  
So... the line for GitHub 14821 should probably come out.

cadawson: Removed it.

PR #2721, opened 2021-09-25
Swope known

swopebe:


almen-ng:
I noticed that this PR is for 2.3 (which another doc issue was created already - https://github.com/open-cluster-management/backlog/issues/16159). This doc issue should be for 2.1.

PR #2724, opened 2021-09-27
Update docs on enabling single node openshift feature

jamestalton:
Signed-off-by: James Talton <jtalton@redhat.com>

### DIFF
--- a/clusters/single_node_openshift.adoc
+++ b/clusters/single_node_openshift.adoc
@@ -1,5 +1,5 @@
 [#single-node-openshift]
-= Creating a Single-Node OpenShift cluster (Development Preview)
+= Creating a Single-Node OpenShift cluster (Technology Preview)
 
 You can create a Single-Node OpenShift (SNO) cluster that contains only a single node. In a standard cluster, there are master nodes and separate worker nodes. This configuration can be helpful when you have many clusters that need to minimize their resource footprint. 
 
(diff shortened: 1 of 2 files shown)


cadawson:
Thanks @jamestalton . I made a couple of changes/additions to this:
1. I think it is planned to be Technology Preview in 2.4, rather than Dev Preview. 
2. Is this still going to be disabled, by default, even though it is Tech Preview? The way I am interpreting it is that there is no feature gate, but you still have to enable the feature.
3. I bumped the version of ACM to 2.4. Since 2.3 was dev preview, my guess is that we will need 2.4, or later. 
4. I bumped the OpenShift version up to 4.9, since this is 2.4. I left it as 4.8 for the release image, as I think you might be able to create a 4.8 cluster as a SNO cluster with 4.8. That is what we are using for the Dev Preview version on 2.3.
Let me know if I am off base on any of these. We can change them, as needed.

cadawson (changes requested):
Looks good! Added a few comments to the PR.

cadawson:
@jamestalton : I had some questions/comments. Can you take a look at the earlier comment? Thanks!

jamestalton:
OCP 4.8+ for SNO support

I think there are two paths for SNO
- CIM/AI
- HIVE

I believe the SNO we are talking about here is the HIVE path through cluster create without CIM.
It is disable by default and must be enabled using the console configmap.

cadawson (approved):
Looks good. Thanks!

cadawson:
Merging

PR #2749, opened 2021-10-06
Improve the IAM Policy Documentation

mprahl:
This document the IAMPolicy spec.clusterRole field. This was added in 2.3 in the following commit but was not documented:
https://github.com/open-cluster-management/iam-policy-controller/commit/23b2c7d876ec16bafcb6f65b4b80bd4eb925faa1

This also removes the IamPolicy namespaceSelector field. This field is here for legacy reasons but it does not affect the behavior of the IAM policy controller, as the controller only handles cluster roles (i.e. not namespaced roles). Removing the documentation for this field makes things less confusing to the user.

This also fixes a couple small grammatical errors.

### DIFF
--- a/governance/create_iam_policy.adoc
+++ b/governance/create_iam_policy.adoc
@@ -35,6 +35,7 @@ metadata:
   label:
     category: "System-Integrity"
 spec:
+  clusterRole: cluster-admin
   remediationAction: inform
   disabled: false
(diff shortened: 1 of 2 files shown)


dockerymick (approved):
@mprahl Great job with the changes! This looks good to me !

PR #2750, opened 2021-10-06
Mj pol generator 16333

dockerymick:
https://github.com/open-cluster-management/backlog/issues/16333


governance/config_policy_generator.adoc
```diff
@@ -0,0 +1,81 @@
+[#config-policy-generator]
```
mprahl: The user does not need to install the Policy Generator unless they are using it on their laptop. That workflow is not officially supported. The only officially supported way of using it is through RHACM Application Lifecycle, which comes preinstalled in RHACM 2.4, so this whole file can be removed. Sorry about the misunderstanding!

mprahl:
Hey @dockerymick, could you please schedule a call with @dhaiducek and I to review this all together? I think it'd be easier to do it this way.

mprahl:
@dockerymick Dale and I posted another commit with our updates. Could you please take a look?

mprahl:
@dockerymick, Jaya had requested I add a section describing how to generate a policy to install an Operator. I just added this in the last commit.

dockerymick:
@mprahl Thanks for the update and your patience. I am taking a look at this PR after the lunch hour

dockerymick:
@mprahl Thanks again for your patience and edits. Can you take another look at the PR when you have an opportunity. A change that I want to bring to your attention is the way we used inline links in this doc. There were more where links were disruptive as I read. Therefore, I moved some inline links to the end of a section. Let me know your thoughts, or we can meet sometime this week

mprahl (approved):
This looks great! Thanks!

PR #2771, opened 2021-10-13
open-cluster-management/backlog#14603 add volsync

cadawson:


services/volsync.adoc
```diff
@@ -0,0 +1,214 @@
...
+To enable VolSync on two clusters in your environment, you must install it on both the source and the target managed clusters. Complete the following steps to leverage the policy function in {product-title-short} to install VolSync:
+
+. Deploy the required VolSync resources to the clusters.
+
+.. In the {product-title-short} console navigation, select *Applications*. 
+
+.. Select *Create application* to start a new application. The application that you create uses an existing policy to install the VolSync resources.
+
+.. Complete the required information for your application.
+
```
tesshuflower: When I try this, I have to pick `Subscription` as the type in the dropdown before I can continue

services/volsync.adoc
```diff
@@ -0,0 +1,214 @@
...
+* At least two configured clusters that are managed by the same {product-title-short} hub cluster.
+
+[#volsync-install-clusters]
+== Installing VolSync on the managed clusters
+
+To enable VolSync on two clusters in your environment, you must install it on both the source and the target managed clusters. Complete the following steps to leverage the policy function in {product-title-short} to install VolSync:
+
+. Deploy the required VolSync resources to the clusters.
+
```
tesshuflower: I've discussed a bit with @jnpacker and I think to install volsync, rather than being a guided experience via deploying an application, the user would instead need to create the policy using the sample one here:

https://github.com/open-cluster-management/policy-collection/blob/main/community/CM-Configuration-Management/policy-persistent-data-management.yaml

If they create this policy on the hub cluster it should then create a volsync subscription on the hub which then targets managed clusters and deploys volsync to them

services/volsync.adoc
```diff
@@ -0,0 +1,214 @@
+[#volsync]
+= Replicating persistent volumes with VolSync (Technology Preview)
+
+VolSync is a Kubernetes operator that enables asynchronous replication of persistent volumes within a cluster, or across clusters with storage types that are not otherwise compatible for replication. It uses the Container Storage Interface (CSI) to overcome the compatibility limitation. After deploying the VolSync operator in your {ocp} environment where you have your {product-title} hub cluster, you can leverage it to create and maintain backup versions of your {product-title-short} managed clusters.
```
JohnStrunk: - I don't think we're advocating always installing on the hub, just on clusters that are source or destination.
- Original phrasing makes it sound like we're backing up the cluster instead of just the data

```suggestion
VolSync is a Kubernetes operator that enables asynchronous replication of persistent volumes within a cluster, or across clusters with storage types that are not otherwise compatible for replication. It uses the Container Storage Interface (CSI) to overcome the compatibility limitation. After deploying the VolSync operator in your {ocp} environment, you can leverage it to create and maintain backup versions of your persistent data.
```

services/volsync.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+**Note:** VolSync does not meet the requirements of the FIPS standard. 
+
+There are three methods that you can use to replicate when you use VolSync, which depend on the number of synchronization locations that you have. The Rsync method is used for this example. For information about the other methods and more information about Rsync, see https://volsync.readthedocs.io/en/latest/usage/index.html[Usage] in the VolSync documentation.  
+
+* Rsync replication is a one-to-one replication of persistent volumes, and is likely to be the most commonly used. This is used for typical backups at a remote site or mirroring an environment.
```
JohnStrunk:
```suggestion
* Rsync replication is a one-to-one replication of persistent volumes, and is likely to be the most commonly used. This is used for replicating data to a remote site. 
```
(removing "backup" wording to avoid confusion w/ the restic-based mover which is specifically for backup)

services/volsync.adoc
```diff
@@ -0,0 +1,214 @@
...
+kubectl annotate sc/gp2 storageclass.kubernetes.io/is-default-class="false" --overwrite
+----
+
+... Redirect the service to CSI and start the service: 
++
+----
+kubectl annotate sc/gp2-csi storageclass.kubernetes.io/is-default-class="true" --overwrite
+----
+
+... Install a `VolumeSnapshotClass` by running the following command:
```
JohnStrunk: I believe OCP 4.9 has started shipping w/ a default VolumeSnapshotClass that is compatible w/ the ebs csi driver, so these instructions may not be necessary depending on the OCP version.

services/volsync.adoc
```diff
@@ -0,0 +1,214 @@
...
+
+... Make sure you select the option to *Deploy to all online clusters and local-cluster*. This ensures that it is included on all of your managed clusters. 
+
+.. Select `Save` to save your application. Within a few minutes, the application will start deploying to all of the clusters that are managed by your {product-title-short} hub cluster. 
+
+. Change your default CSI configuration to your VolSync CSI instance. 
++
+* Your steps might resemble the following procedure for an Amazon Web Services environment:
+
+... Stop the current service:
```
JohnStrunk: The stop/start terminology doesn't quite match what's actually going on...
We're changing which StorageClass is marked as the default one (i.e., if the user doesn't specify a specific StorageClass in their PVC, it will be assigned to the SC that is currently marked as default.)

Same goes for the VolumeSnapshotClass explanation below

services/volsync.adoc
```diff
@@ -0,0 +1,214 @@
...
+  sourcePVC: <persistent_volume_claim>
+  trigger:
+    schedule: "*/5 * * * *"
+  rsync:
+    sshKeys: <volsync-rsync-destination-src-database-destination>
+    address: <source.host.com>
+    copyMethod: Clone
+----
++
+Replace `source_name` with the name of your source volume.
```
JohnStrunk: This can be an arbitrary name, unrelated to the volume (PVC) name

services/volsync.adoc
```diff
@@ -0,0 +1,214 @@
...
++
+Replace `source_namespace` with the name of the namespace where your source is located.
++
+Replace `persistent_volume_claim` with the name of your source claim.
++
+Replace `source.host.com` with the host address of your source cluster.
```
JohnStrunk: The `sshKeys` and `address` fields need to be taken from the ReplicationDestination.

I suggest re-arranging so that the user creates the destination before the source.
Upstream rsync mover usage: https://volsync.readthedocs.io/en/latest/usage/rsync/index.html

Once the ReplicationDestination is created, the operator will populate its `.status.rsync.address` field, and that info needs to be copied here.

The ReplicationDestination will also write `.status.rsync.sshKeys` which is the name of a Secret that the operator has created. This Secret needs to be exported from the destination cluster and imported into the source cluster in the same namespace as the ReplicationSource object. The name of this imported Secret should then be provided in the ReplicationSource's `.spec.rsync.sshKeys` field.

services/volsync.adoc
```diff
@@ -0,0 +1,214 @@
...
+Replace `destination_namespace` with the name of the namespace where your destination is located.
+
+Your can now set up the synchronization method of the persistent volume.
+
+[#volsync-start]
+== Starting your synchronization
+
+You have a few options to select from when determining how you start your replications: always running, on a schedule, or manually. Implement the method from the following list that best meets your needs:  
+
+* *Always:* Runs a constant replication. When a change is detected on the source persistent volume, a replication is automatically started. Your setting for this replication option might resemble the following content:
```
JohnStrunk: I suggest we leave out the "always" docs. We really want to guide users toward scheduled replication for most use cases.

services/volsync.adoc
cadawson: Changed the procedure.

services/volsync.adoc
cadawson: Discussed with Tesshu, and updated this procedure.

services/volsync.adoc
cadawson: Thank you for the clarification. I rearranged this procedure.

services/volsync.adoc
```diff
@@ -0,0 +1,202 @@
+[#volsync]
+= Replicating persistent volumes with VolSync (Technology Preview)
+
+VolSync is a Kubernetes operator that enables asynchronous replication of persistent volumes within a cluster, or across clusters with storage types that are not otherwise compatible for replication. It uses the Container Storage Interface (CSI) to overcome the compatibility limitation. After deploying the VolSync operator in your {ocp} environment, you can leverage it to create and maintain backup versions of your persistent data.
```
JohnStrunk:
```suggestion
VolSync is a Kubernetes operator that enables asynchronous replication of persistent volumes within a cluster, or across clusters with storage types that are not otherwise compatible for replication. It uses the Container Storage Interface (CSI) to overcome the compatibility limitation. After deploying the VolSync operator in your {ocp} environment, you can leverage it to create and maintain copies of your persistent data.
```

services/volsync.adoc
```diff
@@ -0,0 +1,202 @@
...
+      vendor: OpenShift
+----
+
+.. Apply the policy while you are logged in to your hub cluster by entering the following command:
++
+----
+oc -n <namespace> apply -f ./policy-persistent-data-management.yaml
+----
++
+Replace `namespace` with the namespace of your hub cluster nad managed clusters.
```
JohnStrunk:
```suggestion
Replace `namespace` with the namespace of your hub cluster and managed clusters.
```

services/volsync.adoc
```diff
@@ -0,0 +1,202 @@
...
+
+. If your default `StorageClass` is not a CSI configuration, change your default CSI configuration to your VolSync CSI instance.
```
JohnStrunk:
```suggestion
. Your default `StorageClass` should refer to a CSI-based storage driver, so it may be necessary to change the default. 
```

services/volsync.adoc
```diff
@@ -0,0 +1,202 @@
...
+* Your steps might resemble the following procedure for an Amazon Web Services environment:
+
+.. Change your default `StorageClass` by entering the following commands: 
++
+----
+kubectl annotate sc/gp2 storageclass.kubernetes.io/is-default-class="false" --overwrite
+kubectl annotate sc/gp2-csi storageclass.kubernetes.io/is-default-class="true" --overwrite
+----
+
+.. Install a `VolumeSnapshotClass` by running the following command:
```
JohnStrunk:
```suggestion
.. If a corresponding `VolumeSnapshotClass` does not already exist, one may be created by running the following command:
```

services/volsync.adoc
```diff
@@ -0,0 +1,202 @@
...
+---
+apiVersion: volsync/v1alpha1
+kind: ReplicationDestination
+metadata:
+  name: <destination_name>
+  namespace: <destination_namespace>
+spec:
+  rsync:
+    copyMethod: Snapshot
+    capacity: 10Gi
```
JohnStrunk: Note: This capacity should match the capacity of the PVC that is being replicated

services/volsync.adoc
```diff
@@ -0,0 +1,202 @@
...
+[source,yaml]
+----
+...
+status:
+  rsync:
+    address: 10.01.101.001
+    sshKeys: volsync-rsync-dest-src-test
+----
+
+... Copy the name of the secret that is provided as the value of `.status.rsync.sshKeys`. You will have to enter it when you configure the source.
```
JohnStrunk: In addition to the name of the Secret, the Secret itself must be copied to the source cluster.

services/volsync.adoc
tesshuflower: I think namespace really just needs to be a namespace on the hub cluster itself - the policy will be created in this namespace on the hub.  We should probably remove mention of managed clusters on this line.

services/volsync.adoc
tesshuflower: Wondering if we should be clear here where we should be checking for these storageclass/volumesnapshotclass requirements - This should be done on the clusters where VolSync is intended to be used. So for example if the customer has deployed (via the policy) to all their managed clusters with the intention of using Volsync there (as a replicationsource or replicationdestination) then they need to do these steps on those managed clusters.

They would potentially also need to do this on the hub cluster if they intend to use volsync on the hub cluster itself.

services/volsync.adoc
```diff
@@ -0,0 +1,204 @@
...
+----
+kubectl create -f - << SNAPCLASS
+---
+apiVersion: snapshot.storage.k8s.io/v1beta1
+kind: VolumeSnapshotClass
+metadata:
+  name: gp2-csi
+driver: ebs.csi.aws.com
+deletionPolicy: Delete
+SNAPCLASS
```
JohnStrunk: Looks like this snapclass is for AWS, not GCP. Here's one for GCP:

```suggestion
kubectl create -f - << SNAPCLASS
---
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshotClass
metadata:
  name: standard-csi
driver: pd.csi.storage.gke.io
deletionPolicy: Delete
SNAPCLASS
```

JohnStrunk (commented):
Sorry... caught one more issue. :disappointed:

cadawson:
Good catch, @JohnStrunk! Merged the change. Thanks!

cadawson:
Are we ok with this one @tesshuflower and @JohnStrunk? We can always make changes during one of our weekly refreshes after release. if we need to. It is not locked up. Thanks!

JohnStrunk:
Works for me.

tesshuflower:
Same here. Thanks Christopher!

JohnStrunk (approved):

tesshuflower (approved):

cadawson:
Thanks for all of your help with this @tesshuflower and @JohnStrunk!

PR #2788, opened 2021-10-15
open-cluster-management/backlog#15799 Auto import

cadawson:


itdove:
Yes we should keep the manual method as in some cases the communication in the direction hub-mc doesn't work but well mc-hub and in that case the auto-import can not work but the manual import will.

cadawson:
Does this look ok @itdove ?

clusters/import_cli.adoc
```diff
@@ -50,7 +50,45 @@ oc new-project ${CLUSTER_NAME}
...
+metadata:
+  name: auto-import-secret
+  namespace: <cluster_name>
+stringData:
+  autoImportRetry: 5
+  token: <Token to access the cluster>
+  server: <cluster_api_url>
+type: Opaque
+----
+
```
itdove, line 84: Wonder if for both way (kubeconfig and token/server pair) we don't have to detail the steps like it is done for the managed-cluster.yaml.
a. Edit this template.
b. save it
c. Apply it `oc apply -f....`
d....

cadawson:
Hi @itdove. I think I made the change that you were talking about to merge the two files. See if it works better. 

Does adding the klusterlet add-on still have to be done with the auto import? If so, I can have them skip to that step. Otherwise, we can end the procedure there. Thanks!

clusters/import_cli.adoc
```diff
@@ -65,25 +65,21 @@ metadata:
...
+  # that has the current context set to the cluster to import:
+  kubeconfig: |- <kubeconfig_file>
+  # If you are using the token/server pair, add the following two values:
+  token: <Token to access the cluster>
+  server: <cluster_api_url>
 type: Opaque
 ----
 
-** To add the server/token pair of the cluster that you are importing, add the following YAML content: 
+.. Save the file as `managed-cluster.yaml`.
```
itdove: save the file as `auto-import-secret.yaml`

clusters/import_cli.adoc
```diff
@@ -65,25 +65,21 @@ metadata:
...
-kind: Secret
-metadata:
-  name: auto-import-secret
-  namespace: <cluster_name>
-stringData:
-  autoImportRetry: 5
-  token: <Token to access the cluster>
-  server: <cluster_api_url>
-type: Opaque
+oc apply -f managed-cluster.yaml
```
itdove: oc apply -f auto-import-secret.yaml

itdove:
yes addon still have to be done with auto-import

itdove (approved):

PR #2791, opened 2021-10-18
FIPS doc

dockerymick:
https://github.com/open-cluster-management/backlog/issues/14818


release_notes/fips_readiness.adoc
```diff
@@ -0,0 +1,20 @@
...
+
+Read the following limitations with {product-title-short} and FIPS.
+
+* The following Technology preview components are not FIPS ready:
++
+** Hub cluster backup and restore
+** {assist-install}
+** Submariner
+** VolSync
+//what is storage layer? how about, "..components must be encrypted when you configure the storage?
```
gparvin: Maybe "storage provider" or "provided storage" makes more sense?  ACM just uses storage the customer brings to the product, ACM doesn't manage the storage, ACM only uses it.  So the storage that is "brought" to ACM needs to be encrypted...

gparvin (approved):
I approve the changes.  I added a comment about "storage layer" in case you decide you want to reword that.  Thanks!

release_notes/fips_readiness.adoc
dockerymick: Thanks for clarifying and approving!

PR #2792, opened 2021-10-18
Removing troubleshoot topic for observability

dockerymick:
https://github.com/open-cluster-management/backlog/issues/16081#issuecomment-943450834


cadawson (approved):
Looks like that should work! Thanks!

PR #2805, opened 2021-10-20
add import by kubeconfig--rhacm change, issue or info pending

ldpliu:


swopebe:
I was not aware of these changes but I need more information and an issue to point to. This change is going into the RHACM dc in an area I don't normally cover. Normally we have a doc issue, so I would like to know more about these changes @ldpliu -- the doc issue listed is for the MCE docs. Currently I am working on RHACM docs for GA next week so I would like to understand this change. @nathanweatherly FYI

nathanweatherly:
My understanding is this:

We have a PR for how to import clusters via the CLI in the MCE doc. At one point, @ldpliu mentioned we could also doc how to import a cluster w/ the kubeconfig (see comment [here](https://github.com/open-cluster-management/mce-docs/pull/11#discussion_r735239774))

I suggest we don't focus on adding this import strategy to RHACM doc since (1) doc is already deep into RHACM doc review and (2) the import doc will only exist in MCE docs after RHACM 2.4.

So @ldpliu how would you feel about the following:

@swopebe hasn't reviewed the MCE PR yet (see her note above concerning RHACM docs). If you'd like, can you please copy your draft text into the MCE import git issue ([here](https://github.com/open-cluster-management/backlog/issues/16088)) similar to what you did for creating a cluster ([here](https://github.com/open-cluster-management/backlog/issues/16088#issuecomment-947411667)). I can then add your new content to the `import-cli` draft (similar to what I did with your create cluster content) and will also test it. 

With that being said, I'm copying this message into the git issue, as requested by @swopebe 

@ldpliu can you please respond in the issue? Thanks!

ldpliu:
Thanks @nathanweatherly , you are right, I agree with you to add the content to MCE doc and I will append the draft to https://github.com/open-cluster-management/backlog/issues/16088

swopebe:
closing as it was for MCE

PR #2808, opened 2021-10-21
open-cluster-management/backlog#16043--rework and whats new for this …

swopebe:
…issue


applications/app_console.adoc
```diff
@@ -10,10 +10,11 @@ See the following application console capabilities:
 * Visualize deployed applications across your clusters, including any associated resource repositories, subscriptions, and placement configurations.
 
 * Create and edit applications, and subscribe resources. By default, the hub cluster can manage itself and is named the `local cluster`. You can choose to deploy application resources to this local cluster, though deploying applications on the local cluster is not best practice. 
+//can we still say this ^^^^^ line 12 at the end
 
-* From the _Overview_, click an application to view topology, which encompasses application resources, including resource repositories, subscriptions, placement rules, and deployed resources, including any optional pre deployment and post deployment hooks using Ansible Tower tasks (for Git repositories).
+* From the _Overview_, click on an application name to view details and topology, which encompasses application resources, including resource repositories, subscriptions, placements, and deployed resources, including any optional pre deployment and post deployment hooks using Ansible Tower tasks (for Git repositories). Access the `Editor` tab from this same view.
```
chenz4027: Subscription deployment still uses placement rules so we can keep that. No need to use placement here.

chenz4027: Or we can keep both if it's general to all related resources.

chenz4027 (changes requested):

PR #2810, opened 2021-10-21
open-cluster-management/backlog#16593 no fips pwr z

cadawson:


release_notes/fips_readiness.adoc
```diff
@@ -8,13 +8,15 @@ FIPS readiness has been completed for {product-title}. {product-title-short} use
 
 Read the following limitations with {product-title-short} and FIPS.
 
+* {ocp} does not support FIPS on the IBM Power (`pp264le`) and IBM Z (`s390x`) architectures.
```
rahularora19911: I guess it should be 'ppc64le' instead of 'pp264le'. @jaywcarman  can confirm this.
rest looks fine.
Thanks!

rahularora19911 (changes requested):

release_notes/fips_readiness.adoc
jaywcarman: @cadawson - confirming that it should be `ppc64le`. (thanks @rahularora19911 !)

Also agree that it the change looks good to go once the typo is fixed.

cadawson: Updated that. Thanks!

PR #2827, opened 2021-10-26
open-cluster-management/backlog#12254 Add sno z

cadawson:


cadawson:
Hi @CrystalChun and @rahularora19911 . I added some sizing and performance information for SNO clusters and IBM Z. If you will review the piece that you are most familiar with, I would appreciate it.

install/plan_capacity.adoc
```diff
@@ -94,11 +101,18 @@ See the following https://docs.openshift.com/container-platform/4.6/installing/i
...
+* Node count: 3							
 * Instance size:
-** Memory: 16 GB ??
-** Storage size: 120 GB ??
-** vCPU: 16 ??
+** Memory: 16 GB 
+** Storage size: 100 GB
+** vCPU: 10
++
+IBM Z systems provide the ability to configure simultaneous multithreading (SMT), which extends the number of vCPUs that can run on each core. If you configured SMT, your SMT level determines how you satisfy the 4 vCPU requirement. The most common configurations are:
```
rahularora19911: Hi @cadawson,
Can we replace this SMT stuff with the one below:

```
IBM Z systems provide the ability to configure simultaneous multithreading (SMT), which extends the number of vCPUs that can run on each core. If you configured SMT, One physical core (IFL) provides two logical cores (threads). The hypervisor can provide two or more vCPUs.

One vCPU is equivalent to one physical core when simultaneous multithreading (SMT), or hyperthreading, is not enabled. When enabled, use the following formula to calculate the corresponding ratio: (threads per core × cores) × sockets = vCPUs.

For more information about SMT, see https://www.ibm.com/docs/en/aix/7.2?topic=concepts-simultaneous-multithreading[Simultaneous multithreading].
```

I am just trying to be in sync with the OCP documentation here. 
https://access.redhat.com/documentation/en-us/openshift_container_platform/4.9/html-single/installing/index#minimum-resource-requirements_installing-ibm-z

Please let me know. Thanks!

rahularora19911 (changes requested):

install/perform_scale.adoc
```diff
@@ -45,6 +45,20 @@ The following table shows the configuration information for the clusters on the
...
 
+[#perf-1k]
+== Provisioning 1000 Single Node OpenShift clusters
+
+The provisioning of 1000 Single Node OpenShift (SNO) bare metal clusters with a standard 3-node {ocp} hub cluster completed in just over four hours. The provisioning of the clusters took approximately 3 hours and 45 minutes. Approximately 15 minutes after that, all of the clusters were under {product-title-short} management. 
+
+Peak usage of the control plane nodes and worker nodes on the hub cluster during the provision is shown in the following table:
+
+|===
+|Node | Number | CPU: utilization | CPU: Logical cores | Memory: Utilization | Memory: GB consumed
```
CrystalChun: I would recommend including the actual CPU and Memory values. Do you have these numbers Chris?

CrystalChun: Also I don't think we should include the `Number` column. In this context, it doesn't mean much since we're only looking at the peak utilization of any node type.

install/plan_capacity.adoc
cadawson: Yep. I updated it.

install/perform_scale.adoc
cadawson: @CrystalChun: After our discussion the other day, I think we decided to remove this until we have more current information. Let me know if that is not correct.

CrystalChun: Yep, thanks Chris!

rahularora19911 (approved):
LGTM, Thanks!

PR #2832, opened 2021-10-27
Updating the default for the interval parameter

dockerymick:
https://github.com/open-cluster-management/backlog/issues/16354


dockerymick:
@marcolan018 FYI, since this is a small change I am going to move forward an merge this

PR #2833, opened 2021-10-27
Updating interval parameter for observability API

dockerymick:
https://github.com/open-cluster-management/backlog/issues/16354


dockerymick:
@marcolan018 FYI, this is a small change so moving forward with merging this

PR #2848, opened 2021-10-28
Gatekeeper image references

dockerymick:
https://github.com/open-cluster-management/backlog/issues/16320

### DIFF
--- a/governance/create_gatekeeper.adoc
+++ b/governance/create_gatekeeper.adoc
@@ -45,9 +45,6 @@ kind: Gatekeeper
 metadata:
   name: gatekeeper
 spec:
-  # Add fields here
-  image:
-    imagePullPolicy: Always
   audit:
     replicas: 1
     logLevel: DEBUG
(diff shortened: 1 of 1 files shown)


gparvin (commented):
I was expecting more changes in the section: https://github.com/open-cluster-management/rhacm-docs/blob/c434855f3191a28c6aed5bc4222a99dd5354da57/governance/create_gatekeeper.adoc#upgrading-gatekeeper-and-the-gatekeeper-operator

Does that section go away now? (question for @thomasmckay )
It seems like the upgrading section either needs to be deleted or reworked if there isn't an image reference anymore.

governance/create_gatekeeper.adoc
```diff
@@ -47,7 +47,6 @@ metadata:
 spec:
   # Add fields here
   image:
-    image: docker.io/openpolicyagent/gatekeeper:v3.2.2
     imagePullPolicy: Always
```
thomasmckay: lines 49-51 should be removed (the entire `image:` section

thomasmckay:
Yes, the manual step in the docs should be removed. The docs should reflect "normal" operator upgrade, where the upgrade plan is either automatic or manually approved.

PR #2862, opened 2021-11-02
Remove mention of deleting ClusterPool namespace

KevinFCormier:
Signed-off-by: Kevin Cormier <kcormier@redhat.com>

As part of https://github.com/open-cluster-management/backlog/issues/16764, the functionality to delete a ClusterPool namespace under certain conditions is completely removed.

### DIFF
--- a/clusters/cluster_pool_destroy.adoc
+++ b/clusters/cluster_pool_destroy.adoc
@@ -20,6 +20,8 @@ To destroy a cluster pool, complete the following steps:
 . Enter the name of the cluster pool to confirm.
 
 . Select *Destroy* to remove the cluster pool. Any unclaimed clusters in the cluster pool are destroyed. It might take some time for all of the resources to be deleted, and the cluster pool remains visible in the console until all of the resources are deleted. 
++
+The namespace that contained the ClusterPool will not be deleted. Deleting the namespace will destroy any clusters that have been claimed from the ClusterPool, since the ClusterClaim resources for these clusters are created in the same namespace.
 
 *Tip:* You can destroy multiple cluster pools with one action by selecting the box for each of the the cluster pools and using the _Actions_ menu to destroy the selected cluster pools.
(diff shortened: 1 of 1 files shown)


cadawson (approved):
Looks good to me. Would we need to tell them to delete the namespace manually, or is that not necessary?

KevinFCormier:
We could say something like "The namespace that contained the ClusterPool will not be deleted. A ClusterPool does not require exclusive use of a namespace, so if you wish to delete the namespace, be aware that it could contain other resources, including any ClusterClaims created from the ClusterPool. Deleting the namespace will also delete any such ClusterClaims, destroying clusters that have been claimed from the pool." 

But then again, I think it's the norm that we don't automatically delete namespaces. There are only rare cases we can, like the ManagedCluster namespaces.

@dtthuynh what do you think?

dtthuynh:
I think what you wrote is good - It's true, it's not _really_ necessary if we assume we rarely delete namespaces IMO but I think it's worth noting (more like a warning) that by choosing to delete that namespace, you might delete some cluster claims and inadvertently delete some claimed clusters.

cadawson:
Rewording possibility:

The namespace that contained the ClusterPool is not be deleted. A ClusterPool does not require exclusive use of a namespace, so deleting the namespace might delete other resources, including any ClusterClaims that were created by the ClusterPool. Deleting the namespace will also destroy clusters that have been claimed from the ClusterPool.

KevinFCormier:
Hey @cadawson, I like that your version is shorter, but there are a couple issues with it. How about simply:

> The namespace that contained the ClusterPool will not be deleted. Deleting the namespace will destroy any clusters that have been claimed from the ClusterPool, since the ClusterClaim resources for these clusters are created in the same namespace.

cadawson:
I like your suggestion @KevinFCormier! Thanks!

PR #2870, opened 2021-11-03
Add doc for policy_governance_info metric

dockerymick:
https://github.com/open-cluster-management/backlog/issues/17201


governance/policy_governance_info.adoc
```diff
@@ -0,0 +1,25 @@
+[#gov-metric]
+= Governance metric
+
+The policy framework exposes metrics that show policy distribution and compliance. Use the `policy_governance_info` metric on the hub cluster to view trends and analyze any policy failures.
+
+[#metric-overview]
+== Metric overview
+
+When you create a policy, you are creating a _root_ policy. The framework watches for root policies as well as `PlacementRules` and `PlacementBindings`, to determine where to create _propagated_ policies in order to distribute the policy to managed clusters.
+For both root and propagated policies, a metric of `0` is displayed if the policy is compliant, and `1` if it is non-compliant.
```
JustinKuli: Nitpick: I'm not sure if "displayed" is the best word. Maybe "recorded" is better?

governance/policy_governance_info.adoc
```diff
@@ -0,0 +1,25 @@
...
+- `type`: The label values are `root` or `propagated`.
+- `policy`: The name of the associated root policy.
+- `policy_namespace`: The namespace on the hub cluster where the root policy was defined.
+- `cluster_namespace`: The namespace for the cluster where the policy is distributed.
+
+These labels and values enable queries that can show us many things happening in the cluster that might be difficult to track.
+
+*Note*: If the metrics are not needed, and there are any concerns about performance or security, this feature can be disabled. Set the `DISABLE_REPORT_METRICS` environment variable to `true` in the propagator deployment.
+
+//is there a yaml example that can be displayed?
```
JustinKuli: The metric doesn't produce a YAML. If we needed an example, we could share a snippet of what it exposes at the metrics endpoint, but although it's readable, I don't think it's a format most users would be familiar with.

A screenshot from the OCP prometheus UI would be possible, and would be more representative of how a user would interact with the information. Or since that information is in a table, we could reproduce the table here without a screenshot. Any thoughts on what would be best?

governance/policy_governance_info.adoc
```diff
@@ -0,0 +1,25 @@
...
+- `policy_namespace`: The namespace on the hub cluster where the root policy was defined.
+- `cluster_namespace`: The namespace for the cluster where the policy is distributed.
+
+These labels and values enable queries that can show us many things happening in the cluster that might be difficult to track.
+
+*Note*: If the metrics are not needed, and there are any concerns about performance or security, this feature can be disabled. Set the `DISABLE_REPORT_METRICS` environment variable to `true` in the propagator deployment.
+
+//is there a yaml example that can be displayed? 
+// Do we want to reference the user to the observability doc?
+//is it appropriate or accurate to mention the policy_governanace_info metric in the observability doc?
```
JustinKuli: Since I don't see other metrics specifically mentioned there, I don't think it's necessary. Especially since we don't have a pre-built dashboard that would pull together the information in an ideal format for the user.

JustinKuli (approved):
LGTM! Thanks for hopping on a call to help review it quickly :)

dockerymick:
@JustinKuli Thanks for the live review and flexibility Justin!

PR #2896, opened 2021-11-08
open-cluster-management/backlog#16328--secondary channel

swopebe:


release_notes/whats_new.adoc
```diff
@@ -58,6 +58,8 @@ You can choose a type of application to create from the Application console `Cre
 
 With the YAML Editor on, you can see the file update as you create or edit your applications as you modify the fields.
 
+You can now subscribe to a secondary channel if you have a mirrored channel (application source repository). Specify a `secondaryChannel` in the subscription YAML when an application subscription fails to connect to the repository server using the primary channel.
```
rokej: You can now ~~subscribe to~~ specify a secondary channel in the subscription YAML if you have a mirrored channel (application source repository). When an application subscription fails to connect to the repository server using the primary channel, it will use the secondary channel for application deployment and reconciliation.

PR #2929, opened 2021-11-10
Updates related to comment on closed issue

dockerymick:
https://github.com/open-cluster-management/backlog/issues/16821#issuecomment-961166156

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -676,7 +676,7 @@ For instance, namespace `charts-v1` is used by the installer as a Helm type chan
 
 When the Ansible Automation Platform (early access) is installed, `AnsibleJobs` fails to run. To submit prehook and posthook `AnsibleJobs` through {product-title-short}, use the `early-access-cluster-scoped` option. The option is available in Ansible Automation Platform (early access) version `2.0.1+0.1635279521` and later.
 
-The previously mentioned restrictions affect all components that need to perform ansible jobs.
+The previously mentioned restrictions affect all components that need to perform Ansible jobs.
 
 [#ansible-automation-operator-access]
 === Ansible Automation Platform operator access Ansible Tower outside of a proxy
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -676,6 +676,8 @@ For instance, namespace `charts-v1` is used by the installer as a Helm type chan
 
 When the Ansible Automation Platform (early access) is installed, `AnsibleJobs` fails to run. To submit prehook and posthook `AnsibleJobs` through {product-title-short}, use the `early-access-cluster-scoped` option. The option is available in Ansible Automation Platform (early access) version `2.0.1+0.1635279521` and later.
 
+The previously mentioned restrictions affect all components that need to perform ansible jobs.
+
 [#ansible-automation-operator-access]
 === Ansible Automation Platform operator access Ansible Tower outside of a proxy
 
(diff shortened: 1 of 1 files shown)


release_notes/known_issues.adoc
```diff
@@ -676,6 +676,8 @@ For instance, namespace `charts-v1` is used by the installer as a Helm type chan
 
 When the Ansible Automation Platform (early access) is installed, `AnsibleJobs` fails to run. To submit prehook and posthook `AnsibleJobs` through {product-title-short}, use the `early-access-cluster-scoped` option. The option is available in Ansible Automation Platform (early access) version `2.0.1+0.1635279521` and later.
 
+The previously mentioned restrictions affect all components that need to perform Ansible jobs.
```
dockerymick: @swopebe @drkho is this change still required? Do you both agree with this change?

drkho: I believe so...the option of "early-access-cluster-scoped" should still be available. @xiangjingli , can you confirm please?

drkho: And GRC still depends on AAP operator for AnsibleJobs

xiangjingli: @almen-ng could you help confirm if the `early-access-cluster-scoped` option is still available when install the AAP? Has AAP released a new version including the fix?

almen-ng: Currently the channel that should be used when installed `Ansible Automation Platform` (do not need the `(early access)` anymore) are the channels that have `-cluster-scoped` appended at the end. 

The latest release of AAP is `stable-2.1-cluster-scoped` that has the new fix in place.

dockerymick: @almen-ng thank you for your feedback! I have made a couple of changes. When you have an opportunity, may you re-review the PR?

almen-ng: No problem! 

So essentially what is happening is that there are 4 channels currently and two of them are not compatible with AAP (the ones that do not have `-cluster-scoped` appended to them. 

A suggestion would perhaps be rewording this entire known issue as perhaps an fyi that Ansible will only work if the `-cluster-scoped` options are chosen or else AnsibleJobs will not be able to run.

Here are the channels for your reference :)
image

dockerymick: @almen-ng Thank you, thank you! Just made an update

almen-ng: @dockerymick np! 

> The option is available in Ansible Automation Platform version `stable-2.1` and later.

As shown in the picture above, it is not stable-2.1 that the option is available. It just as long as it is a cluster scoped update channel (in this case, identified by `-cluster-scoped` appended at the end, so right now only early-access-cluster-scoped and stable-2.1-cluster-scoped should be used, but there will be new releases of AAP from time to time so the list will expand) then AAP functionalities will work.

Sorry for the confusion!

dockerymick:
I think we are good now! @almen-ng let me know if you think otherwise

almen-ng:
That parts that were changed look good!

almen-ng:
The parts that were changed look good @dockerymick :)

release_notes/known_issues.adoc
```diff
@@ -765,6 +765,14 @@ Use the command line interface (CLI) to make updates to the policies when you us
...
 When you install the gatekeeper operator on {ocp} version 4.9, the installation fails. Before you upgrade {ocp-short} to version 4.9.0., you must upgrade the gatekeeper operator to version 0.2.0. See link:../governance/create_gatekeeper.adoc#upgrading-gatekeeper-gatekeeper-operator[Upgrading gatekeeper and the gatekeeper operator] for more information.
 
+
+[#grc-ansible-automation-platform-fail]
+=== Ansible Automation Platform (early access) job fail
+
+When the Ansible Automation Platform (early access) is installed, `AnsibleJobs` fails to run. To submit prehook and posthook `AnsibleJobs` through {product-title-short}, use the `early-access-cluster-scoped` option. The option is available in Ansible Automation Platform (early access) version `2.0.1+0.1635279521` and later.
+
+The configuration of ansible automation job for governance policies are not functional.
+
```
almen-ng: I believe this part is not necessary anymore as it is covered by line 675-677. 

@drkho would you be able to confirm?

drkho: 👍🏼

dockerymick: Makes sense! Removing that section now. Thanks! @almen-ng @drkho

PR #2953, opened 2021-11-18
Refresh 2.3 11/18/21

cadawson:


dockerymick (approved):
No conflicts! One thing i did notice was the use of ConfigMap. In our source, some are monospaced, some are not. I thought we agreed not to use monospace for ConfigMap, but this can be addressed added as technical debt

PR #2957, opened 2021-11-22
open-cluster-management/backlog#16623 --

swopebe:
--new stuff

### DIFF
--- a/applications/app_model.adoc
+++ b/applications/app_model.adoc
@@ -11,8 +11,8 @@ View the following application resource sections:
 
 * <<applications,Applications>>
 * <<subscriptions,Subscriptions>>
-* <<channels,Channels>>
-* <<placement-rules,Placement rules>>
+* <<applicationset,ApplicationSet>>
+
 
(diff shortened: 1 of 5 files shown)


applications/app_model.adoc
```diff
@@ -10,17 +10,33 @@ image:../images/basic-application-model2.1.png[Application model]
 View the following application resource sections:
 
 * <<applications,Applications>>
-* <<channels,Channels>>
 * <<subscriptions,Subscriptions>>
+* <<channels,Channels>>
 * <<placement-rules,Placement rules>>
 
 [#applications]
```
chenz4027, line 17: I think the organizing strategy is good, but do we consider to reiterate that we would have 3 types of applications?

chenz4027 (commented):
Just one comment

applications/app_model.adoc
swopebe: after quick meeting, we made some other decisions, merging to send another review.

PR #2963, opened 2021-11-24
Updating the conref

dockerymick:
https://github.com/open-cluster-management/backlog/issues/18140


dockerymick:
Verified that there is not a conref name `{olm-short}`. There is only `{olm}`
Screen Shot 2021-11-24 at 3 20 49 PM

dockerymick:
I am comfortable with merging this PR since it is a small change

PR #2973, opened 2021-11-29
few minor edits

berenss:


berenss:
the word `project` is uniquely OpenShift. It is used to indicate a Kubernetes namespace, hence we should be consistent about writing 'project' and it does not need to be referred to as a 'project namespace'.

berenss:
Is there any verbiage being added to instruct the user about how to install on infrastructure nodes, or, is that in another topic?

swopebe:
Thanks @berenss for these improvements; merged.

PR #2998, opened 2021-12-03
open-cluster-management/backlog#16626-install stuck in status

swopebe:


troubleshooting/trouble_install_status.adoc
```diff
@@ -0,0 +1,24 @@
+[#troubleshooting-stuck-pending]
+= Troubleshooting installation status stuck in installing or pending
+
+When installing {product-title-short}, the `MultiClusterHub` remains in `Installing` phase, or multiple pods maintain a `Pending`` status.
```
JakobGray:
```suggestion
When installing {product-title-short}, the `MultiClusterHub` remains in `Installing` phase, or multiple pods maintain a `Pending` status.
```

troubleshooting/trouble_install_status.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+[#symptom-stuck-pending]
+== Symptom: Stuck in Pending status 
+
+More than ten minutes passed since you installed `MultiClusterHub` and one or more components from the `status.components` field of the `MultiClusterHub` resource report `ProgressDeadlineExceeded`. Resource constraints on the cluster might be the issue. 
+
+Check the pods in the `Multiclusterhub` namespace. You might see `Pending` with a status similar to the following:
```
JakobGray:
```suggestion
Check the pods in the `Multiclusterhub`'s namespace. You might see `Pending` with a status similar to the following:
```

troubleshooting/trouble_install_status.adoc
```diff
@@ -0,0 +1,24 @@
...
+message: '0/6 nodes are available: 3 Insufficient cpu, 3 node(s) had taint {node-role.kubernetes.io/master:
+        }, that the pod didn't tolerate.'
+----
+
+In this case, the worker nodes resources are not sufficient in the cluster to run the product.
+
+[#resolving-the-problem-adjust-sizing]
+== Resolving the problem: Adjust worker node sizing
+
+If you have this problem, then your cluster needs to be updated with either larger or more worker nodes. See link:../install/installing#sizing-your-cluster[Sizing your cluster] for guidelines on sizing your cluster.
```
JakobGray: This link doesn't resolve for me in Github. Does it only resolve once published?

troubleshooting/trouble_install_status.adoc
swopebe: @JakobGray in tech writing, we avoid possessives, per IBM/Red Hat standards.

troubleshooting/trouble_install_status.adoc
swopebe: I think I was changing this as you were in it. So this is resolved.

troubleshooting/trouble_install_status.adoc
swopebe: Fixed!

troubleshooting/trouble_install_status.adoc
JakobGray: I still see an extra tick mark after Pending

troubleshooting/trouble_install_status.adoc
JakobGray: Do you have a preferred way to address the possessive without using possessives? The code block style makes it sound like you need to check the namespace called `Multiclusterhub`. But what you really need to do is check the namespace the multiclusterhub is installed in, which is by default `open-cluster-management`.

swopebe: Check the pods in the namespace where `Multiclusterhub` was installed.

JakobGray (approved):

PR #3001, opened 2021-12-03
open-cluster-management/backlog#18278 Add errata 2.3.5

cadawson:


chuckbrant:
This is a container updates only, and there is not feature or functional changes

PR #3009, opened 2021-12-07
Edit Submariner overview and prereqs docs

dfarrell07:
Batch of small edits to the Submariner overview and prerequisites docs.

Signed-off-by: Daniel Farrell <dfarrell@redhat.com>

### DIFF
--- a/services/submariner.adoc
+++ b/services/submariner.adoc
@@ -30,7 +30,7 @@ Ensure that you have the following prerequisites before using Submariner:
 * A {product-title-short} hub cluster that is running on {ocp} version 4.6, or later, with Kubernetes version 1.19, or later.
 * A credential to access the hub cluster with `cluster-admin` permissions.
 * Two or more {ocp-short} managed clusters that are running on {ocp-short} version 4.4, or later, with Kubernetes version 1.17, or later, and are managed by the {product-title-short} hub cluster.
-* Non-overlaping Pod and Service Classless Inter-Domain Routing (CIDR) ranges for clusters. The upstream https://submariner.io/getting-started/architecture/globalnet/[Globalent] feature, which allows overlaping CIDRs, is not currently supported.
+* Non-overlaping Pod and Service Classless Inter-Domain Routing (CIDR) ranges for clusters. The upstream Globalent feature, which allows overlaping CIDRs, will be supported in future releases.
 * IP connectivity must be configured between the Gateway nodes. When connecting two clusters, at least one of the clusters must be accessible to the Gateway node using its public or private IP address designated to the Gateway node. See https://submariner.io/operations/nat-traversal[Submariner NAT Traversal] for more information.
 * Firewall configuration across all nodes in each of the managed clusters must allow 4800/UDP in both directions.
 * Firewall configuration on the Gateway nodes must allow ingress 8080/TCP so the other nodes in the cluster can access it.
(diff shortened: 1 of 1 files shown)


services/submariner.adoc
```diff
@@ -31,17 +30,17 @@ Ensure that you have the following prerequisites before using Submariner:
...
 * A credential to access the hub cluster with `cluster-admin` permissions.
 * Two or more {ocp-short} managed clusters that are running on {ocp-short} version 4.4, or later, with Kubernetes version 1.17, or later, and are managed by the {product-title-short} hub cluster.
-* Pod and Service Classless Inter-Domain Routing (CIDR) between the clusters that do not overlap.
-* IP connectivity must be configured between the Gateway nodes. When connecting two clusters, at least one of the clusters must be accessible to the Gateway node using its public or private IP address designated to the Gateway node. See https://submariner.io/operations/nat-traversal[Submariner Nat Traversal] for more information.
-* Firewall configuration across all nodes in each of the managed clusters must allow 4800/UDP in both directions. 
-* Firewall configuration on the Gateway nodes must allow ingress 8080/TCP so the other nodes in the cluster can access it. 
+* Non-overlapping Pod and Service Classless Inter-Domain Routing (CIDR) ranges for clusters. The upstream https://submariner.io/getting-started/architecture/globalnet/[Globalent] feature, which allows overlapping CIDRs, is not supported.
```
cadawson: If this feature is not supported, then I suggest we don't provide the link. Best case would be not to mention that feature at all, as we said that they could not be overlapping. If there is a compelling reason to mention it (like customers will definitely know about it and wonder specifically whether it is supported), we can state that it is not supported. I think the link provides an opportunity for them to link away to something that is not useful. Thoughts?

cadawson (changes requested):
This looks good. Thank you. I made some minor edits and added a comment. Let me know if you have questions or concerns. Thanks!

cadawson:
Is there an issue that we can use for this? If not, will you create an issue using the doc template for this, and assign it to me @dfarrell07 ? Thanks!

services/submariner.adoc
dfarrell07: Thanks for the review @cadawson! I do think users will know that Globalnet is a feature of Submariner, and will wonder if it's supported and why not. Fortually, as I understand it (@nyechiel can confirm), Globalnet is planned for ACM 2.5. I updated the proposed wording to remove the link and try to clarify for users that support coming, but without being too committed to details.

dfarrell07:
> Is there an issue that we can use for this? If not, will you create an issue using the doc template for this, and assign it to me @dfarrell07 ? Thanks!

No Issue, I was just planning on sending some drive-by contributions to the Submariner docs. :nerd_face: 

We're talking about [this template](https://github.com/open-cluster-management/rhacm-docs/blob/2.4_stage/.github/ISSUE_TEMPLATE.md) I assume? Or should I follow that internal user process? (why is there a difference between the two? :sweat_smile:)

services/submariner.adoc
cadawson: Thanks for the explanation @dfarrell07 . We will leave the mention, then. In the formal docs, we are very careful about preannouncing features and support. The reason is that we might decide in the end that we don't want to support a certain feature when the market changes. If we said we would, then someone might be able to claim that we misled them into a purchase based on that statement. 
Anyway, I tweaked the wording just a little. It mentions that it is not supported, but doesn't hint that it will be. When it is, we will just add it then. Will this work? Thanks!

cadawson:
Hi @dfarrell07 . Sorry to scare you. :-) That wasn't the intent. Typically we tie our doc PRs back to an issue. That might be a dev issue or, most often, a doc issue. I wasn't sure if you had one. I created one for this, just so we have a record. It is issue #18505 (https://github.com/open-cluster-management/backlog/issues/18505).

services/submariner.adoc
dfarrell07: That makes sense @cadawson, thanks for the explanation! Yes the new wording looks good to me. :heavy_check_mark:

dfarrell07:
> I created one for this, just so we have a record. It is issue #18505 ([open-cluster-management/backlog#18505](https://github.com/open-cluster-management/backlog/issues/18505)).

Thanks @cadawson! Can I use that same Issue for other small Submariner docs fixes?

cadawson:
Yes, feel free to use it for additional minor updates. Thanks!

PR #3017, opened 2021-12-09
open-cluster-management/backlog#18387 ki subnet req

cadawson:


clusters/create_cluster_on_prem.adoc
```diff
@@ -58,6 +58,8 @@ The locations of the available hosts that meet the requirements for this cluster
 +
 *Note:* This value must match the name that you used to create the DNS records listed in the prerequisites section. If not provided, the DNS must be pre-configured so that `test.apps.` resolves correctly.
 
+. Select a subnet from the list of available subnets. 
+
```
jelkosz: actually, reading it again, its already documented above: 
`After your hosts finish binding and display a `Bound` status, select a subnet for your cluster from the list of options for the bound host`

release_notes/known_issues.adoc
```diff
@@ -160,6 +160,12 @@ When observability is enabled or disabled on {ocp-short} managed clusters that a
 [#cluster-management-issues]
 == Cluster management known issues
 
+[#subnet-required-on-prem-clust-create]
+=== Selecting a subnet is required when creating an on-premises cluster
+//2.4:18387
+
+When you create an on-premises cluster using the {product-title-short} console, you must select an available subnet for your cluster. It is not marked as a required field.
```
jelkosz, line 167: Its correct, but FYI this is planned to be fixed in 2.5 where we will add way more validations and will mark those fields as mandatory. But yes, currently this is correct.

clusters/create_cluster_on_prem.adoc
cadawson: Good point. That should work.

cadawson: Removed the duplicate.

release_notes/known_issues.adoc
cadawson: We will have the opportunity to scrub this before 2.5 releases. We can remove it at that time.

jelkosz:
lgtm

PR #3038, opened 2022-01-07
fix alertforward support version.

morvencao:
The alert forward feature support OCP 4.8 and later, the 2.3 doc is correct, but it is wrong for 2.4 doc.

Signed-off-by: morvencao <lcao@redhat.com>


marcolan018 (approved):
/lgtm

PR #3053, opened 2022-01-10
Updating the splash page with brief descriptions

dockerymick:
https://github.com/stolostron/backlog/issues/11423

### DIFF
--- a/install/docinfo.xml
+++ b/install/docinfo.xml
@@ -2,7 +2,7 @@
 <productnumber>{product-version}</productnumber>
 <subtitle>Install</subtitle>
 <abstract>
-    <para>Installing instructions for {product-title}</para>
+    <para>Read more for installing instructions on connected and disconnected networks, such as requirements and recommendations, sizing your cluster, cluster scalability, multicluster advanced configurations, upgrades, and uninstalling. </para>
 </abstract>
 <authorgroup>
     <orgname>{product-title} Team</orgname>
--- a/install/docinfo.xml
+++ b/install/docinfo.xml
@@ -2,7 +2,7 @@
 <productnumber>{product-version}</productnumber>
 <subtitle>Install</subtitle>
 <abstract>
-    <para>Read more for installing instructions on connected and disconnected networks, such as requirements and recommendations, sizing your cluster, cluster scalability, multicluster advanced configurations, upgrades, and uninstalling. </para>
+    <para>Read more about installing on connected and disconnected networks, requirements and recommendations for installation, multicluster advanced configurations, and instructions for upgrading and uninstalling. </para>
 </abstract>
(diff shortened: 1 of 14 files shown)


services/serv_overview.adoc
```diff
@@ -5,3 +5,10 @@ You can add services to use with {product-title} that can improve some area of p
 
 The following sections provide a summary of the types of services that are available for {product-title-short}.
 
+// Assuminng that the files in this folder should be listed on this page
+
+
+//* xref:../services/deploy_submariner_api.adoc[Deploying Submariner with APIs]
```
dockerymick, line 11: @cadawson It seemed like reference links for this book needed to be added. Let me know if this is not correct.

cadawson: @dockerymick : I think that was left off because right now it is only Submariner. For consistency, we can go ahead and add it. It will be helpful when we start adding other services. Thank you!

install/docinfo.xml
```diff
@@ -2,7 +2,7 @@
 <productnumber>{product-version}</productnumber>
 <subtitle>Install</subtitle>
 <abstract>
-    <para>Installing instructions for {product-title}</para>
+    <para>Read more for installing instructions on connected and disconnected networks, such as requirements and recommendations, sizing your cluster, cluster scalability, multicluster advanced configurations, upgrades, and uninstalling. </para>
```
cadawson: This seems a little long. I suggest shortening to: 

Read more about installing on connected and disconnected networks, requirements and recommendations for installation, multicluster advanced configurations, and instructions for upgrading and uninstalling.

Or something like that. Maybe even leave it at: ...installing on connected and disconnected networks, upgrading and uninstalling.

cadawson (approved):
This looks great! Thanks for adding these, Mikela! I think it will improve how it looks on the splash page. I added a suggestion, but I like it.

dockerymick:
Thanks for your review Chris and thanks for catching my typos! I agree with your suggestion. Moving forward with the merge since there is at least one approval.

cadawson:
Thanks again, @dockerymick ! Nice improvement!

observability/docinfo.xml
```diff
@@ -2,7 +2,7 @@
 <productnumber>{product-version}</productnumber>
 <subtitle>Observability</subtitle>
 <abstract>
-    <para>Read more to learn how to optimiize your your managged clusters by enabling and customizing the observability service.</para>
+    <para>Read more to learn how to optimize your your managged clusters by enabling and customizing the observability service.</para>
 </abstract>
```
swopebe, line 6: Sorry @dockerymick to get to this so late, not sure if this was fixed.... two gs in managed on line 5.

PR #3057, opened 2022-01-12
Adding req for observability

dockerymick:
https://github.com/stolostron/backlog/issues/17858


marcolan018 (changes requested):
storageclass is optional in MCO CR. 
How about using this statement?
You must define a storage class in the `MultiClusterObservability` CR if there is no default storage class specified

dockerymick:
Thanks for your review @marcolan018, I agree with your suggestion and have updated the file. Moving forward to merge the PR

PR #3060, opened 2022-01-13
stolostron/backlog#19002 volsync updates

cadawson:


tesshuflower:
@cadawson the changes are looking really good to me!

It's a bit hard for me to tell looking at the PR, so I may have just missed it - but I think you have things like a source.yaml (for ReplicationSource) in the docs itself which is really nice - do you do the same for the ReplicationDestination?  I still see some references that talk about taking examples from `examples/rsync`.  Will we expect the user to pull files from github or will it all be converted to text in the doc?

cadawson:
Thanks @tesshuflower . I have examples of the text at the beginning of the steps. My thought was: Introduce step > show yaml sample for a look at what it should look like > provide the steps to get there. Does that sound reasonable? Thanks!

tesshuflower:
> Thanks @tesshuflower . I have examples of the text at the beginning of the steps. My thought was: Introduce step > show yaml sample for a look at what it should look like > provide the steps to get there. Does that sound reasonable? Thanks!

That sounds great!  Let me go through and look at it in more detail

clusters/volsync.adoc
```diff
@@ -45,147 +51,260 @@ oc -n <namespace> apply -f ./policy-persistent-data-management.yaml
 +
 Replace `namespace` with a namespace on your hub cluster.
 
-. If your default `StorageClass` on each of the clusters where you plan to configure VolSync does not refer to a CSI-based storage driver, change the default configuration to refer to a CSI-based storage driver. 
+. For Rsync-based replication, configure custom resources on the source and destination clusters. The custom resources use the `address` value to connect the source to the destination, and the `sshKeys` to ensure that the transferred data is secure.
```
tesshuflower: Can we move this entire section outside of the "Installing VolSync on the managed clusters" section?  I think this is essentially a walkthrough of an example scenario of setting up pvc replication between managed clusters using rsync.

tesshuflower: Is it possible to also add a description of the scenario?  something along the lines of "This example configures PVC replication from a managed cluster named `source` in namespace `source` to a managed cluster named `destination` in namespace `dest`. Wondering if we should actually name the clusters `source-cluster` and `destination-cluster`.

clusters/volsync.adoc
```diff
@@ -45,147 +51,260 @@ oc -n <namespace> apply -f ./policy-persistent-data-management.yaml
...
-
-.. If a corresponding `VolumeSnapshotClass` does not exist, create one by running the following command:
+.. Configure your destination cluster named `destination`.
 
+... Run the following commands on the destination cluster to create and prepare the namespace:
++
 ----
-kubectl create -f - << SNAPCLASS
+$ kubectl create ns dest
+$ kubectl create -n dest -f examples/rsync/volsync_v1alpha1_replicationdestination_remotecluster.yaml
```
tesshuflower: This line is telling the users to create the replicationdestination from a yaml file that doesn't exist.  Can we move this to after the explanation of the yaml in the next lines?  I.e. Show the user the replicationdestination - let them copy this yaml file and edit it, and then tell them to run the cmd to create the replicationdestination.

clusters/volsync.adoc
```diff
@@ -45,147 +51,260 @@ oc -n <namespace> apply -f ./policy-persistent-data-management.yaml
...
+Replace `destination_name` with the name of your destination volume.
++
+Replace `dest` with the name of the namespace where your destination is located.
++
+For this example, the `ServiceType` value of `LoadBalancer` is used. The load balancer service is created by the source cluster to enable your source managed cluster to transfer information to a different destination managed cluster. You can use `ClusterIP` as the service type if your source and destinations are on the same cluster. Note the address of the Load Balancer and the name of the secret to refer to when you configure the source cluster.
++ 
+The `storageClassName` and `volumeSnapshotClassName` are optional parameters. If you are using a storage class and volume snapshot class name that are different than the defaults for your environment, specify those values. 
 
-.. Set `gp2-csi` as your default `VolumeSnapshotClass` by running the following command:
+... Run the following command to copy the value of `.status.rsync.address`:
```
tesshuflower: Maybe add some explanation here - after the replication destination is created, it should fill out the status with an address.  You can see in the volsync docs there is a bit of an explanation about replication destination and its fields - something like this would be helpful here.  https://volsync.readthedocs.io/en/stable/usage/rsync/index.html#destination-status

clusters/volsync.adoc
```diff
@@ -45,147 +51,260 @@ oc -n <namespace> apply -f ./policy-persistent-data-management.yaml
...
 
-.. Set `standard-csi` as your default `VolumeSnapshotClass` by running the following command:
+.. Deploy the database. For this example, a database is copied. You might use an application. 
+
+... On the source cluster, `source-cluster`, enter the following commands to create and prepare the database: 
 +
 ----
-kubectl annotate volumesnapshotclass/standard-csi snapshot.storage.kubernetes.io/is-default-class="true"
+$ kubectl create ns source
+$ kubectl create -n source -f examples/source-database
```
tesshuflower: If we're embedding the files in the doc, then we don't have this source database (user doesn't have these examples files).  Maybe a decision needs to be made here whether we run through this entire database example or simply tell them they can deploy their app which creates their source PVC on the source-cluster.

clusters/volsync.adoc
```diff
@@ -45,147 +51,260 @@ oc -n <namespace> apply -f ./policy-persistent-data-management.yaml
...
 +
 Replace `source.host.com` with the host address that you copied from the `.status.rsync.address` field of the `ReplicationDestination` when you configured it. 
++
+If your storage driver supports cloning, using `Clone` as the value for `copyMethod` might be a more streamlined process for the replication.
++ 
+The `storageClassName` and `volumeSnapshotClassName` are optional parameters. If you are using a storage class and volume snapshot class name that are different than the defaults for your environment, specify those values. 
 
 You can now set up the synchronization method of the persistent volume.
 
+... Copy the SSH secret from the destination cluster by entering the following command on the destination cluster:
```
tesshuflower: Need to be very explicit that this command should be done against the destination-cluster.  Maybe some consistent marker in the doc to indicate what commands run against what cluster?  I guess I don't like the wording "enter the command ON the destination cluster".  This command is assuming it's running on some system that is able to access both source and destination cluster.

Some docs mention explicitly things like:

```
Run command against the source cluster
# export KUBECONFIG=source-cluster.config
# kubectl get ....

Run command against destination-cluster
# export KUBECONFIG=destination-cluster.config
# kubectl get
```

^ Just as an example - is it possible to see what is generally done in the ACM docs for this kind of thing?

clusters/volsync.adoc
```diff
@@ -45,147 +51,260 @@ oc -n <namespace> apply -f ./policy-persistent-data-management.yaml
...
+----
+$ kubectl create -f /tmp/secret.yaml
+----
++
+*Note:* In this example, the source and the destination PVs are on the same server. If your PVs are not on the same server, then copy your secret file to the server of your source PV. 
+
+... On the source cluster, modify the `volsync_v1alpha1_replicationsource_remotecluster.yaml` file by replacing the value of the `address` and `sshKeys` in the `ReplicationSource` object with the values that you noted from the destintation cluster by entering the following commands:
++
+----
+$ sed -i "s/my.host.com/$ADDRESS/g" examples/rsync/volsync_v1alpha1_replicationsource_remotecluster.yaml
```
tesshuflower: these files in examples do not exist, so instructions would need to be modified to tell the user to copy the sample replicationsource.yaml above, edit it and then run the sed and create commands

clusters/volsync.adoc
cadawson: I created a new section out of this, and made it a separate list of steps

cadawson: Added a description.

clusters/volsync.adoc
cadawson: Moved some things around here. I think I understood what you meant.

clusters/volsync.adoc
cadawson: Tried to make this change.  Please review.

clusters/volsync.adoc
cadawson: Updated this to be against.

clusters/volsync.adoc
```diff
@@ -36,156 +38,281 @@ spec:
...
+. Configure your destination cluster named `destination`.
 
+.. Run the following command on the destination cluster to create the namespace:
++
 ----
-kubectl create -f - << SNAPCLASS
+$ kubectl create ns dest
 ----
 
+.. Copy the following YAML content to create a new file called `replication_destination.yaml` on the destination cluster:
```
tesshuflower: The file itself most likely won't be created on the destination cluster - can probably just end with at "... create a new file called replication_destination.yaml"

clusters/volsync.adoc
```diff
@@ -36,156 +38,281 @@ spec:
...
 oc -n <namespace> apply -f ./policy-persistent-data-management.yaml
 ----
 +
 Replace `namespace` with a namespace on your hub cluster.
 
-. If your default `StorageClass` on each of the clusters where you plan to configure VolSync does not refer to a CSI-based storage driver, change the default configuration to refer to a CSI-based storage driver. 
-+
-* Your steps might resemble the following procedure for an Amazon Web Services environment:
+[#volsync-rsync-clusters]
+== Configiring Rsync replication across managed clusters
```
tesshuflower: typo "Configiring"

clusters/volsync.adoc
```diff
@@ -36,156 +38,281 @@ spec:
...
-.. Change your default `StorageClass` by entering the following commands: 
+For Rsync-based replication, configure custom resources on the source and destination clusters. The custom resources use the `address` value to connect the source to the destination, and the `sshKeys` to ensure that the transferred data is secure.
 +
-----
-kubectl annotate sc/gp2 storageclass.kubernetes.io/is-default-class="false" --overwrite
-kubectl annotate sc/gp2-csi storageclass.kubernetes.io/is-default-class="true" --overwrite
-----
+**Note:** You must copy the values for `address` and `sshKeys` from the destination to the source, so configure the destination before you configure the source.
+
+This example provides the steps to configure an Rsync replication from a PVC on the `source` cluster in the `source` namespace to a PVC on a `destination` cluster in the `dest` namespace.
```
tesshuflower: Possible to move this up to be the first thing under the "Configuring rsync replication accross managed clusters" header?

clusters/volsync.adoc
```diff
@@ -36,156 +38,281 @@ spec:
...
 
-.. If a corresponding `VolumeSnapshotClass` does not exist, create one by running the following command:
+. Configure your destination cluster named `destination`.
```
tesshuflower: Think we need to be consistent here throughout the example - later on we simply call this the `destination cluster`.  So maybe we can always refer to it as "the destination cluster" rather than saying "...destination cluster named destination".  This is probably my fault with some initial wording I gave you.

clusters/volsync.adoc
```diff
@@ -36,156 +38,281 @@ spec:
...
+$ echo $SSHKEYS
 ----
 +
-Your CSI is now configured on a Google Cloud Platform environment.
-
-. For Rsync-based replication, configure custom resources on the source and destination clusters. The custom resources use the `address` value to connect the source to the destination, and the `sshKeys` to ensure that the transferred data is secure. **Note:** You must copy the values for `address` and `sshKeys` from the destination to the source, so configure the destination before you configure the source. 
+You will have to enter it on the source cluster when you configure the source. The output should be the name of your SSH keys secret file, which might resemble the following name:
++
+----
+volsync-rsync-dst-src-database-destination
```
tesshuflower: Let's change this example name to `volsync-rsync-dst-src-destination-name`

clusters/volsync.adoc
```diff
@@ -36,156 +38,281 @@ spec:
...
 +
 [source,yaml]
 ----
+---
 apiVersion: volsync.backube/v1alpha1
 kind: ReplicationSource
 metadata:
-  name: <name>
-  namespace: <source_namespace>
+  name: <database-source>
```
tesshuflower: can we change this to "name: <source_name>" to match what we have for the replicationdestination.yaml?

clusters/volsync.adoc
```diff
@@ -36,156 +38,281 @@ spec:
...
     sshKeys: <volsync-rsync-destination-src-database-destination>
     address: <source.host.com>
-    copyMethod: Clone
+    copyMethod: Snapshot
+    storageClassName: gp2-csi
+    volumeSnapshotClassName: gp2-csi
 ----
 +
-Replace `name` with a unique name for your replication.
+Replace `database_source` with a unique name for your replication.
```
tesshuflower: change to `source_name`

clusters/volsync.adoc
```diff
@@ -36,156 +38,281 @@ spec:
...
+
+.. On the source cluster, create the secret file by entering the following command on the source cluster:
++
+----
+$ kubectl create -f /tmp/secret.yaml
+----
++
+*Note:* In this example, the source and the destination PVs are on the same server. If your PVs are not on the same server, then copy your secret file to the server of your source PV. 
+
+.. On the source cluster, modify the `volsync_v1alpha1_replicationsource_remotecluster.yaml` file by replacing the value of the `address` and `sshKeys` in the `ReplicationSource` object with the values that you noted from the destintation cluster by entering the following commands:
```
tesshuflower: should be "..modify the `replication_source.yaml` file ..."

clusters/volsync.adoc
```diff
@@ -36,156 +38,281 @@ spec:
...
++
+----
+$ sed -i "s/my.host.com/$ADDRESS/g" replication_source.yaml
```
tesshuflower: Our example yaml has <source.host.com> as the example address - so this instruction would need to change (same with sshkeys below).

Perhaps in the example.yaml we should use the following:

```
<my.host.com>
<mysshkeys>
```

And then the sed commands would be:
```
$ sed -i "s/<my.host.com>/$ADDRESS/g" replication_source.yaml
$ sed -i "s/<mysshkeys>/$SSHKEYS/g" replication_source.yaml
```

Note also above the instructions say to replace those fields already - we probably need to say at that point that there are additional instructions here to do that.

clusters/volsync.adoc
```diff
@@ -36,156 +38,281 @@ spec:
...
 +
-Replace `name` with a unique name for your replication.
+Replace `database_source` with a unique name for your replication.
 +
-Replace `source_namespace` with the name of the namespace where your source is located.
+Replace `source` with the name of the namespace where your source is located.
 +
 Replace `persistent_volume_claim` with the name of your source claim.
 +
 Replace `volsync-rsync-destination-src-database-destination` with the keys that you copied from the `.status.rsync.sshKeys` field of the `ReplicationDestination` when you configured it.
```
tesshuflower: Should mention there are detailed steps for this below (the sed commands)

clusters/volsync.adoc
```diff
@@ -36,156 +38,281 @@ spec:
...
+Replace `database_source` with a unique name for your replication.
 +
-Replace `source_namespace` with the name of the namespace where your source is located.
+Replace `source` with the name of the namespace where your source is located.
 +
 Replace `persistent_volume_claim` with the name of your source claim.
 +
 Replace `volsync-rsync-destination-src-database-destination` with the keys that you copied from the `.status.rsync.sshKeys` field of the `ReplicationDestination` when you configured it. 
 +
 Replace `source.host.com` with the host address that you copied from the `.status.rsync.address` field of the `ReplicationDestination` when you configured it.
```
tesshuflower: Should mention there are detailed steps for this below (the sed commands)

clusters/volsync.adoc
```diff
@@ -36,155 +38,292 @@ spec:
...
-* The steps might resemble the following for a Google Cloud Platform environment:
+| `.status.rsync.address` | IP address of the destination cluster that is used to enable the source and destination clusters to communicate.
+| `.status.rsync.sshKeys` | Name of the SSH key file that enables secure data transfer from the source cluster to the destination cluster. 
+|==========
 
-.. Change your default `StorageClass` by entering the following commands: 
+.. Run the following command to copy the value of `.status.rsync.address` to use on the source cluster:
++
+----
+$ ADDRESS=`kubectl get replicationdestination database-destination -n dest --template={{.status.rsync.address}}`
```
tesshuflower: in this command `database-destination` should be the name they used in their replication destination.  I suggested above we hardcode the name in the example to something like `test-replication-destination`

clusters/volsync.adoc
```diff
@@ -36,155 +38,292 @@ spec:
...
+    copyMethod: Snapshot
+    capacity: 2Gi
+    accessModes: [ReadWriteOnce]
+    storageClassName: gp2-csi
+    volumeSnapshotClassName: gp2-csi
 ----
++
+*Note:* The `capacity` value should match the capacity of the persistent volume claim that is being replicated.
++
+Replace `destination_name` with the name of your destination volume.
```
tesshuflower: This actually isn't the name of the destination volume, it's the name they want to name their replication destination CR.  Maybe we should just hardcode a name here and use it throughout the rest of this doc?  Something like `test-replication-destination`.  we can just mention they can change this to whatever they like.

clusters/volsync.adoc
```diff
@@ -36,155 +38,292 @@ spec:
...
 [source,yaml]
 ----
+---
 apiVersion: volsync.backube/v1alpha1
-kind: ReplicationDestination
+kind: ReplicationSource
 metadata:
-  name: <destination_name>
-  namespace: <destination_namespace>
+  name: <source_name>
```
tesshuflower: Maybe we should harcode a name here to make the example easier like I suggested with the replicationdestination.  something like `test-replication-source`

cadawson: As above: I'm not sure I understand the advantage of using a name here, rather than a variable. If we keep it consistent and hit all of the instances, the variables should be ok, no? See if I caught them all. Thanks!

tesshuflower: The problem is we have things like "Replace source_name with the name for your replication source CR. See step 3f for instructions on how to replace this automatically." - but there are no steps to replace the source_name.

Additionally other steps refer where we provide commands to extract fields from the replicationsource/destination, those commands need to use the name of the replication source -so you'd have to explain every time that the user needs to replace the name with whatever they provided earlier as `source_name` or `dest_name`.

An example is step 3g:

```
$ kubectl describe ReplicationSource -n source application-source
```

The name here needs to match the name they used for their replication source CR.  The example uses `application-source` which I don't think we've used anywhere else?

cadawson:
@tesshuflower : Can we run through these instructions to determine if they work? If you think it is better, we can merge this and open a new issue with anything else that causes a problem. Thanks!

tesshuflower:
> @tesshuflower : Can we run through these instructions to determine if they work? If you think it is better, we can merge this and open a new issue with anything else that causes a problem. Thanks!

@cadawson the doc still references files in examples - I think we need to address that before we can merge it.  Could you look at the comments above related to examples?  The text would need to tell the user more generically to deploy their app that creates the PVC etc rather than telling them to deploy an example yaml we aren't providing in the doc.

clusters/volsync.adoc
```diff
@@ -36,155 +38,267 @@ spec:
...
+
+[#volsync-convert-backup-pvc]
+== Converting a replicated image to a usable persistent volume claim
+
+There might be a time when you need to use the replicated image. The image must be converted to a persistent volume claim before it can be used. To convert a replicated image to a persistent volume claim, complete the following steps:
+
+. When the replication is complete, identify the latest snapshot from the `ReplicationDestination` object by entering the following command:
++
+----
+$ kubectl get replicationdestination database-destination -n dest --template={{.status.latestImage.name}}
```
tesshuflower: `database-destination` should be the name of their replication destination, so need to say something like `<destination_name>` and tell them to replace.

Also, `dest` is the destination namespace, so same explanation needed - something like `<dest>` and instructions to replace. - edit: we call this <dest> in the replicationdestination example yaml, but then also use <dest_namespace> sometimes - should be consistent.

clusters/volsync.adoc
```diff
@@ -36,155 +38,267 @@ spec:
...
 ----
 +
-Replace `name` with a unique name for your replication.
+Replace `pvc-name` with a name for your new persistent volume claim.
 +
-Replace `source_namespace` with the name of the namespace where your source is located.
+Replace `dest-namespace` with the namespace of the namespace where the persistent volume claim is located.  
 +
-Replace `persistent_volume_claim` with the name of your source claim.
+Replace `snapshot_to_replace` with the snapshot name that you found in the previous step.
```
tesshuflower: Can we use `VolumeSnapshot` instead of `snapshot`?

clusters/volsync.adoc
```diff
@@ -36,155 +38,267 @@ spec:
...
 +
-Replace `volsync-rsync-destination-src-database-destination` with the keys that you copied from the `.status.rsync.sshKeys` field of the `ReplicationDestination` when you configured it. 
-+
-Replace `source.host.com` with the host address that you copied from the `.status.rsync.address` field of the `ReplicationDestination` when you configured it. 
+You can update `resources.requests.storage` with a different value, but it is best proctice when the value is at least the same size as the initial source persistent volume claim.
 
-You can now set up the synchronization method of the persistent volume.
+. Validate that your persistent volume claim pod is running in the environment by entering the following command:
```
tesshuflower: these instructions are about creating a PVC, so they need to validate that their "persistent volume clam has been created in the environment".  Please remove any mention of "pod".

clusters/volsync.adoc
```diff
@@ -36,155 +38,267 @@ spec:
...
++
+----
+$ kubectl get pods -n <dest-namespace>
```
tesshuflower: should be `kubectl get pvc -n <dest-namespace>`

clusters/volsync.adoc
```diff
@@ -36,155 +38,267 @@ spec:
...
 ----
+---
 apiVersion: volsync.backube/v1alpha1
-kind: ReplicationDestination
+kind: ReplicationSource
 metadata:
-  name: <destination_name>
-  namespace: <destination_namespace>
+  name: <source_name>
+  namespace: <source>
```
tesshuflower: It's confusing here as later on we mention `source` is the name of the source pvc and not the namespace?  Can we globally use something consistent for the source/dest namespaces?  I'd suggest using something like:
`source-namespace`, `destination-namespace`

clusters/volsync.adoc
```diff
@@ -36,155 +38,267 @@ spec:
...
-    capacity: 10Gi
-    accessModes: ["ReadWriteOnce"]
+    storageClassName: gp2-csi
+    volumeSnapshotClassName: gp2-csi
 ----
 +
-*Note:* The `capacity` value should match the capacity of the persistent volume claim that is being replicated.
+Replace `source_name` with the name for your replication source CR. See step 3f for instructions on how to replace this automatically.
++
+Replace `source` with the name of the persistent volume claim where your source is located. See step 3f for instructions on how to replace this automatically.
```
tesshuflower: This is actually the namespace, should be about replacing `source/source-namespace` with the namespace

clusters/volsync.adoc
```diff
@@ -36,155 +38,267 @@ spec:
...
+    storageClassName: gp2-csi
+    volumeSnapshotClassName: gp2-csi
 ----
 +
-*Note:* The `capacity` value should match the capacity of the persistent volume claim that is being replicated.
+Replace `source_name` with the name for your replication source CR. See step 3f for instructions on how to replace this automatically.
++
+Replace `source` with the name of the persistent volume claim where your source is located. See step 3f for instructions on how to replace this automatically. 
++
+Replace `persistent_volume_claim` with the name of your source claim.
```
tesshuflower: I think the edit meant for here ended up on line 169 - can we call this "your source persistent volume claim" instead of "source claim"?

clusters/volsync.adoc
```diff
@@ -36,155 +38,267 @@ spec:
...
+Replace `mysshkeys` with the keys that you copied from the `.status.rsync.sshKeys` field of the `ReplicationDestination` when you configured it.
++
+Replace `source` with the name of the persistent volume claim where your source is located.
++
+*Note:* You must create the the file in the same namespace as the persistent volume claim that you want to replicate. 
+
+.. Verify that the replication completed by running the following command on the `ReplicationSource` object:
++
+----
+$ kubectl describe ReplicationSource -n source <pvc-source>
```
tesshuflower: Another example here of the namespace (the -n parameter) being hardcoded - since we've given them instructions on how to change, we need to let them know to replace on every single command.  So source should be something like `<source>` (or `<source-namespace>` if we rename to that)

clusters/volsync.adoc
```diff
@@ -36,155 +38,267 @@ spec:
...
+.. Copy the SSH secret from the destination cluster by entering the following command against the destination cluster:
 +
-[source,yaml]
 ----
-...
-status:
-  rsync:
-    address: 10.01.101.001
-    sshKeys: volsync-rsync-dest-src-test
+$ kubectl get secret -n dest $SSHKEYS -o yaml > /tmp/secret.yaml
```
tesshuflower: Another example where dest namespace is just hardcoded in the cmd line example

tesshuflower (changes requested):
Added some comments

clusters/volsync.adoc
```diff
@@ -36,155 +38,286 @@ spec:
...
+Replace `pvc-name` with a name for your new persistent volume claim.
 +
-Replace `persistent_volume_claim` with the name of your source claim.
+Replace `destination-ns` with the namespace where the persistent volume claim is located.  
 +
-Replace `volsync-rsync-destination-src-database-destination` with the keys that you copied from the `.status.rsync.sshKeys` field of the `ReplicationDestination` when you configured it. 
+Replace `snapshot_to_replace` with the `VolumeSnapshot` name that you found in the previous step.
 +
-Replace `source.host.com` with the host address that you copied from the `.status.rsync.address` field of the `ReplicationDestination` when you configured it. 
+You can update `resources.requests.storage` with a different value, but it is best proctice when the value is at least the same size as the initial source persistent volume claim.
```
tesshuflower: typo: `proctice`

clusters/volsync.adoc
```diff
@@ -36,155 +38,286 @@ spec:
...
-status:
-  rsync:
-    address: 10.01.101.001
-    sshKeys: volsync-rsync-dest-src-test
+$ vi /tmp/secret.yaml
+----
+
+.. In the open secret file on the destination cluster, make the following changes:
++
+* Change the namespace to the namespace of your source cluster. For this example, it is `source`.
```
tesshuflower: source probably should be source-ns to match the rest

tesshuflower (changes requested):
looking really good! I only found a couple really small items

clusters/volsync.adoc
cadawson: Missed that one.

tesshuflower (approved):
/lgtm

cadawson:
Thanks @tesshuflower !

clusters/volsync.adoc
```diff
@@ -36,155 +38,286 @@ spec:
...
+* Change the namespace to the namespace of your source cluster. For this example, it is `source-ns`.
+* Remove the owner references (`.metadata.ownerReferences`).
+
+.. On the source cluster, create the secret file by entering the following command on the source cluster:
++
+----
+$ kubectl create -f /tmp/secret.yaml
+----
++
+*Note:* In this example, the source and the destination persistent volume claims are on the same server. If your persistent volume claims are not on the same server, then copy your secret file to the server of your source persistent volume claims.
```
tesshuflower: @cadawson Sorry, I missed this before - I think this note is most likely not needed and may be confusing to the reader.

The example is talking about separate clusters, and we do give instructions on how to copy the secret from one cluster to the other - can we just remove this note (line 221)

tesshuflower (commented):
sorry I missed it before, please see my comment about line 221

cadawson:
Removed the note from the source. Merging.

PR #3084, opened 2022-01-25
adding deprecated label to custom policy controller

dockerymick:
https://github.com/stolostron/backlog/issues/19019#issuecomment-1020452149


gparvin (approved):
Thanks for adding this!

PR #3097, opened 2022-01-26
stolostron/backlog#18380 (insufficient content)

swopebe:
This may not be enough information for a good release note -- please look at the template to see what we need in doc issues.

**How does the user change to the right channel, also I am not sure I worded everything correctly, so please review that as well?**

Thank you!

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -236,6 +236,12 @@ When observability is enabled or disabled on {ocp-short} managed clusters that a
 [#cluster-management-issues]
 == Cluster management known issues
 
+[#clusterimageset-fast-channel]
+=== 
+//2.4:17790
+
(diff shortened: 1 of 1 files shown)


ldpliu (approved):

ldpliu:
This is the detail problem in https://coreos.slack.com/archives/CTDEY6EEA/p1643831842776819, it may help you understand the issue.
How about change to : 

--------------
**## The `Channel` is not right in cluster detail page.**

Currently, The `clusterimageset` is in `"fast"` channel by default, but the provisioned cluster is in `"stable"` channel by default.  the product does not sync the `"channel"` to the provisioned {ocp-short} cluster. 
You could manually change to the right channel in the {ocp-short} console `Administration->Cluster Settings->Details Channel.` 

--------------

swopebe:
Added that little change, merging, closing.

PR #3117, opened 2022-01-31
stolostron/backlog#18619 cim upd vol_size

cadawson:


jelkosz:
@filanov can you take a look plz? I have no idea :)

cadawson:
Hi @filanov . Adding you as an assignee for this, hoping you can help with it or identify someone who can. Thank you!

filanov:
Sorry for the delay, filesystem storage will keep the logs and manifests, kubeconfig etc. for the clusters. 
at the beginning we kept the images there as we but now it's handled by a different PV that is handled by the image service, it will be added in 2.5 and backported to 2.4 as well. @carbonin can explain the details and the requirements for the images service (not sure that it's related to this question) 

Regarding the 20G i think that we can reduce it, depend on how many clusters we need to support. 
logs from the controller and the agents are probably the biggest part at the moment.

clusters/create_infra_env.adoc
```diff
@@ -70,9 +71,9 @@ spec:
       cpuArchitecture: "x86_64"
 ----
 +
-Replace `db_volume_size` with the volume size for the `databaseStorage` field, for example `10Gi`.
+Replace `db_volume_size` with the volume size for the `databaseStorage` field, for example `10Gi`. This value specifies how much storage is is allotted for storing files like database tables and database views for the clusters. You might need to use a higher value if there are many clusters.
 +
-Replace `fs_volume_size` with the volume size for the volume size for the `filesystemStorage` field, for example `20Gi`.
+Replace `fs_volume_size` with the size of the volume for the `filesystemStorage` field, for example `2Gi`. This value specifies how much storage is allotted for storing logs, manifests, and `kubeconfig` files for the clusters. You might need to use a higher value if there are many clusters.
```
filanov: suggesting to set it to 10Gi, to be on the safe side. 
@carbonin @romfreiman correct me if i'm wrong

cadawson: Made that update.

clusters/create_infra_env.adoc
```diff
@@ -70,9 +71,9 @@ spec:
       cpuArchitecture: "x86_64"
 ----
 +
-Replace `db_volume_size` with the volume size for the `databaseStorage` field, for example `10Gi`.
+Replace `db_volume_size` with the volume size for the `databaseStorage` field, for example `10Gi`. This value specifies how much storage is is allotted for storing files like database tables and database views for the clusters. You might need to use a higher value if there are many clusters.
 +
-Replace `fs_volume_size` with the volume size for the volume size for the `filesystemStorage` field, for example `20Gi`.
+Replace `fs_volume_size` with the size of the volume for the `filesystemStorage` field, for example `10Gi`. This value specifies how much storage is allotted for storing logs, manifests, and `kubeconfig` files for the clusters. You might need to use a higher value if there are many clusters.
```
romfreiman: allocated

filanov:
will try to check the avg size for each installed cluster. then maybe we can give guidance about the estimation of the size

filanov:
Looked a bit at our log server and looks like we need few MB for each cluster. in worst case lets say 20MB for each cluster. 
It depend on how many host there are but it's hard to say.

filanov:
I forgot about the must gather. so lets make it 50 per cluster

cadawson:
@filanov @romfreiman Made the updates, and changed the required amount to 50MB for each. I think we should be set here. Thanks!

filanov:
LGTM

cadawson:
Thanks @filanov ! I will merge. If additional changes are required, please open a new issue.

PR #3123, opened 2022-02-03
stolostron/backlog#18411 add Hive and ocp

cadawson:


clusters/create_infra_env.adoc
```diff
@@ -3,6 +3,8 @@
 
 You can use the {product-title} console to create an infrastructure environment to manage your hosts and create clusters on those hosts.
 
+When you create a cluster, the creation process uses the {ocp-short} installer with the Hive resource. If you have questions about cluster creation after completing this procedure, see   https://access.redhat.com/documentation/en-us/openshift_container_platform/4.9/html/installing/installing-on-bare-metal[Installing on bare metal] in the {ocp-short} documentation for more information about the process.
+
```
showeimer: CIM doesn't use OpenShift installer I think, so we may want to remove the statement here.

@jelkosz ?

jelkosz:
AFAIK it is. @filanov please keep me honest here.

filanov:
> AFAIK it is. @filanov please keep me honest here.

openshift installer is used when we install the cluster (installation actually starting), but it's not triggered by hive, at least not when we are talking about assisted-installer. 
With assisted-installer flow, hive just watch the agent cluster install CRs and copy the status to Hive's cluster deployment.

cadawson:
Removed the statement from infrastructure, @showeimer . Thanks!

showeimer (approved):

PR #3131, opened 2022-02-08
Bypass template processing grc

dockerymick:
https://github.com/stolostron/backlog/issues/18218

### DIFF
--- a/governance/custom_template.adoc
+++ b/governance/custom_template.adoc
@@ -8,7 +8,7 @@ Configuration policies support the inclusion of Golang text templates in the obj
 * <<hub-templates,Support for hub cluster templates in configuration policies>>
 ** <<template-processing,Template processing>>
 ** <<special-annotation-processing,Special annotation for reprocessing>>
-** <<modify-template-processing,Modify template processing>>
+** <<bypass-template-processing,Bypass template processing>>
 
 [#prerequisites-templatized]
 == Prerequisite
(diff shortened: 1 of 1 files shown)


governance/custom_template.adoc
```diff
@@ -453,3 +454,12 @@ See the following table for a comparison of hub cluster and managed cluster temp
 |===
 
 
+[#disable-template-processing]
+=== Disable template processing
+
+You might create a policy that contains a template that is not intended to be processed by {product-title-short}. By default, {product-title-short} processses all templates. 
+
+To disable template processing for your hub cluster, you must change `{{ template content }}` to `{{ '{{ template content }}' }}`.
```
gparvin: I have a small concern here that this wording sounds like the template processing is being disabled for the entire hub cluster.  The intent is for the user to modify the template so it does not get processed by ACM.  You need to modify each template in the policy that you don't want ACM to process in this way (or you do the alternative workaround that is mentioned after this)

gparvin (commented):
See comment on line 462 about a phrase that could be confusing.

governance/custom_template.adoc
dockerymick: @gparvin Thanks Gus for your feedback! I made some updates. When you have a chance, let me know your thoughts

governance/custom_template.adoc
```diff
@@ -453,3 +454,12 @@ See the following table for a comparison of hub cluster and managed cluster temp
 |===
 
 
+[#modify-template-processing]
+=== Modify template processing
+
+You might create a policy that contains a template that is not intended to be processed by {product-title-short}. By default, {product-title-short} processses all templates. 
+
+To modify template processing for your hub cluster, you must change `{{ template content }}` to `{{ '{{ template content }}' }}`.
```
gparvin: Would it be ok to say:
To bypass {product-title-short}'s template processing, you must ...

gparvin (commented):
I'm not sure if "bypass" is an acceptable word to use, but I think something more like the suggestion I added would help make it a little more clear what we're doing.

governance/custom_template.adoc
dockerymick: Yea, I think that its fine to say! I checked the IBM Style Guide and there were no mentions of it. Making the update now

gparvin (approved):

PR #3133, opened 2022-02-10
Bypass template processing 2.3

dockerymick:
https://github.com/stolostron/backlog/issues/18218

### DIFF
--- a/governance/custom_template.adoc
+++ b/governance/custom_template.adoc
@@ -5,7 +5,6 @@ Configuration policies support the inclusion of Golang text templates in the obj
 
 * <<prerequisites-templatized,Prerequisite>>
 * <<template-functions,Template functions>>
-* <<managed-templates,Support for managed cluster templates in configuration policies>>
 ** <<template-processing,Template processing>>
 ** <<special-annotation-processing,Special annotation for reprocessing>>
 ** <<bypass-template-processing,Bypass template processing>>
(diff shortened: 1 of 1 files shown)


gparvin (commented):
We didn't have hub templates support in ACM 2.3.  We did have the managed cluster templates support.  I'll have to let @ckandag comment on whether the annotation to disable the template processing was backported to 2.3.

ckandag:
yes, it was backported to 2.3

> We didn't have hub templates support in ACM 2.3. We did have the managed cluster templates support. I'll have to let @ckandag comment on whether the annotation to disable the template processing was backported to 2.3.

yes, it was backported to 2.3, while the hub templates are not in 2.4,  managedcluster templates can still be disabled by setting the annotation on the configuration policy definition.

but looks like you have ported entire hub template section as well which needs to be fixed. pls see inline comments

governance/custom_template.adoc
```diff
@@ -361,3 +368,72 @@ spec:
         {{ (fromConfigMap "site-config" "site1" "enabled")  | toBool }}
 ----
 
+[#managed-templates]
```
ckandag: this entire section needs to be removed for 2.3. as we do not support hub templates in 2.3, only the "bypassing" section should be included for 2.3

governance/custom_template.adoc
```diff
@@ -361,3 +368,72 @@ spec:
...
+| You can reference any resource on the cluster.
+
+| Processing errors
+| Errors from the hub cluster templates are displayed as violations on the managed clusters the policy applies to.
+| Errors from the managed cluster templates are displayed as violations on the specific target cluster where the violation occurred.
+|===
+
+
+[#bypass-template-processing]
+=== Bypass template processing
```
ckandag, line 423: this section is the only thing that needs to be added to 2.3

governance/custom_template.adoc
```diff
@@ -361,3 +368,72 @@ spec:
...
+
+You might create a policy that contains a template that is not intended to be processed by {product-title-short}. By default, {product-title-short} processses all templates. 
+
+To bypass template processing for your hub cluster, you must change `{{ template content }}` to `{{ '{{ template content }}' }}`.
```
ckandag: to bypass policy template processing, you must change ....

no need to say "for your hub cluster". - it actually isnt just for hub cluster isnt for every /any cluster

ckandag (changes requested):
pls see inline comments

ckandag (approved):

dockerymick:
Thanks for the approval @ckandag !

PR #3170, opened 2022-03-02
Adding the metadataComplianceType field

dockerymick:
https://github.com/stolostron/backlog/issues/18649

### DIFF
--- a/governance/custom_template.adoc
+++ b/governance/custom_template.adoc
@@ -103,7 +103,6 @@ spec:
     - default
   object-templates:
   - complianceType: musthave
-    metadataComplianceType: musthave
     objectDefinition:
       kind: ConfigMap
       apiVersion: v1
@@ -145,7 +144,6 @@ spec:
     - default
   object-templates:
   - complianceType: musthave
-    metadataComplianceType: musthave
     objectDefinition:
       kind: ConfigMap
       apiVersion: v1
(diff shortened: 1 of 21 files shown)


governance/custom_template.adoc
```diff
@@ -63,6 +63,7 @@ spec:
     - default
   object-templates:
   - complianceType: musthave
+    metadataComplianceType: musthave
```
willkutler: `metadataComplianceType` defaults to the same value as `complianceType` if it is not set, so it might be less confusing if we omit it in cases where it's not needed

governance/ocp_cis_policy.adoc
```diff
@@ -24,6 +24,7 @@ spec:
   severity: high
   object-templates:
     - complianceType: musthave # this template creates ScanSettingBinding:cis
+      metadataComplianceType: musthave # metadata will be processed as musthave, spec as mustonlyhave
```
willkutler: this comment seems to be duplicated a few times, I don't think it's needed since metadataComplianceType and complianceType are the same

governance/custom_template.adoc
dockerymick: @willkutler Should i add "metadataComplianceType defaults to the same value as complianceType if it is not set` as a comment in the sample? Or remove the entire parameter?

willkutler (approved):

dockerymick:
Thanks for your feedback and approval Will! @willkutler

PR #3218, opened 2022-03-14
stolostron/backlog#20089 firmware host

cadawson:


achuzhoy (commented):
/lgtm
nit: I'd personally use a different wording, because we have 2 sentences beginning with the same "If you do not have" wording. But the message is right.

cadawson:
@achuzhoy Thanks for the quick review. I changed the wording a bit. This is only 2.5, correct?

achuzhoy:
> @achuzhoy Thanks for the quick review. I changed the wording a bit. This is only 2.5, correct?

Looks good!
Thanks.

achuzhoy:
I saw it in 2.4

cadawson:
Made change to 2.4, as well.

PR #3250, opened 2022-03-16
stolostron/backlog#20374--new content for MCE

swopebe:


multicluster_engine/add-on_managed_service.adoc
```diff
@@ -0,0 +1,72 @@
+[#managed-serviceaccount-addon]
+= Enabling Managed-ServiceAccount add-ons
+
+By default, the `Managed-ServiceAccount` add-on is disabled. This component when enabled allows you to create or delete a service account on a managed cluster. 
+
+When a `ManagedServiceAccount` custom resource is created in the `<managed_cluster>` namespace on the hub cluster, a `ServiceAccount` is created on the managed cluster. 
+
+A `TokenRequest` is made with the `ServiceAccount` on the managed cluster to the Kubernetes API server on the managed cluster. The token is then stored in a `Secret` in the `<target_managed_cluster>` namespace on the hub cluster.
+
```
TheRealHaoLiu, line 11: we should put a note here that the Request Token have no audiance meaning that it can only be use in cluster

nathanweatherly: Text from doc input issue: 

```
The resulting token saved in the `Secret` is not the `ServiceAccount`'s token. It can be rotated and expire, unlike the `ServiceAccount`'s token.

The resulting token has a default audience of `empty` or `none`, meaning that it can only be used within the managed cluster or from the hub via the cluster proxy. It cannot authenticate with the external endpoint of the Kubernetes API server directly.
```

So looking at the existing draft from Brandi, I think we should add a Note

```
**Note**: The token can be rotated and expire. It has a default audience of `none`, meaning it cannot authenticate with the external endpoint fo the Kubernetes API server directly. It can only be used within the managed cluster.
```

(took out line about cluster proxy, since cluster proxy is ACM feature, not MCE feature)

This note is important to explain to users that using the managed serviceaccount component does not result in a privilege escalation.

nathanweatherly: @TheRealHaoLiu mentioned we might could want to add a link so users can learn more about tokenrequests and audiences

https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/

I can't remember how external links are handled in our doc, so i'm adding this comment for us to review later

nathanweatherly:
Brandi and I talked some via slack yesterday afternoon. To reiterate here so I don’t forget, we would very much like an API doc for the `ManagedServiceAccount` custom resource. A link to the CRD and also the RBAC support for the CRD is in the doc input issue: https://github.com/stolostron/backlog/issues/20374

nathanweatherly:
Other note concerning the ManagedServiceAccount API: The attribute `ttlSecondsAfterCreation` mentions the necessity of the EphemeralIdentity feature gate, which for MCE users is enabled by default. 

With that being said, we don't mention the feature gate anywhere else, and if a user looks at the ManagedServiceAccount doc, I'd hate for them to get confused by that explanation. 

Would a release note be appropriate here? 

Something like:

```
The EphemeralIdentity feature gate, which is required by the Managed ServiceAccount Tech Preview component, is enabled by default.
```

multicluster_engine/add-on_managed_service.adoc
swopebe: @TheRealHaoLiu I don't understand what you are saying here?

nathanweatherly: We should add a note here saying

> **Note**: The token can be rotated and expire. It has a default audience of `none`, meaning it cannot authenticate with the external endpoint fo the Kubernetes API server directly. It can only be used within the managed cluster.

nathanweatherly (approved):

nathanweatherly:
I guess /approve is different from "Approving"

nathanweatherly:
@swopebe Can/should we add a banner saying explicitly that this is Tech Preview?

swopebe:
@nathanweatherly we have a rule that we write it as such:

**Technology Preview**

nathanweatherly:
That works!

swopebe:
This was reviewed, approved, built, and discussed in a meeting. Closing.

PR #3302, opened 2022-04-01
stolostron/backlog#20771 dynamic plugin

swopebe:
--new doc for dynamic plugin

### DIFF
--- a/console/console.adoc
+++ b/console/console.adoc
@@ -19,7 +19,7 @@ image:../images/nav-2.5.png[Navigation]
 [#ocp-plug-ins]
 == {ocp-short} hybrid plug-ins
 
-**Technology Preview:** With {ocp-short} 4.10 and {product-title-short} 2.5, {product-title-short} web console can be integrated into the {ocp-short} console. To use this feature, the console plug-ins must remain enabled.
+With {ocp-short} 4.10 and {product-title-short} 2.5, {product-title-short} web console can be integrated into the {ocp-short} console. To use this feature, the console plug-ins must remain enabled.
 
 For new installations, the console plug-ins are enabled by default. If you upgraded from a previous version of {product-title-short} and want to enable the plug-ins, or if you want to disable the plug-ins, see the following instructions:
 
--- a/console/console.adoc
+++ b/console/console.adoc
@@ -16,6 +16,21 @@ image:../images/nav-2.5.png[Navigation]
 * <<governance-nav,Governance>>
 * <<credentials-nav,Credentials>>
 
+**Technology Preview:** With {ocp-short} 4.10 and {product-title-short} 2.5, {product-title-short} web console can be integrated into the OpenShift console as a technology preview. This requires both the {product-title-short} and mce console plugins to be enabled.
+
+For new installations, the {product-title-short} and mce console plugins are enabled by default. If you upgraded from a previous version of {product-title-short} and want to enable the plugins, or if you'd prefer to disable them, from the 
+
+. To disable the plugin, be sure you are in the _Administrator_ perspective in the {ocp-short} console.
+. Find *Adminstration* in the navigation and click *Cluster Settings*, then click _Configuration_ tab. 
+. From the list of _Configuration resources_, click the **Console** resource, which contains cluster-wide configuration for the web console. 
+. Click on the _Console plugins_ tab. Both the `acm` and `mce` plugins are listed. 
+. Modify plugin status from the table. In a few moments, you are prompted to refresh the console.
+
+For {ocp-short} 4.10 with the plugins enabled, you can access {product-title-short} within the {ocp-short} console from the cluster switcher by selecting *All clusters* from the drop-down menu. Without plugins enabled on {ocp-short} versions 4.8 through 4.10, {product-title-short} is available in the perspective switcher.
+
+
+You can also click *Networking* > *Routes* to access the standalone or separate {product-title-short} web console.
+
 [#home-page]
 == Home
 
--- a/console/console.adoc
+++ b/console/console.adoc
@@ -16,20 +16,20 @@ image:../images/nav-2.5.png[Navigation]
 * <<governance-nav,Governance>>
 * <<credentials-nav,Credentials>>
 
-[#ocp-plugins]
-== {ocp-short} hybrid plugins
+[#ocp-plug-ins]
+== {ocp-short} hybrid plug-ins
 
-**Technology Preview:** With {ocp-short} 4.10 and {product-title-short} 2.5, {product-title-short} web console can be integrated into the {ocp-short} console. To use this feature, the console plugins must remain enabled.
+**Technology Preview:** With {ocp-short} 4.10 and {product-title-short} 2.5, {product-title-short} web console can be integrated into the {ocp-short} console. To use this feature, the console plug-ins must remain enabled.
 
-For new installations, the console plugins are enabled by default. If you upgraded from a previous version of {product-title-short} and want to enable the plugins, or if you want to disable the plugins, see the following instructions:
+For new installations, the console plug-ins are enabled by default. If you upgraded from a previous version of {product-title-short} and want to enable the plug-ins, or if you want to disable the plug-ins, see the following instructions:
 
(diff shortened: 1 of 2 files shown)


console/console.adoc
```diff
@@ -16,6 +16,23 @@ image:../images/nav-2.5.png[Navigation]
 * <<governance-nav,Governance>>
 * <<credentials-nav,Credentials>>
 
+[#ocp-plug-ins]
+== {ocp-short} hybrid plug-ins
```
KevinFCormier: I'm not sure if the word "hybrid" will mean anything to the user here. I would probably say "console" instead.

console/console.adoc
```diff
@@ -16,6 +16,23 @@ image:../images/nav-2.5.png[Navigation]
...
+
+With {ocp-short} 4.10 and {product-title-short} 2.5, {product-title-short} web console can be integrated into the {ocp-short} console. To use this feature, the console plug-ins must remain enabled.
+
+For new installations, the console plug-ins are enabled by default. If you upgraded from a previous version of {product-title-short} and want to enable the plug-ins, or if you want to disable the plug-ins, see the following instructions:
+
+. To disable the plug-in, be sure you are in the _Administrator_ perspective in the {ocp-short} console.
+. Find *Adminstration* in the navigation and click *Cluster Settings*, then click _Configuration_ tab. 
+. From the list of _Configuration resources_, click the **Console** resource, which contains cluster-wide configuration for the web console.
```
KevinFCormier: There are 2 **Console** resources in the list.
image

We must specify that it is the second one here, with the API group `operator.openshift.io`

console/console.adoc
```diff
@@ -16,6 +16,23 @@ image:../images/nav-2.5.png[Navigation]
...
+== {ocp-short} hybrid plug-ins
+
+With {ocp-short} 4.10 and {product-title-short} 2.5, {product-title-short} web console can be integrated into the {ocp-short} console. To use this feature, the console plug-ins must remain enabled.
+
+For new installations, the console plug-ins are enabled by default. If you upgraded from a previous version of {product-title-short} and want to enable the plug-ins, or if you want to disable the plug-ins, see the following instructions:
+
+. To disable the plug-in, be sure you are in the _Administrator_ perspective in the {ocp-short} console.
+. Find *Adminstration* in the navigation and click *Cluster Settings*, then click _Configuration_ tab. 
+. From the list of _Configuration resources_, click the **Console** resource, which contains cluster-wide configuration for the web console. 
+. Click on the _Console plug-ins_ tab. Both the `acm` and `mce` plug-ins are listed.
```
KevinFCormier, line 29: I know the preferred form uses the hyphen for "plug-ins" but this is not what is used in the OCP UI:
image

KevinFCormier: I have been meaning to open an issue for the mismatch between their UI and their documentation. I will do that and see which they intend to change.

console/console.adoc
```diff
@@ -16,6 +16,23 @@ image:../images/nav-2.5.png[Navigation]
...
+
+For new installations, the console plug-ins are enabled by default. If you upgraded from a previous version of {product-title-short} and want to enable the plug-ins, or if you want to disable the plug-ins, see the following instructions:
+
+. To disable the plug-in, be sure you are in the _Administrator_ perspective in the {ocp-short} console.
+. Find *Adminstration* in the navigation and click *Cluster Settings*, then click _Configuration_ tab. 
+. From the list of _Configuration resources_, click the **Console** resource, which contains cluster-wide configuration for the web console. 
+. Click on the _Console plug-ins_ tab. Both the `acm` and `mce` plug-ins are listed. 
+. Modify plug-in status from the table. In a few moments, you are prompted to refresh the console.
+
+For {ocp-short} 4.10 with the plug-ins enabled, you can access {product-title-short} within the {ocp-short} console from the cluster switcher by selecting *All clusters* from the drop-down menu. Without plug-ins enabled on {ocp-short} versions 4.8 through 4.10, {product-title-short} is available in the perspective switcher.
```
KevinFCormier: In the UI, they use a capital C for "All Clusters"
image

release_notes/whats_new.adoc
```diff
@@ -26,7 +26,7 @@ Learn more about what is new this release:
...
+* For {ocp-short} 4.10, with the https://docs.openshift.com/container-platform/4.10/release_notes/ocp-4-10-release-notes.html#ocp-4-10-dynamic-plugin-technology-preview[{ocp} 4.10 release] and a more hybrid console, you can use dynamic plug-ins. See the {ocp-short} documentation about https://docs.openshift.com/container-platform/4.10/web_console/dynamic-plug-ins.html#dynamic-plug-ins_dynamic-plug-ins[Adding a dynamic plug-in to the OpenShift Container Platform web console] to create and deploy a dynamic plug-in on your cluster that is loaded at run-time. Learn how to enable plug-ins for {product-title-short} in the xref:../console/console.adoc#console-overview[Console overview].
```
KevinFCormier: I'm not sure that it makes sense to link to this documentation here. ACM delivers a console plugin that customers can use. That documentation tells customers how to create their own plugins. Secondly, plugins are Tech Preview in general, but they are fully supported for Red Hat products like us. This link includes the term "technology preview" so it might make it seem like our plugins are a tech preview.

@showeimer do you agree?

KevinFCormier (changes requested):
Thanks @swopebe. Just a few corrections to make. Please see my comments.

console/console.adoc
KevinFCormier: Opened BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2075634

console/console.adoc
swopebe: Makes sense -- thanks, sorry for the delay, I was on PTO.

console/console.adoc
swopebe: Thanks @KevinFCormier sometimes, especially with UI, the words don't match the style manual. Style manual rules what we do, so they need to change that and I will mention to OCP writers (we often discuss style on our channel). I will leave as-is for now. In this case, they should change it, but we are correct.

console/console.adoc
swopebe: Again, style says use sentence format but OCP does headline format at times. I will also ask about this when I get a chance... not a big deal to change this one.

release_notes/whats_new.adoc
swopebe: I can be clearer that it is tech preview and to go here. We don't usually make a What's new entry without linking to a doc, but if you feel strongly about it...

console/console.adoc
swopebe: Changed to:

From the list of _Configuration resources_, click the **Console** resource with the `operator.openshift.io` API group, which contains cluster-wide configuration for the web console.

console/console.adoc
swopebe: *All Clusters* from the drop-down menu

release_notes/whats_new.adoc
swopebe: @KevinFCormier I edited a bit here. OCP team asked for us to link to their doc from our release note, and we agreed. I separated their plug and ours. See if this works.

console/console.adoc
KevinFCormier: Good enough. Sounds like they might change it in the UI at some point.

release_notes/whats_new.adoc
KevinFCormier: @swopebe May I ask who from OCP requested this, or can I get some more explanation for the rationale? It's good that you removed the mention of tech preview since it does not apply to our use-case, but I still don't understand the motivation for linking to that documentation.

swopebe: @KevinFCormier the documentation team and manager wants us to point to theirs because they start reaching dynamic plug in from the OCP console and they want to point to our doc when we GA. I think it's not necessary but also not threatening in anyway.

swopebe: Can we merge, build, close?

KevinFCormier (approved):

PR #3304, opened 2022-04-01
stolostron/backlog#19748 add taint tol

cadawson:


haoqing0110 (changes requested):
Could we also add taint & toleration reference on the placement page? as it's also part of the placement feature and users should know it.

cadawson:
Hello @haoqing0110 . We can certainly add a reference. The way that it is structured now, is that the taints and tolerations topic will be a subtopic within the placement topic. It will be at the end of it. It is hard to see that the way in the PR. Does that address your concern? Thanks!

haoqing0110:
@cadawson yes, a subtopic of taints and tolerations within the placement topic would be good. 

We have a similar doc in 
https://open-cluster-management.io/concepts/managedcluster/#taints-of-managedclusters
https://open-cluster-management.io/concepts/placement/#taintstolerations ,mentioned taints and tolerations in both managedcluster and placement page, hope this can help express my thoughts.

clusters/man_clust_taints_tol.adoc
```diff
@@ -0,0 +1,151 @@
+[#taints-tolerations-manclust]
+= Using taints and tolerations to place managed clusters
+
+You can control the placement of your managed clusters or managed cluster sets by using taints and tolerations. Taints and tolerations provide a way to prevent managed clusters from being placed on certain placements. This control can be helpful if you want to prevent certain managed clusters from being included in some placements. You can add a taint to the managed cluster, and add a toleration to the placement. If the taint and the toleration do not match, then the managed cluster is not placed in that placement.
```
elgnay: Some minor comments. I'm not sure about if `place` is the right verb in the sentences below. I prefer to use `select` instead. What do you think?

> Taints and tolerations provide a way to prevent managed clusters from being placed on certain placements.

Taints and tolerations provide a way to prevent managed clusters from being __selected__ by certain placements.

> If the taint and the toleration do not match, then the managed cluster is not placed in that placement.

If the taint and the toleration do not match, then the managed cluster is not __selected__ by that placement.

cadawson:
Thank you @haoqing0110 and @elgnay . I appreciate your review. I think the current structure will work well with its relationship to ManagedClusterSet. We can make sure after we merge that it looks ok, but I think it is similar to the example that you sent.
I agree that "selected" fits better in those cases than "placed," and I made those updates. I think we are set with this one. Thank you!

cadawson:
Are you ok with this @haoqing0110 ? Thanks!

haoqing0110 (approved):
/lgtm
/approved

PR #3307, opened 2022-04-05
Update gitops_config.adoc (Planned for Sprint 6--Brandi)

rokej:
https://github.com/stolostron/backlog/issues/21254

### DIFF
--- a/applications/gitops_config.adoc
+++ b/applications/gitops_config.adoc
@@ -65,3 +65,31 @@ spec:
 ----
 
 . Save your changes. You can now follow the GitOps workflow to manage your applications. See link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.9/html-single/cicd/index#understanding-openshift-gitops[About GitOps] to learn more.
+
+
+*Important security notes:*
+
+When integrating with GitOps operator, for every managed cluster bound to the gitops namespace (through the placement and ManagedClusterSetBinding custom resources ), a secret with a token to access the MangedCluster is created in that namespace. This is required for the Gitops controller to sync resources to the remote managed cluster. When a user is given admin access to a gitops namespace to perform Application LifeCycle operations, the user will also gain access to this secret and thereby admin level to the managed cluster. If this is not acceptable, instead of binding the user to the namespace-scoped admin role, a more restrictive custom role with just the permissions required to work with App lifecycle specific resources can be create and used to bound the user. For example, 
+
++
+[source,yaml]
+----
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: application-set-admin
+rules:
+- apiGroups:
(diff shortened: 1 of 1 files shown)


applications/gitops_config.adoc
```diff
@@ -65,3 +65,31 @@ spec:
 ----
 
 . Save your changes. You can now follow the GitOps workflow to manage your applications. See link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.9/html-single/cicd/index#understanding-openshift-gitops[About GitOps] to learn more.
+
+
+*Important security notes:*
+
+When integrating with GitOps operator, for every managed cluster bound to the gitops namespace (through the placement and ManagedClusterSetBinding custom resources ), a secret with a token to access the MangedCluster is created in that namespace. This is required for the Gitops controller to sync resources to the remote managed cluster. When a user is given admin access to a gitops namespace to perform Application LifeCycle operations, the user will also gain access to this secret and thereby admin level to the managed cluster. If this is not acceptable, instead of binding the user to the namespace-scoped admin role, a more restrictive custom role with just the permissions required to work with App lifecycle specific resources can be create and used to bound the user. For example, 
+
```
dockerymick, line 73: Thanks for pulling me in. I will leave the updates to Brandi. Here are my suggestions:


When you integrate with the GitOps operator for every managed cluster bound to the GitOps namespace, through the placement and `ManagedClusterSetBinding` custom resources, a secret with a token to access the `ManagedCluster` is created in the namespace. This is required for the GitOps controller to sync resources to the remote managed cluster. When a user is given administrator access to a GitOps namespace to perform application lifecycle operations, the user also gains access to this secret and `admin` level to the managed cluster.

If this is not desired, instead of binding the user to the namespace-scoped `admin` role, use a more restrictive custom role with permissions required to work with application resources that can be created and used to bound the user. View the following example:

is remote cluster the same as saying local cluster?

dockerymick (changes requested):
Your addition looks great! Thanks for including me, I made a suggestion

applications/gitops_config.adoc
swopebe: Thanks @dockerymick 

I made the changes, but I still do agree that as far as we can tell "remote" and "managed" are the same. (local cluster is the hub when managed)....

@rokej can we remove `remote` or is there a reason for it? Thanks!

rokej: You mean `remote managed cluster`? Yes we can remove `remote` and just say `managed cluster`

dockerymick (approved):
Thanks for accepting my suggestions. The changes look good to me! I will leave the merge to Brandi or Roke if all is well

applications/gitops_config.adoc
swopebe: Done, thanks @rokej

PR #3308, opened 2022-04-06
Add new fields to advanced configuration sections (Sprint 6--Brandi)

JakobGray:
Issue: https://github.com/stolostron/backlog/issues/21434

Signed-off-by: Jakob Gray <20209054+JakobGray@users.noreply.github.com>

### DIFF
--- a/multicluster_engine/adv_config_install.adoc
+++ b/multicluster_engine/adv_config_install.adoc
@@ -120,9 +120,9 @@ spec:
 
 The `Managed-ServiceAccount` add-on can be enabled after creating `MultiClusterEngine` by editing the resource on the command line, setting the `managedserviceaccount-preview` component to `enabled: true`. Alternatively, you can run the following command and replace <multiclusterengine-name> with the name of your `MultiClusterEngine` resource.
 
-```
+----
 oc patch MultiClusterEngine <multiclusterengine-name> --type=json -p='[{"op": "add", "path": "/spec/overrides/components/-","value":{"name":"managedserviceaccount-preview","enabled":true}}]'
-```
+----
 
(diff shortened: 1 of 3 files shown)


multicluster_engine/adv_config_install.adoc
```diff
@@ -58,11 +58,104 @@ spec:
...
+The above infra-node toleration is set on pods by default without specifying any tolerations in the config. Customizing tolerations in the config will replace this default.
+
+== Enabling and Disabling Components
+
+Some components are not enabled by default and must be enabled in the `MultiClusterEngine` spec to use. These include the ManagedServiceAccount and Hypershift add-ons. These components are in a tech-preview state and not fully supported as of {product-title-short} 2.0. 
+
 [#managed-serviceaccount-addon-intro]
-== ManagedServiceAccount add-on
+=== Enabling ManagedServiceAccount add-on
+
```
swopebe, line 105: @JakobGray I edited a bit and this header will need to be parallel to the others. All lists either start with a verb or a noun, so I will make this one match the others.

JakobGray: `Enabling ClusterProxyAddon` and `Enabling ManagedServiceAccount add-on` are subsections of `Enabling and Disabling Components`. Which is why I gave it three '='s (I think I saw that somewhere else)

swopebe: We cannot have a section "enabling" because it does not fit the purpose and other headers. I called them non-default. We can have a meeting to discuss what you are going for here but the header for enabling is removed.

swopebe (approved):
live meeting and changes to structure for both

PR #3316, opened 2022-04-08
stolostron/backlog#21588 Add 2.2.12 errata

cadawson:


juliana-hsu (approved):
I assume the Errata 2.2.11 text is removed?

cadawson:
We just add the new errata to the list, so it provides a list for each of the z-streams. Thank you! @juliana-hsu

PR #3320, opened 2022-04-11
adding obs what's new

dockerymick:
https://app.zenhub.com/workspaces/engineering-backlog-do-not-delete-604fab62d4b98d00150a2854/issues/stolostron/backlog/19745


release_notes/whats_new.adoc
```diff
@@ -32,8 +32,8 @@ Additionally, _Search_ is available in the navigation from the _Home_ tab and no
 
 [#observability-whats-new]
 === Observability
-
-//New content here
+// what doc should the user be referenced to? Should we reference the Designing your Grafana dashboard section? 
+* {product-title-short} supports {ocp-short} version 3.11 Grafana dashboards. See, link:../.../[] for more details.
```
morvencao: @dockerymick OCP 3.11 dashboards are delivered out-of-the-box, so we don't need to provide the `Designing your Grafana dashboard section` link, that's is for customizing the dashboard.
Maybe a better solution would be describe where are the OCP 3.11 dashboards. So I suggest to change: `See, link:../.../[] for more details` to:

After you enable Observability services, open the Grafana dashboards URL: `https://$ACM_URL/grafana/dashboards` with your browser, then you should find a folder named `OCP 3.11`, all the OCP 3.11 dashboards are located in that folder.

dockerymick: Thanks for your patience and feedback @morvencao. i think it would be best to add your suggested statement as a note on the [Enable observability service](https://github.com/stolostron/rhacm-docs/blob/2.5_stage/observability/observability_enable.adoc#enabling-observability) section. Thoughts?

morvencao: @dockerymick That makes sense to me.

dockerymick:
@morvencao Thanks for your patience! I have made a couple of more updates. When you have an opportunity may you re-review the PR?

morvencao (commented):
thanks @dockerymick 
/LGTM

PR #3324, opened 2022-04-12
update the klusterletaddonconfig for 2.4

zhiweiyin318:
Signed-off-by: Zhiwei Yin <zyin@redhat.com>

remove deprecated fields in klusterletaddonconfig

### DIFF
--- a/clusters/import_cli.adoc
+++ b/clusters/import_cli.adoc
@@ -198,15 +198,10 @@ metadata:
   name: <cluster_name>
   namespace: <cluster_name>
 spec:
-  clusterName: <cluster_name>
-  clusterNamespace: <cluster_name>
   applicationManager:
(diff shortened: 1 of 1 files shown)


cadawson (changes requested):
This looks good, @zhiweiyin318 . Just to be sure, was this change made to all of the 2.4 versions, or was it changed in a z-stream? If it was in a z-stream, we might want to mention when we removed it. If it was all of 2.4, and we just did not remove it when it released, then we should be ok. Thanks!

zhiweiyin318:
@cadawson this change is for all 2.4 version. these fields are deprecated from 2.4.0.

cadawson:
Perfect! Thank you @zhiweiyin318 ! I will merge.

PR #3329, opened 2022-04-13
service account as subscription admin

rokej:

https://github.com/stolostron/backlog/issues/21309

Add doc to show that a service account can be used in cluster role binding for subscription admin.


dockerymick (approved):
Changes look good! Approved

PR #3339, opened 2022-04-18
stolostron/backlog#20395 add default mgdclstset

cadawson:


ldpliu (approved):

ldpliu:
@cadawson Thanks for your work, and I approved the pr.

And as the default ManagedCluster set name is "default".
So I am not sure if it's better to highlight the default  set name in doc like: `default`

cadawson:
We were thinking about the same thing @ldpliu . I made a few changes that now calls out the name of the managed cluster set as 'default'. Thanks!

PR #3352, opened 2022-04-20
stolostron/backlog#20928 rm fips note volsync

cadawson:


tesshuflower (approved):
/lgtm

Thanks for the quick update Chris!

release_notes/fips_readiness.adoc
```diff
@@ -12,11 +12,13 @@ Read the following limitations with {product-title-short} and FIPS.
 
 * The following Technology preview components are not FIPS ready:
 +
-** Hub cluster backup and restore
-** {assist-install}
-** Submariner
-** VolSync
+** Integrity Shield
+** Hub of hubs (Developer preview)
```
dockerymick: Can you explain why we are placing a developer preview component in the product doc? It is also being listed after stating "which Technology preview" components are not FIPS ready

release_notes/fips_readiness.adoc
```diff
@@ -12,11 +12,13 @@ Read the following limitations with {product-title-short} and FIPS.
...
 
 * Persistent Volume Claim (PVC) and S3 storage that is used by the search and observability components must be encrypted when you configure the provided storage. {product-title-short} does not provide storage encryption, see the {ocp-short} documentation, link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.9/html/installing/installing-fips[Support for FIPS cryptography].
 
-* When you provision managed clusters from {product-title-short}, you must set `fips: true` in the `install-config.yaml` file before you deploy the new managed cluster.
+* When you provision managed clusters using the {product-title-short} console, select the following check box in the _Cluster details_ section of the managed cluster creation to enable the FIPS standards: 
++
+----
+FIPS with information text: Use the Federal Information Processing Standards (FIPS) modules provided with RHCOS instead of the default Kubernetes cryptography suite file before you deploy the new managed cluster
```
dockerymick: Should it be RHOCS? Should there also be a period at the end of this statement ?

dockerymick (changes requested):
Thanks for pulling me in Chris. I left a couple of comments and questions

release_notes/fips_readiness.adoc
cadawson: I know that we don't document dev preview in the product docs. In this case, I thought that FIPS compliance would be a pretty big deal, even if you are using a dev preview version. You bring up a good point, though. They should not be using it for any production content if it is dev preview (or tech preview, for that matter). So maybe it is not necessary to mention here.

release_notes/fips_readiness.adoc
cadawson: In the UI today, it is RHOCS. You are correct that it should be Red Hat Enterprise Linux CoreOS. Randy made an update when we went through the UI, but I haven't seen it in the playback instance yet. I went ahead and changed it in the docs, as I think it will make it into the UI by GA. 
Yes, I added a period at the end. 
Thanks!

release_notes/fips_readiness.adoc
gparvin: Yes, I think it's find to take it out.  I think we have other dev preview components too so since they likely wouldn't be FIPS compliant either we should probably not list dev preview items or have a general statement indicating dev preview items are not reviewed for FIPS readiness.

dockerymick (approved):
Thanks for answering my questions and pulling me in for the review. I believe the last change left is to remove the dev preview. Other than that, this looks good to me!

cadawson:
Hi @gparvin . I removed the Hub of hubs developer preview item, as we don't document dev preview items in the formal documentation. Did you want to take another look at this one before we merge? Thanks!

gparvin:
@cadawson We can only support fips on `x86_64`.  See the first sentence here: https://access.redhat.com/documentation/en-us/openshift_container_platform/4.10/html/installing/installing-fips
I think we should reword the sentence indicating that we only support x86_64 matching OpenShift instead of listing platforms we don't support FIPS on.

Other than that I think this looks good!  Thanks

cadawson:
Thanks @gparvin . I updated that bullet to which architectures it is supported on. Want one last look?

gparvin (approved):

dockerymick:
Based on the approvals and confirmation from Chris via Slack, I am merging this issue.

PR #3373, opened 2022-04-26
stolostron/backlog#20366--addon

swopebe:


multicluster_engine/addon_managed_service.adoc
```diff
@@ -1,7 +1,7 @@
 [#managed-serviceaccount-addon]
 = Enabling ManagedServiceAccount add-ons
```
swopebe, line 3: Should this topic be tech preview?

JakobGray: The managedserviceaccount is tech preview, yes

swopebe: Changing that, then will merge and close. THANKS!

PR #3377, opened 2022-04-29
Update create_policy.adoc

dockerymick:
very small change
https://app.zenhub.com/workspaces/engineering-backlog-do-not-delete-604fab62d4b98d00150a2854/issues/stolostron/backlog/21357


dockerymick:
Removed the space in between `link:` and the real url

PR #3382, opened 2022-05-02
stolostron/backlog#21091

swopebe:
--changing a note

### DIFF
--- a/clusters/import_cli.adoc
+++ b/clusters/import_cli.adoc
@@ -116,7 +116,7 @@ type: Opaque
 oc apply -f auto-import-secret.yaml
 ----
 +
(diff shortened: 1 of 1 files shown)


cadawson (changes requested):
By default, the auto-import secret is used one time and deleted when the import process completes. If you want to keep the auto import secret, add `managedcluster-import-controller.open-cluster-management.io/keeping-auto-import-secret` to the secret. You can add it by running the following command:

cadawson:
Thanks for picking this up, @swopebe . Added a suggestion for the text. Otherwise, it looks good to me!

swopebe:
@cadawson can you look again?

cadawson (approved):
Works for me! Thanks a lot!

PR #3390, opened 2022-05-04
stolostron/backlog#21314

swopebe:
--removal


xiangjingli:
@swopebe  @qiujian16   I double checked the `Applications` part is correct.  The rest parts needs foundation squad check

PR #3391, opened 2022-05-04
Updating backup and restore docs

dockerymick:
https://app.zenhub.com/workspaces/engineering-backlog-do-not-delete-604fab62d4b98d00150a2854/issues/stolostron/backlog/21861


clusters/backup_and_restore.adoc
```diff
@@ -142,58 +174,37 @@ spec:
...
-
-[#managed-cluster-activation-data]
-=== Managed cluster activation data
+[#clean-hub-restore]
+=== Clean the hub cluster before restore
 
-Managed cluster activation data is the backup data that is being actively managed by the hub cluster, when restored on the hub cluster. Activation data resources are stored by the managed clusters backup and by the resource-generic backup, using the `cluster.open-cluster-management.io/backup: cluster-activation` label. 
+Velero currently skips existing backed up resources on the hub cluster. This limits the scenarios that can be used when you restore hub clsuter data on a new hub cluster. If the new hub cluster is not used and the restore is applied only once, the hub cluster is not the recommended option to use for passive configuration. The data on the new hub cluster is not reflective of the data available with the restored resources.
```
sahare: typo in `clsuter data` on line 190

clusters/backup_and_restore.adoc
```diff
@@ -142,58 +174,37 @@ spec:
 
 **Notes**:
 
-** Only managed clusters created using the `hive` API are automatically imported when the backup is restored on another hub cluster. All other managed clusters show up as _Pending Import_ and must be manually imported back on the new hub cluster.
+** Only managed clusters created by the Hive API are automatically connected with the new hub cluster when the `acm-managed-clusters` backup, from the _managed clusters_ backup is restored on another hub cluster. All other managed clusters are in the `Pending Import` state and must be imported back onto the new hub cluster. The Hive API managed clusters can bbe connected with the new hub cluster because Hive provides the `kubeconfig` file to connect to the managed cluster. This is backed up and restored on the new hub cluster. The import controller updates the bootstrap `kubeconfig` file on the managed cluster using the restored configuration. The `kubeconfig` file is only available for managed clusters created by using the Hive API.
```
sahare: typo in `can bbe connected`

clusters/backup_and_restore.adoc
```diff
@@ -233,6 +243,55 @@ spec:
...
+
+[#managed-cluster-activation-data]
+=== Managed cluster activation data
+
+Managed cluster activation data is the backup data that is being actively managed by the hub cluster, when restored on the hub cluster. Activation data resources are stored by the managed clusters backup and by the resource-generic backup, using the `cluster.open-cluster-management.io/backup: cluster-activation` label. 
+
+[#resources-restored-managed-cluster]
+=== Resources restored at managed activation time
+
+When the `cluster.open-cluster-management.io/backup: cluster-activation` label is added to a resource the resource is automatically backed up in the `acm-resources-generic-schedule` backup resource. If any of these resources need to be restored, only when the managed clusters are moved to the new hub cluster, you have to set the label value to `cluster-activation` when the `veleroManagedClustersBackupName:latest` label is used in the restored resource.
```
sahare: The text that starts on line 264 and ends before the sample is very confusing, the right description can be:

`Managed clusters activation data or activation data, is backup data which when restored on a new hub will result in managed clusters being actively managed by the hub where the restore was executed. Activation data resources are stored by the managed clusters backup, and by the resources-generic backup, using the cluster.open-cluster-management.io/backup: cluster-activation label.

When you add the cluster.open-cluster-management.io/backup label to a resource, this resource is automatically backed up under the acm-resources-generic-schedule backup. If any of these resources need to be restored only when the managed clusters are moved to the new hub, so when the veleroManagedClustersBackupName:latest is used on the restored resource, then you have to set the label value to cluster-activation. This will ensure the resource is not restored unless the managed cluster activation is called.`

release_notes/known_issues.adoc
```diff
@@ -840,3 +840,16 @@ Avoid backup collisions by setting the `BackupSchedule.cluster.open-cluster-mana
...
+
+* Since Velero skips existing resources, the policy on the new hub cluster is unchanged. Therefore, the policy is not the same as the one backed up on the initial hub cluster.
+
+* The new hub cluster has a different configuration from the active hub cluster when a user reapplies the backup on the new hub cluster. Since there is an existing policy on the hub cluster from a previous restore, it is not restored again. Even when the backup contains the expected updates, the policy contents are not updated by Velero on the new hub cluster. 
+
+To address the previously mentioned limitations, create a `restore.cluster.open-cluster-management.io` resource for the cluster backup and restore operator runs a set of steps, which cleans up the hub cluster before Velero restore begins. For more information, see link:../clusters/backup_and_restore.adoc#clean-hub-restore[Clean the hub cluster before restore].
```
sahare: Not sure if we should add this restore limitation part under known issues, it's basically explaining a limitation in Velero restore operation, so we have implemented a solution to resolve it, these limitations are taken from the section that you describe this Velero restore limitation and the cleanup options, but these are examples of that Velero limitation and doesn't make sense to be here on its own, the text below explains it better:

`Velero currently skips backed up resources if they are already installed on the hub. This limits the scenarios that can be used when restoring hub data on a new hub. Unless the new hub is not used and the restore is applied only once, the hub could not be relibly used as a passive configuration: the data on this hub is not reflective of the data available with the restored resources.

Restore limitations examples:

A Policy exists on the new hub, before the backup data is restored on this hub. After the restore of the backup resources, the new hub is not identical with the initial hub from where the data was restored. The Policy should not be running on the new hub since this is a Policy not available with the backup resources.

A Policy exists on the new hub, before the backup data is restored on this hub. The backup data contains the same Policy but in an updated configuration. Since Velero skips existing resources, the Policy will stay unchanged on the new hub, so the Policy is not the same as the one backed up on the initial hub.

A Policy is restored on the new hub. The primary hub keeps updating the content and the Policy content changes as well. The user reapplies the backup on the new hub, expecting to get the updated Policy. Since the Policy already exists on the hub - created by the previous restore - it will not be restored again. So the new hub has now a different configuration from the primary hub, even if the backup contains the expected updated content; that content is not updated by Velero on the new hub.

To address above limitations, when a Restore.cluster.open-cluster-management.io resource is created, the Cluster Back up and Restore Operator runs a prepare for restore set of steps which will clean up the hub, before Velero restore is called.`

clusters/backup_and_restore.adoc
```diff
@@ -1,7 +1,7 @@
...
+The cluster backup and restore operator provides disaster recovery solutions for when {product-title} hub cluster goes down and needs to be recreated. It runs on the hub cluster and depends on the link:https://github.com/openshift/oadp-operator[OADP Operator] to install Velero, and to create a connection from the hub cluster to the backup storage location where the data is stored. Velero is the component that runs the backup and restore operations. The cluster backup and restore support for all {product-title-short} hub cluster resources, like managed clusters, applications, policies, and bare metal assets.
```
sahare: It should be `The cluster backup and restore operator solution provides backup and restore support for all`

sahare (changes requested):
Thanks @dockerymick, I've added some comments.

clusters/backup_and_restore.adoc
dockerymick: Great catch, thank you!

release_notes/known_issues.adoc
dockerymick: What are your thoughts about me specifying that these are Velero restore limitations. I disagree with listing limitations in the topic. i think it is best to let the user know the recommended step rather than just stating what is wrong

clusters/backup_and_restore.adoc
dockerymick: Updated

release_notes/known_issues.adoc
sahare: Sure, just this text still needs to be updated:
`To address the previously mentioned limitations, create a `restore.cluster.open-cluster-management.io` resource for the cluster backup and restore operator runs a set of steps, which cleans up the hub cluster before Velero restore begins. `

dockerymick: @sahare thanks for your feedback. Why do you all have "prepare for restore" set of steps? Is that the name of a script? Should the following be monospaced: `prepare-for-restore` set of steps

sahare: @dockerymick No, it's just a process in our code which runs a set of steps to prepare for restore by cleaning up the hub cluster before the Velero restore begins.

dockerymick: @sahare Thanks for explaining! I think we are good on this PR now. May you do another review and let me know if i missed anything. If all is well, will you approve the PR?

sahare (approved):
Looks good to me, Thanks!

PR #3401, opened 2022-05-09
Document that the lookup template function can return an empty map

mprahl:
Relates:
https://github.com/stolostron/backlog/issues/21729

### DIFF
--- a/governance/custom_template.adoc
+++ b/governance/custom_template.adoc
@@ -163,7 +163,7 @@ spec:
 [#lookup-func]
 === lookup function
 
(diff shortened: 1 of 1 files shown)


mprahl:
@dockerymick this will also need to be added to the 2.4 and 2.3 documentation.

dockerymick (approved):
Thanks for this update Matt! Approved !

mprahl:
Thanks @dockerymick! Could you please backport this to 2.4 and 2.3?

PR #3415, opened 2022-05-13
Copy Hypershift and ManagedServiceAccount to MCH advanced config

JakobGray:
Issue: https://github.com/stolostron/backlog/issues/22534

Signed-off-by: Jakob Gray <20209054+JakobGray@users.noreply.github.com>


swopebe:
@JakobGray can you explain what's going on here before I merge so I understand what I am putting in the doc and why, the issue is not very descriptive.

JakobGray:
@swopebe When installing ACM a user can enable MCE components in the multiclusterhub resource and they will be propagated to the multiclusterengine resource. Therefore we should include MCE toggles in the multiclusterhub advanced configuration doc as well.

Hypershift and MSA are MCE components, but they can be included in the MCH spec and then when ACM creates it's own MCE under the covers it will create MCE with those components enabled. For people who aren't using MCE standalone they should only be handling the MCH resource.

swopebe (approved):

PR #3444, opened 2022-05-19
updating observability export to external endpoints

dockerymick:
https://app.zenhub.com/workspaces/engineering-backlog-do-not-delete-604fab62d4b98d00150a2854/issues/stolostron/backlog/20624

### DIFF
--- a/apis/observability.json.adoc
+++ b/apis/observability.json.adoc
@@ -109,6 +109,17 @@ __required__|Parameters describing the MultiClusterObservability resource to be 
       "metricObjectStorage": {
         "name": "thanos-object-storage",
         "key": "thanos.yaml"
+    "url": 
+    "http_client_config":
+      "basic_auth":
+        "username":
+        "password":
+      "tls_config":
+        "secret_name":
+        "ca_file_key":
+        "cert_file_key":
+        "key_file_key":
+        "insecure_skip_verify":
        }
    }
 }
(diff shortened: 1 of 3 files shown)


marcolan018:
@dockerymick 

The confgiuration for the endpoint is not part of MCO CR. It's a separate secret. So I think we should not put it in the observability API doc? Or add it as an appendix of the API doc?

And also we really need update in MCO API side, sorry I missed that part in my original post. Please check the content below:

In `storageConfig`, adds a new item `writeStorage `
Name | Description | Schema
 --- | --- | --- 
writeStorage(optional) | A list access info of write endpints | []WriteStorage

And the definition `WriteStorage` is as below:
Name | Description | Schema
 --- | --- | --- 
name(required)|The name of the secret with endpoint access info|string
key(required)|The key of the secret to select from|string

apis/observability.json.adoc
```diff
@@ -109,6 +109,17 @@ __required__|Parameters describing the MultiClusterObservability resource to be
...
+    "http_client_config":
+      "basic_auth":
+        "username":
+        "password":
+      "tls_config":
+        "secret_name":
+        "ca_file_key":
+        "cert_file_key":
+        "key_file_key":
+        "insecure_skip_verify":
```
marcolan018: should not be added here

apis/observability.json.adoc
```diff
@@ -284,8 +295,49 @@ __optional__|The advanced configuration settings for observability. |<<_rhacm-do
...
+_optioal_|Name of the secret that contains certificates.|string
+|**ca_file_key** +
+_optional_|Key of the CA certificate in the secret.|string
+|**cert_file_key** +
+_optional_|Key of the client certificate in the secret.|string
+|**key_file_key** +
+_optional_|Key of the client key in the secret.|string
+|**insecure_skip_verify** +
+_optional_|Parameter to skip the verification for target certificate.|bool
+|===
```
marcolan018: should not be added here

marcolan018 (approved):

dockerymick:
@marcolan018 thanks for your explanation. In my opinion, the table for writeStorage is not necessary. The table is disruptive. I think it is best to keep the explanation in the docs. However, I did add the others tables

dockerymick:
Made a couple of updates. Merging this PR

marcolan018:
Hi @dockerymick 
the changes of writeStorage in observability API doc is necessary.  It’s a real change on the API, cannot be ignored. From user perspective, if they want to enable remote endpoint export, they need to firstly create the secret. For the content of the secret, you already added the description. Secondly, they also need to add the writeStorage part in MCO CR. So we need to add the related description about this. How to specify the secret name in MCO CR and the related key.
The similar case is the object storage configuration. In our sample, the secret thanos-object-stoage contains all necessary info about how to access S3 object storage. And in MCO CR, also need to specify the secret info as below, And this should be included in observability api doc

```
    metricObjectStorage:
      name: thanos-object-storage
      key: thanos.yaml
```

So the desired update of observability API doc is as below:

In `storageConfig`, adds a new item `writeStorage `
Name | Description | Schema
 --- | --- | --- 
writeStorage(optional) | A list access info of write endpints | []WriteStorage

And the definition `WriteStorage` is as below:
Name | Description | Schema
 --- | --- | --- 
name(required)|The name of the secret with endpoint access info|string
key(required)|The key of the secret to select from|string

PR #3446, opened 2022-05-19
removing the important notice from policy tables

dockerymick:
https://app.zenhub.com/workspaces/engineering-backlog-do-not-delete-604fab62d4b98d00150a2854/issues/stolostron/backlog/22430

### DIFF
--- a/governance/etcd_encryption_policy.adoc
+++ b/governance/etcd_encryption_policy.adoc
@@ -75,7 +75,6 @@ The `exclude` parameter specifies the namespaces you explicitly do not want to a
 | Optional.
 Specifies the remediation of your policy.
 The parameter values are `enforce` and `inform`.
-*Important*: Some policies might not support the enforce feature.
 
 | disabled
 | Required.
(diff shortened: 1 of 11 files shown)


drkho (commented):
Hi, @dockerymick . In governance/policy_controllers.adoc, the configuration policy controller is the one to support "enforce" and "inform". However, the controller may not be able to enforce with some policies or policy templates.

Hi, @gparvin , please correct me if I am wrong here.

gparvin:
The "important" note regarding remediation actions in the Policy Controllers section I think should be kept.  It would be nice if it was moved down to just after the list of Policy Controllers -- that may help it make more sense.

I agree that we don't need to state this limitation in the config samples.  So the following places (tables) should have the "important" note removed inside the tables detailing the fields:
- Memory usage policy table
- Namespace policy YAML table
- Image vulnerability policy YAML table
- Pod policy table
- Pod security policy table
- Role policy table
- Role binding policy table

I think that will improve the clarity, but @drkho feel free to look at his more -- I just did a quick look at where I thought this was confusing.
Thanks!

dockerymick:
@drkho @gparvin thanks for both reviews and clarification. I have updated the files as suggested.

dockerymick:
@drkho @gparvin thanks for both reviews and clarification. I have updated the files as suggested.

dockerymick:
Moving forward with this merge

PR #3458, opened 2022-05-25
update to scale

swopebe:


swopebe:
Should prob move to the sizing doc...

PR #3487, opened 2022-05-31
Document the evaluationInterval never value

mprahl:


governance/config_policy_ctrl.adoc
```diff
@@ -136,7 +136,7 @@ The `exclude` parameter specifies the namespaces you explicitly do not want to a
 `mustnothave`: Indicated that an object with the same name or labels cannot exist and need to be deleted, regardless of the specification or rules.
 
 | spec.evaluationInterval.compliant
-| Optional. Used to define how often the policy is evaluated when it is in the compliant state. The values must be in the format of a duration which is a sequence of numbers with time unit suffixes. For example, `12h30m5s` represents 12 hours, 30 minutes, and 5 seconds.
+| Optional. Used to define how often the policy is evaluated when it is in the compliant state. The values must be in the format of a duration which is a sequence of numbers with time unit suffixes. For example, `12h30m5s` represents 12 hours, 30 minutes, and 5 seconds. It can also be set to `never` so that the policy is not reevaluated on that cluster once it is compliant unless the policy spec is updated.
```
dhaiducek: This should also be added to `spec.evaluationInterval.noncompliant`, correct?

dhaiducek: (Or at least summarized? Something like: "...time unit suffixes, and also accepts `never`.")

mprahl: You're right. I originally left it thinking it's not usually a good idea to set `evaluationInterval.noncompliant` to `never`, but we should leave that up to the user.

governance/config_policy_ctrl.adoc
```diff
@@ -136,10 +136,10 @@ The `exclude` parameter specifies the namespaces you explicitly do not want to a
...
```
dockerymick: It can also be set to `never` so that the policy is not reevaluated on the compliant  cluster, unless the policy `spec` is updated.

governance/config_policy_ctrl.adoc
```diff
@@ -136,10 +136,10 @@ The `exclude` parameter specifies the namespaces you explicitly do not want to a
...
 
 | spec.evaluationInterval.noncompliant
-| Optional. Used to define how often the policy is evaluated when it is in the non-compliant state. Similar to the `evaluationInterval.compliant` parameter, the values must be in the format of a duration which is a sequence of numbers with time unit suffixes.
+| Optional. Used to define how often the policy is evaluated when it is in the non-compliant state. Similar to the `evaluationInterval.compliant` parameter, the values must be in the format of a duration which is a sequence of numbers with time unit suffixes. It can also be set to `never` so that the policy is not reevaluated on that cluster once it is noncompliant unless the policy spec is updated.
```
dockerymick: It can also be set to `never` so that the policy is not reevaluated on the compliant  cluster, unless the policy `spec` is updated.

dockerymick (changes requested):
Thanks for making these updates @mprahl I made a couple of comments. Is there an issue related to this?

mprahl:
Thanks for the review @dockerymick! I just updated the PR based on your comments.

This is a followup on https://github.com/stolostron/backlog/issues/20519. I had just forgotten to tell you about the "never" option at the time.

dockerymick (approved):
Looks good !

PR #3489, opened 2022-06-01
stolostron/backlog#20292 Minor (do not merge)

cadawson:


install/cluster_size.adoc
```diff
@@ -1,7 +1,7 @@
 [#sizing-your-cluster]
```
cadawson, line 1: I think we talked about changing this to hub cluster, if it is applicable. Based on what Randy mentioned, that might still be a good idea. 

I see where "lower bound" could be confused as a minimum requirement. Should we explain it?

Should "Gb" be "GB"?

Will there be a counterpart to "Product environment"?

cadawson (commented):
I think the list format looks good for this content. I added some questions that I had, based on what Randy mentioned, and what I noticed. Thanks!

install/cluster_size.adoc
swopebe: @cadawson you likely saw that @randymgeorge already addresses this in the issue comment I sent you, so I feel we should remain in the issue for now for efficiency. Please continue the discussion in the issue so we don't have multiple branches addressing this.
.

swopebe:
closing, I did a lot of work in the cluster doc to clean this up: https://github.com/stolostron/backlog/issues/20292

swopebe:
Best practice: Slack/GitHub to communicate and avoid redundant work.

PR #3503, opened 2022-06-02
Updating certs doc

dockerymick:
https://app.zenhub.com/workspaces/engineering-backlog-do-not-delete-604fab62d4b98d00150a2854/issues/stolostron/backlog/21352

### DIFF
--- a/governance/certificates.adoc
+++ b/governance/certificates.adoc
@@ -5,20 +5,14 @@ Various certificates are created and used throughout {product-title}.
 
 You can bring your own certificates. You must create a Kubernetes TLS Secret for your certificate. After you create your certificates, you can replace certain certificates that are created by the {product-title-short} installer.
 
-*Required access*: Cluster administrator or team administrator.
+*Required access*: Cluster administrator
 
-*Note:* Replacing certificates is supported only on native {product-title-short} installations.
+All certificates required by services that run on {product-title-short} are created during the installation of {product-title-short}. Certificates are created and managed by the following components:
 
-All certificates required by services that run on {product-title-short} are created during the installation of {product-title-short}.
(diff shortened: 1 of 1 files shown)


governance/certificates.adoc
```diff
@@ -167,20 +161,20 @@ oc get secret -n open-cluster-management -o custom-columns=Name:.metadata.name,E
 [#refresh-hub-cluster-managed-certificates]
```
gparvin, line 161: Use the proposed command instead of the one above.  It should be:
```
for ns in multicluster-engine open-cluster-management ; do echo "$ns:" ; oc get secret -n $ns -o custom-columns=Name:.metadata.name,Expiration:.metadata.annotations.service\\.beta\\.openshift\\.io/expiry | grep -v '<none>' ; echo ""; done
```

gparvin (commented):
I only saw one command that didn't get updated.  It didn't let me mark the exact line -- it's the command just before the comment I added that needs to be updated due to having mce now along with ACM.

governance/certificates.adoc
dockerymick: Should the following statement be added on Line 16:

After the rotation is complete, apply the new certificates to all of the services with the following command:

```
oc get secret -n $ns -o custom-columns=Name:.metadata.name,Expiration:.metadata.annotations.service\\.beta\\.openshift\\.io/expiry | grep -v '<none>' ; echo ""; done
```

gparvin: Not on line 16, the command you are showing isn't complete  The command in my comment should go on line 156.  Make sure you get the full command.

dockerymick:
@gparvin Thanks for clarifying! I made an update this morning. When you have a chance , will you review again please

gparvin (approved):
Looks good.  Thanks

PR #3538, opened 2022-06-06
Clarify hub template docs

dhaiducek:
Clarify that:
- Hub cluster templates (`{{hub`) are restricted to the namespace of the policy
- `fromSecret` and `protect` are now available to encrypt sensitive data from the hub

### DIFF
--- a/governance/custom_template.adoc
+++ b/governance/custom_template.adoc
@@ -22,8 +22,6 @@ Template functions, such as resource-specific and generic `lookup` template func
 
 To conform templates with YAML syntax, templates must be set in the policy resource as strings using quotes or a block character (`|` or `>`). This causes the resolved template value to also be a string. To override this, consider using `toInt` or `toBool` as the final function in the template to initiate further processing that forces the value to be interpreted as an integer or boolean respectively.
 
-*Note*: If the string value is more than 80 characters, the block character `|` needs to be used to avoid YAML parsing errors.
-
 Continue reading to view descriptions and examples for some of the custom template functions that are supported:
 
 * <<fromsecret-func,fromSecret function>>
@@ -436,9 +434,9 @@ In addition to managed cluster templates that are dynamically customized to the 
 
 Hub cluster templates are based on Golang text template specifications, and the `{{hub … hub}}` delimiter indicates a hub cluster template in a configuration policy.
 
-For security, both resource-specific and the generic lookup functions in hub cluster templates are restricted to the namespace of the policy on the hub.
+For security, both resource-specific and the generic lookup functions in hub cluster templates are restricted to the namespace of the policy on the hub cluster.
 
-*Important:* If you use hub cluster templates to propagate secrets or other sensitive data, the sensitive data exists in the managed cluster namespace on the hub cluster and on the managed clusters where that policy is distributed. The template content is expanded in the policy, and policies are not encrypted by the {ocp-short} ETCD encryption support. To address this, use `fromSecret`, which will automatically encrypt the values from the Secret, or `protect` to encrypt other values.
+*Important:* If you use hub cluster templates to propagate secrets or other sensitive data, the sensitive data exists in the managed cluster namespace on the hub cluster and on the managed clusters where that policy is distributed. The template content is expanded in the policy, and policies are not encrypted by the {ocp-short} ETCD encryption support. To address this, use `fromSecret`, which automatically encrypts the values from the Secret, or `protect` to encrypt other values.
 
 [#template-processing]
 === Template processing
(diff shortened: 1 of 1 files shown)


governance/custom_template.adoc
```diff
@@ -18,11 +18,11 @@ Configuration policies support the inclusion of Golang text templates in the obj
...
 
-To conform templates into YAML syntax, templates must be set in the policy resource as strings using quotes or a block character (`|` or `>`). This causes the resolved template value to also be a string. To override this, consider using `toInt` or `toBool` as the final function in the template to initiate further processing that forces the value to be interpreted as an integer, or boolean.
+To conform templates with YAML syntax, templates must be set in the policy resource as strings using quotes or a block character (`|` or `>`). This causes the resolved template value to also be a string. To override this, consider using `toInt` or `toBool` as the final function in the template to initiate further processing that forces the value to be interpreted as an integer or boolean respectively.
 
-*Note*: If the string value is more than 80 characters, this block character, `|` needs to be used to avoid YAML parsing errors.
+*Note*: If the string value is more than 80 characters, the block character `|` needs to be used to avoid YAML parsing errors.
```
mprahl: In 2.4+ this no longer applies, so we can just delete this line.

ckandag:
@dhaiducek , thanks for making the updates,  there is a table further down in the same page where this is stated as part of the AccescControl but what you added puts it upfront and makes it much clearer.

While looking into this ,  I noticed an error  introduced by another PR   in 2.5  ... where the table that shows differences between hub and managedcluster got  messed up  ( contents of the columns got switched for some of the rows) 

Could you pls fix that as well when you are at it.

governance/custom_template.adoc
```diff
@@ -432,11 +432,13 @@ To force an immediate rotation, delete the `policy.open-cluster-management.io/la
...
+In addition to managed cluster templates that are dynamically customized to the target cluster, {product-title-short} also supports hub cluster templates to define configuration policies using values from the hub cluster. This combination reduces the need to create separate policies for each target cluster or hardcode configuration values in the policy definitions. 
 
 Hub cluster templates are based on Golang text template specifications, and the `{{hub … hub}}` delimiter indicates a hub cluster template in a configuration policy.
 
-*Important:* If you use hub cluster templates to propagate secrets or other sensitive data, the sensitive data exists in the managed cluster namespace on the hub cluster and on the managed clusters where that policy is distributed. The template content is expanded in the policy, and policies are not encrypted by the {ocp-short} ETCD encryption support.
+For security, both resource-specific and the generic lookup functions in hub cluster templates are restricted to the namespace of the policy on the hub.
```
dockerymick: Add cluster after hub:

For security, both resource-specific and the generic lookup functions in hub cluster templates are restricted to the namespace of the policy on the hub cluster.

governance/custom_template.adoc
```diff
@@ -432,11 +432,13 @@ To force an immediate rotation, delete the `policy.open-cluster-management.io/la
...
+
+*Important:* If you use hub cluster templates to propagate secrets or other sensitive data, the sensitive data exists in the managed cluster namespace on the hub cluster and on the managed clusters where that policy is distributed. The template content is expanded in the policy, and policies are not encrypted by the {ocp-short} ETCD encryption support. To address this, use `fromSecret`, which will automatically encrypt the values from the Secret, or `protect` to encrypt other values.
```
dockerymick: Change: 

To address this, use `fromSecret`, which will automatically encrypt the values from the Secret, or `protect` to encrypt other values.

TO

To address this, use `fromSecret`, which automatically encrypts the values from the secret, or `protect` to encrypt other values.

dockerymick (changes requested):
Thanks for making these changes. Is there a doc issue associated with this PR?

governance/custom_template.adoc
dhaiducek: So this entire PR should also be applied to `2.4`?

mprahl: Yes, except for the `fromSecret` and `protect` portions.

dhaiducek:
@dockerymick Sure thing! Just created one: 
- https://github.com/stolostron/backlog/issues/23008

dhaiducek:
@dockerymick Would you please confirm that the published "Comparison table of hub cluster and managed cluster" table has the columns in the correct place? It looks like GitHub isn't handling the pipe character that appears on this line in the row:
```
The equivalent call might use the folllowing syntax: `{{hub "(lookup "v1" "Secret" "default" "my-hub-secret").data.message | protect hub}}`
```

image

Do we need to escape it? (i.e. `|` --> `\|`)

dhaiducek:
I pushed the change to escape the pipe character since it seemed to resolve it--please confirm this is correct!

dockerymick (approved):
Looks good from my perspective! I also learned something new. I didn't realize how the line bar interfered with the table. Definitely mentioning this to Chris and Brandi 
Screen Shot 2022-06-06 at 1 08 43 PM 1

PR #3540, opened 2022-06-06
Adding known issue for gov - ARM

dockerymick:
https://app.zenhub.com/workspaces/engineering-backlog-do-not-delete-604fab62d4b98d00150a2854/issues/stolostron/backlog/21496

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -866,13 +866,14 @@ When you have a configuration policy that is configured with `mustnothave` for t
 
 [#operators-deployed-with-policies]
 === Operators deployed with policies do not support ARM
-//What does ARM stand for again?
-While installation into an ARM environment is supported, operators that are deployed with policies might not support ARM environments. The following policies that install operators do not support ARM environments:
+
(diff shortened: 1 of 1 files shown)


gparvin (approved):
These changes look good to me.  I don't see anywhere in our doc where we cover ARM support in general -- which you could link to if we have it.  ARM support was part of EPIC https://github.com/stolostron/backlog/issues/19061.

FYI: ARM is a different processor than amd/intel.  It means all images need to be compiled for that processor -- which our CICD team does for us.  For GRC Policies, things we compile support this different ARM processor -- such as gatekeeper -- but if we don't compile it, then when you install something like the Compliance Operator it may not have images compiled for the ARM processor.  So our policy that installs the Compliance Operator won't work if you are trying to make it work on an ARM system -- because the Compliance Operator doesn't have images for the ARM platform/processor.

dockerymick:
@gparvin thanks for explaining Gus and thanks for the review! I reached out to my team and confirmed that info about ARM support will be added to the What's new topic. Moving forward with the merge

PR #3541, opened 2022-06-07
Swope new

swopebe:


cadawson (approved):
Thanks Brandi and Mikela, but especially Brandi on this one! It looks great! I made a few updates, so I'm good.

release_notes/whats_new.adoc
```diff
@@ -88,20 +86,32 @@ See the following cluster features that are available as Technology Preview:
   
   ** See the {product-title-short} documentation at link:../install/adv_config_install.adoc#managedserviceaccount-add-on-technical-preview[Enabling ManagedServiceAccount add-ons] to learn more.
 
-* The `hypershift` add-on helps you host OpenShift control planes at scale, separating management from workloads. The component is disabled by default.
+* Use the `hypershift` add-on to manage and provision clusters with hosted control planes at scale, separating management from workloads. The component is disabled by default.
 
   ** See the _multicluster engine operator_ documentation at link:../multicluster_engine/addon_managed_service.adoc#hypershift-addon-intro[Hypershift add-on (Technology Preview)] to learn more.
```
dockerymick, line 91: I don't see this file in the clusters folder: `hypershift_intro.adoc`

dockerymick: I don't see this file in the clusters folder: `hypershift_intro.adoc`

swopebe: @dockerymick this is a section in a file in MCE, that's why. The folder is MCE. There should be some hypershift doc but we will prob have to deal with that when it comes up if it doesn't exist.

swopebe: https://github.com/stolostron/rhacm-docs/blob/2.5_stage/multicluster_engine/adv_config_install.adoc#hypershift-add-on-technology-preview

dockerymick: @swopebe thanks! I think we should reference the user to the link from your previous comment, rather than pointing users to the addon_managed_service.adoc file

swopebe: Not sure what you are saying.

dockerymick: Reference users to `multicluster_engine/adv_config_install.adoc` instead of `multicluster_engine/addon_managed_service.adoc`

swopebe: Oh, I see what you are saying...sorry, way too many links. I think it is correct now.

PR #3588, opened 2022-06-10
Refresh link issues 06/10/22

cadawson:


add-ons/submariner/submariner_deploy_mcaddon_api.adoc
```diff
@@ -3,7 +3,7 @@
 
 To deploy Submariner by using the `ManagedClusterAddOn` API, complete the following steps:
 
-. Create a `ManagedClusterSet` on the hub cluster by using the instructions provided in link:../../clusters/managedclustersets.adoc#managedclustersets[Creating and managing ManagedClusterSets]. Your entry for the `ManagedClusterSet` should resemble the following content:
+. Create a `ManagedClusterSet` on the hub cluster by using the instructions provided in link:../../clusters/managing-your-clusters/managedclustersets.adoc#managedclustersets[Creating and managing ManagedClusterSets]. Your entry for the `ManagedClusterSet` should resemble the following content:
```
dockerymick, line 6: Received 404. Change `managing-your-clusters` to `managedclustersets.adoc`

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -17,7 +16,7 @@ Before installing VolSync on your clusters, you must have the following requirem
 
 * At least two configured clusters that are managed by the same {product-title-short} hub cluster
 
-* Network connectivity between the clusters that you are configuring with VolSync; If the clusters are not on the same network, you can configure the link:../submariner/submariner.adoc#submariner[Submariner multicluster networking and service discovery] and use the `ClusterIP` value for `ServiceType` to network the clusters, or use a load balancer with the `LoadBalancer` value for `ServiceType`. 
+* Network connectivity between the clusters that you are configuring with VolSync; If the clusters are not on the same network, you can configure the link:../add-ons/submariner/submariner.adoc#submariner[Submariner multicluster networking and service discovery] and use the `ClusterIP` value for `ServiceType` to network the clusters, or use a load balancer with the `LoadBalancer` value for `ServiceType`.
```
dockerymick, line 19: Received 404. Remove `add-ons`

release_notes/whats_new.adoc
```diff
@@ -86,13 +86,13 @@ See the following cluster features that are available as Technology Preview:
 
 * The `Managed-ServiceAccount` component allows you to create and delete a Service Account on a managed cluster. The component is disabled by default.
 
-  ** See the _multicluster engine operator_ documentation at link:../multicluster_engine/addon_managed_service.adoc#managed-serviceaccount-addon[Enabling ManagedServiceAccount add-ons (Technology Preview)] to learn more.
+  ** See the _multicluster engine operator_ documentation at link:../multicluster_engine_operator/addon_managed_service.adoc#managed-serviceaccount-addon[Enabling ManagedServiceAccount add-ons (Technology Preview)] to learn more.
```
dockerymick, line 89: Received 404 for both links in this file. the folder name is only `multicluster_engine`

dockerymick (changes requested):

add-ons/submariner/submariner_deploy_mcaddon_api.adoc
cadawson: Problem is that is what it was before that is resulting in a 404 on the main site. It seems to be linking to it on the staging site with this voodoo.

add-ons/volsync/volsync_replicate.adoc
cadawson: This one also does not seem to be working the way it is supposed to. I had to hack it this way. I think it will work, according to some manual tests that I did. I know it is not working now.

release_notes/whats_new.adoc
cadawson: Again, your logic and understanding of the way things should work is right on. I realize now that this is why I couldn't figure out how to fix them the other day. They were fine, according to how it should work. For some reason, these are not working that way. This is the link that it is actually creating. I can show you what is happening, if you are interested.

cadawson:
Thanks for the review, @dockerymick . Your reasoning is sound, but this is Pantheon. I am not sure why, but these links are acting differently than the other ones. I can show you how, if you are interested. The links seem to work this way on staging, though we won't be certain until they get on prod. Feel free to play with them, if you want to. If you can make them work some other way, I am happy to go with that. I was going for "They appear to be working this way on staging, and I have already spent too much time with them."

release_notes/whats_new.adoc
cadawson: I think I figured out why this one is happening. The `Title URL` for the book is set to `multicluster_engine_operator`. When it builds, it puts that name in the URL. When we add the links, we use the folder name `multicluster_engine`. When they both get built, they don't match. I see 2 options: 
1. We change the name of the title URL to be multicluster_operator, and things should be fixed. Though I think there might be a link from the installation build that will no longer work.
2. Leave this fix for now and try to rename at a z-stream and update the link at that time. 
3. Leave this fix and not worry about it for now. We will be moving the MCE stuff out soon, and it can be addressed then. 
Thoughts?

dockerymick: i would update the title url

dockerymick (approved):
Thanks for explaining!

PR #3593, opened 2022-06-13
Adding upgrade step for backup and restore

dockerymick:
https://github.com/stolostron/backlog/issues/22776

### DIFF
--- a/clusters/backup_restore_enable.adoc
+++ b/clusters/backup_restore_enable.adoc
@@ -8,7 +8,6 @@ Enable the cluster backup and restore operator to schedule back up and restore f
 * <<prereq-backup-restore-enable,Prerequisites>>
 * <<enabling-backup-restore,Enabling the backup and restore operator>>
 * <<using-backup-restore,Using the backup and restore operator>>
-** <<upgrading-backup-restore,Upgrading the backup and restore operator>>
 * <<view-restore-events,View restore events>>
 
 [#prereq-backup-restore-enable]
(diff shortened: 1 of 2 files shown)


clusters/backup_restore_enable.adoc
```diff
@@ -85,7 +85,7 @@ apiVersion: operator.open-cluster-management.io/v1
     namespace: open-cluster-management
   spec:
     availabilityConfig: High
-    enableClusterBackup: false << this is deprecated as of 2.5, no longer used
+    cluster-backup: false
```
cameronmwall: So, this is not quite what are spec looks like. `enableClusterBackup ` will still exist, it just won't do anything because it has been deprecated. Instead, enabling or disabling cluster backup will look like the following

```
  overrides:
    components:
      - enabled: true
        name: multiclusterhub-repo
      - enabled: true
        name: search
      - enabled: true
        name: management-ingress
      - enabled: true
        name: console
      - enabled: true
        name: insights
      - enabled: true
        name: grc
      - enabled: true
        name: cluster-lifecycle
      - enabled: true
        name: volsync
      - enabled: true
        name: multicluster-engine
      - enabled: false
        name: cluster-proxy-addon
      - enabled: false
        name: cluster-backup
 ```

File: release_notes/known_issues.adoc [hunk 2]
```diff
@@ -932,7 +932,21 @@ View the following restore limitations:
 
 To address the previously mentioned limitations, when a `restore.cluster.open-cluster-management.io` resource is created, the cluster backup and restore operator runs a set of steps to prepare for restore by cleaning the hub cluster before Velero restore begins. For more information, see link:../clusters/backup_and_restore.adoc#clean-hub-restore[Clean the hub cluster before restore].
 
+[#upgrade-limitation]
+=== Cluster backup and restore upgrade limitation
+
+If you upgrade your cluster from {product-version-prev} to {product-version} with the `cluster-backup` parameter set to `true`, the following message appears:
```
cameronmwall: So, because in this case the upgrade is being blocked we are still using the 2.4 `enableClusterBackup` parameter to unblock it

PR #3601, opened 2022-06-14
Remove outdated Submariner prerequisites

dfarrell07:
Closes stolostron/backlog#23455

Signed-off-by: Daniel Farrell <dfarrell@redhat.com>

### DIFF
--- a/add-ons/submariner/submariner.adoc
+++ b/add-ons/submariner/submariner.adoc
@@ -29,10 +29,7 @@ You can enable Submariner on {ocp-short} clusters that are hosted in the followi
 
 Ensure that you have the following prerequisites before using Submariner:
 
-* A {product-title-short} hub cluster that is running on {ocp} version 4.6, or later, with Kubernetes version 1.19, or later.
 * A credential to access the hub cluster with `cluster-admin` permissions.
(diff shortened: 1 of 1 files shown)


aswinsuryan (approved):

cadawson (commented):
Thanks @dfarrell07 . I understand the need to remove the Globalnet note. Is the requirement for 4.6, and later no longer an issue? Maybe this is because versions earlier than 4.6 are no longer supported, but does that mean that people still won't be running an older version and try this? 
Same question for the 4.4 statement. 
If there is no concern about people running and trying to add Submariner to an older version, it is right to remove it. Just wanted to make sure that there might not be someone who tries something unexpected and expects support.

dfarrell07:
Yeah I think the assumption is that people should not be running these older versions by the time they get to the Submariner docs because they have already validated their setup with the support matrix. @nyechiel might have more to say.

nyechiel:
In general, these are very old k8s versions which are not supported by ACM. I don't see any value keeping it here (I believe this was just copied from the upstream docs).

cadawson (approved):
Looks good. Thanks!

cadawson:
Opened issue #23508 for this.

PR #3626, opened 2022-06-22
adding backup and restore known issue

dockerymick:
https://github.com/stolostron/backlog/issues/22098


sahare (approved):
Thanks @dockerymick , I'm approving with a minor comment:
I think this is more clear that we are talking about manually imported clusters:
Managed clusters that are imported on the primary hub cluster using the <i>Import cluster</i> operation will show only when the activation data is restored on the passive hub.

PR #3627, opened 2022-06-22
https://github.com/stolostron/backlog/issues/23463--remove hacky file

swopebe:


clusters/remove_managed_cluster.adoc
```diff
@@ -80,12 +76,15 @@ open-cluster-management-agent         Active   10m
...
 +
 ----
-./cleanup-managed-cluster.sh
+oc delete namespaces open-cluster-management-agent open-cluster-management-agent-addon --wait=false
+oc get crds | grep open-cluster-management.io | awk '{print $1}' | xargs oc delete crds --wait=false
+oc get crds | grep open-cluster-management.io | awk '{print $1}' | xargs oc patch crds --type=merge -p '{"metadata":{"finalizers": []}}'
 ----
+
 . Run the following command to ensure that both namespaces are removed: 
 +
```
skeeey, line 89: I think we may enhance this part, sometimes if there are remaining crds in the managed cluster, it will cause some problems, so we need ensure the crds are removed, thanks

4. Run the following command to ensure that both namespaces and all open cluster management crds are removed:
```
 oc get crds | grep open-cluster-management.io | awk '{print $1}'
 oc get ns | grep open-cluster-management-agent
```

skeeey (approved):

clusters/remove_managed_cluster.adoc
swopebe: that change is in

PR #3648, opened 2022-06-30
Update release_images.adoc

oafischer:
Fix error in apiVersion
https://github.com/stolostron/backlog/issues/22189#issue-1224320884


dockerymick (approved):
Great job making this change Oliver. Let's reach out to a member on the Cluster squad to verify if the change should also be applied to 2.5. I assume yes, but let's double check. Approving

PR #3655, opened 2022-07-01
removing the exclude parameter from the example

dockerymick:
https://app.zenhub.com/workspaces/engineering-backlog-do-not-delete-604fab62d4b98d00150a2854/issues/stolostron/backlog/21997


dhaiducek (approved):
LGTM! I'll also look at adding some additional examples (and/or updating these) that both apply to `2.6` and that we can backport to earlier versions.

dockerymick:
@dhaiducek Sounds good, thanks for reviewing!

PR #3663, opened 2022-07-05
Add link to disableHubSelfManagement

dockerymick:
Live review to add reference link w/Oliver
https://app.zenhub.com/workspaces/engineering-backlog-do-not-delete-604fab62d4b98d00150a2854/issues/stolostron/backlog/21822

### DIFF
--- a/clusters/import_gui.adoc
+++ b/clusters/import_gui.adoc
@@ -155,6 +155,3 @@ oc delete po -n open-cluster-management `oc get pod -n open-cluster-management |
 ----
 
 You can change the value of the hub cluster to not import automatically by changing the `disableHubSelfManagement` value to `true`. For more information, see the link:../install/adv_config_install.adoc#disable-hub-self-management[disableHubSelfManagement] topic.
-
-// as described in link:../install/install_connected.adoc#installing-while-connected-online[Installing while connected online].
-// suggesting to remove this to minimize clicks. Let's take user directly to the topic 
--- a/clusters/import_gui.adoc
+++ b/clusters/import_gui.adoc
@@ -154,4 +154,7 @@ On the _Clusters_ page, click *Actions* > *Detach cluster* to remove your cluste
 oc delete po -n open-cluster-management `oc get pod -n open-cluster-management | grep multiclusterhub-operator| cut -d' ' -f1`
 ----
 
-You can change the value of the hub cluster to not import automatically by changing the `disableHubSelfManagement` value to `true`, as described in link:../install/install_connected.adoc#installing-while-connected-online[Installing while connected online].
+You can change the value of the hub cluster to not import automatically by changing the `disableHubSelfManagement` value to `true`. For more information, see the link:../install/adv_config_install.adoc#disable-hub-self-management[disableHubSelfManagement] topic.
+
+// as described in link:../install/install_connected.adoc#installing-while-connected-online[Installing while connected online].
+// suggesting to remove this to minimize clicks. Let's take user directly to the topic 


birsanv:
@dockerymick I am not the right person to review this.. it took me some time to understand why I was tagged here :) I think it's because I made a comment somewhere that the option `disableHubSelfManagement` in the context of the doc is confusing. I still find it confusing though .. I'll let @zkayyali812 help since he works on the MCH 

The issue I was reporting : 
- on the _Removing an imported cluster_ doc section it's discussed a disableSelfManagement option. It doesn't say what CRD defines the option and is not clear from the context. I honestly wanted to use that option but did not know where to look for it
- I guessed it's on the MCH ( from the link you have in the PR looks like is coming from there ). But I have a MCH 2.5 and Isee no such spec option .. @zkayyali812 can you confirm this is a valid option now ?

The confusing text below : ( The Note suddenly talks about an option but it doesn't say where it is coming from )

_Removing an imported cluster
Complete the following procedure to remove an imported cluster and the open-cluster-management-agent-addon that was created on the managed cluster.

On the Clusters page, click Actions > Detach cluster to remove your cluster from management.

>>>>>>Note: If you attempt to detach the hub cluster, which is named local-cluster, **be aware that the default setting of disableHubSelfManagement is false**._  ....

I see you added a link to the option , which is good; but I still can't find the option with MCH 2.5
```
You can change the value of the hub cluster to not import automatically by changing the disableHubSelfManagement value to true. For more information, see the [disableHubSelfManagement](https://github.com/stolostron/rhacm-docs/blob/f958f0c826334b761e946bc688def0d21c29bdc2/install/adv_config_install.adoc#disable-hub-self-management) topic.
```

oafischer (approved):
Link looks good!

dockerymick:
@birsanv Thanks so much for clarifying Valentina. Reaching out to Zack today

zkayyali812 (approved):

zkayyali812:
Hey all,
This PR looks great to me. Perhaps someone from @acm-installer-team may want to sign off as well.

We have `disableHubSelfManagement` as an option in our spec [here](https://github.com/stolostron/multiclusterhub-operator/blob/main/api/v1/multiclusterhub_types.go#L79).

birsanv:
> Hey all, This PR looks great to me. Perhaps someone from @acm-installer-team may want to sign off as well.
> 
> We have `disableHubSelfManagement` as an option in our spec [here](https://github.com/stolostron/multiclusterhub-operator/blob/main/api/v1/multiclusterhub_types.go#L79).

thank you @zkayyali812 I didn't see this option with the MCH CR so I was not aware of it. So if the disableHubSelfManagement is missing it defaults to false , right ?

I find this still confusing because it implies that you should see a `disableHubSelfManagement` option under MCH which is either true or false. 
> You can change the value of the hub cluster to not import automatically by changing the disableHubSelfManagement value to true.

It could actually not exist at all , as is the case by default, in which case we should probably say : 
> You can change the value of the hub cluster to not import automatically **by adding the disableHubSelfManagement property under the MCH spec** and setting it to  true

zkayyali812:
@birsanv 

> thank you @zkayyali812 I didn't see this option with the MCH CR so I was not aware of it. So if the disableHubSelfManagement is missing it defaults to false , right ?

Yes, if it is not set, disableHubSelfManagement is set to false, and the hub is imported as the local-cluster

> I find this still confusing because it implies that you should see a disableHubSelfManagement option under MCH which is either true or false.

Since disableHubSelfManagement is a boolean, if it is not set(null), it defaults to false, otherwise it is set to true and displayed in the spec. This could be updated to optionally show this value in the spec, even when it is `false`

> You can change the value of the hub cluster to not import automatically by adding the disableHubSelfManagement property under the MCH spec and setting it to true

This is a great point, since it may not exist in the users spec, I think this would be best to guide the user to update or add the key/value if they don't exist

dockerymick:
@zkayyali812 Thanks for confirming and approving this Zack!

dockerymick:
@oafischer I removed the statement that we had as a comment. Moving forward with the merge

oafischer:
This thread was hard to find because it was missing a link back to the original report on GitHub. I'll add it here for future reference so that it is easier to find: https://github.com/stolostron/backlog/issues/21822

PR #3668, opened 2022-07-06
fix Git SSH channel sample (unplanned sprint 11)

rokej:

Remove incorrect fields from the sample channel YAML.


dockerymick:
@rokej Thanks for making this update Roke, is there a doc issue related to this?

dockerymick (approved):
Looks good to me, will leave for Brandi to approve if needed

rokej:
@dockerymick There is no doc issue for this change. Do you need one?

PR #3671, opened 2022-07-07
stolostron/backlog#23648--working on deprecation links, info, etc...b…

swopebe:
…aremetal


clusters/bare_assets.adoc
```diff
@@ -1,7 +1,7 @@
 [#creating-and-modifying-bare-metal-assets]
 = Creating and modifying bare metal assets
 
-**Deprecation notice:** The procedure for creating bare metal clusters using bare metal assets is deprecated. Bare metal assets will be removed from a future release.
```
swopebe, line 4: removing announcement to remove, we go into this in the deprecated topic.

release_notes/deprecate_remove.adoc
```diff
@@ -113,19 +113,19 @@ A _deprecated_ component, feature, or service is supported, but no longer recomm
...
 | 2.5
-| None
-| See link:../clusters/bare_assets.adoc#creating-and-modifying-bare-metal-assets[Creating and modifying bare metal assets] and link:../clusters/create_bare.adoc#creating-a-cluster-on-bare-metal[Creating a cluster on bare metal] for the affected areas.
+| Create an infrastructure environment with the console
+| See link:../clusters/bare_assets.adoc#creating-and-modifying-bare-metal-assets[Creating and modifying bare metal assets] and link:../clusters/create_bare.adoc#creating-a-cluster-on-bare-metal[Creating a cluster on bare metal] for the affected deprecated areas. See link:../clusters/create_cluster_on_prem.adoc#creating-a-cluster-on-premises[Creating a cluster in an on-premises environment] for the proceding process.
 
 | Add-on operator
-| Installation of built-in managed cluster add-ons.
+| Installation of built-in managed cluster add-ons
 | 2.5
```
swopebe, line 123: removing periods because 1. consistently not using them and 2. incomplete sentences

clusters/create_infra_env.adoc
```diff
@@ -8,8 +8,6 @@ You can use the {product-title} console to create an infrastructure environment
 * <<creating-your-infra-env-with-the-console,Creating your infrastructure environment with the console>>
 * <<accessing-your-infra-env,Accessing your infrastructure environment>>
 
-//-
```
cadawson, line 11: This resolves an issue that I ran into where it was not ending the previous list. It is formatting correctly with this. If it works fine without it, I have no problem with removing it. Given that it didn't work before, and I had to research how to make it work correctly, I would prefer to leave it, if it is not tested.

clusters/single_node_openshift_cluster_pool_create.adoc
```diff
@@ -1,5 +1,6 @@
 [#creating-a-single-node-openshift-clusterpool]
 = Creating a single node OpenShift cluster pool (Technology Preview)
+//see GitHub issue #23648 about inaccuracy and possible removal, this topic and the issue is very confusing.
```
cadawson: This file is not included in the build right now. Addressing the other SNO file.

cadawson (approved):
Looks good. Made some changes from your suggestions. Thanks!

clusters/create_infra_env.adoc
swopebe: I disagree. Our source should should be cleanly built so if we have to use hacky comment tags, we should review the way we are building it this list, for instance. I have never seen this issue in all the doc. We really should not clutter the source with comment tags that cannot easily be explained later when the content is considered legacy...I have worked in that environment and it is terrible to have to go remove notes and hacks from others a year later. We should keep our code clean.

clusters/single_node_openshift_cluster_pool_create.adoc
swopebe: It was building and appears to have been inaccurate. Thanks for removing it.

PR #3683, opened 2022-07-11
Clarify headless service use-case with Globalnet (Sprint 13 work)

sridhargaddam:
Signed-off-by: Sridhar Gaddam <sgaddam@redhat.com>


swopebe:
please assign a reviewer and link the issue so that we know what we are adding to the documentation. This will be slated for next Sprint.

dfarrell07 (approved):
This is the result of a report by a user here: https://app.slack.com/client/T027F3GAJ/C0134E73VH6/thread/C0134E73VH6-1656396243.142199

Doesn't the change itself, written by the main SME, capture what we're adding the documentation? Wouldn't it be a better use of our time to just review it for docs style and get it merged, shipped, and to the customers who are stumbling on this?

dfarrell07:
Issue: https://github.com/stolostron/backlog/issues/24159

sridhargaddam:
Thanks for reporting the issue @dfarrell07

vthapar (approved):

release_notes/known_issues.adoc
```diff
@@ -1085,9 +1085,11 @@ Run the `subctl verify` command to confirm that the end-to-end tests pass.
 Automatic cloud preparation for Red Hat OpenStack clusters is not supported for Submariner from the product-title-short} console. You can use the {product-title-short} APIs to prepare the clouds manually.
 
 [#submariner-headless-services-globalnet]
-=== Submariner does not support headless services with Globalnet
+=== Local pods cannot access exported headless services residing in same cluster when using Globalnet
 
-When using headless services with Globalnet, DNS resolution for local pods does not work when using the `clusterset.local` domain name.
+Submariner supports headless services with Globalnet. However, when you access the exported headless service from a client that resides in the same cluster using `clusterset.local` domain name, the `globalIP` that is associated with the headless service is returned to the client, which is not routable in the cluster. 
+
```
swopebe, line 1129: Please see that I changed to **active voice** (industry standards) and **removed preannounce** to fix (industry standard). One more change that I was not sure about and need to clear up (because we have to be as clear as possible):

...in the same cluster **that uses the** `clusterset.local` domain name?

...in the same cluster **by using** `clusterset.local` domain name?

It can be read both ways, which is why I am calling for an edit here. I think you are saying the first option, let me know.

swopebe (changes requested):
see changes and comment

release_notes/known_issues.adoc
dfarrell07: It's the second. The idea is that `app0.clusterset.local` will always resolve as a global IP, including from the cluster that hosts `app0`. That global IP will not be routable in the host cluster.

I'd suggest:

> ...in the same cluster __by using a__ `clusterset.local` domain name

sridhargaddam: As @dfarrell07 mentioned, its the second option.

swopebe (approved):

release_notes/known_issues.adoc
swopebe: asked for peer review, but taking a while. Merging.

PR #3684, opened 2022-07-11
Add new known issue in 2.5

oafischer:
https://github.com/stolostron/backlog/issues/22760#issue-1248630248

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -923,7 +923,7 @@ The _Console_ and _Topology_ for _Application_ changes for the {product-version}
 === ApplicationSet resources do not show status in topology
 // 2.5:22760
 
-When you create ApplicationSet applications that deploy resources to a different namespace than the namespace defined in the ApplicationSet YAML, the resource status does not appear in the topology.
+When you create `ApplicationSet` applications that deploy resources to a different namespace than the namespace defined in the `ApplicationSet` `YAML`, the resource status does not appear in the topology.
 
 [#governance-known-issues]
 == Governance known issues
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -923,7 +923,7 @@ The _Console_ and _Topology_ for _Application_ changes for the {product-version}
 === ApplicationSet resources do not show status in topology
 // 2.5:22760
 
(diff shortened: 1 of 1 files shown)


release_notes/known_issues.adoc
```diff
@@ -919,6 +919,12 @@ See the following limitations to various _Application_ tables in the console:
 
 The _Console_ and _Topology_ for _Application_ changes for the {product-version}. There is no filtering capability from the console Topology page.
 
+[#appset-resources-no-topolgy]
+=== ApplicationSet resources do not show status in topology
+// 2.5:22760
+
+When you create ApplicationSet applications that deploy resources to a different namespace than the namespace defined in the ApplicationSet YAML, the resource status does not appear in the topology.
```
swopebe: @dockerymick can you show where we use `these` for component names?

dockerymick: @swopebe I'm not sure what component names you're referring to

dockerymick: Change:

When you create ApplicationSet applications that deploy resources to a different namespace than the namespace defined in the ApplicationSet YAML, the resource status does not appear in the topology.

TO

When you create `ApplicationSet` applications that deploy resources to a different namespace than the namespace defined in the `ApplicationSet` YAML, the resource status does not appear in the topology.

monospace the application names of the resources/files

dockerymick (changes requested):

release_notes/known_issues.adoc
oafischer: Thanks for the clarifying. Also changed YAML because I saw it is monospaced in other areas of the documentation as well.

dockerymick: When YAML is mentioned in all capitals, you shouldn't monospace it because it's the name of the programming language. However, when it is lowercase and mentioned as a file extension, it should be doc'ed as `.yaml`. 

Another great topic to bring up during our doc arch call. Let's add it to the agenda when we have a chance

oafischer: Thanks for clarifying! Yeah let's definitely talk about it during our call so I can get up to speed on these things. Went ahead and changed it back.

dockerymick (approved):
Looks good!

swopebe:
@dockerymick we did discuss this in an arch call a few releases ago and decided to handle them as such:

`ApplicationSet` unless we are talking about something generic like "application set" (as in a set of applications).

..outside of a title, which can cause build issues maybe?


We definitely have voted on this in the past.

PR #3688, opened 2022-07-11
stolostron/backlog#23648 mod sno create

cadawson:


clusters/single_node_openshift.adoc
```diff
@@ -1,34 +1,26 @@
 [#single-node-openshift]
 = Creating a single node OpenShift cluster
 
-You can create a single node OpenShift (SNO) cluster that contains only a single node. In a standard cluster, there are control plane nodes and separate worker nodes. This configuration can be helpful when you have many clusters that need to minimize their resource footprint. 
+You can create a single node OpenShift (SNO) cluster that contains only a single node in a bare metal environment. In a standard cluster, there are control plane nodes and separate worker nodes. This configuration can be helpful when you have many clusters that need to minimize their resource footprint.
```
berenss: It stood out to me that, within the stream of text that describes SNO, we pivot and talk about a "standard cluster"

> In a standard cluster, there are control plane nodes and separate worker nodes.

Perhaps it would be better to state:
> The single node cluster is used for both the control plane services as well as user workloads.

clusters/single_node_openshift.adoc
```diff
@@ -1,34 +1,26 @@
...
-** VMware vSphere
+* Credentials for accessing the bare metal server where you want your SNO cluster.
 
 [#sno_create]
 == Create a Single-Node OpenShift cluster
 
 To create a SNO cluster, complete the following steps:
 
-Complete the steps in the {product-title-short} console for creating a cluster on one of the providers that are supported to run a SNO cluster in xref:../clusters/create.adoc#creating-a-cluster[Creating a cluster].
+Complete the steps in the {product-title-short} console for creating a cluster in the bare metal environment. Because the bare metal assets feature is deprecated, it is best practice to use the Central Infrastructure Management method of creating a SNO cluster. The procedure to create a SNO cluster is essentially the same as creating an on-premises cluster. See xref:../clusters/create_cluster_on_prem.adoc#creating-a-cluster-on-premises[Creating a cluster in an on-premises environment].
```
berenss: I would suggest we leave out the:

> Because the bare metal assets feature is deprecated,

as it will potentially become something we need to come back to in the future and remove it (when indeed it gets pulled out of the code)

berenss (approved):
Suggest we include review from a member of the Central Infrastructure Management team, perhaps @filanov or Avishay Traeger

clusters/single_node_openshift.adoc
cadawson: I see your point. I reworded a little to avoid this.

clusters/single_node_openshift.adoc
cadawson: I prefer your idea, as I thought the same way. I thought it would be better to lead them over during the transition, but I agree that it is somewhere that we can miss and update. I removed it. Thanks!

clusters/single_node_openshift.adoc
```diff
@@ -45,8 +37,6 @@ Add the `spec:skipMachinePools` parameter in the YAML content with a value of `t
 skipMachinePools: true
 ----
```
berenss: @filanov @jelkosz please review lines above to ensure technically accurate for CIM SNO flow. thx!

clusters/single_node_openshift.adoc
filanov: Why only bare metal? vm can be an sno as well

berenss: I see your point - the OCP doc seems updated now, to reflect it with OCP 4.10:
https://docs.openshift.com/container-platform/4.10/installing/installing_sno/install-sno-preparing-to-install-sno.html

> Supported platforms: Installing OpenShift Container Platform on a single node is supported on bare metal, vSphere, Red Hat OpenStack, and Red Hat Virtualization platforms. In all cases, you must specify the platform.none: {} parameter in the install-config.yaml configuration file.

clusters/single_node_openshift.adoc
```diff
@@ -1,34 +1,26 @@
 [#single-node-openshift]
 = Creating a single node OpenShift cluster
 
-You can create a single node OpenShift (SNO) cluster that contains only a single node. In a standard cluster, there are control plane nodes and separate worker nodes. This configuration can be helpful when you have many clusters that need to minimize their resource footprint. 
+In a bare metal environment, you can create a single node OpenShift (SNO) cluster. A SNO cluster contains only a single node, which serves as the control plane node and the user workloads. This configuration can be helpful when you have many clusters that need to minimize their resource footprint.
```
avishayt: > A SNO cluster contains only a single node, which serves as the control plane node and the user workloads.
The wording is confusing here - to me, it reads that the node serves as the user workloads.  Maybe "which hosts the control plane and user workloads"?

avishayt: > This configuration can be helpful when you have many clusters that need to minimize their resource footprint.

Why many clusters?  It can be helpful when you have even one cluster that requires a small resource footprint.

clusters/single_node_openshift.adoc
```diff
@@ -1,34 +1,26 @@
...
 
 * A {product-title} 2.4, or later, hub cluster that is running on {ocp} version 4.9, or later.
-* Credentials to access the provider where you want to create your SNO cluster. The supported providers include: 
-+
-** Amazon Web Services
-** Microsoft Azure
-** Google Cloud Platform
-** Red Hat OpenStack Platform
-** VMware vSphere
+* Credentials for accessing the bare metal server where you want your SNO cluster.
```
avishayt: Same note as above about bare-metal focus.  Also, CIM doesn't require credentials for the servers as long as the user can boot the server with the discovery image.

clusters/single_node_openshift.adoc
```diff
@@ -1,34 +1,26 @@
...
 
 [#sno_create]
 == Create a Single-Node OpenShift cluster
 
 To create a SNO cluster, complete the following steps:
 
-Complete the steps in the {product-title-short} console for creating a cluster on one of the providers that are supported to run a SNO cluster in xref:../clusters/create.adoc#creating-a-cluster[Creating a cluster].
+Complete the steps in the {product-title-short} console for creating a cluster in the bare metal environment. It is best practice to use the Central Infrastructure Management method of creating a SNO cluster. The procedure to create a SNO cluster is essentially the same as creating an on-premises cluster. See xref:../clusters/create_cluster_on_prem.adoc#creating-a-cluster-on-premises[Creating a cluster in an on-premises environment].
```
avishayt: Seems weird to contrast an SNO cluster with an on-premises cluster, because SNO is on-prem in this case as well.

clusters/single_node_openshift.adoc
berenss: Agree with your rewrite, as it captures our intent: 
"which hosts the control plane services and also the user workloads"

berenss: agree - strikeout 'many' and we should be good.

clusters/single_node_openshift.adoc
berenss: Agree. perhaps all we need here then, is a reminder to define the On-Premises credential: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/credentials/credentials#creating-a-credential-for-an-on-premises-environment

clusters/single_node_openshift.adoc
berenss: It's true. a great point overall. we have 2 topics,
[1.7.9. Creating a cluster in an on-premises environment](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/clusters/managing-your-clusters#creating-a-cluster-on-premises)

[1.7.10. Creating a single node OpenShift cluster](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/clusters/managing-your-clusters#single-node-openshift)

but ultimately, I'm not sure why. Perhaps the only need we have is to update 1.7.9 with a short message about how to deploy the SNO, which is to ensure the release image >=4.9 and check the box.

clusters/single_node_openshift.adoc
```diff
@@ -1,34 +1,25 @@
...
 
 * A {product-title} 2.4, or later, hub cluster that is running on {ocp} version 4.9, or later.
-* Credentials to access the provider where you want to create your SNO cluster. The supported providers include: 
-+
-** Amazon Web Services
-** Microsoft Azure
-** Google Cloud Platform
-** Red Hat OpenStack Platform
-** VMware vSphere
+* Credentials for accessing the where you want your SNO cluster.
```
dockerymick: Credentials for accessing the provider where you want your SNO cluster.

clusters/single_node_openshift.adoc
```diff
@@ -1,34 +1,25 @@
...
+* An On-Premises credential defined and available to select when you create the cluster. See link:../credentials/credential_on_prem.adoc##creating-a-credential-for-an-on-premises-environment[Creating a credential for an on-premises environment] for more information about the credential.
```
dockerymick: An on-premises credential that is defined and available to select when you create the cluster. 

Remove the extra hashtag the comes before the anchor name 

change `##creating` to `#creating`

clusters/single_node_openshift.adoc
```diff
@@ -1,34 +1,25 @@
 [#single-node-openshift]
 = Creating a single node OpenShift cluster
```
dockerymick: Later in the doc, you write out single node OpenShift cluster as _Single-Node OpenShift cluster_

I took a look at OpenShift docs and they doc it as : `single-node OpenShift cluster`. I do not see the abbreviation used at all in their doc

To be consistent with OCP doc, I recommend updating the title from `Creating single node OpenShift cluster` to `Creating single-node OpenShift cluster`

clusters/single_node_openshift.adoc
```diff
@@ -1,34 +1,25 @@
...
 
-You can create a single node OpenShift (SNO) cluster that contains only a single node. In a standard cluster, there are control plane nodes and separate worker nodes. This configuration can be helpful when you have many clusters that need to minimize their resource footprint. 
+On VMware vSphere, Red Hat OpenStack, Red Hat Virtualization Platform, and in the bare metal environment, you can create a single node OpenShift (SNO) cluster. A SNO cluster contains only a single node which hosts the control plane services and the user workloads. This configuration can be helpful when you want to minimize the resource footprint of the cluster.
```
dockerymick: I would switch the wording around a bit: 

You can create a single-node OpenShift cluster on VMware vSphere, Red Hat OpenStack, Red Hat Virtualization Platform, and in the bare metal environment. 

I suggest that we don't use SNO at all to remain consistent with OCP docs

dockerymick (changes requested):

clusters/single_node_openshift.adoc
berenss: people will be quickly searching for the word SNO. I suggest we have it somewhere

clusters/single_node_openshift.adoc
cadawson: Changed the later one. Only an interface text reference remains. Thanks!

clusters/single_node_openshift.adoc
cadawson: SNO appears to be one of those inconsistent things. There are some uses of it in the OpenShift documentation, but I don't see many. I will define it on the first occurrence and then just use SNO. Then we have both.

clusters/single_node_openshift.adoc
cadawson: Correct. Thanks

clusters/single_node_openshift.adoc
```diff
@@ -11,8 +11,8 @@ This procedure explains how to deploy a single SNO cluster. You can also test th
 == Prerequisites
 
 * A {product-title} 2.4, or later, hub cluster that is running on {ocp} version 4.9, or later.
-* Credentials for accessing the where you want your SNO cluster.
-* An On-Premises credential defined and available to select when you create the cluster. See link:../credentials/credential_on_prem.adoc##creating-a-credential-for-an-on-premises-environment[Creating a credential for an on-premises environment] for more information about the credential.
+* Credentials for accessing the provider where you want your SNO cluster.
```
avishayt: Why are provider credentials required?  It's none platform.

berenss: Essentially it is only these things:
- pull secret
- ssh
- base domain is optional
image

In the AI SaaS those are fields that the user populates in-line as they create the cluster... in ACM, we have Credentials that are defined for all infra types, On-Premises means they will be using ACM+AI in the CIM flow.

avishayt: OK, I understand.  I read "provider credentials" as vSphere credentials, for example.

berenss:
My overall comment about the SNO topic remains:
- why do we insist on keeping the single_node_openshift.adoc topic? 
- to me, SNO is just a selection on the CIM create flow, and may not need to be covered as extensively in ACM, as it is very well covered with many details in the OCP doc.

clusters/create_cluster_on_prem.adoc
```diff
@@ -3,6 +3,10 @@
 
 You can use the {product-title} console to create an on-premises {ocp} cluster. This process is recommended instead of the bare metal process that is deprecated.
 
+It is best practice for you to use this procedure for creating a single-node OpenShift (SNO) cluster. You can create a single-node OpenShift cluster on VMware vSphere, Red Hat OpenStack, Red Hat Virtualization Platform, and in a bare metal environment. A single-node OpenShift cluster contains only a single node which hosts the control plane services and the user workloads. This configuration can be helpful when you want to minimize the resource footprint of the cluster.
```
dockerymick: Add a comma in the following statement: A single-node OpenShift cluster contains only a single node, which hosts the control plane services and the user workloads.

clusters/create_cluster_on_prem.adoc
```diff
@@ -62,6 +80,91 @@ After the hosts are bound, and the validations pass, complete the networking inf
 +
 *Note:* This value must match the name that you used to create the DNS records listed in the prerequisites section. If not provided, the DNS must be pre-configured so that `test.apps.` resolves correctly.
 
-You can view the status of the installation on the _Clusters_ navigation page. 
+You can view the status of the installation on the _Clusters_ navigation page.
 
 Continue with xref:../clusters/access_cluster.adoc#accessing-your-cluster[Accessing your cluster] for instructions for accessing your cluster. 
+
+[#dynamic-metrics-for-sno]
```
dockerymick: As discussed through Slack


"I agree with your suggestion to add Dynamic metrics for SNO clusters section to the  Observing environments documentation, It might be best to add the section after this section: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/observability/observing-environments-intro#external-metric-query "

dockerymick (approved):
Made a couple of suggestions, thanks for pulling me in

clusters/create_cluster_on_prem.adoc
cadawson: Thanks @dockerymick. Moved to suggested location.

cadawson:
I see your point @berenss and @avishayt . It was initially created as a separate file when it was tech preview, then became dev preview. We had to do some swizzling of some content, but I think we have it straight now. Moved SNO content to on premises cluster, and shifted the SNO data information to observability. See what you think. Thanks!

berenss (approved):
I like it!

dockerymick (approved):
I updated the link for the entry in What's new. This looks good, thanks Chris!

cadawson:
@filanov : Does this look good from a CIM standpoint, now? Thanks!

cadawson:
Adding @omertuc as a reviewer for Michael Filanov.

clusters/create_cluster_on_prem.adoc
```diff
@@ -3,6 +3,10 @@
 
 You can use the {product-title} console to create an on-premises {ocp} cluster. This process is recommended instead of the bare metal process that is deprecated.
 
+It is best practice for you to use this procedure for creating a single-node OpenShift (SNO) cluster. You can create a single-node OpenShift cluster on VMware vSphere, Red Hat OpenStack, Red Hat Virtualization Platform, and in a bare metal environment. A single-node OpenShift cluster contains only a single node, which hosts the control plane services and the user workloads. This configuration can be helpful when you want to minimize the resource footprint of the cluster.
```
omertuc: >You can create a single-node OpenShift cluster on VMware vSphere, Red Hat OpenStack, Red Hat Virtualization Platform, and in a bare metal environment. 

It's worth mentioning that even though you can deploy single node on all of these platforms, it will be deployed with absolutely no platform integration ("none platform")

clusters/single_node_openshift.adoc
omertuc: I agree, "credentials for accessing the provider" doesn't sound right to refer to pullsecret/ssh/base domain

clusters/create_cluster_on_prem.adoc
```diff
@@ -42,7 +47,20 @@ Every managed cluster must be associated with a managed cluster set. If you do n
...
 
-A single-node OpenShift (SNO) cluster contains a single node which hosts the control plane services and the user workloads. This configuration can be helpful when you want to minimize the resource footprint of the cluster. If you want your cluster to be a SNO cluster, select the SNO option. SNO clusters are only supported on {ocp-short} version 4.9, and later. If you did not select a release image of {ocp-short} version 4.9, or later, the option for selecting a SNO cluster is not available. 
+The *OpenShift version* identifies the version of the {ocp-short} image that is used to create the cluster. If the version that you want to use is available, you can select the image from the list of images. If the image that you want to use is not a standard image, you can enter the URL to the image that you want to use. See xref:../clusters/release_images.adoc#release-images[Release images] for more information about release images. 
+
+When you select an OpenShift version that is 4.9 or later, an option to select *Install single node OpenShift (SNO)* is displayed. A single-node OpenShift cluster contains a single node which hosts the control plane services and the user workloads.
```
omertuc: If these docs are for 2.5, it's probably worth mentioning that adding additional nodes to a single node cluster is not supported, as that is only supported starting ACM 2.6 and OCP 4.11

clusters/create_cluster_on_prem.adoc
```diff
@@ -42,7 +47,20 @@ Every managed cluster must be associated with a managed cluster set. If you do n
...
+
+If you want your cluster to be a single-node OpenShift cluster, select the single-node OpenShift option. If you are creating a single-node OpenShift cluster, add the `spec:skipMachinePools` parameter in the YAML content with a value of `true` by completing the following steps:
```
omertuc: >add the `spec:skipMachinePools` parameter in the YAML content with a value of `true` by completing the following steps

Which YAML is this referring to? I've never heard about such requirement. Could you give me some more details about this?

clusters/single_node_openshift.adoc
```diff
@@ -43,8 +36,6 @@ Add the `spec:skipMachinePools` parameter in the YAML content with a value of `t
 skipMachinePools: true
 ----
 
-*Important:* When you create an {ocp-short} SNO cluster on GCP, you must configure more CPU and memory resources than for a standard GCP cluster. The minimum requirements for a SNO cluster on GCP is 8 CPUs and 32 GB of memory. The standard requirements of 4 CPUs and 16 GB of memory is not stable for this type of cluster.
```
omertuc, line 46: This paragraph was deleted, but I don't see it's getting re-added anywhere? Was it deemed unimportant? Because it seems important to me

cadawson: @omertuc GCP is not listed as one of the supported platforms. Should we add it to the list and add this note back? Thanks! Added the note back, for now.

omertuc: Oh sorry, I only skimmed the paragraph and didn't notice it was talking about GCP specifically, I thought it was talking generally about SNO's exceptional requirements.

omertuc: Still I would expect there to be a general paragraph about how a SNO control-plane node needs 4 more CPU cores than a control-plane node in a multi-node control-plane cluster (8 cores rather than 4 cores)

omertuc: Not sure where this GCP being exceptional stuff came from, I've never heard of that

clusters/single_node_openshift.adoc
berenss: I don’t think we need this and suggest we remove it.

clusters/single_node_openshift.adoc
berenss: All of it should be removed. 

We had a dev preview flag that would allow a checkbox for SNO to be created via Hive. It was never a customer facing flag, just for internal/lab use. I think this is where the guidance of this section must have come from (engineering may have suggested it not knowing for sure that it was never intended to be customer facing)

clusters/create_cluster_on_prem.adoc
berenss: I have no idea. Suggest we remove it.

clusters/create_cluster_on_prem.adoc
berenss: To be explicit, `platform=none`

clusters/single_node_openshift.adoc
cadawson: Changed the phrasing to make it clearly a credential.

clusters/single_node_openshift.adoc
cadawson: Removed the GCP specific note. Added the note about the additional core requirement. 
Removed the skipMachinePools steps.

cadawson:
Thanks for your input on this @omertuc . I know you were pulled in at the end of it. I addressed your comments, and I think we are ok to move forward with this. Let me know if you see any other issues or if it looks ok. Thanks!

Thanks for your input, as well @berenss . Particularly because you are out of the office.

clusters/create_cluster_on_prem.adoc
```diff
@@ -51,18 +51,9 @@ The *OpenShift version* identifies the version of the {ocp-short} image that is
...
-
-. Find the `spec` section for your cluster and add the following flag to the `spec` section:
-+
-[source,yaml]
-----
-skipMachinePools: true
-----
-
-*Important:* When you create a single-node OpenShift cluster on Google Cloud Platform, you must configure more CPU and memory resources than for a standard GCP cluster. The minimum requirements for a single-node OpenShift cluster on GCP is 8 CPUs and 32 GB of memory. The standard requirements of 4 CPUs and 16 GB of memory is not stable for this type of cluster. 
+*Note:* The single-node OpenShift control plane requires 4 additional CPU cores than a control plane node requires for a multinode control plane cluster. For example, the single-node OpenShift cluster requires 8 cores while a multinode cluster requires 4 cores.
```
omertuc: The +4 thing is not a "rule" and is not true in general, it just currently happens to be 4 more in SNO. So perhaps change the phrasing to something like:

```suggestion
*Note:* The single-node OpenShift control plane requires 8 CPU cores as opposed to a control plane node for a multinode control plane cluster which only requires 4 CPU cores.
```

cadawson: Got it. Content updated. Thanks @omertuc .

clusters/single_node_openshift.adoc
omertuc: I'm confused, this comment thread and the docs it comments on seem to be mixing the concepts of "ACM credentials" and "provider credentials".

SNO doesn't need any "provider credentials", because it has no platform integration

omertuc: I believe the credentials that @berenss mentioned in https://github.com/stolostron/rhacm-docs/pull/3688/files#r930100187 are the credentials this document should be documenting, not provider credentials

cadawson: Hi @omertuc . You are correct. Scott is referring to the ACM credential. I removed the reference to the "provider credentials" and now have a reference to the ACM credential with a link to a topic that provides instructions about how to create one.

omertuc: The line still says "Credentials **for accessing the provider** where you want your SNO cluster."

PR #3690, opened 2022-07-11
stolostron/backlog#23198--adding content but needs work and explanation

swopebe:


cadawson (approved):
Looks good. Thanks! Merging.

cadawson:
This is already added in the prerequisites section.

clusters/hypershift_configure.adoc
```diff
@@ -15,6 +15,8 @@ true
 [#hypershift-convert-mgt-cluster]
 == Converting an existing cluster to a hosting service cluster
 
+//Chris please add description
+
```
swopebe, line 19: @cadawson we need to break up headers with a description about why the user needs the section. This is considered a short description and makes is easier to chunk content, also good for accessibility.

swopebe:
This should not have been merged without discussion, clarity, organization

PR #3691, opened 2022-07-11
stolostron/backlog#21770 proc must be done

cadawson:


oafischer (commented):
Looks good to me! I wonder if it might also make sense to switch to active voice while making this change. Something like "You must do/complete this procedure before you can create..."

credentials/credential_ansible.adoc
```diff
@@ -5,7 +5,7 @@ You need a credential to use {product-title} console to deploy and manage an {oc
 
 **Required access**: Edit
 
-*Note:* This procedure must be done before you can create a cluster with {product-title}.
+*Note:* This procedure must be done before you can create an Ansible job template to enable automation on a {product-title-short} cluster.
```
swopebe, line 8: @cadawson @oafischer @dockerymick  (I didn't get to see this but would like to use this):

If you revisit the active voice section on the IBM style manual, you see examples of both active voice and tense, as well as concise writing. This should be:

You must complete...

Or (to get rid of the must)

Complete this procedure before you create..

Yes, you may get this in the draft from devs but we know the style we should write in and why (this is not a new change to the manual), so let's work as a team to rewrite in active voice, avoiding unnecessary words and such.

No need to change, I see this in places in the doc for 2.5. I would just like to get better in this area for 2.6. #backtobasics in 2.6.

PR #3692, opened 2022-07-11
Add new known issue in 2.5

oafischer:
https://github.com/stolostron/backlog/issues/23631#issue-1276971319


cadawson (approved):
Looks good. Do we need any explanation about how to disable and enable the addon?

oafischer:
@cadawson Good point, let me reach out to the dev and ask about that, maybe we can link if there's already something on the doc (I searched and didn't find anything)

PR #3696, opened 2022-07-12
stolostron/backlog#22678 restate ki RHEL worker

cadawson:


release_notes/known_issues.adoc
```diff
@@ -1080,9 +1080,9 @@ If you have manually installed OADP, you must manually uninstall OADP before you
 Only OpenShiftSDN is supported as a CNI network provider. OVN is currently not supported.
 
 [#submariner-no-rhel]
-=== Submariner does not support Red Hat Enterprise Linux nodes as worker nodes
+=== Submariner does not support some Red Hat Enterprise Linux nodes as worker nodes
 
-When deploying Submariner on a cluster that includes Red Hat Enterprise Linux worker nodes, application workloads fail to communicate with remote clusters. Submariner only supports {ocp-short} clusters that use Red Hat Enterprise Linux CoreOS nodes.
+When deploying Submariner on a cluster that includes Red Hat Enterprise Linux worker nodes with the kernel version between 4.18.0-359.el8.x86_64 and 4.18.0-372.11.1.el8_6.x86_64, application workloads fail to communicate with remote clusters. Submariner only supports {ocp-short} clusters that use Red Hat Enterprise Linux CoreOS nodes or Red Hat Enterprise Linux worker nodes with a kernel outside of the specified range.
```
dfarrell07: > Submariner only supports {ocp-short} clusters that use Red Hat Enterprise Linux CoreOS nodes or Red Hat Enterprise Linux worker nodes with a kernel outside of the specified range.

I think we only support CoreOS nodes outside of this kernel range too, it's just that we don't expect users to run into them.

Maybe better to drop this second sentence, just have the line about the kernel versions with the bug.

sridhargaddam: > I think we only support CoreOS nodes outside of this kernel range too, it's just that we don't expect users to run into them.

Correct, as users will be using the official OCP-4.11 images which has the proper kernel version.
At the same time, if users use "4.18.0-372.11.1.el8_6.x86_64" kernel, there will not be any issue as this kernel includes the necessary fix.


Also, I'm trying to reach out to the QE team which originally reported the RHEL issue to understand the different flavours of RHEL. I'll keep this thread updated if we need this clarification or if we can remove this section all-together.

cadawson: Excellent. Thank you. I've updated the content. Why don't we update this now, and you can open a new issue if we need to change or remove it after you hear from QE? Will that work, @sridhargaddam ? @dfarrell07

nyechiel (approved):

release_notes/known_issues.adoc
sridhargaddam: Makes sense, thanks @cadawson

PR #3702, opened 2022-07-13
stolostron/backlog#22489 rem bm 2.4 on 4.8 note 2.4

cadawson:


cadawson:
Hi @elgnay . This issue was opened as a BZ issue to raise that it is no longer true. It looks like this known issue was added back in version 2.2, and has not been removed. Eveline confirmed that QE tested it, and it worked ok. I just wanted to make sure that someone in dev had look at it. I also removed the known issue from 2.5, as well. Thank you!

PR #3716, opened 2022-07-15
Cleaning up

oafischer:

Added some stuff and removed some stuff. Not sure if the simplified GitHub procedures doc is useful/something we want to pull into the actual pages at some point.


dockerymick (approved):
Looks good!

PR #3711, opened 2022-07-15
Clarify route opening in 2.4

oafischer:
https://github.com/stolostron/backlog/issues/21616#issue-1200212785

PR #3717, opened 2022-07-18
Clarify opening in 2.5

oafischer:
https://github.com/stolostron/backlog/issues/21616

Same as https://github.com/stolostron/rhacm-docs/pull/3711 but for 2.5

### DIFF
--- a/install/requirements.adoc
+++ b/install/requirements.adoc
@@ -48,7 +48,7 @@ For the hub cluster networking requirements, see the following table:
 
 | Outbound to managed cluster
 | HTTPS
-| Retrieval of logs dynamically from Search console for the pods of the managed cluster. This connection uses a service running on the managed cluster called - `klusterlet-addon-workmgr`
+| Retrieval of logs dynamically from Search console for the pods of the managed cluster. This connection creates a route called `klusterlet-addon-workmgr` in the `open-cluster-management-agent-addon` namespace of the managed cluster. The host of the route is `<route name>-<namespace>.apps.<cluster domain>`.
 | 443
 
 | Outbound to managed cluster


install/requirements.adoc
```diff
@@ -48,7 +48,7 @@ For the hub cluster networking requirements, see the following table:
 
 | Outbound to managed cluster
 | HTTPS
-| Retrieval of logs dynamically from Search console for the pods of the managed cluster. This connection uses a service running on the managed cluster called - `klusterlet-addon-workmgr`
+| Retrieval of logs dynamically from Search console for the pods of the managed cluster. This connection creates a route called `klusterlet-addon-workmgr` in the `open-cluster-management-agent-addon` namespace of the managed cluster. The host of the route is `<route name>-<namespace>.apps.<cluster domain>`.
```
cadawson, line 51: Not sure if something like this might help to clarify for the second sentence, if I am reading the request correctly: If your environment requires the full URI to open this connection, the route name is: `klusterlet-addon-workmgr-open-cluster-management-agent-addons.apps.<managed-cluster-domain>`. Not sure.

cadawson (changes requested):
One thing that struck me was that the originator said, "The existing text also mentions a service running on the managed cluster while the connection description is outbound to the hub cluster." 
I don't know if this was an error in the copy, but none of them read, "Outbound to the hub cluster." This makes me wonder if it was a mistake in copying the text, or if the text should actually read, "Outbound to the hub cluster", or if there is some error there. Maybe Zhiwei knows? 
Otherwise, looks good! Thanks!

oafischer:
@zhiweiyin318 Can you shed light on the theory about the outbound cluster as suggested by @cadawson [here](https://github.com/stolostron/rhacm-docs/pull/3717#pullrequestreview-1042538739?)?

install/requirements.adoc
oafischer: Not sure either, but I think I'll leave it as is since we already merged it in 2.4 with Zhiwei approving the changes as presented here. I did also pull in @zhiweiyin318 again to see if he can answer your other question/thought. Thanks.

zhiweiyin318:
@cadawson The host of the route is `klusterlet-addon-workmgr-open-cluster-management-agent-addon.apps.<managed-cluster-domain`. in the host name `klusterlet-addon-workmgr` is the route name and `open-cluster-management-agent-addon` is the namespace.

"The existing text also mentions a service running on the managed cluster while the connection description is outbound to the hub cluster."

yes, there is a service `klusterlet-addon-workmgr` running in the ns `open-cluster-management-agent-addon` on the managed cluster, the service is exposed via route if the cluster is OCP,  via LoadBalancer if the cluster is non-OCP. 
currently only the hub cluster can access the service.

cadawson (approved):
After quick discussion, looks good. Thank you!

PR #3721, opened 2022-07-18
Add new known Submariner issue in 2.5

oafischer:
https://github.com/stolostron/backlog/issues/24258

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -1029,7 +1029,7 @@ Submariner is not validated for clusters that are provisioned in an air-gapped e
 You cannot deploy multiple gateways.
 
 [#submariner-vxlan]
-=== Submariner does not support VXLAN with NAT enabled
+=== Submariner does not support VXLAN when NAT is enabled
 //2.5:24258
 
-Submariner with the VXLAN cable driver is not supported with NAT enabled. Submariner with the VXLAN cable driver is currently only supported in non-NAT deployments.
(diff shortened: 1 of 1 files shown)


release_notes/known_issues.adoc
```diff
@@ -1027,3 +1027,9 @@ Submariner is not validated for clusters that are provisioned in an air-gapped e
 //2.5:23706
 
 You cannot deploy multiple gateways.
+
+[#submariner-vxlan]
+=== Submariner does not support VXLAN with NAT enabled
```
cadawson: How about "Submariner does not support VXLAN when NAT is enabled"?

release_notes/known_issues.adoc
```diff
@@ -1027,3 +1027,9 @@ Submariner is not validated for clusters that are provisioned in an air-gapped e
...
+//2.5:24258
+
+Submariner with the VXLAN cable driver is not supported with NAT enabled. Submariner with the VXLAN cable driver is currently only supported in non-NAT deployments.
```
cadawson: It sounds like we are saying the same thing in these sentences. If it works, consider changing the second sentence to read more like a workaround. For example: "To use Submariner with the VXLAN cable driver, deploy it in an environment where NAT is not enabled." Or similar. Maybe even remove the first sentence and just keep the second one.

cadawson (approved):
Looks good, Oliver! Thanks for documenting that. I added a couple of considerations, but I don't need to see it again.

PR #3731, opened 2022-07-20
stolostron/backlog#23682 Add S3 credential

cadawson:


clusters/hosted_control_plane_configure.adoc
```diff
@@ -79,14 +79,31 @@ oc apply -f <filename>
 +
 Replace `filename` with the name of the file that you created. 
 
-. Because hosted control planes is a Technology Preview feature only on AWS, create an OIDC S3 credentials secret named `hypershift-operator-oidc-provider-s3-credentials` for the HyperShift operator. Save the secret in the managed cluster namespace, which is the namespace of the managed cluster that is used as the hosting service cluster. If you used `local-cluster`, then create the secret in the `local-cluster` namespace
+. Because hosted control planes is a Technology Preview feature only on AWS, create an OIDC S3 credentials secret named `hypershift-operator-oidc-provider-s3-credentials` for the HyperShift operator. HyperShift operatorSave the secret in the managed cluster namespace, which is the namespace of the managed cluster that is used as the hosting service cluster. If you used `local-cluster`, then create the secret in the `local-cluster` namespace.
```
dockerymick, line 82: seems like the term `HyperShift operatorSave` is misplaced. Do you mean to say, "Save the HyperShift operator secret in the managed cluster namespace, which is the..."?

dockerymick (changes requested):

clusters/hosted_control_plane_configure.adoc
cadawson: Good catch. That was a miss on my part. Thanks!

clusters/hosted_control_plane_configure.adoc
```diff
@@ -79,14 +79,31 @@ oc apply -f <filename>
...
 * `region`: Region of the s3 bucket
 
++
+If you are deploying the `hypershift-operator` using mulicluster engine, you must have an S3 secret file available for the configuration on any platform. If you are not running AWS, complete the following steps to create a substitute file: 
+
+.. Create a file named `s3-credentials-placeholder` that contains the following content:
++
+----
+[default]
+aws_access_key_id=ABCDEFGHIJKLMNOPQRST
```
jnpacker, line 97: Don't we usually include some text instead of a FAKE key?

clusters/hosted_control_plane_configure.adoc
jnpacker, line 82: Instead of "save" we should use "apply" or "create"

cadawson:
It was determined that his was not required. It is fixed in 2.5.2 and in 2.6.

PR #3734, opened 2022-07-20
Change key name in 2.4

oafischer:
https://github.com/stolostron/backlog/issues/22211

### DIFF
--- a/clusters/upgrade_cluster_discon.adoc
+++ b/clusters/upgrade_cluster_discon.adoc
@@ -133,7 +133,7 @@ See https://docs.openshift.com/container-platform/4.9/registry/configuring-regis
 
 . Create a ConfigMap in the `openshift-config` namespace. 
 
-. Add your CA certificate under the key `cincinnati-registry`. {cincinnati-short} uses this setting to locate your certificate:
+. Add your CA certificate under the key `updateservice-registry`. {cincinnati-short} uses this setting to locate your certificate:
 +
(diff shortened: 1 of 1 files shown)


clusters/upgrade_cluster_discon.adoc
```diff
@@ -133,7 +133,7 @@ See https://docs.openshift.com/container-platform/4.9/registry/configuring-regis
 
 . Create a ConfigMap in the `openshift-config` namespace. 
 
-. Add your CA certificate under the key `cincinnati-registry`. {cincinnati-short} uses this setting to locate your certificate:
+. Add your CA certificate under the key `updateservice-registry`. {cincinnati-short} uses this setting to locate your certificate:
```
mikeshng, line 136: `{cincinnati-short}`

Maybe this needs to be updated? I am not familiar with the doc so not sure if this is a key and there is a value associate with it. If there is a value then that needs to be change too.

mikeshng (changes requested):

clusters/upgrade_cluster_discon.adoc
oafischer: `{cincinnati-short}` is just a common attribute, a way to define a name once globally. You can find it along with the other attributes [here](https://github.com/stolostron/rhacm-docs/blob/2.5_stage/modules/common-attributes.adoc).

It shouldn't have anything to do with the actual key value.

PR #3737, opened 2022-07-20
Adding known issue for backup and restore 2.5

dockerymick:
https://github.com/stolostron/backlog/issues/22402


swopebe (approved):
made a small commit. Your comment on 997 is up to you and dev, simply remove comment if not needed.

sahare (approved):
Thanks Mikela, looks good, and I'm not aware of any example we can provide.

PR #3739, opened 2022-07-21
Add note in 2.5

oafischer:
https://github.com/stolostron/backlog/issues/23931

### DIFF
--- a/clusters/create_azure.adoc
+++ b/clusters/create_azure.adoc
@@ -36,6 +36,8 @@ If you need to create a credential, see link:../credentials/credential_azure.ado
 
 The name of the cluster is used in the hostname of the cluster.
 
+*Note:* The Red Hat Advanced Cluster Managament for Kubernetes controller creates a namespace for clusters you create. Destroying the cluster deletes the namespace. Only use resources of the cluster instance for this namespace.
+
 *Tip:* Select *YAML: On* to view content updates as you enter the information in the console. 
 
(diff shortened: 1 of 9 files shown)


clusters/create_azure.adoc
```diff
@@ -36,6 +36,8 @@ If you need to create a credential, see link:../credentials/credential_azure.ado
 
 The name of the cluster is used in the hostname of the cluster.
 
+*Note:* The Red Hat Advanced Cluster Managament for Kubernetes controller creates a namespace for clusters you create. Destroying the cluster deletes the namespace. Only use resources of the cluster instance for this namespace.
```
cadawson: This should probably be an important note. There is a chance that they can lose data. 

*Important:* When you create a cluster, the {product-title-short} controller creates a namespace for the cluster and its resources. Ensure that you include only resources for that cluster instance in that namespace. Destroying the cluster deletes the namespace and all of the resources in it.

cadawson (changes requested):
Comment added. Should this also be added to create_bare.adoc? Thanks!

oafischer:
> Comment added. Should this also be added to create_bare.adoc? Thanks!

Good catch on create_bare.adoc @cadawson, that should be included too. I'll add the change there too and address your comment, thanks!

clusters/create_azure.adoc
oafischer: I like your version, thanks for providing that. I'll go ahead and implement it everywhere as an important note.

cadawson (approved):
Looks good. Thank you!

PR #3740, opened 2022-07-21
https://github.com/stolostron/backlog/issues/24449 -- 2.2 notice

swopebe:


cadawson (changes requested):
Made a minor update, but looks good. Thanks for putting this together. 
I am torn about the third sentence...should it be removed as redundant? 

Feel free to merge when you are ready. What you decide is fine with me. I don't need to see it again. Just requested changes so you would know that you would know that I made some changes. Thanks!

PR #3746, opened 2022-07-22
Updates for must-gather

dockerymick:
https://github.com/stolostron/backlog/issues/22770

### DIFF
--- a/troubleshooting/must_gather.adoc
+++ b/troubleshooting/must_gather.adoc
@@ -30,13 +30,6 @@ See the following procedure to start using the `must-gather` command:
 ----
 oc adm must-gather --image=registry.redhat.io/rhacm2/acm-must-gather-rhel8:v2.5.0 --dest-dir=<directory>
 ----
-+
-*Note:* To receive a complete data set, the following `must-gather` command must be ran for the multicluster engine, as well:
-+
-----
-oc adm must-gather --image=registry.redhat.io/multicluster-engine/must-gather-rhel8:v2.0.0 --dest-dir=<directory>
-----
-+
 
 . Log in to your cluster. For the usual use-case, you should run the `must-gather` while you are logged into your hub cluster. 
 
--- a/troubleshooting/must_gather.adoc
+++ b/troubleshooting/must_gather.adoc
@@ -38,7 +38,7 @@ oc adm must-gather --image=registry.redhat.io/multicluster-engine/must-gather-rh
 ----
 +
 
-. Log in to your cluster. For the usual use-case, you should run the `must-gather` while you are logged into your _hub_ cluster. 
+. Log in to your cluster. For the usual use-case, you should run the `must-gather` while you are logged into your hub cluster. 
 
 +
 *Note:* If you want to check your managed clusters, find the `gather-managed.log` file that is located in the the `cluster-scoped-resources` directory:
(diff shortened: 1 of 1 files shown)


troubleshooting/must_gather.adoc
```diff
@@ -25,33 +25,32 @@ See the following procedure to start using the `must-gather` command:
...
+. Add the {product-title} image that is used for gathering data and the directory. Run the following command, where you insert the image and the directory for the output:
 +
-*Note:* If you want to check your managed clusters, find the `gather-managed.log` file that is located in the the `cluster-scoped-resources` directory:
+----
+oc adm must-gather --image=registry.redhat.io/rhacm2/acm-must-gather-rhel8:v2.5.0 --dest-dir=<directory>
+----
 +
-
+*Note:* To receive a complete data set, the following `must-gather` command must be ran for the multicluster engine, as well:
 +
```
swopebe: This should be in the MCE docs; we need to keep RHACM and MCE separated.

swopebe: The multicluster engine operator (see that folder for the doc where this should go).

dockerymick: Ok, ,got it. So this note should most likely be removed?

swopebe:
Though I do not think many are using it @dockerymick MCE has it's own here: https://github.com/stolostron/rhacm-docs/blob/2.6_stage/multicluster_engine/must_gather.adoc

If that is sufficient, then I think you can keep your doc RHACM-centric without changing the MCE doc.

swopebe (approved):
approving based on separation on MCE and RHACM references.

troubleshooting/must_gather.adoc
```diff
@@ -25,11 +25,16 @@ See the following procedure to start using the `must-gather` command:
...
 . Learn about the `must-gather` command and install the prerequisites that you need at https://docs.openshift.com/container-platform/4.9/support/gathering-cluster-data.html[Gathering data about your cluster] in the {ocp} documentation.
 
-. Log in to your cluster. For the usual use-case, you should run the `must-gather` while you are logged into your _hub_ cluster. 
+. Add the {product-title} image that is used for gathering data and the directory. Run the following command, where you insert the image and the directory for the output:
++
+----
+oc adm must-gather --image=registry.redhat.io/rhacm2/acm-must-gather-rhel8:v2.5.0 --dest-dir=<directory>
+----
+
+. Log in to your cluster. For the usual use-case, you should run the `must-gather` while you are logged into your hub cluster.
```
dislbenn: @dockerymick should we include `Log in to your cluster` in step two? I can scenarios where people would attempt to run step two before logging into the cluster.

dislbenn (approved):

troubleshooting/must_gather.adoc
dockerymick: @dislbenn To be sure that I am understanding your suggestion.  Add the instruction to "Log in.." should be moved to step 2? 

2. Log in to your cluster. Add the {product-title} image that is used for gathering data and the directory. Run the following command, where you insert the image and the directory for the output:

   ```
   oc adm must-gather --image=registry.redhat.io/rhacm2/acm-must-gather-rhel8:v2.5.0 --dest-dir=<directory>
   ```

3. For the usual use-case, you should run the `must-gather` while you are logged into your hub cluster.

dislbenn: Yes. That is what I would recommended.

PR #3751, opened 2022-07-22
stolostron/backlog#24234 updates

cadawson:


cadawson:
Re-review for headers and such. Thanks @swopebe !

clusters/hosted_control_planes_configure.adoc
```diff
@@ -1,6 +1,8 @@
-[#hosted-control-plane-configure]
+[#hosted-control-planes-configure]
 = Configuring hosted control planes
 
+Configuring hosted control planes requires a hosting service cluster and a hosted cluster. By deploying the HyperShift operator on an existing cluster, you can make that cluster into a hosting service cluster and start the creation of the hosted cluster. 
+
 Hosted control planes is a Technology Preview feature, so the related components are disabled by default. Enable the feature by editing the `multiclusterengine` custom resource to set the `spec.overrides.components[?(@.name=='hypershift-preview')].enabled` to `true`.
```
swopebe, line 6: We need to tag this as such in the title...I believe I held and architecture meeting and we all agreed on this maybe for 2.0 and after, I think we documented that, as well.

swopebe:
Few small comments. I put the issue into awaiting.

clusters/hosted_control_planes_configure.adoc
```diff
@@ -85,10 +87,10 @@ The secret must contain 3 fields:
 +
 * `bucket`: An S3 bucket with public access to host OIDC discovery documents for your HyperShift clusters
 * `credentials`: A reference to a file than contains the credentials of the `default` profile that can access the bucket. By default, HyperShift only uses the `default` profile to operate the `bucket`.
-* `region`: Region of the s3 bucket
+* `region`: Region of the S3 bucket
```
rokej: This whole block about creating an OIDC S3 credentials secret should go before creating the ManagedClusterAddOn. Also, we should remove `Because hosted control planes is a Technology Preview feature only on AWS,` and instead say something like `If you plan to create and manage hosted clusters on AWS`.

clusters/remove_hosted_control_planes.adoc
```diff
@@ -0,0 +1,22 @@
...
+To destroy a HyperShift hosted cluster, delete the `HypershiftDeployment` resource by running the following command: 
+
+----
+oc delete hypershiftdeployment hypershift-demo -n default
+---- 
+
+[#hypershift-uninstall-operator]
+== Uninstalling the HyperShift operator
+
+To uninstall the HyperShift operator, delete the `hypershift-addon` by running the following command:
```
rokej: To uninstall the HyperShift operator from a managed or hosting service cluster, delete the hypershift-addon ManagedClusterAddon from the managed cluster.

clusters/remove_hosted_control_planes.adoc
```diff
@@ -0,0 +1,22 @@
+[#remove-hosted-control-planes]
+= Removing the hosted control plane resources
+
+When removing the hosted control plane cluster feature, you must destroy the HyperShift hosted cluster and uninstall the HyperShift operator. 
+
+[#hypershift-cluster-destroy]
+== Destroying a HyperShift hosted cluster
+
+To destroy a HyperShift hosted cluster, delete the `HypershiftDeployment` resource by running the following command:
```
rokej: oc delete -f HypershiftDeployment_YAML_file_name

OR

oc delete hd HypershiftDeployment_resource_name

clusters/remove_hosted_control_planes.adoc
```diff
@@ -0,0 +1,22 @@
+[#remove-hosted-control-planes]
+= Removing the hosted control plane resources
+
+When removing the hosted control plane cluster feature, you must destroy the HyperShift hosted cluster and uninstall the HyperShift operator.
```
rokej: I am not sure what this means.

rokej:
@philipwu08 Could you please review this as well?

clusters/remove_hosted_control_planes.adoc
zanetworker: Do you mean management instead of managed?

clusters/remove_hosted_control_planes.adoc
zanetworker: Did you mean "When disabling the hosted control planes preview feature"?

clusters/remove_hosted_control_planes.adoc
cadawson: Made this management. Let me know if that is not what was intended.

clusters/remove_hosted_control_planes.adoc
cadawson: I changed this to disable, rather than remove. That sounds like it would be more accurate.

cadawson: Changed it to disable, instead of removing. Maybe that helps? @rokej

clusters/hosted_control_planes_configure.adoc
cadawson: Removed it from the text. It is in the parent topic, so I think we are ok there.

clusters/disable_hosted_control_planes.adoc
```diff
@@ -0,0 +1,28 @@
...
+To destroy a HyperShift hosted cluster, delete the `HypershiftDeployment` resource by running one of the following commands: 
+
+----
+oc delete -f <HypershiftDeployment_yaml_file_name>
+----
+
+or 
+
+----
+oc delete hd <HypershiftDeployment_resource_name>
```
philipwu08: For consistency, maybe we should also include the namespace?

oc delete hd -n <HypershiftDeployment_namespace> <HypershiftDeployment_resource_name>

PR #3754, opened 2022-07-25
stolostron/backlog#21400 klusterlet operator added (2.4 +)

swopebe:


cadawson (approved):
Looks good. I changed klusterler to klusterlet, which I think was reasonable. Hope that wasn't intentional. Thanks!

PR #3769, opened 2022-07-29
Removing 2.3+ notice with refreshing 2.5 notice (all in one)

swopebe:
https://github.com/stolostron/backlog/issues/21400 

Since I am updating the deprecation for 2.5 and earlier, I went ahead and confirmed that for 2.6 that this would be removed.


dockerymick (approved):
Looks good! Thanks @swopebe

PR #3772, opened 2022-07-29
Fix azure cloud prepare steps

aswinsuryan:
Add network security group rules to allow AH and ESP protocol

Signed-off-by: Aswin Suryanarayanan <aswinsuryan@gmail.com>

### DIFF
--- a/add-ons/submariner/submariner_prepare_hosts.adoc
+++ b/add-ons/submariner/submariner_prepare_hosts.adoc
@@ -31,7 +31,7 @@ Replace `gw-vm-nic` with the interface address.
 az network nsg create --name <gw-nsg-name> --resource-group <res-group>
 ----
 
-. Create network security groups rules in your Azure environment to open the tunnel port (4500/UDP by default), NAT discovery port (4490/UDP by default) and metrics port (8080/TCP and 8081/TCP by default) for Submariner. This needs to be created in both inbound and outbound directions for each of the ports.
+. Create network security groups rules in your Azure environment to open the tunnel port (4500/UDP by default), NAT discovery port (4490/UDP by default) and metrics port (8080/TCP and 8081/TCP by default) for Submariner. These rules need to be created in both inbound and outbound directions for each of the ports.
 +
 ----
 az network nsg rule create --resource-group <res-group> \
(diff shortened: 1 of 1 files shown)


add-ons/submariner/submariner_prepare_hosts.adoc
```diff
@@ -45,6 +45,20 @@ az network nsg rule create --resource-group <res-group> \
 --protocol <Protocol> --destination-port-ranges <port>
 ----
 
+. Create network security groups rules to allow ESP and AH protocols. This needs to be created in both inbound and outbound directions for both the protocols.
```
sridhargaddam: The Azure documentation shows Esp and Ah (i.e., the only first letter is capital) as protocol values. Can you please check if ESP/AH is an acceptable value?

https://docs.microsoft.com/en-us/cli/azure/network/nsg/rule?view=azure-cli-latest

cadawson: I don't think this is a question for me, but I will provide the docs answer in case it helps. We use the full caps version after spelling it out the first time.

cadawson:
Thank you @aswinsuryan . I did a quick review of it, and added a few minor edits. Questions that I had for this:
1. Is this an optional step?
2. Does it apply to 2.6, as well? 
Thanks!

cadawson:
I don't know if @aswinsuryan is getting notifications from this. @sridhargaddam , can you help? Thanks!

sridhargaddam:
> Thank you @aswinsuryan . I did a quick review of it, and added a few minor edits. Questions that I had for this:
> 
> 1. Is this an optional step?

Initially @aswinsuryan suspected that these rules are mandatory, but on further analysis we realised that its not the case. We can still go ahead with the PR, address any review comments and merge it - as it will help deployments where NAT is not used between the clusters.

> 2. Does it apply to 2.6, as well?

In Submariner 0.13.0 release (ACM 2.6) we support auto-cloud preparation for Azure clusters from UI (and as part of subctl cloud prepare ....), so these manual steps are not required.

cadawson:
Thanks @sridhargaddam and @aswinsuryan . The review comments should be addressed.  If it is ok, we can go ahead and merge, if you will approve the PR. Thanks!

nyechiel:
@aswinsuryan @sridhargaddam do we actually want to merge this in? My understanding is that this is not the root cause for the issue Maxim reported.

aswinsuryan:
@nyechiel it would still good to have for the cases where UDP node is not enabled for ipsec

sridhargaddam:
Yes, to be aligned with what we do in cloud prepare, I feel we should add these rules as it will help when there is no NAT involved between the clusters.

PR #3777, opened 2022-08-01
Fix the OpenShift GitOps init command

mprahl:
Some users report that the other command didn't work. This new approach
seems to work for everyone.

This also changes it so that the image tags use the `{product-version}`
variable.

Relates:
https://github.com/stolostron/backlog/issues/24747

### DIFF
--- a/governance/policy_generator.adoc
+++ b/governance/policy_generator.adoc
@@ -430,7 +430,7 @@ oc -n openshift-gitops edit argocd openshift-gitops
 ----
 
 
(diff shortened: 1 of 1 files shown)


dockerymick (approved):
Thanks for this update Matt! You used the conref/variable perfectly! Approving and will approve your other PR

cadawson (commented):
Looks good, @dockerymick . My only suggestion might have been "YAML content" instead of "YAML", but not a big deal. Thanks!

dockerymick:
@cadawson Thanks! I like your suggestion

PR #3779, opened 2022-08-01
Instruct the user to set the version when installing the PG

mprahl:
The `{product-version}` variable does not get replaced in the source
YAML, so instruct the user to replace the version.

Relates:
https://github.com/stolostron/backlog/issues/24747


dockerymick (approved):
Looks good

PR #3782, opened 2022-08-01
stolostron/backlog#20569 hub cluster network

swopebe:
--working

### DIFF
--- a/networking/network_config_hub.adoc
+++ b/networking/network_config_hub.adoc
@@ -1,13 +1,13 @@
 [#hub-network-config]
 = Hub cluster network configuration
 
-You can refer to the following configuration for your hub cluster network.
+You can refer to the following configuration for your hub cluster network:
 
 [#hub-network-table]
 == Hub cluster network configuration table
 
 |===
-|Direction | Protocol | Connection | Port (if specified) | Source address | Destination address
+| Direction | Protocol | Connection | Port (if specified) | Source address | Destination address
 
 | Outbound to managed cluster 
 | HTTPS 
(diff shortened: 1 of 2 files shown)


networking/network_config_hub.adoc
```diff
@@ -1,56 +1,63 @@
...
+
 |===
-|Direction | Protocol | Connection | Port (if specified) | Source address | Destination address
+| Direction | Protocol | Connection | Port (if specified) | Source address | Destination address
 
-| Outbound to managed cluster 
+| Outbound to the managed cluster 
 | HTTPS 
-| Retrieval of logs dynamically from Search console for the pods of the managed cluster. This connection uses a service running on the managed cluster called - klusterlet-addon-workmgr 
+| Retrieval of logs dynamically from Search console for the pods of the managed cluster, uses a service that is running on the managed cluster, the `klusterlet-addon-workmgr`
```
cadawson: This cell is a little confusing to me. It looks like you are avoiding having a sentence in there, for parallelism. 

If I am not misunderstanding, maybe something like: 
Retrieval of logs dynamically from Search console for the pods of the managed cluster, which uses the `klusterlet-addon-workmgr` service that is running on the managed cluster

cadawson (changes requested):
One suggestion, but otherwise looks good. Thanks!

networking/network_config_hub.adoc
swopebe: It's two different concepts, but I will check with developers to see how best to present it without a run-on sentence.

swopebe: Sending of logs dynamically from Search console for the pods of the managed cluster, uses the `klusterlet-addon-workmgr` service that is running on the managed cluster

cadawson: Better than the original, but looks like 2 list items joined by a comma. Maybe split them into 2 list items? Don't spend much time on it. If you are good with it, then roll with it. It's a challenge.

swopebe: listing in a cell I am against unless absolutely necessary. It is exactly two items separated by a comma, which we do sometimes. Similar to a csv.

cadawson (approved):
Looks good. Thanks!

PR #3787, opened 2022-08-01
Adding the configurationPolicyAnnotations parameter

dockerymick:
https://github.com/stolostron/backlog/issues/24114

### DIFF
--- a/governance/policy_generator.adoc
+++ b/governance/policy_generator.adoc
@@ -543,113 +543,142 @@ Note that all the fields in the `policyDefaults` section except for `namespace` 
 | Field | Description
 
 | apiVersion
-| Required.
+| Required
+
 Set the value to `policy.open-cluster-management.io/v1`.
 
 | complianceType
-| Optional.
+| Optional
+
 Determines the policy controller behavior when comparing the manifest to objects on the cluster. The parameter values are `musthave`, `mustonlyhave`, or `mustnothave`. The default value is `musthave`.
 
 | kind
-| Required.
+| Required
+
 Set the value to `PolicyGenerator` to indicate the type of policy.
 
 | metadata
-| Required.
+| Required
+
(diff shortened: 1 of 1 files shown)


mprahl (approved):

governance/policy_generator.adoc
```diff
@@ -578,6 +578,10 @@ Any default value listed here is overridden for an entry in the policies array e
 | Optional.
 Array of categories to be used in the `policy.open-cluster-management.io/categories` annotation. The default value is `CM Configuration Management`.
 
+| policyDefaults.configurationPolicyAnnotations
+| Optional.
+Key-value pairs of annotations to set on generated configuration policies. For example, you can disable policy templates by defining the following parameter: `{"policy.open-cluster-management.io/disable-templates": "true"}`. The default value is empty (`{}`).
```
swopebe, line 593: no need for period on line 582

dockerymick: Thanks for reviewing Brandi. I think that is a fair change, I will have to remove the period from the Optional or Required lines. I'll create a new break line between parameter description and requirement status. Thoughts?

swopebe: I handle two incomplete sentences in one cell the same as comma separated values, however, on this you may consider another column for Requirements.

If not, I do a comma in between the two. I like the way your strategy looks and if it works, we can make this change across tables. (I don't like too many bullets in a list if we can avoid it, this looks good!)

swopebe (approved):

governance/policy_generator.adoc
dockerymick: Thanks Brandi! Likewise, I am not a fun of too many bullets. I would like to make the change across the doc. I'll bring this up in the channel or next meeting

dockerymick:
Merging after receiving approval

PR #3791, opened 2022-08-02

swopebe:
none of these tools are used, we replaced this with the pantheon build and the link plugin in Chrome.


oafischer (approved):
Good to know that we don't use these tools anymore, thanks.

PR #3792, opened 2022-08-02
https://github.com/stolostron/backlog/issues/24307 we don't do videos…

swopebe:
… here, removing


oafischer (approved):
No videos, noted.

PR #3794, opened 2022-08-02

swopebe:
We don't need this, any guidance would go in general writing guidance


oafischer (approved):
OK, consolidate writing guidance.

PR #3813, opened 2022-08-04
Clarify APIs and admin role reqs in sub policies in 2.5

oafischer:
https://github.com/stolostron/backlog/issues/22781

### DIFF
--- a/governance/deploy_gitops.adoc
+++ b/governance/deploy_gitops.adoc
@@ -142,9 +142,10 @@ View the following explanations for each argument:
 After you run the `deploy.sh` script, any user with access to the repository can commit changes to the branch, which pushes changes to existing policies on your clusters.
 
 *Note*: To deploy policies with subscriptions, complete the following steps:
-+
+
 . Bind the `open-cluster-management:subscription-admin` ClusterRole to the user that will create the subscription.
 . If you are using an allow list in the subscription, include the following entries:
+
 +
 [source,yaml]
 ----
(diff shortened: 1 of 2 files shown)


cadawson (approved):
Looks good! I fixed a couple of older typos that I saw. Also, for consistency, we try to put the colon inside of the asterisks in *Note:*. Not something to take time changing, but I went ahead and changed them in here. I think you were following what was there, but we want use the *Note:* when you happen to notice it. Thanks!

governance/deploy_gitops.adoc
```diff
@@ -141,6 +141,20 @@ View the following explanations for each argument:
...
 
+*Note:* To deploy policies with subscriptions, complete the following steps:
+
+. Bind the `open-cluster-management:subscription-admin` ClusterRole to the user creating the subscription.
+. If you are using an allow list in the subscription, include the following entries:
+
++
+[source,yaml]
+----
+  allow:
```
gparvin: I think we would want to show an allow list like this:

```
    - apiVersion: policy.open-cluster-management.io/v1
      kinds:
        - "*"
    - apiVersion: policy.open-cluster-management.io/v1beta1
      kinds:
        - "*"
    - apiVersion: apps.open-cluster-management.io/v1
      kinds:
        - PlacementRule
    - apiVersion: cluster.open-cluster-management.io/v1beta1
      kinds:
        - Placement

```

gparvin (commented):
This looks good -- I just had the one comment about the allow list update.  Thanks

dockerymick:
Remember to also make this update for 2.6 as well

PR #3822, opened 2022-08-08
Primary cluster shut down

dockerymick:
https://github.com/stolostron/backlog/issues/24161

### DIFF
--- a/clusters/backup_and_restore.adoc
+++ b/clusters/backup_and_restore.adoc
@@ -306,7 +306,7 @@ Use the link:https://github.com/stolostron/cluster-backup-operator/blob/release-
 [#primary-cluster-shut-down]
 ==== Primary cluster must be shut down
 
-When you restore a backup on a new hub cluster, make sure that the previous hub cluster, where the backup was created, is shut down. If it is running, the previous hub cluster tries to reimport the managed clusters when the managed cluster reconciliation finds that the managed clusters are no longer available.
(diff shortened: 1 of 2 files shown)


birsanv (approved):
lgtm

clusters/backup_and_restore.adoc
```diff
@@ -301,6 +302,12 @@ The cleanup option uses the `cleanupBeforeRestore` property to identify the subs
 
 Use the link:https://github.com/stolostron/cluster-backup-operator/blob/release-2.5/config/samples/cluster_v1beta1_restore_passive_activate.yaml[`restore-passive-activate`] sample when you want the hub cluster to manage the clusters. In this case it is assumed that the other data has been restored already on the hub cluster that using the passive resource.
 
+
+[#primary-cluster-shut-down]
+==== Primary cluster must be shut down
+
+When you restore a backup on a new hub cluster, make sure that the previous hub cluster, where the backup was created, is shut down. If it is running, the previous hub cluster tries to reimport the managed clusters when the managed cluster reconciliation finds that the managed clusters are no longer available.
```
swopebe: I don't think you need commas here. And we try not to use "it" and use what "it" is.

When you restore a backup on a new hub cluster, ensure that the previous hub cluster where the backup was created is shut down. If that cluster is running, the previous hub cluster tries to reimport the managed clusters when the managed cluster reconciliation finds that the managed clusters are not available.

swopebe (approved):

PR #3852, opened 2022-08-11
Adding steps about cleaning resources from policies'

dockerymick:
https://github.com/stolostron/backlog/issues/24368


governance/create_policy.adoc
```diff
@@ -299,3 +300,10 @@ Click the *Delete* button.
 
 To manage other policies, see xref:../governance/create_policy.adoc#managing-security-policies[Managing security policies] for more information.
 Refer to xref:../governance/grc_intro.adoc#governance[Governance] for more topics about policies.
+
+[#cleaning-up-resources-from-policies]
+== Cleaning up resources that are created by policies
+
+NEED PROCEDURE HERE
```
willkutler: @dockerymick do you need some input from Justin or I here?

dockerymick: Yes can you provide the steps please?

willkutler: Sorry for the delayed response - when you create the policy, you will be able to set a new field called `spec.pruneObjectBehavior` in a configurationPolicy that will determine the behavior for the resource cleanup. It defaults to `None`, which keeps the behavior the same as previous releases, but setting it to `DeleteIfCreated` or `DeleteAll` will clean up any resources created by the policy or all resources managed by the policy, respectively. This field can be set directly in the yaml when creating policies by CLI, and there is a new field in the UI where users can set it as well. Once the field is set, all you need to do to initiate resource cleanup is to delete the policy, and any managed resources that need to be cleaned up will be deleted as well.

governance/create_policy.adoc
```diff
@@ -299,3 +301,17 @@ Click the *Delete* button.
...
 Refer to xref:../governance/grc_intro.adoc#governance[Governance] for more topics about policies.
+
+[#cleaning-up-resources-from-policies]
+== Cleaning up resources that are created by policies
+
+Use the `pruneObjectBehavior` parameter to clean up resources that are creted by policies. 
+
+
+//As I wait for the playback recording or feedback from dev
+//Can you enable/disable this feature?
```
JustinKuli: Yes. If the `pruneObjectBehavior` parameter is not set, or is set to "None", then the previous behavior is preserved. If the parameter is set to "DeleteIfCreated" or "DeleteAll" then related resources may be cleaned up when the policy is deleted.

governance/create_policy.adoc
```diff
@@ -299,3 +301,17 @@ Click the *Delete* button.
...
+//How is it displayed from the console?
```
JustinKuli: It appears when creating or editing a policy, for example:

Screen Shot 2022-08-22 at 3 02 11 PM
:

governance/create_policy.adoc
```diff
@@ -299,3 +301,17 @@ Click the *Delete* button.
...
+//are there any notes that need to be added about this? 
+//do all policies support this feature?
```
JustinKuli: Only ConfigurationPolicies. The parameter can be applied and will operate on any ConfigurationPolicy. However, for policies that manage operators, there may still be manual cleanup from an administrator required. For example, a policy might install an operator by specifying the Subscription and OperatorGroup, but even if those objects are pruned automatically by the policy, a ClusterServiceVersion might still be on the cluster and need to be removed (like you mention below).

JustinKuli (commented):
Just to add to Will's response

governance/create_policy.adoc
dockerymick: @willkutler How do you start/initiate the clean up? Does the user have to run a command ? From the console, would the user only update the parameter and then save the policy?

governance/create_policy.adoc
```diff
@@ -299,3 +301,18 @@ Click the *Delete* button.
...
+[#cleaning-up-resources-from-policies]
+== Cleaning up resources that are created by policies
+
+Use the `pruneObjectBehavior` parameter in a configuration policy to clean up resources that are creted by the policy. The following values can be used for the parameter, `DeleteIfCreated`, `DeleteAll`, and `None`. View the following descriptions of the previously mentioned values:
+
+
+* `DeleteIfCreated`: Cleans up any resources created by the policy
+* `DeleteAll`: Cleans up all resources managed by the policy.
+* `None`: This is the default value, which maintains the same behavior from previous releases.
+//can we describe what was done in previous releases? Just a brief statement? Was this parameter used in 2.5 and I missed it? ^
```
dockerymick: @willkutler @JustinKuli comment here

JustinKuli: Previously, no resources other than the policy itself were deleted.

```
* `None`: This is the default value, which maintains the same behavior from previous releases, where no related resources are deleted.
```

governance/create_policy.adoc
```diff
@@ -170,6 +171,7 @@ Your YAML file might resemble the following policy:
    namespaces:
      exclude: ["kube*"]
      include: ["default"]
+     pruneObjectBehavior: DeleteAll
```
JustinKuli: Unless this is an example of using the new feature, I would prefer this to be set to the default of `None`.

Outside of the content of this PR, I wonder why this policy has the categories/controls/standards that it does? They are not the standard ones found in any of our pre-built policies. A similar policy for a pod created from the wizard would set:
- Standards: `NIST SP 800-53`
- Categories: `CM Configuration Management`
- Controls: `CM-2 Baseline Configuration`

Is there a reason for this policy to use the current settings? I just worry it could be confusing.

governance/create_policy.adoc
```diff
@@ -299,3 +301,17 @@ Click the *Delete* button.
 
 To manage other policies, see xref:../governance/create_policy.adoc#managing-security-policies[Managing security policies] for more information.
 Refer to xref:../governance/grc_intro.adoc#governance[Governance] for more topics about policies.
+
+[#cleaning-up-resources-from-policies]
+== Cleaning up resources that are created by policies
+
+Use the `pruneObjectBehavior` parameter in a configuration policy to clean up resources that are creted by the policy. The following values can be used for the parameter, `DeleteIfCreated`, `DeleteAll`, and `None`. View the following descriptions of the previously mentioned values:
```
oafischer: Missing "a" in created

oafischer, line 308: I would add a colon after "The following values can be used for the parameter". I don't think you need to create a bulleted list after that. You can keep it as is. There is at least one other spot in the doc where it is done that way too.

governance/create_policy.adoc
JustinKuli: Also, we've updated the wizard to no longer set the `clusterConditions` in the PlacementRule like what is in this example. The new PlacementRule YAML would look like:

```yaml
 kind: PlacementRule
 metadata:
   name: placement-pod
spec:
  clusterSelector:
    matchExpressions:
      - key: cloud
        operator: In
        values:
          - IBM
  clusterConditions: []
```

(Note that I believe the current PlacementRule has invalid syntax - `clusterLabels` is not an accepted field and would be ignored)

governance/create_policy.adoc
```diff
@@ -299,3 +301,17 @@ Click the *Delete* button.
...
+
+
+* `DeleteIfCreated`: Cleans up any resources created by the policy
```
JustinKuli: Nit: add a period at the end of the line.

governance/create_policy.adoc
oafischer: Do we need the 3 values listed twice each? Not sure if the bulleted list with descriptions is enough as is. I'll leave it up to you.

governance/create_policy.adoc
```diff
@@ -299,3 +301,17 @@ Click the *Delete* button.
...
+* `DeleteAll`: Cleans up all resources managed by the policy.
+* `None`: This is the default value and maintains the same behavior from previous releases, where no related resources are deleted.
+
+You can set the value directly in the YAML as you create a policy from the CLI. From the console, you can select the value in the _Prune Object Behavior_ section of the the _Policy templates_ step.
+
+*Note:* If a policy that installs an operator with the `pruneObjectBehavior` field deleted, additional clean up is needed to complete the operator uninstall. Additional clean up might include deleting the operator `ClusterServiceVersion` object.
```
oafischer: If a policy that installs an operator with the `pruneObjectBehavior` field is deleted (?)

JustinKuli (commented):
Content looks right to me, just a few things I noticed (most of which is not really in the new stuff), thanks!

oafischer (approved):
See comments. Going to approve since it's just typos. Left one suggestion but I don't feel strongly on that and will leave it up to you.

governance/create_policy.adoc
willkutler: I would like to specify here that when pruneObjectBehavior is set, the related objects will only be cleaned up when the configurationPolicy (or parent policy) associated with them is deleted

dockerymick: You are right. I will update the statement to read:

Use the `pruneObjectBehavior` parameter in a configuration policy to clean up resources that are created by the policy. View the following descriptions of the values that can be used for the parameter:

governance/create_policy.adoc
dockerymick: Thanks, here's my suggestion

If a policy that installs an operator with the `pruneObjectBehavior` parameter undefined, additional clean up is needed to complete the operator uninstall.

governance/create_policy.adoc
dockerymick: @JustinKuli Should `clusterLabels` be mentioned in our [Deprecations and removals topic](https://github.com/stolostron/rhacm-docs/blob/2.6_stage/release_notes/deprecate_remove.adoc)? What is the name of the parameter that handles cluster labels? Should users only use `machLabels`?

governance/create_policy.adoc
dockerymick: Great catch! Definitely helps with consistency

governance/create_policy.adoc
JustinKuli: Even if the `pruneObjectBehavior` field is defined, additional clean up might be required to complete the operator uninstall.

governance/create_policy.adoc
JustinKuli: This is awkward... it looks like `clusterLabels` were removed from PlacementRule in 2019: https://github.com/stolostron/multicloud-operators-placementrule/commit/c9a4d98e0b6afcd73bd5d083a094a87c41e56725#diff-20a1d0b52c6870815c0163de228c933b750e2dcb9b533610390dfe3889f1cbc8

It was (basically) replaced by `clusterSelector`.

My suggested PlacementRule above is what the policy wizard would generate, but this is also valid and equivalent, and more similar to what was there previously:
```yaml
 kind: PlacementRule
 metadata:
   name: placement-pod
spec:
  clusterConditions: []
  clusterSelector:
     matchLabels:
       cloud: "IBM"
```

So really the only changes needed were removing the ClusterConditions, and renaming `clusterLabels` to `clusterSelector`

governance/create_policy.adoc
```diff
@@ -86,7 +87,7 @@ spec:
 ----
 
 . Define a `PlacementRule`.
-Be sure to change the `PlacementRule` to specify the clusters where the policies need to be applied, either by `clusterNames`, or `clusterLabels`. View link:../applications/placement_sample.adoc#placement-rule-samples[Placement rule samples overview]
+Be sure to change the `PlacementRule` to specify the clusters where the policies need to be applied by `clusterNames`. View link:../applications/placement_sample.adoc#placement-rule-samples[Placement rule samples overview]
```
JustinKuli:
```suggestion
Be sure to change the `PlacementRule` to specify the clusters where the policies need to be applied by adjusting the `clusterSelector`. View link:../applications/placement_sample.adoc#placement-rule-samples[Placement rule samples overview]
```

I think specifying `clusterNames` would work, but I believe that we try to encourage using the selector instead (since that will scale much to larger fleets).

governance/create_policy.adoc
```diff
@@ -103,7 +104,6 @@ spec:
   clusterNames:
   - "cluster1"
   - "cluster2"
-  clusterLabels:
```
JustinKuli, line 106: Don't delete this, I think just rename it to `clusterSelector`.

governance/create_policy.adoc
```diff
@@ -299,3 +301,16 @@ Click the *Delete* button.
...
+
+Use the `pruneObjectBehavior` parameter in a configuration policy to clean up resources that are created by the policy. When `pruneObjectBehavior` is set, the related objects are only cleaned up after the configuration policy (or parent policy) associated with them is deleted. View the following descriptions of the values that can be used for the parameter:
+
+* `DeleteIfCreated`: Cleans up any resources created by the policy.
+* `DeleteAll`: Cleans up all resources managed by the policy.
+* `None`: This is the default value and maintains the same behavior from previous releases, where no related resources are deleted.
+
+You can set the value directly in the YAML as you create a policy from the CLI. From the console, you can select the value in the _Prune Object Behavior_ section of the the _Policy templates_ step.
+
+*Note:* If a policy that installs an operator with the `pruneObjectBehavior` parameter undefined, additional clean up is needed to complete the operator uninstall. Additional clean up might include deleting the operator `ClusterServiceVersion` object.
```
JustinKuli:
```suggestion
*Note:* If a policy that installs an operator uses the `pruneObjectBehavior` parameter, then additional clean up is needed to complete the operator uninstall. Additional clean up might include deleting the operator `ClusterServiceVersion` object.
```

Just to remind: if the parameter is undefined, then no clean up will be done automatically, so of course additional clean up is needed. I think the point of this note is to mention that even when it *is* defined, the automatic cleanup is insufficient.

JustinKuli (changes requested):
I'm getting side-tracked with the `clusterSelector` stuff, and I don't mean it to hold up this PR. We should look through it separately - I'd be happy to do the legwork and check through the other governance sections for it.

dockerymick:
Thanks Justin! I appreciate you, you aren't holding it up! I appreciate your attention to detail and rather we doc what is best for the reader/user

JustinKuli (approved):

dockerymick:
Thanks for yall reviews and help! @JustinKuli @oafischer @willkutler 

Merging this pr

PR #3871, opened 2022-08-15
Specify proxy config as part of outbound config in 2.3

oafischer:
https://github.com/stolostron/backlog/issues/23059

### DIFF
--- a/observability/customize_observability.adoc
+++ b/observability/customize_observability.adoc
@@ -294,6 +294,15 @@ receivers:
     text: 'https://internal.myorg.net/wiki/alerts/{{ .GroupLabels.app }}/{{ .GroupLabels.alertname }}'
 ----
 
+. If your {product-title-short} hub cluster has a cluster-wide proxy configuration, add the following `global` entry to the `alertmanager-config` YAML file: 
++
+----
+global:
+  slack_api_url: '<slack_webhook_url>'
+  http_config:
+    proxy_url: http://****
+----
+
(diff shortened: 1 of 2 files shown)


observability/customize_observability.adoc
```diff
@@ -294,6 +294,15 @@ receivers:
     text: 'https://internal.myorg.net/wiki/alerts/{{ .GroupLabels.app }}/{{ .GroupLabels.alertname }}'
 ----
 
+. If your {product-title-short} hub cluster has a cluster-wide proxy configuration, add the following `global` entry to the `alertmanager-config` YAML file:
```
marcolan018: the proxy setting for alert forwarding has no relationship with the cluster-wide proxy configuration.

oafischer: @marcolan018 What exactly does that mean?

marcolan018: "If your {product-title-short} hub cluster has a cluster-wide proxy configuration", this is not required here because the proxy config of alert forwarding is totally different than the cluster-wide proxy configuration, and those two configurations have no relationship.

Need to update the content as something like the below:

If you want to configure a proxy for alert forwarding, you can add the following `global` entry to the `alertmanager-config` YAML file:

oafischer: Thanks @marcolan018 for explaining that. I misunderstood the original issue. I will add your suggestion, much appreciated!

marcolan018 (approved):
lgtm

dockerymick (approved):
Looks good to me! Thanks Oliver

PR #3872, opened 2022-08-15
Add bind/view permission in 2.6

oafischer:
https://github.com/stolostron/backlog/issues/23820

### DIFF
--- a/multicluster_engine/cluster_lifecycle/managedclustersets_assign_role.adoc
+++ b/multicluster_engine/cluster_lifecycle/managedclustersets_assign_role.adoc
@@ -11,6 +11,9 @@ The `ManagedClusterSet` API offers three levels of RBAC permissions:
 ** Full access permissions to all of the cluster and cluster pool resources that are assigned to the managed cluster set.
 ** Permission to create clusters, import clusters, and create cluster pools. The permissions must be assigned to the managed cluster set when the managed cluster set is created.
 * Cluster set `bind`
+** Permission to bind the cluster set to a namespace by creating a `ManagedClusterSetBinding`.
++
+**Note:** The user or group must also have permission to create the `ManagedClusterSetBinding` in the target namespace.
 ** Read only permissions to all of the cluster and cluster pool resources that are assigned to the managed cluster set.
 ** No permission to create clusters, import clusters, or create cluster pools.
 * Cluster set `view`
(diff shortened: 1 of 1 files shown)


multicluster_engine/cluster_lifecycle/managedclustersets_assign_role.adoc
```diff
@@ -5,15 +5,20 @@ You can assign users or groups to your cluster set that are provided by the conf
 
 *Required access:* Cluster administrator
 
-The `ManagedClusterSet` API offers two levels of RBAC permissions:
+The `ManagedClusterSet` API offers three levels of RBAC permissions:
 
 * Cluster set `admin`
 ** Full access permissions to all of the cluster and cluster pool resources that are assigned to the managed cluster set.
 ** Permission to create clusters, import clusters, and create cluster pools. The permissions must be assigned to the managed cluster set when the managed cluster set is created.
+* Cluster set `bind`
```
KevinFCormier, line 13: You missed the line that @ldpliu added here: `Bind the cluster set to a namespace.`

This is important because it is the main point of this role. I think we should be a bit more explicit... something like: `Permission to bind the cluster set to a namespace by creating a ManagedClusterSetBinding. (The user or group must also have permission to create the ManagedClusterSetBinding in the target namespace.)`

KevinFCormier (changes requested):
Please see inline comment.

multicluster_engine/cluster_lifecycle/managedclustersets_assign_role.adoc
oafischer: Thanks for explaining that in more detail @KevinFCormier. Would you say we should keep the other two bullet points about permissions?

KevinFCormier: Yes, but I will leave to @ldpliu to confirm that the role also includes those 2 other permissions.

KevinFCormier (approved):
Approving content-wise. I'm not really sure what the formatting will look like.

oafischer:
Thanks for your help on this and my other PRs @KevinFCormier! Much appreciated. Wanted to get the content correct first before opening this up for peer review for formatting etc.

multicluster_engine/cluster_lifecycle/managedclustersets_assign_role.adoc
ldpliu: Yes, The current change is ok for me

PR #3873, opened 2022-08-15
Add automatic global set creation in 2.6

oafischer:
Link to the original issue: https://github.com/stolostron/backlog/issues/23495

### DIFF
--- a/multicluster_engine/cluster_lifecycle/managedclustersets_add_cluster.adoc
+++ b/multicluster_engine/cluster_lifecycle/managedclustersets_add_cluster.adoc
@@ -4,6 +4,8 @@
 
 After you create your `ManagedClusterSet`, you must add one or more managed clusters. You can add managed clusters to your managed cluster set by using either the console or the command line.
 
+**Note:** <<#adding-clusters-to-a-managedclusterset,Adding a cluster to a ManagedClusterSet>> does not apply to the global cluster set. The `global` set always includes all managed clusters. See xref:../multicluster_engine/cluster_lifecycle/managedclustersets_intro.adoc#managedclustersets_global[Global ManagedClusterSet] for more details.
+
 [#adding-clusters-to-a-managedclusterset-console]
 == Adding clusters to a ManagedClusterSet by using the console
 
--- a/multicluster_engine/cluster_lifecycle/managedclustersets_add_cluster.adoc
+++ b/multicluster_engine/cluster_lifecycle/managedclustersets_add_cluster.adoc
@@ -4,8 +4,6 @@
 
 After you create your `ManagedClusterSet`, you must add one or more managed clusters. You can add managed clusters to your managed cluster set by using either the console or the command line.
 
-**Note:** <<#adding-clusters-to-a-managedclusterset,Adding a cluster to a ManagedClusterSet>> does not apply to the global cluster set. The `global` set always includes all managed clusters. See xref:../cluster_lifecycle/managedclustersets_intro.adoc#managedclustersets_global[Global ManagedClusterSet] for more details.
-
 [#adding-clusters-to-a-managedclusterset-console]
 == Adding clusters to a ManagedClusterSet by using the console
 
--- a/multicluster_engine/cluster_lifecycle/managedclustersets_add_cluster.adoc
+++ b/multicluster_engine/cluster_lifecycle/managedclustersets_add_cluster.adoc
@@ -4,7 +4,7 @@
 
 After you create your `ManagedClusterSet`, you must add one or more managed clusters. You can add managed clusters to your managed cluster set by using either the console or the command line.
 
-**Note:** <<#adding-clusters-to-a-managedclusterset,Adding a cluster to a ManagedClusterSet>> does not apply to the global cluster set. The `global` set always includes all managed clusters. See xref:../multicluster_engine/cluster_lifecycle/managedclustersets_intro.adoc#managedclustersets_global[Global ManagedClusterSet] for more details.
+**Note:** <<#adding-clusters-to-a-managedclusterset,Adding a cluster to a ManagedClusterSet>> does not apply to the global cluster set. The `global` set always includes all managed clusters. See xref:../cluster_lifecycle/managedclustersets_intro.adoc#managedclustersets_global[Global ManagedClusterSet] for more details.
 
 [#adding-clusters-to-a-managedclusterset-console]
 == Adding clusters to a ManagedClusterSet by using the console
--- a/multicluster_engine/cluster_lifecycle/managedclustersets_intro.adoc
+++ b/multicluster_engine/cluster_lifecycle/managedclustersets_intro.adoc
@@ -16,7 +16,7 @@ When you create a managed cluster, the following are automatically created to ea
 * The namespace called `open-cluster-management-global-set`.
 * A `ManagedClusterSetBinding` called `global` to bind the `global` `ManagedClusterSet` to the `open-cluster-management-global-set` namespace.
 
-You cannot delete or update the `global` managed cluster set. The `global` managed cluster set includes all managed clusters. See the following example:
+**Important:** You cannot delete, update or edit the `global` managed cluster set. The `global` managed cluster set includes all managed clusters. See the following example:
 
 [source,yaml]
 ----
--- a/multicluster_engine/cluster_lifecycle/managedclustersets_intro.adoc
+++ b/multicluster_engine/cluster_lifecycle/managedclustersets_intro.adoc
@@ -12,7 +12,7 @@ Each managed cluster must be a member of a `ManagedClusterSet`. When you install
 
 When you create a managed cluster, the following are automatically created to ease management.
 
-* A `ManagedClusterSet` called `global.
+* A `ManagedClusterSet` called `global`.
 * The namespace called `open-cluster-management-global-set`.
 * A `ManagedClusterSetBinding` called `global` to bind the `global` `ManagedClusterSet` to the `open-cluster-management-global-set` namespace.
 
--- a/multicluster_engine/cluster_lifecycle/managedclustersets_intro.adoc
+++ b/multicluster_engine/cluster_lifecycle/managedclustersets_intro.adoc
@@ -3,7 +3,7 @@
 
 A `ManagedClusterSet` is a group of managed clusters. With a managed cluster set, you can manage access to all of the managed clusters in the group together. You can also create a `ManagedClusterSetBinding` resource to bind a `ManagedClusterSet` resource to a namespace.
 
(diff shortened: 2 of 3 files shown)


oafischer:
@ldpliu I have added the global managed cluster set section to the file you originally wanted it in, but due to the MCE move, I wonder if it would be better to have this section here instead: https://github.com/stolostron/rhacm-docs/blob/2.6_stage/multicluster_engine/cluster_lifecycle/managedclustersets_create.adoc

Let me know what you think. Feel free to weigh in on this as well @KevinFCormier

multicluster_engine/cluster_lifecycle/managedclustersets_add_cluster.adoc
```diff
@@ -4,6 +4,8 @@
 
 After you create your `ManagedClusterSet`, you must add one or more managed clusters. You can add managed clusters to your managed cluster set by using either the console or the command line.
 
+**Note:** <<#adding-clusters-to-a-managedclusterset,Adding a cluster to a ManagedClusterSet>> does not apply to the global cluster set. The `global` set always includes all managed clusters. See xref:../cluster_lifecycle/managedclustersets_intro.adoc#managedclustersets_global[Global ManagedClusterSet] for more details.
```
dockerymick: Should these be two separate notes? If yes, I recommend reorganizing this statement to read as the following content:

**Notes**: 
+
* When you add a cluster to a managed cluster set, the action does not apply to the global cluster set. See Adding a cluster to a ManagedClusterSet for more details. 

* The `global` set always includes all managed clusters. See Global ManagedClusterSet.

multicluster_engine/cluster_lifecycle/managedclustersets_intro.adoc
```diff
@@ -5,7 +5,29 @@ A `ManagedClusterSet` is a group of managed clusters. With a managed cluster set
 
 Each managed cluster must be a member of a `ManagedClusterSet`. When you install the hub cluster, a default `ManagedClusterSet` is created called `default`. All managed clusters that are not specifically assigned to a managed cluster set are automatically assigned to the `default` managed cluster set. To ensure that the default managed cluster set is always available, you cannot delete or update the `default` managed cluster set.
```
dockerymick: Each managed cluster must be a member of a managed cluster set. When you install the hub cluster, a default `ManagedClusterSet` resource is created called `default`.

multicluster_engine/cluster_lifecycle/managedclustersets_intro.adoc
```diff
@@ -5,7 +5,29 @@ A `ManagedClusterSet` is a group of managed clusters. With a managed cluster set
...
 
-**Note:** Cluster pools that are not specifically added to a `ManagedClusterSet` are not added to the default `ManagedClusterSet`. After a managed cluster is claimed from a cluster pool, it is added to the default `ManagedClusterSet` if it is not specifically added to another `ManagedClusterSet`. 
+**Note:** Cluster pools that are not specifically added to a `ManagedClusterSet` are not added to the default `ManagedClusterSet`. After a managed cluster is claimed from a cluster pool, it is added to the default `ManagedClusterSet` if it is not specifically added to another `ManagedClusterSet`.
```
dockerymick: Cluster pools that are not specifically added to a managed cluster set are not added to the default `ManagedClusterSet` resource. After a managed cluster is claimed from a cluster pool, it is added to the default `ManagedClusterSet` resource if it is not specifically added to another managed cluster set.

multicluster_engine/cluster_lifecycle/managedclustersets_intro.adoc
```diff
@@ -5,7 +5,29 @@ A `ManagedClusterSet` is a group of managed clusters. With a managed cluster set
...
+
+[#managedclustersets_global]
+== Global ManagedClusterSet
+
+When you create a managed cluster, the following are automatically created to ease management.
```
dockerymick: Add a colon ( : ) instead of a period at the end of this statement

multicluster_engine/cluster_lifecycle/managedclustersets_remove_cluster.adoc
```diff
@@ -5,6 +5,8 @@ You might want to remove a managed cluster from a managed cluster set to move it
 
 **Note:** Every managed cluster must be assigned to a managed cluster set. If you remove a managed cluster from a `ManagedClusterSet` and do not assign it to a different `ManagedClusterSet`, it is automatically added to the `default` managed cluster set.
 
+**Note:** <<#removing-a-managed-cluster-from-a-managedclusterset,Removing a managed cluster from a ManagedClusterSet>> does not apply to the global cluster set. The `global` set always includes all managed clusters. See xref:../cluster_lifecycle/managedclustersets_intro.adoc#managedclustersets_global[Global ManagedClusterSet] for more details.
```
dockerymick: Similar to my previous comment, rearrange this statement. 

**Notes**: 
+
* When you remove a managed cluster from the managed cluster set, the action does not apply to the global cluster set. See Removing a managed cluster from a ManagedClusterSet for more details. 
* The `global` set always includes all managed clusters. See Global ManagedClusterSet for more details.

dockerymick (changes requested):

multicluster_engine/cluster_lifecycle/managedclustersets_add_cluster.adoc
KevinFCormier: I think the way this was stated initially makes more sense. This is not 2 separate notes. You cannot add a cluster to the global cluster set because it automatically includes all managed clusters.

multicluster_engine/cluster_lifecycle/managedclustersets_remove_cluster.adoc
KevinFCormier: I think the way this was stated initially makes more sense. This is not 2 separate notes. You cannot remove a cluster from the global cluster set because it automatically includes all managed clusters.

KevinFCormier (approved):
Thanks. Your changes addressed my concerns.

multicluster_engine/cluster_lifecycle/managedclustersets_remove_cluster.adoc
dockerymick: We typically do not lead statements with links. For accessibility reasons ([reading order/meaningful sequence](https://www.ibm.com/able/requirements/requirements)), we are making the user aware of important info and then pointing them to a reference. Additionally, we don't want to take the user/reader away from the doc too much. Also see [2.4.4 Link purpose](https://www.ibm.com/able/requirements/requirements). I recommend the following: 

When you remove a managed cluster from the managed cluster set, the action does not apply to the global cluster set. See Removing a managed cluster from a ManagedClusterSet for more details.
The global set always includes all managed clusters. See Global ManagedClusterSet for more details.

multicluster_engine/cluster_lifecycle/managedclustersets_add_cluster.adoc
dockerymick: When you add a cluster to a managed cluster set, the action does not apply to the global cluster set. See Adding a cluster to a ManagedClusterSet for more details. The global set always includes all managed clusters. See Global ManagedClusterSet.

KevinFCormier: @dockerymick The first sentence doesn't make sense to me. This section of the doc ("Adding a cluster to a ManagedClusterSet") does not apply to the global cluster set. When you are adding a cluster to a managed cluster set, of course the action does not apply to the global set because you cannot add a cluster to the global cluster set. All managed clusters are automatically members of the global set.

The second sentence is not needed because that is the section this note appears in.

oafischer: @KevinFCormier Can you explain the original statements here in more detail? https://github.com/stolostron/backlog/issues/23495#issuecomment-1191510476

Might be more helpful since @dockerymick is trying to fix what I have in this PR, and what I have in this PR is based on that comment I linked. May be useful to go back to the source and make sure we understand that first.

KevinFCormier: Within this section there are 2 subsections describing how to add a cluster to a managed cluster set using the console or command line. In the instructions for the console, notice that it mentions the **Manage resource assignments** action. You can do this on cluster sets you created your self, like `example` in the following screenshots, or even `default`, but it is not an option for the `global` set.

image
image
image

Via command-line, it appears there is no error if you add the label `cluster.open-cluster-management.io/clusterset: global` however, this has no effect on the cluster aside from removing it from a different cluster set if the label was already there with a different value. FYI @ldpliu

oafischer: @KevinFCormier Thank you for explaining in detail, I think I understand what you meant with your original comment now.

Not sure what the best way to implement this is. Do we want a separate note for both GUI and command line procedures? A note for GUI, as far as I am concerned, would look something like this:

The **Manage resource assignments** option is not available for the `global` managed cluster set. See [Global ManagedClusterSet]() for more details.

A global note at the beginning of the topic might look something like this:

You cannot modify the `global` managed cluster set.

Not sure what the best way to express the limitation is. I want it to make sense for someone who doesn't have your detailed explanation above. Do you have some input @KevinFCormier?

My personal opinion is that we don't need any additional notes at all because we already explain that the `global` managed cluster set cannot be deleted or updated. This is stated in the intro topic. What I would do though is make that statement into an important note and expand it to explain that editing is also not possible.

dockerymick: @KevinFCormier Thanks for explaining, I appreciate you. @oafischer I really like your suggestion to use the **Important** notice for the explanation

oafischer: Given that this is the cleanest solution from a doc and user perspective, I'm going to move ahead with it. Thanks everyone for helping with this PR.

multicluster_engine/cluster_lifecycle/managedclustersets_intro.adoc
```diff
@@ -3,9 +3,31 @@
...
+[source,yaml]
+----
+apiVersion: cluster.open-cluster-management.io/v1beta1
+kind: ManagedClusterSetBinding
+metadata:
+  name: global
+  namespace: open-cluster-management-global-set
+spec:
+  clusterSet: global
+----
```
dockerymick, line 30: To align the notice after the last bullet point, add a plus sign on line 18, and another on line 20.

multicluster_engine/cluster_lifecycle/managedclustersets_intro.adoc
```diff
@@ -3,9 +3,31 @@
...
```
dockerymick, line 31: I recommend prompting the user for your upcoming list of references. I would add a statement that is similar to the following:

Continue reading to learn more on updating manage cluster sets:

dockerymick (approved):
Last couple of suggestions

PR #3876, opened 2022-08-16
Specify proxy config as part of outbound config in 2.4

oafischer:
https://github.com/stolostron/backlog/issues/23059


swopebe (approved):
perfection

PR #3887, opened 2022-08-17
Product-wide update sizing and scaling in 2.6

oafischer:
https://github.com/stolostron/backlog/issues/24810

### DIFF
--- a/install/cluster_size.adoc
+++ b/install/cluster_size.adoc
@@ -10,14 +10,14 @@ Each {product-title} cluster is unique and the following guidelines provide samp
 * Memory reservations and limits, which establish memory capacity on a worker node to assign to a container. 
 * Persistent data, which is managed by the product and stored in the etcd cluster that is used by Kubernetes. 
 
-*Best practice:* For {ocp-short}, distribute the main nodes of the cluster across three (3) availability zones.
+*Best practice:* For {ocp-short}, distribute the master nodes of the cluster across three (3) availability zones.
 
 [#product-environment]
 == Product environment
 
 **Note:** The following requirements are _not_ minimum requirements.
 
-.Main node
+.Master node
 |===
 | Availability zones | etcd | Total reserved memory | Total reserved CPU
 
--- a/install/cluster_size.adoc
+++ b/install/cluster_size.adoc
@@ -91,6 +91,7 @@ See the following https://docs.openshift.com/container-platform/4.10/installing/
 | 4
 | 16 GB
 | 120 GB
+|===
  
 [#openshift-cluster-on-vmware-vsphere]
 === OpenShift cluster on VMware vSphere
--- a/install/cluster_size.adoc
+++ b/install/cluster_size.adoc
@@ -133,7 +133,7 @@ A {product-title} hub cluster can be installed and supported on {ocp-short} bare
 
 See example requirements for creating and managing 2200 single node OpenShift Container Platform clusters. See the minimum requirements for using {product-title-short} to create single node OpenShift (SNO) clusters (230 and more provisioned at the same time), and manage those SNO clusters with a hub cluster:
 
-.Main (schedulable)
+.Master (schedulable)
 |===
 | Node count | Memory (peak cluster usage) | Memory (single node max) | CPU cluster max | CPU single node max
 
--- a/install/cluster_size.adoc
+++ b/install/cluster_size.adoc
@@ -17,19 +17,17 @@ Each {product-title} cluster is unique and the following guidelines provide samp
 
 **Note:** The following requirements are _not_ minimum requirements.
 
-.Master node
+.Product environment
 |===
-| Availability zones | etcd | Total reserved memory | Total reserved CPU
+| Node type | Availability zones | etcd | Total reserved memory | Total reserved CPU
 
+| Master
 | 3
 | 3
 | Per {ocp-short} sizing guidelines
 | Per {ocp-short} sizing guidelines
-|===
 
-.Worker node
-|===
-| Availability zones | redisgraph/redis | Total reserved memory | Total reserved CPU
+| Worker
 | 3
 | 1
 | 12 GB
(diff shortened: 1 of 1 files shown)


oafischer:
Still need to fix the formatting, maybe you can help @dockerymick.

EDIT: I think it's OK now.

install/cluster_size.adoc
```diff
@@ -10,24 +10,31 @@ Each {product-title} cluster is unique and the following guidelines provide samp
 * Memory reservations and limits, which establish memory capacity on a worker node to assign to a container. 
 * Persistent data, which is managed by the product and stored in the etcd cluster that is used by Kubernetes. 
 
-*Best practice:* For {ocp-short}, distribute the master nodes of the cluster across three (3) availability zones.
+*Best practice:* For {ocp-short}, distribute the main nodes of the cluster across three (3) availability zones.
```
dockerymick: This should be _master_ nodes. I haven't seen anything about that being renamed. I do recall the conscious language effort. @randymgeorge can you confirm if this should be _master nodes_ or _main nodes_

randymgeorge: master

dockerymick:
@oafischer thanks for starting this. I recommend that we make one large table. I recommend that we add a column before the _Node_ column that is named _Cloud service_. Then add a new column after the _Storage size_ column that is named _Resources_.  

Also there should be one table for the _Product environment_ section.  Here is my recommendation:


Product environment

|====
| Node type | Availability zones | etcd | Total reserved memory | Total reserved CPU 

| Master 
| 3
| 3
| Per {ocp-short} sizing guidelines
| Per {ocp-short} sizing guidelines

| Worker
| 3
| 1
| 12 GB
| 6
|====


|====
| Cloud service | Node count |  Availability zones | Instance size | vCPUs | Memory | Storage size | Resources

| {ocp-short} on Amazon Web Services
| 3
| 3
| m5.xlarge
| 4
| 16 GB 
| 120 GB
| See the [Amazon Web Services information in the {ocp-short} product documentation](https://docs.openshift.com/container-platform/4.10/installing/installing_aws/installing-aws-customizations.html#installing-aws-customizations) for more information.

Also learn more about [machine types](https://aws.amazon.com/ec2/instance-types/m5/).

.
.
.
.
.
|====

Should _Cores per socket_ be an added column? I see that OpenShift cluster on VMware vSphere is the only cloud service that has a value for that column. The table in _Creating and managing single node {ocp-short} clusters_ sectionn should remain the same bc the column names are so different from the previous sections

oafischer:
I removed the cores per socket column and added the info to the regular vCPU column. Let me know if that works. Some of the rows are also really tall due to the information in Resources, not sure if we want to keep it that way.

install/cluster_size.adoc
```diff
@@ -17,135 +17,132 @@ Each {product-title} cluster is unique and the following guidelines provide samp
...
+| Per {ocp-short} sizing guidelines
+
+| Worker
+| 3
+| 1
+| 12 GB
+| 6
+|===
 
 In addition to {product-title-short}, the {ocp-short} cluster runs additional services to support cluster features. The following node sizes (3 nodes of the types noted in the information that follows, distributed evenly across 3 availability zones) are recommended.
```
dockerymick: Change: 

In addition to {product-title-short}, the {ocp-short} cluster runs additional services to support cluster features. The following node sizes (3 nodes of the types noted in the information that follows, distributed evenly across 3 availability zones) are recommended.
Write Preview


to 

In addition to {product-title-short}, the {ocp-short} cluster runs additional services to support cluster features. 

**Best practice:** The recommended node size is 3 nodes.


Thoughts??

oafischer: I like how much shorter your version is. I think it's OK to use the **Best practice** tag, but I would argue that since we are supposed to avoid tags when possible anyway, we can probably just put that sentence into the main paragraph. With your shorter version, the paragraph is not as cluttered anymore so it should be easier to see that information, even without a special tag.

dockerymick: Yes, you are right. Sounds good! I really like your idea to place the Core per socket in parentheses!

install/cluster_size.adoc
```diff
@@ -17,135 +17,134 @@ Each {product-title} cluster is unique and the following guidelines provide samp
...
+| Worker
+| 3
+| 1
+| 12 GB
+| 6
+|===
+
+In addition to {product-title-short}, the {ocp-short} cluster runs additional services to support cluster features. The recommended node size is three nodes.
+
+//is the second statement in line 37 neccessary since we have the Best practice notice o line 13?
```
dockerymick, line 39: @oafischer FYI
@randymgeorge when you have an opportunity, will you select to _View file_ and let us know if any information needs to be updated? From my perspective, the table looks great!

dockerymick (approved):

install/cluster_size.adoc
oafischer: The statement in line 37 is referring to the additional services, while 13 is referring to OCP. It's true that they both happen to use 3 as a recommendation. I would leave it as is.

dockerymick:
After Randy's review or review from development, I am comfortable with merging this

randymgeorge:
Not sure why the bare metal at the end of the table is marked "deprecated".   Also, having 3 avail zones does not make sense for this row.

oafischer:
@randymgeorge Bare metal assets is deprecated in 2.6: https://github.com/stolostron/rhacm-docs/blob/2.6_stage/release_notes/deprecate_remove.adoc#api-deprecations

How many avail zones would you say make sense for this row?

randymgeorge:
@oafischer Correct.  That is the Hive IPI API which is deprecated.  It is superceded by the Assisted Installer API.  Thus, the function to deploy on bare metal is still available, the API differs.   With that said, this section of the documentation is about the Hub Cluster and its requirements to run in different environments, e.g. bare metal.   The deprecated API is for ACM to deploy managed clusters onto bare metal.  This section of the doc is not about deploying managed clusters.

oafischer:
@randymgeorge Thanks for clarifying, I'll go ahead and remove the label. Let me know about the avail zones as well.

oafischer:
@randymgeorge Once we get the info on the recommended number of avail zones, we can go ahead with the merge.

randymgeorge:
avail zones can be removed as well

oafischer:
Done, going ahead with merge, thanks for you help on this @randymgeorge and @dockerymick!

PR #3901, opened 2022-08-19
Add HyperShift known issue in 2.6

oafischer:
https://github.com/stolostron/backlog/issues/25156

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -526,6 +526,12 @@ To work around this issue, complete the following steps:
 . Delete the `cluster-proxy-signer` secret from the `open-cluster-management` namespace.
 . Enable `cluster-proxy-addon`.
 
+[#hypershift-proxy-install-not-supported-ocp-410z]
+=== {ocp-short } 4.10.z does not support HyperShift clusters with proxy configuration
+// 2.6:25156
+
+ When you create a HyperShift cluster with a cluster-wide proxy configuration on {ocp-short] 4.10.z, the `nodeip-configuration.service` service does not start on worker nodes.
+
 [#application-management-known-issues]
 == Application management known issues
 
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -527,10 +527,10 @@ To work around this issue, complete the following steps:
 . Enable `cluster-proxy-addon`.
 
 [#hypershift-proxy-install-not-supported-ocp-410z]
-=== {ocp-short} 4.10.z does not support HyperShift clusters with proxy configuration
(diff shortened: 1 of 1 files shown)


release_notes/known_issues.adoc
```diff
@@ -526,6 +526,12 @@ To work around this issue, complete the following steps:
 . Delete the `cluster-proxy-signer` secret from the `open-cluster-management` namespace.
 . Enable `cluster-proxy-addon`.
 
+[#hypershift-proxy-install-not-supported-ocp-410z]
+=== {ocp-short} 4.10.z does not support HyperShift clusters with proxy configuration
```
cadawson: Make sure that we mean "HyperShift" specifically here. We are using "hosted control planes" for the overall concept. "HyperShift" is the operator that we are using to create the "hosting service cluster" (which is the hosted control plane equivalent of the hub cluster). We are using HyperShift to make that work right now, there might be other options in the future. For consistency, and to avoid getting trapped into using only HyperShift, we want to make sure that we are using the right terminology in the right place. We are referring to the managed clusters in hosted control planes as "hosted clusters". It might be fine, but just want to be sure that we are using the right terms.

release_notes/known_issues.adoc
```diff
@@ -526,6 +526,12 @@ To work around this issue, complete the following steps:
...
+// 2.6:25156
+
+ When you create a HyperShift cluster with a cluster-wide proxy configuration on {ocp-short} 4.10.z, the `nodeip-configuration.service` service does not start on worker nodes.
```
cadawson: Hosting service cluster? 

...on the worker nodes?

cadawson (changes requested):
A couple of comments, but nothing major. Thanks!

release_notes/known_issues.adoc
oafischer: Might be a better question for @ldpliu. Can you confirm you meant HyperShift here?

release_notes/known_issues.adoc
oafischer: I'll implement the second suggestion, thanks. Going to ping @ldpliu for help on the first suggestion.

release_notes/known_issues.adoc
ldpliu: Thanks, I think @cadawson is right. and we should change "HyperShift" to "hosted control planes"

release_notes/known_issues.adoc
ldpliu: Yes, we should change it.

oafischer:
Thanks for chiming in @ldpliu, you can review the changes again when you have a chance @cadawson.

oafischer:
Made the suggested changes, going ahead with merge.

PR #3910, opened 2022-08-19
stolostron/backlog#24960 merge mce intro and main

cadawson:


cadawson:
HI @swopebe.  think I have these mostly consistent now. There are some comments that you had in there that I left alone, but hopefully you can work with it easier. I had these files confirmed during the meeting, but they agreed that a technical review is needed. One thing that I am not sure of is whether we can use more than one main file. I think we can only have one per book, and we break it up that way. We can have as many intro docs for a toc as we want. I wonder if multiple main files is why it is breaking the build. I can look into merging them into one, if you agree. Thanks!

PR #3949, opened 2022-08-24
Link .0 ext ca certs in 2.6

oafischer:
https://github.com/stolostron/backlog/issues/25246


abutcher:
@oafischer Looks good! There is a line slightly above this change that could further indicate that there may be multiple certificates with a `.0` extension. This best practice suggestion might make more sense given that context.

Something like this:

```
Use the certificate from the `certs/<platform>` directory that has a `.0` extension. *Tip:* You can use the `ls certs/<platform>` command to list all of the available certificates for your platform.
------
Use the certificates from the `certs/<platform>` directory that have a `.0` extension. *Tip:* You can use the `ls certs/<platform>` command to list all of the available certificates for your platform.
```

oafischer:
@abutcher Thanks for the input! I went ahead and implemented your suggestion. This is ready for peer review now.

abutcher (approved):

dockerymick (approved):
Looks good to me! thanks for the update!

PR #3950, opened 2022-08-24
Doc steps for discon inst img in 2.5

oafischer:
https://github.com/stolostron/backlog/issues/24308

### DIFF
--- a/clusters/create_openstack.adoc
+++ b/clusters/create_openstack.adoc
@@ -78,6 +78,8 @@ Proxy information that is provided in the credential is automatically added to t
 
 * Additional trust bundle: Specify the contents of the certificate file that is required to access the mirror registry. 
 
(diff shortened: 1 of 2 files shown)


clusters/create_openstack.adoc
```diff
@@ -78,6 +78,8 @@ Proxy information that is provided in the credential is automatically added to t
 
 * Additional trust bundle: Specify the contents of the certificate file that is required to access the mirror registry. 
 
+You can define the disconnected installation image by clicking *Disconnected installation*. See link:../release_notes/known_issues.adoc#create-with-disconnected[Disconnected installation settings for cluster creation cannot be entered or are ignored if entered] for more details on limitations.
+
```
swopebe, line 82: Looks great. One thing we need to remember is that we should only sparingly point to known issues in the documentation. The reason is, they resolve and another maintenance point. If we are pointing to a known issue in a main topic, it's usually an indication that the known issue is actually part of the process and should be written as such.

swopebe (approved):

clusters/create_openstack.adoc
berenss: great advice and I recognize that my own comment stirred that link.

berenss (approved):
/approve

PR #3954, opened 2022-08-25
Add Ansible credentials known issue in 2.6

oafischer:
https://github.com/stolostron/backlog/issues/25265

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -98,13 +98,21 @@ oc edit pvc <storageclassname>-search-redisgraph-0
 === Search query parsing error
 //2.5:22391 
 
-If an environment becomes large and requires more tests for scaling, the search queries can timeout which results in an parsing error message being displayed. This error is displayed after 30 seconds of waiting for a search query.
+If an environment becomes large and requires more tests for scaling, the search queries can timeout which results in a parsing error message being displayed. This error is displayed after 30 seconds of waiting for a search query.
 
 Extend the timeout time with the following command:
 
 ----
 kubectl annotate route multicloud-console haproxy.router.openshift.io/timeout=Xs
 ----
+
+[#ansible-tower-credentials-need-secure-host-automation-templates]
+=== Cannot use Ansible Tower credentials with host URL containing insecure protocol
+//2.6:25265
+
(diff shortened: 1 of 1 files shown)


release_notes/known_issues.adoc
```diff
@@ -98,14 +98,22 @@ oc edit pvc <storageclassname>-search-redisgraph-0
...
 
 ----
 kubectl annotate route multicloud-console haproxy.router.openshift.io/timeout=Xs
 ----
 
+[#ansible-tower-credentials-need-secure-host-automation-templates]
+=== Cannot use Ansible Tower credentials with host URL containing insecure protocol
+//2.6:25265
+
+You cannot use Ansible Tower credentials with a host URL containing an insecure protocol (http) to create automation templates. When you use an Ansible Tower credential with an insecure protocol in the automation wizard, the job list cannot be populated in the dropdown. To work around this issue, use a secure protocol (https) in the host URL.
```
cadawson: Suggestion here for translation ease. Also "dropdown" should be "drop-down menu".

When you create an Ansible Tower automation template with the console automation flow, the Ansible Tower credential that you use must specify a host URL that uses a secure protocol (https). When you try to use an Ansible Tower credential that specifies an insecure protocol, the job list cannot be populated in the drop-down menu. To work around this issue, use a secure protocol (https) in the host URL.

release_notes/known_issues.adoc
```diff
@@ -98,14 +98,22 @@ oc edit pvc <storageclassname>-search-redisgraph-0
...
+
+*Note:* The console prevents you from creating Ansible Tower credentials with a host URL containing an insecure protocol, but you can still manually create or patch a resource using an insecure protocol.
```
cadawson: Suggestion:
While you cannot create Ansible Tower templates from credentials with a host URL that contains an insecure protocol using the console, you can still manually create or patch a resource using an insecure protocol.

CD: Check this one. It sounds like the first statement said that you can't create templates, and the Note says that you can't create a credential. At least that is how I read it.

cadawson (changes requested):
Thanks Oliver! A couple of suggestions, and something to make sure we are saying the same thing across the statement.

release_notes/known_issues.adoc
oafischer: Yes that's correct. I wasn't sure if it's worth including the note at all because this known issue can only occur if the user creates the credential manually to begin with. I figured it still provides good context though. We can leave it out but I'll let @Randy424 weigh in on that.

Randy424 (approved):
/lgtm

PR #3960, opened 2022-08-25
no issue, update based on scrum chat

swopebe:


oafischer (approved):
Thanks for adding that, I prefer this for my workflow. Others can weigh in if they disagree.

PR #3964, opened 2022-08-26
stolostron/backlog#24637 add rclone restic fips

cadawson:


cadawson:
Hi @tesshuflower. Can you look through this, when you have a chance? Thanks!

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -16,19 +12,30 @@ Before installing VolSync on your clusters, you must have the following requirem
...
 
+
+[#rsync-backup-volsync]
+== Configuring an Rsync backup
+
+You can create a 1:1 asynchronous replication of persistent volumes by using an Rsync backup. You can use Rsync-based replication for disaster recovery or sending data to a remote site.
+
+The following example shows how to configure by using the Rsync method. For additional information about Rsync, see https://volsync.readthedocs.io/en/latest/usage/index.html[Usage] in the VolSync documentation.  
+
+VolSync does not create its own namespace, so it is in the same namespace as other {ocp-short} all-namespace operators. Any changes that you make to the operator settings for VolSync also affects the other operators in the same namespace, such as if you change to manual approval for channel updates.
```
tesshuflower, line 24: I think this (line 27) should be somewhere under the installation section for volsync

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -16,19 +12,30 @@ Before installing VolSync on your clusters, you must have the following requirem
 
 * At least two configured clusters that are managed by the same {product-title-short} hub cluster
 
-* Network connectivity between the clusters that you are configuring with VolSync; If the clusters are not on the same network, you can configure the link:../add-ons/submariner/submariner.adoc#submariner[Submariner multicluster networking and service discovery] and use the `ClusterIP` value for `ServiceType` to network the clusters, or use a load balancer with the `LoadBalancer` value for `ServiceType`. 
+* Network connectivity between the clusters that you are configuring with VolSync. If the clusters are not on the same network, you can configure the xref:../submariner/submariner.adoc#submariner[Submariner multicluster networking and service discovery] and use the `ClusterIP` value for `ServiceType` to network the clusters, or use a load balancer with the `LoadBalancer` value for `ServiceType`. 
 
 * The storage driver that you use for your source persistent volume must be CSI-compatible and able to support snapshots. 
 
+
+[#rsync-backup-volsync]
```
tesshuflower: rsync replication (for rsync I'd rather call it replication) was already documented back in acm 2-5 - can we still use that? We could still send them to the volsync docs for extra info.

 I think also I'd like to start with the install instructions before specific steps for rsync/rclone/restic

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -328,3 +335,222 @@ Status:
...
+Replace `destination` with the name of your replication destination CR.
++
+Replace `restic-repo` with the path to your repository where the source is stored.
++
+Replace `pvc-name` with the name of the new persistent volume claim where you want to restore the data. Use an existing persistent volume claim for this, rather than provisioning a new one. 
+
+The restore process only needs to be completed once, and this example restores the most recent backup. For more information about restore options, see https://volsync.readthedocs.io/en/latest/usage/restic/index.html#restore-options[Restore options] in the VolSync documentation. 
+
+[#rclone-backup-volsync]
+== Configuring an Rclone backup
```
tesshuflower: I think for rsync/rclone we should probably use the term `replication` instead of `backup`.  restic is the one where `backup` is appropriate.

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -328,3 +335,222 @@ Status:
...
++
+Replace `source` with the persistent volume claim that you are backing up.
++
+Replace the value for `schedule` with how often to run the backup. This example has the schedule for every 6 minutes. This value must be within quotation marks. See xref:../volsync/volsync_schedule.adoc#volsync-schedule[Scheduling your synchronization] for more information.
++
+Replace `intermediate-s3-bucket` with the path to the configuration section of the Rclone configuration file. This value must be within quotation marks. 
++
+Replace `destination-bucket` with the path to the object bucket where you want your replicated files copied. This value must be contaned within quotation marks. 
++
+Replace `rclone-secret` with the name of the secret that contains your Rclone configuration information. This value must be contained within quotation marks.
```
tesshuflower: I'm not sure we need to say it needs to be within quotation marks?  I think this part is just yaml standards - so for example for this secret name they would not need to use quotation marks normally.

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -328,3 +335,222 @@ Status:
...
+    schedule: "3,9,15,21,27,33,39,45,51,57 * * * *"
+  rclone:
+    rcloneConfigSection: "intermediate-s3-bucket"
+    rcloneDestPath: "destination-bucket"
+    rcloneConfig: "rclone-secret"
+    copyMethod: Snapshot
+    accessModes: [ReadWriteOnce]
+    capacity: 10Gi
+    storageClassName: <my-sc>
+    volumeSnapshotClassName: my-vsc
```
tesshuflower: This is called `my-vsc-name` on line 556.  Maybe to be consistent just use `<my-vsc>`  and use `<>`?

tesshuflower (changes requested):
Looks great in general Chris!  I've just made some small suggestions.

add-ons/volsync/volsync_replicate.adoc
cadawson: Yes, we can. I left the same information in rsync from the previous version. I just moved it around a little to work with the others, as well. 

Moved the install to earlier.

add-ons/volsync/volsync_replicate.adoc
cadawson: If they don't need the quotation marks, then I would suggest we leave them off of the example. They might have been used to show a variable, but we show it other ways. Let me update.

cadawson:
Thanks for the review @tesshuflower. I've updated the content with your comments. Please take a look, when you can. Thanks!

tesshuflower (approved):
/lgtm

Thanks a lot Chris!  Looks great!

PR #3968, opened 2022-08-26
stolostron/backlog#24470 add rel img conn

cadawson:


swopebe (approved):

cadawson:
Hi @swopebe . I know we are trying to keep the clusters section to a minimum. These are the release images files that we talked about that do not apply to MCE, but only to ACM. I've added them here. One of them also resolved a missing link issue in a file that I reviewed, so I went ahead and moved them and copied the main file. I wanted to let you know and see if you had other ideas about how to include these, other than how they are now. Thanks!

PR #3970, opened 2022-08-26
SNO workers in 2.6

oafischer:
https://github.com/stolostron/backlog/issues/24569


jelkosz:
I dont think this is written to the correct place. The feature is about the CIM flow only while the doc seems to be adding it to the IPI (IIUC).
So I'd suggest to remove this and add to release notes that if you create an SNO cluster using CIM (e.g. using the "host inventory" option in cluster creation wizard), in the details screen you will be able to add addition worker nodes to it.

oafischer:
Closing to start new PR.

PR #3975, opened 2022-08-29
stolostron/backlog#25426 errata docs 2.3.12

cadawson:


cadawson:
Confirmed only container updates by Juliana in the issue.

PR #3979, opened 2022-08-29
remove objectstore support in allow deny list feature

rokej:

https://github.com/stolostron/backlog/issues/22763

The subscription allow, deny list feature is not supported for object storage subscription. Remove it from the doc.


oafischer (approved):
Looks good, thanks. Does this need to go into known issues as well?

PR #3993, opened 2022-08-30
URL fragment update

cadawson:


cadawson:
Hi @swopebe . Finally got backup running. There were a couple of things happening, so I am not completely sure what it was. I believe the target URL has to be based on what the name is above it. It populates it automatically. I think the build expects it to be the same, and fails if it is not. 

When I tried to change it, though, something happened to break in the build process, so it would not complete the change. I had to open a ticket to get that fixed. Once he fixed that with the new target URL fragment, it is now running correctly. 

This is the update to the doc that includes this. If it looks ok, merge it, and we can move on. Make any changes that you want to. Thanks!

PR #4009, opened 2022-08-31
New

swopebe:


release_notes/whats_new.adoc
```diff
@@ -24,14 +24,24 @@ Learn more about what is new this release:
...
+
+----
+https://grafana-open-cluster-management-observability.{OPENSHIFT_INGRESS_DOMAIN}
+----
+
+See link:../observability/design_grafana.adoc#console-overview[Designing your Grafana dashboard] for more information. 
+
+Learn more about the console in the link:../console/console.adoc#console-overview[Console overview].
 
 [#observability-whats-new]
```
dockerymick: Seems like this section should be removed and only web console should be updated?

release_notes/whats_new.adoc
```diff
@@ -24,14 +24,24 @@ Learn more about what is new this release:
 [#web-console-whats-new]
 == Web console
 
-Learn more in the link:../console/console.adoc#console-overview[Console overview].
+You can now access Grafana through the [ocp-short] route with the following URL:
```
dockerymick: use curly brackets for conref

dockerymick (changes requested):
Verified links and left comments

release_notes/whats_new.adoc
swopebe: ugh

release_notes/whats_new.adoc
swopebe: Right, @dockerymick -- I asked about this on Slack so if you have no updates for this section, I will remove.

dockerymick: Sounds good, yea the current entry is related to observability, but i understand why we would keep it in Web console

dockerymick (approved):
looks good, thanks for the update!

PR #3870, opened 2022-08-15
GRC Updates - Draft 2

dhaiducek:
Alternative to:
- #3857

Addresses:
- https://github.com/stolostron/backlog/issues/23525
- [Bug 2090781](https://bugzilla.redhat.com/show_bug.cgi?id=2090781) - [RFE DOC] Highlight difference between "musthave" and "mustonlyhave"

Closes #3857

PR #4015, opened 2022-09-01
Remove Zettaset page

dhaiducek:
We don't have much concrete information on the deployment since it requires authorization we don't have. But this update makes it consistent with other pages and clarifies it a bit.

This PR has a dependency on:
- #3870


swopebe:
This is a huge PR, was this pulled from a recent 2.6 stage?

swopebe:
We GA next week, so a PR like this is a bit alarming, looking for context.

dhaiducek:
Once #3870 merges, this will be less huge--this branch came from that one, so all those commits will go away, leaving only the last two.

governance/zettaset_policy.adoc
```diff
@@ -1,9 +1,9 @@
 [#zettaset-policy-intro]
 = Zettaset policy
 
-The Kubernetes configuration policy controller monitors the status of your `ZettaSet` policy. Apply a `ZettaSet` policy to deploy the Zettaset application. The `Zettaset` policy uses the ConfigMap reloader to restart the deployment after the ConfigMap changes. The policy deploys a replica of the `zts-cmc-deployment` policy.
+The Kubernetes configuration policy controller monitors the status of your ZettaSet policy. Apply a ZettaSet policy to deploy the Zettaset application alongside a ConfigMap reloader. The ZettaSet application uses the ConfigMap reloader to restart the application after the referenced ConfigMap changes. This policy has a remediation action set to `inform` by default. Set the remediation action to `enforce` to deploy the application to the managed clusters.
```
dhaiducek: I took out the `` `ZettaSet` `` backticks because ZettaSet is a partner brand. I wasn't sure backticks were needed in that case?

dockerymick (approved):
Thanks for the quick update

PR #4016, opened 2022-09-01
Document cluster import connection to pull secrets in 2.5

oafischer:
https://github.com/stolostron/backlog/issues/24422
https://mail.google.com/chat/u/0/#chat/space/AAAAWskU424/QY5pSR7GO1U
https://bugzilla.redhat.com/show_bug.cgi?id=2109390

### DIFF
--- a/clusters/import_gui.adoc
+++ b/clusters/import_gui.adoc
@@ -148,7 +148,7 @@ Your cluster is imported. You can import another by selecting *Import another*.
 
 When you deploy a new hub, you must define the pull secret. The pull secret you define during the hub install is used for importing clusters that were not created by {ocp-short}. {ocp-short} clusters do not use the pull secret you defined during the hub install.
 
-Complete the following steps to import clusters that were not created by {ocp-short} with the correct pull secret:
+To import clusters that were not created by {ocp-short} using the correct pull secret, complete the following steps:
 
 . Run the following command to create a secret that is copied to the managed cluster:
 +
--- a/clusters/import_gui.adoc
+++ b/clusters/import_gui.adoc
@@ -142,6 +142,52 @@ spec:
 Replace the value of the URL with the URL that provides external access to the managed cluster that you are importing.
 
 Your cluster is imported. You can import another by selecting *Import another*.
+
+[#importing-a-cluster-non-ocp]
+=== Importing a cluster that was not created by {ocp-short}
+
+When you deploy a new hub, you must define the pull secret. The pull secret you define during the hub install is used for importing clusters that were not created by {ocp-short}. {ocp-short} clusters do not use the pull secret you defined during the hub install.
+
+Complete the following steps to import clusters that were not created by {ocp-short} with the correct pull secret:
+
+. Run the following command to create a secret that is copied to the managed cluster:
++
+----
+oc create secret generic pull-secret -n open-cluster-management --from-file=.dockerconfigjson=../../pull-secret --type=kubernetes.io/dockerconfigjson
+----
+
+. Apply the operator CRD by running the following command:
++
+----
+ oc apply -f v2_multiclusterhub.yaml -n open-cluster-management
+----
+
+The `operator.yaml` file might resemble the following example:
+
+[source,yaml]
+----
+apiVersion: operator.open-cluster-management.io/v1
+  kind: MultiClusterHub
+  metadata:
+    name: multiclusterhub
+  spec:
+    availabilityConfig: High
+    hive:
+    backup:
+    velero: {}
+    failedProvisionConfig: {}
+    imagePullSecret: pull-secret
+    ingress:
+    sslCiphers:
+      - ECDHE-ECDSA-AES256-GCM-SHA384
+      - ECDHE-RSA-AES256-GCM-SHA384
+      - ECDHE-ECDSA-CHACHA20-POLY1305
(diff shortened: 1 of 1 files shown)


clusters/import_gui.adoc
```diff
@@ -142,6 +142,52 @@ spec:
 Replace the value of the URL with the URL that provides external access to the managed cluster that you are importing.
 
 Your cluster is imported. You can import another by selecting *Import another*.
+
+[#importing-a-cluster-non-ocp]
+=== Importing a cluster that was not created by {ocp-short}
```
berenss, line 147: To be precise, this is about: Importing a non OpenShift cluster
for example Oliver, we're talking about EKS, GKE, AKS, IKS - kubernetes distributions that ACM does not create, so these must get imported.

clusters/import_gui.adoc
```diff
@@ -142,6 +142,52 @@ spec:
...
+
+When you deploy a new hub, you must define the pull secret. The pull secret you define during the hub install is used for importing clusters that were not created by {ocp-short}. {ocp-short} clusters do not use the pull secret you defined during the hub install.
```
berenss, line 149: notice the topic here, in the Install section: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/install/installing#custom-image-pull-secret

clusters/import_gui.adoc
```diff
@@ -142,6 +142,52 @@ spec:
...
+To import clusters that were not created by {ocp-short} using the correct pull secret, complete the following steps:
+
+. Run the following command to create a secret that is copied to the managed cluster:
++
+----
+oc create secret generic pull-secret -n open-cluster-management --from-file=.dockerconfigjson=../../pull-secret --type=kubernetes.io/dockerconfigjson
+----
+
+. Apply the operator CRD by running the following command:
++
```
berenss, line 160: I don't know if this is actually necessary. Perhaps a review from install squad can clarify.

clusters/import_gui.adoc
```diff
@@ -142,6 +142,52 @@ spec:
...
+----
+oc apply -f v2_multiclusterhub.yaml -n open-cluster-management
+----
+
+The `operator.yaml` file might resemble the following example:
+
```
berenss, line 166: What is the `operator.yaml` file? in the line above it is the `v2_multiclusterhub.yaml`
but either way, I'm not sure that this step is necessary.

clusters/import_gui.adoc
oafischer: "Not created by ocp-short" in this case is the same as "non OpenShift cluster" since we can't say the latter phrase.

Is there another way to refer to these clusters if "not created by {ocp-short}" is not precise enough but we can't use "non OpenShift cluster" either?

clusters/import_gui.adoc
oafischer: Thanks for linking that Scott, I had not seen it. That looks exactly like what was requested in the original issue. Can you confirm that it addresses the original issues sufficiently? If so I would close this PR and the issue. No need to duplicate this info IMO.

clusters/import_gui.adoc
oafischer: I am not sure, @ch-stark mentioned the operator.yaml file here: https://github.com/stolostron/backlog/issues/24422

clusters/import_gui.adoc
oafischer: You can find the links to the original issues I'm referring to at the top of this page.

berenss: > 

I think the main concern that causes problems is that the customer will skip the configuration of the pull secret at install time. So they have a hub and life is good, then they decide to import non-OCP cluster and then whammy. 
so it’s almost like we need a note at the top of this Import section that reminds the user about the necessity of having the pull secret defined before attempting non-OCP import. 
and whether or not there are any extra steps needed to reconcile that new pull secret with the hub installer I think that should be deferred to hub installer squad rather than the related links.

berenss: We also have to remember that the person that installs ACM might not be the person that uses ACM
And that ‘install’ and ‘use’ step could be separated by days/months.

oafischer: Thanks for the explanation Scott, it's really helpful to hear these insights and understand things from a customer perspective. As someone who is still learning a lot about the product/docs I really appreciate you providing the context here. It sounds to me as though implementation in this PR could work, but it might be even better to shift the information further up and integrate it into the "Importing a cluster" section above.

Funnily enough that was my original plan but that entire is section is so messy/overloaded (from a formatting standpoint) that I didn't feel like I could add this information effectively.

I think I'll proceed as follows:
1. Move the information into the "Importing a cluster" section, possibly as a [special tag](https://stolostron.github.io/rhacm-docs/wr_special_tag_guidance) or at least have an additional (important) note to draw attention. I'll find a way to get around the mess.
2. Use the install section you linked as a guide for the content in this PR and adjust what I have accordingly.

Let me know what your thoughts on that are and if that sounds like it would address the customer need you had in mind. Thanks again for taking time so late in the day. I'll try my best to get this in on time. Would be great to have this addressed for the release.

oafischer:
Closing in favor of https://github.com/stolostron/rhacm-docs/pull/4017

PR #4017, opened 2022-09-02
stolostron/backlog#24460 upd pull secret

cadawson:


oafischer (approved):
Looks great from a writing perspective! Based on my previous work and @berenss comments, I think the content is complete too. Some additional food for thought here: https://github.com/stolostron/rhacm-docs/pull/4016

Since the original issue comes from the user potentially glossing over the docs, I wonder if we can emphasize the importance of the pull secret when importing a non-ocp cluster. Maybe an **Important:** tag might help, if the formatting allows it.

berenss (approved):
this looks great and informs on a lot of the work @oafischer is looking into.
I agree a Note or 'heads up' would be nice to emphasize the need for that pull secret for non-OCP import.

oafischer:
Added a note to the section in question. I think that's about as far as I would go.

PR #3789, opened 2022-08-01
stolostron/backlog#20969 hyper endpts 25

cadawson:

PR #4018, opened 2022-09-02
stolostron/backlog#20969 hstd ctrl endpts 2.6

cadawson:


cadawson:
Comments from PR #3789 included here.

multicluster_engine/about/requirements.adoc
```diff
@@ -78,3 +78,17 @@ For the managed cluster networking requirements, see the following table:
...
+
+When you use hosted control planes, the `HypershiftDeployment` resource must have connectivity to the following endpoints:
+
+* {ocp-short} control-plane and worker nodes
+* AWS API
+* S3 API
+* Azure API
+* Registration agents for all managed clusters
+* {ocp-short} image repositories that store the ISO images of the coreOS and teh image registry for {ocp-short} pods
+* {product-title-short} hub connects to the klusterlet that runs on a hosted cluster
```
jnpacker: connects to the klusterlet that runs on a hosting cluster.  The klusterlet runs outside the hostedCluster.

multicluster_engine/about/requirements.adoc
```diff
@@ -78,3 +78,17 @@ For the managed cluster networking requirements, see the following table:
...
 
+[#network-configuration-hosted-control-planes]
+=== Hosted control planes networking requirements (Technology Preview)
+
+When you use hosted control planes, the `HypershiftDeployment` resource must have connectivity to the following endpoints:
+
+* {ocp-short} control-plane and worker nodes
+* AWS API
+* S3 API
+* Azure API
```
jnpacker: It's either the first two or Azure.

So maybe in brackets (Amazon Web Service) and (Azure cloud service)

Or (When hosting on AWS) and (When hosting on Azure)

multicluster_engine/about/requirements.adoc
```diff
@@ -78,3 +78,17 @@ For the managed cluster networking requirements, see the following table:
...
+* Registration agents for all managed clusters
+* {ocp-short} image repositories that store the ISO images of the coreOS and teh image registry for {ocp-short} pods
+* {product-title-short} hub connects to the klusterlet that runs on a hosted cluster
+* Local API client of the klusterlet on the hosting cluster communicates with the API of the HyperShift hosted cluster
+
```
jnpacker: I assume there is a section on BM.  Is there any communication we need to disclose for the Agent service that runs on the MCE Hub??

swopebe: I actually am wondering if we can instead have a networking topic so we don't end up with stuff everywhere:

https://github.com/stolostron/rhacm-docs/blob/2.6_stage/networking/networking_intro.adoc

?

jnpacker: Sure.  We would want to probably have a pointer to that though from each section, if that is where the networking for the different topics will land...

Example, App, Policy, Standalone provisioning, Hypershift provisioning (For the two provisioning's across several platform API's, and the list goes on. 😄

cadawson:
Made some updates to this one. Is this what you had in mind @jnpacker and @swopebe?

multicluster_engine/about/networking_mce.adoc
```diff
@@ -0,0 +1,119 @@
...
+
+* <<hub-network-config-mce,Hub cluster network configuration>>
+* <<managed-network-config-mce,Managed cluster network configuration>>
+* <<network-configuration-infra-operator-mce,Additional networking requirements when installing using the infrastructure operator>>
+* <<network-configuration-hive-mce,Additional networking requirements when installing using the Hive Operator>>
+* <<network-configuration-hosted-control-planes-mce,Hosted control planes networking requirements (Technology Preview>>
+
+[#hub-network-config-mce]
+== Hub cluster network configuration
+
```
swopebe, line 14: Why do we have hub cluster in the MCE doc?

multicluster_engine/about/networking_mce.adoc
```diff
@@ -0,0 +1,119 @@
...
+| 6443
+
+| Inbound
+| Post-commit hook from GitHub to the multicluster engine for Kubernetes cluster. This setting is only required when you use certain application management functions.
+| 6443
+
+|===
+
+[#managed-network-config-mce]
+== Managed cluster network configuration
```
swopebe, line 41: One word of caution: This needs to be tested/validated by the foundation team or someone (I believe they were the ones who tested the RHACM doc in this area), if not already. And I wonder if the network topic in RHACM should be revisited bc MCE doc should only contain the MCE stuff and RHACM should only contain RHACM stuff.

multicluster_engine/about/requirements.adoc
zhujian7: I think the hub do not need to connect to the klusterlet, instead the klusterlet agent need to access to the hub

multicluster_engine/about/requirements.adoc
```diff
@@ -78,3 +78,17 @@ For the managed cluster networking requirements, see the following table:
...
+[#network-configuration-hosted-control-planes]
+=== Hosted control planes networking requirements (Technology Preview)
+
+When you use hosted control planes, the `HypershiftDeployment` resource must have connectivity to the following endpoints:
+
+* {ocp-short} control-plane and worker nodes
+* AWS API
+* S3 API
+* Azure API
+* Registration agents for all managed clusters
```
zhujian7: I do not quite understand this `Registration agents for all managed clusters` here. does it mean the Registration agents for all managed clusters need to access the hub?

cadawson:
Hi @rokej. Can you take a look at this update and at Jian's questions, when you can? Thank you!

multicluster_engine/about/requirements.adoc
rokej: HypershiftDeployment does not need network connection to registration agents. We can remove this line.

multicluster_engine/about/requirements.adoc
rokej: I agree with Jian. We can remove this line `hub connects to the klusterlet that runs on a hosted cluster`

multicluster_engine/about/requirements.adoc
rokej: Outbound connection to AWS API and S3 API (Amazon Web Service) for AWS platform hosted clusters
Outbound connection to Azure API (Azure cloud service) for Azure platform hosted clusters

multicluster_engine/about/networking_mce.adoc
cadawson: MCE also has hub cluster.

multicluster_engine/about/requirements.adoc
cadawson: Removed this line

rokej (approved):

multicluster_engine/about/networking_mce.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+| Outbound from the {ocp-short} managed cluster to the hub cluster
+| HTTPS
+| Reports hardware information using the `assistedService` route 
+| 443
+
+| Outbound from the {ocp-short} managed cluster to the ISO/rootfs image repository
+| HTTP
+| Downloads the rootfs image
+| 80
```
jnpacker: Are we 100% sure it doesn't do 443 (https as well or instead?)

cadawson: @rokej , can you help answer this? Thanks!

rokej: @cadawson Sorry I did not test or author this part so I don't know. I would imagine 443 should be the default port though.

jnpacker: Googling around, docker registry port defaults are 443 HTTPS/TLS, but I couldn't find if Quay.io uses anything different.

cadawson: So should we include an entry for Quay.io inbound and outbound on https port 443? @jnpacker ?

jnpacker: We reach out to quay to pull image outbound.  Quay.io does not reach into ACM. So the connection starts outbound to quay.io

cadawson: Added that update to the managed cluster table @jnpacker. Does it look ok? Thanks!

multicluster_engine/about/networking_mce.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+|===
+| Direction | Connection | Port (if specified)
+
+| Outbound
+| API of the cloud provider
+| 
+
+| Outbound
+| Kubernetes API server of the provisioned managed cluster
```
zhujian7, line 27: I think MCE does not need to access the managed cluster.

multicluster_engine/about/networking_mce.adoc
```diff
@@ -0,0 +1,138 @@
...
+[#managed-network-config-mce]
+== Managed cluster network configuration
+
+For the managed cluster networking requirements, see the following table:
+
+|===
+| Direction | Connection | Port (if specified)
+
+| Outbound and inbound
+| Kubernetes API server of the multicluster engine for Kubernetes cluster
```
zhujian7, line 53: I think only `managedcluster -> mce` is required

multicluster_engine/about/networking_mce.adoc
```diff
@@ -0,0 +1,138 @@
...
+| Outbound
+| API of the cloud provider
+| 
+
+| Outbound
+| Kubernetes API server of the provisioned managed cluster
+| 6443
+
+| Outbound and inbound
+| The `WorkManager` service route on the managed cluster
```
zhujian7, line 31: Does `WorkManager` refer to klusterlet-work-agent or workmgr-addon-agent?

multicluster_engine/about/networking_mce.adoc
```diff
@@ -0,0 +1,138 @@
...
+| 443
+
+| Inbound
+| The Kubernetes API server of the multicluster engine for Kubernetes cluster from the managed cluster
+| 6443
+
+| Inbound
```
jnpacker: This one does not apply to MCE.

multicluster_engine/about/networking_mce.adoc
```diff
@@ -0,0 +1,138 @@
...
+See the hub cluster network requirements in the following table:
+
+|===
+| Direction | Connection | Port (if specified)
+
+| Outbound
+| API of the cloud provider
+| 
+
+| Outbound
```
jnpacker, line 26: I think we want to call out that this is optional, and only used by some of our input flows (you enter the ManagedCluster credentials in the Console to import it)

multicluster_engine/about/networking_mce.adoc
jnpacker: Still not sure about Line 81, who can we confirm its 80 and not 443 TLS? @eranco74

multicluster_engine/about/networking_mce.adoc
```diff
@@ -0,0 +1,138 @@
...
+| For hosted clusters on Microsoft Azure cloud services only: Outbound connection to Azure API
+| 
+
+| Outbound
+| {ocp-short} image repositories that store the ISO images of the coreOS and the image registry for {ocp-short} pods
+| 
+
+| Outbound
+| Local API client of the klusterlet on the hosting cluster communicates with the API of the HyperShift hosted cluster
```
jnpacker: I'm not sure this leaves the cluster, @zhujian7 can you comment on if we're using a local Service URL to communicate, so it is NOT external in any way?

zhujian7: yes, we are using the Local Service URL, see https://github.com/stolostron/hypershift-addon-operator/blob/b0183710d1ea7bd15fc3539ac544da2081481206/pkg/agent/agent.go#L280, it is NOT external

multicluster_engine/about/networking_mce.adoc
cadawson: Sent a slack message to follow up with @eranco74 .

eranco74: I'm assuming the ISO/rootfs image repository is the assisted-image-service.
The assisted-image-service is using http, so it should be port 80.

eranco74: OK, seems that the assisted-image-service [supports https as well](https://github.com/openshift/assisted-image-service/blob/95f275cd8f19d972c1a38518595bca17a64bd170/README.md?plain=1#L23-L33).
 
@carbonin what is the default and do we allow users to change it?

multicluster_engine/about/networking_mce.adoc
```diff
@@ -0,0 +1,130 @@
...
+| HTTPS
+| Reports hardware information using the `assistedService` route 
+| 443
+
+| Outbound from the {ocp-short} managed cluster to the ISO/rootfs image repository
+| HTTP
+| Downloads the rootfs image
+| 80
+
+| Outbound from the {ocp-short} managed cluster to the quay.io image repository
```
jnpacker: It's officially `registery.redhat.com` for the images. Even if its backed by quay.io.

multicluster_engine/about/networking_mce.adoc
jnpacker: @cadawson so we can have 80/443

carbonin: Is this for downloading the ISO to the image service?

In that case it depends entirely on what the user puts in `OS_IMAGES` so it can be either http or https.

carbonin: Or is this downloading the rootfs from the image service to the node being deployed?
In which case it should always be https

multicluster_engine/about/networking_mce.adoc
cadawson: Updated. Thanks @jnpacker .

multicluster_engine/about/networking_mce.adoc
```diff
@@ -0,0 +1,130 @@
...
+
+| Outbound from the {ocp-short} managed cluster to the hub cluster
+| HTTPS
+| Reports hardware information using the `assistedService` route 
+| 443
+
+| Outbound from the {ocp-short} managed cluster to the ISO/rootfs image repository
+| HTTP, HTTPS or TLS
+| Downloads the rootfs image
+| HTTP 80, HTTPS 80, or 443 TLS
```
jnpacker: This is HTTP 80 and HTTPS 443.

cadawson:
Ok, @jnpacker, we might be done with this one. Will you check?

PR #4025, opened 2022-09-02
Horizontal scaling of ocp cluster in 2.6

oafischer:
https://github.com/stolostron/backlog/issues/24059

### DIFF
--- a/multicluster_engine/cluster_lifecycle/add_hosts_infra_env.adoc
+++ b/multicluster_engine/cluster_lifecycle/add_hosts_infra_env.adoc
@@ -42,3 +42,11 @@ When you add a host by using the BMC option, the host is automatically approved.
 .. Select *Upload* and navigate to the completed YAML template to add multiple hosts. 
 
 You can now create an on-premises cluster on this infrastructure environment. See xref:../cluster_lifecycle/create_cluster_on_prem.adoc#creating-a-cluster-on-premises[Creating a cluster in an on-premises environment] for more information about creating a cluster.
+
+[#adding-nodes-ocp-clusters-cim]
+== Adding nodes to existing {ocp-short} clusters deployed with CIM
+
+To add more nodes to an existing {ocp-short} cluster deployed with CIM, complete the following steps:
+
+. Add additional hosts to create new Agent custom resources. See <<adding-hosts-to-an-infrastructure-environment, Adding hosts to an infrastructure environment>> for more details.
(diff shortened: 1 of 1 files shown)


multicluster_engine/cluster_lifecycle/add_hosts_infra_env.adoc
```diff
@@ -42,3 +42,11 @@ When you add a host by using the BMC option, the host is automatically approved.
 .. Select *Upload* and navigate to the completed YAML template to add multiple hosts. 
 
 You can now create an on-premises cluster on this infrastructure environment. See xref:../cluster_lifecycle/create_cluster_on_prem.adoc#creating-a-cluster-on-premises[Creating a cluster in an on-premises environment] for more information about creating a cluster.
+
+[#adding-nodes-ocp-clusters-cim]
+== Adding nodes to existing {ocp-short} clusters deployed with CIM
+
+To add more nodes to an existing {ocp-short} cluster deployed with CIM, complete the following steps:
+
+. Add additional hosts to create new Agent custom resources. See <<adding-hosts-to-an-infrastructure-environment, Adding hosts to an infrastructure environment>> for more details.
```
dockerymick: Is "Agent" the name or kind for the custom resource? Consider monospacing Agent. I trust your decision

dockerymick (approved):
Left a comment, looks good!

multicluster_engine/cluster_lifecycle/add_hosts_infra_env.adoc
oafischer: Good question, I've mostly seen them just as Agent and as far as I understand they are custom resources, so I'm inclined to leave them as is. Maybe @mihirlele can clarify.

berenss: Agent, within the context of CIM, is a part of the host discovery mechanism that allows the bare metal host to send it's information back to the ACM Hub and store that into an inventory list. ACM could then use the bare metal host from that inventory to either provision a new cluster or add a new host to an existing cluster.

This is not the same agent as in when we talk about the klusterlet agent.

Agent is a CRD and belongs to agent-install.openshift.io

oafischer: Thanks Scott! Very insightful as always. I appreciate the explanation.

What few examples of CRDs I see are monospaced, so I'll go with that. I also changed CR to CRD.

PR #4026, opened 2022-09-02
Mj improvements

dockerymick:
https://github.com/stolostron/backlog/issues/24816

### DIFF
--- a/about/architecture.adoc
+++ b/about/architecture.adoc
@@ -39,7 +39,9 @@ From the hub cluster console, you can view an aggregation of all cluster health 
 
 Much of the cluster lifecycle function is provided by the multicluster engine for Kubernetes operator that is installed automatically with {product-title-short}. See link:../multicluster_engine/cluster_lifecycle/cluster_lifecycle_intro.adoc#cluster-overview[The multicluster engine operator cluster lifecycle overview] for general information about the cluster lifecycle function. For information about the cluster lifecycle upgrades that are only available with {product-title-short}, see link:../clusters/intro.adoc#managing-your-clusters[Managing your clusters].
 
-//Need to ensure there is an intro file for this ^
+//Need to ensure there is an intro file for this ^ 
+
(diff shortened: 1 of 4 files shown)


clusters/cluster_lifecycle_intro.adoc
```diff
@@ -1,9 +1,9 @@
 [#cluster-lifecycle-architecture]
 = Cluster lifecycle architecture
-
+//can this be renamed to just Cluster lifecycle overview? the cluster_lifecycle_arch.adoc file and this file have the same titles
```
cadawson, line 3: I think it might be a good idea, but I would rather leave it like this so close to the GA.

cadawson (approved):
Good improvements here. I don't use the internal links, unless it is the topic links at the top. Probably to remain consistent with how they came over when we converted. Thanks! Left you a note about the renaming of files. Kinda scary. Thanks!

about/architecture.adoc
```diff
@@ -39,7 +39,9 @@ From the hub cluster console, you can view an aggregation of all cluster health
 
 Much of the cluster lifecycle function is provided by the multicluster engine for Kubernetes operator that is installed automatically with {product-title-short}. See link:../multicluster_engine/cluster_lifecycle/cluster_lifecycle_intro.adoc#cluster-overview[The multicluster engine operator cluster lifecycle overview] for general information about the cluster lifecycle function. For information about the cluster lifecycle upgrades that are only available with {product-title-short}, see link:../clusters/intro.adoc#managing-your-clusters[Managing your clusters].
 
-//Need to ensure there is an intro file for this ^
+//Need to ensure there is an intro file for this ^
```
dockerymick: @cadawson this link is broken, need to verify where to point them too

cadawson: @dockerymick: I gotcha now. I fixed the link. I went ahead and linked to the main page. Thanks!

PR #4030, opened 2022-09-02
Updating links

dockerymick:
https://github.com/stolostron/backlog/issues/24816


oafischer (approved):
Fixed one link, everything else checks out according to the tool. Will merge like you requested.

PR #4046, opened 2022-09-06
Add known issue console in 2.6

oafischer:
https://github.com/stolostron/backlog/issues/25389

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -120,7 +120,7 @@ When you create an Ansible Tower automation template with the console automation
 
 When you edit namespace bindings for a cluster set with the `admin` role or `bind` role, you might encounter the following error:
 
-`ResourceError: managedclustersetbindings.cluster.open-cluster-management.io "<cluster-set>" is forbidden: User "<user>" cannot create/delete resource "managedclustersetbindings" in API group "cluster.open-cluster-management.io" in the namespace "<namespace".`
+`ResourceError: managedclustersetbindings.cluster.open-cluster-management.io "<cluster-set>" is forbidden: User "<user>" cannot create/delete resource "managedclustersetbindings" in API group "cluster.open-cluster-management.io" in the namespace "<namespace>".`
 
 To work around the issue, make sure you also have permission to create or delete a `ManagedClusterSetBinding` resource in the namespace you want to bind. The role bindings only allow you to bind the cluster set to the namespace.
 
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -114,11 +114,11 @@ When you create an Ansible Tower automation template with the console automation
 
 *Note:* While you cannot create Ansible Tower templates from credentials with a host URL that contains an insecure protocol using the console, you can still manually create or patch a resource using an insecure protocol.
 
-[#cannot-edit-namespace-bindings-for-cluster]
-=== Cannot edit namespace bindings for cluster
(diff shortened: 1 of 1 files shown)


release_notes/known_issues.adoc
```diff
@@ -114,6 +114,16 @@ When you create an Ansible Tower automation template with the console automation
...
 
+[#cannot-edit-namespace-bindings-for-cluster]
+=== Cannot edit namespace bindings for cluster
+//2.6:25389
+
+When you edit namespace bindings for a cluster set with the `admin` role or `bind` role, you might encounter the following error:
+
+`ResourceError: managedclustersetbindings.cluster.open-cluster-management.io "<cluster-set>" is forbidden: User "<user>" cannot create/delete resource "managedclustersetbindings" in API group "cluster.open-cluster-management.io" in the namespace "<namespace>".`
+
+To work around the issue, make sure you also have permission to create or delete a `ManagedClusterSetBinding` resource in the namespace you want to bind. The role bindings only allow you to bind the cluster set to the namespace.
```
KevinFCormier: Could we avoid the words "work around" and maybe say "resolve" instead? We usually use that phrase when the problem is a bug. Although we are adding this as a known issue, the issue that we intend to fix is that the UI should not show namespaces for which the user would encounter these errors. 

Also, will you open another PR for 2.5 once we finalize the wording on this one?

oafischer: Yes, I'll change it to resolve. I'll start a PR for 2.5 once this one is merged.

release_notes/known_issues.adoc
```diff
@@ -114,6 +114,16 @@ When you create an Ansible Tower automation template with the console automation
 
 *Note:* While you cannot create Ansible Tower templates from credentials with a host URL that contains an insecure protocol using the console, you can still manually create or patch a resource using an insecure protocol.
 
+[#cannot-edit-namespace-bindings-for-cluster]
+=== Cannot edit namespace bindings for cluster
```
cadawson: Should this be cluster set?

cadawson (approved):
Other than that question, it looks good. Thanks!

release_notes/known_issues.adoc
KevinFCormier: Good catch!

release_notes/known_issues.adoc
```diff
@@ -114,6 +114,16 @@ When you create an Ansible Tower automation template with the console automation
...
+//2.6:25389
+
+When you edit namespace bindings for a cluster set with the `admin` role or `bind` role, you might encounter the following error:
+
+`ResourceError: managedclustersetbindings.cluster.open-cluster-management.io "<cluster-set>" is forbidden: User "<user>" cannot create/delete resource "managedclustersetbindings" in API group "cluster.open-cluster-management.io" in the namespace "<namespace>".`
```
KevinFCormier, line 123: Also note that this error has text in angle brackets: `<cluster-set>`, `<user>`, and `<namespace>`

These are placeholders for the actual values. Not sure if we need to point that out, or say they might see an error similar to the following or something to that effect.

release_notes/known_issues.adoc
oafischer: Good catch, yes. Fixed it, thanks!

release_notes/known_issues.adoc
oafischer: In that case we can say something like "you might encounter an error that resembles the following message". I'll update it, thanks.

KevinFCormier (approved):

PR #4059, opened 2022-09-09
change xref back to link

cadawson:


cadawson:
Changing the link to an xref caused the build to fail. Changing it back to complete a build and test the other possible fix.

PR #4065, opened 2022-09-09
Document Kyverno policies in 2.6

oafischer:
https://github.com/stolostron/backlog/issues/25458

@gparvin The other policies have links to topics about them in the doc. Kyverno has no dedicated topics at the moment. Are you sure we can deliver this to the customer with no additional documentation? Will there be documentation in the future?


gparvin:
We definitely need some documentation to go along with this.  The documentation on the policy-collection repository would be the starting point.  The last 3 entries in this table are the policies: https://github.com/stolostron/policy-collection/tree/main/stable/CM-Configuration-Management#configuration-management

In all 3 policies there is one resource created - a `ClusterPolicy` resource.
1. Kyverno network policy it is: https://github.com/stolostron/policy-collection/blob/main/stable/CM-Configuration-Management/policy-kyverno-add-network-policy.yaml#L28-L63
2. Kyverno quota policy it is: https://github.com/stolostron/policy-collection/blob/main/stable/CM-Configuration-Management/policy-kyverno-add-quota.yaml#L28-L80
3. Kyverno secret sync policy it is: https://github.com/stolostron/policy-collection/blob/main/stable/CM-Configuration-Management/policy-kyverno-sync-secrets.yaml#L28-L57

I can help provide more data but this may give a good overview.

oafischer:
Thanks @gparvin, I'll start working on the additional documentation and pull you in when I need help or additional details.

oafischer:
I have documented the new policies. Please review when you have a chance @gparvin. Thanks!

governance/kyverno_add_network_policy.adoc
```diff
@@ -0,0 +1,69 @@
...
+* <<kyverno-add-network-sample,Kyverno add network policy sample>>
+
+[#kyverno-add-network-policy-yaml-structure]
+== Kyverno add network policy YAML structure
+
+Your `policy-kyverno-add-network-policy` policy might resemble the following YAML file:
+
+[source,yaml]
+----
+     apiVersion: policy.open-cluster-management.io/v1
```
cadawson: Is this lining up correctly?

governance/kyverno_add_network_policy.adoc
```diff
@@ -0,0 +1,69 @@
+[#kyverno-add-network-policy]
+= Kyverno add network policy
+
+The Kyverno add network policy configures a new `NetworkPolicy` resource called `default-deny` which denies all traffic when you create a new namespace.
```
cadawson: Should we set the name of the policy offset on the first version, like: The Kyverno <i>add network</i> policy...or The Kyverno <i>add network policy</i>... Not sure if that would be consistent with similar topics, though.

cadawson (approved):
A couple of considerations. Thanks!

governance/kyverno_add_network_policy.adoc
oafischer: Good catch, I'll fix that.

governance/kyverno_add_network_policy.adoc
oafischer: Might be a good question for @gparvin but like you said, for consistency I'm inclined to leave it as is.

governance/kyverno_sync_secrets_policy.adoc
```diff
@@ -0,0 +1,54 @@
+[#kyverno-sync-secrets-policy]
+= Kyverno sync secrets policy
+
+The Kyverno sync secrets policy copies a secret called `regcred` which exists in the `default` namespace to new namespaces you creat and updates the secret when changes are detected.
```
gparvin: Change `creat` to `create`

oafischer: Good catch!

governance/kyverno_add_network_policy.adoc
gparvin: I think leaving it the way it is works the best.  Policies have lots of names so highlighting a name can be confusing.

gparvin (commented):
The 3 Kyverno policies all have a dependency that is mentioned in the community doc.

Note: Kyverno controller must be installed to use the kyverno policy. See the [Policy to install Kyverno](https://github.com/stolostron/policy-collection/blob/main/community/CM-Configuration-Management/policy-install-kyverno.yaml) in the community folder.

I think a note similar to that should be added to these sections.

oafischer:
@gparvin If it looks good now I will go ahead and merge, then do the same thing for 2.5. I opted against using a **Note:** since we try to [avoid special tags according to our process](https://stolostron.github.io/rhacm-docs/wr_special_tag_guidance) and in this case it works well with the intro paragraph. Thanks for your help on this!

gparvin (approved):
Thanks!

PR #4067, opened 2022-09-09
Add sku while creating public-ip and note about removal of g/w node.

aswinsuryan:
1) The Azure CLI creates a public IP of type basic which conflicts
with the load balance type (Standard) created by the Openshift installer. 
Changed the doc to specify the SKU type as Standard explicitly

2) Add note for Azure g/w node deletion after uninstallation

Signed-off-by: Aswin Suryanarayanan <aswinsuryan@gmail.com>

### DIFF
--- a/add-ons/submariner/submariner_prepare_hosts.adoc
+++ b/add-ons/submariner/submariner_prepare_hosts.adoc
@@ -17,7 +17,7 @@ kubectl label nodes <worker-node-name> "submariner.io/gateway=true" --overwrite
 . Create a public IP and assign it to the VM of the node that was tagged as gateway node by running the following commands:
 +
 ----
-az network public-ip create --name <public-ip-name> --resource-group <res-group>
+az network public-ip create --name <public-ip-name> --resource-group <res-group> -sku Standard
 az network nic ip-config update --name <name>  --nic-name <gw-vm-nic> --resource-group <res-group>  --public-ip-address <public-ip-name>
 ----
 +
--- a/add-ons/submariner/submariner_prepare_hosts.adoc
+++ b/add-ons/submariner/submariner_prepare_hosts.adoc
@@ -65,8 +65,7 @@ az network nsg rule create --resource-group <res-group> \
 --protocol udp --destination-port-ranges <vxlan-port>
 ----
 
-**Important:** Ensure that a new gateway node is tagged as a gateway node when trying a reinstall of Submariner. Reusing the current gateway after uninstalling of submariner can lead the connections to go to an error state.
-This is applicable only to {product-title}, where manual cloud preperation steps are used.
+*Important:* Ensure that a new gateway node is tagged as a gateway node when you reinstall Submariner. Reusing the current gateway after uninstalling Submariner result in the connections displaying an error state. This requirement is only applicable when you are using {product-title} with manual cloud preparation steps.
 
 [#preparing-ibm]
 == Preparing IBM Cloud for Submariner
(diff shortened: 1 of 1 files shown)


add-ons/submariner/submariner_prepare_hosts.adoc
```diff
@@ -17,14 +17,16 @@ kubectl label nodes <worker-node-name> "submariner.io/gateway=true" --overwrite
 . Create a public IP and assign it to the VM of the node that was tagged as gateway node by running the following commands:
 +
 ----
-az network public-ip create --name <public-ip-name> --resource-group <res-group>
+az network public-ip create --name <public-ip-name> --resource-group <res-group> -sku <sku-type>
```
sridhargaddam: Do we need similar change in `subctl cloud prepare azure ...`?

Also, is this a mandatory parameter?

sridhargaddam: Assuming possible values for "sku-type" is basic and standard. Does it work if "\<sku-type\>" is configured as "basic"? If not, I think the example should use "standard" instead of expecting the user to figure it out.

sridhargaddam: Also, can you please include some notes in the commit message why this change is necessary.

aswinsuryan: We already use Standard in cloud-prepare. 

THe Openshift installer always creates the Load Balancer with type SKU Standard and public IP needs to be of the same type. So I changed as you suggested to use ` --sku Standard`

sridhargaddam:
thanks for addressing the comments @aswinsuryan 

/lgtm

add-ons/submariner/submariner_prepare_hosts.adoc
```diff
@@ -65,6 +65,9 @@ az network nsg rule create --resource-group <res-group> \
 --protocol udp --destination-port-ranges <vxlan-port>
 ----
 
+**Important:** Verify that the gateway node created above is deleted and a new one is created when trying a reinstall of Submariner. Reusing an existing gateway after uninstalling of submariner can lead the connections to go
```
sridhargaddam: > Verify that the gateway node created above is deleted and a new one is created when trying a reinstall of Submariner

Since we are not creating a new dedicated Gateway but re-using an existing worker node, shall we include a message saying that a different Gateway node should be used when re-installing Submariner?

>Reusing an existing gateway after uninstalling of submariner can lead the connections to go to error state.

Reusing an existing gateway after uninstalling of submariner can lead the connections failures.

Also, shall we include a message saying that this issue is only applicable for 2.5 (i.e., {product-title}) where manual cloud preparation steps are used and is addressed in the next release?

aswinsuryan: oh I forgot about it, changed the wording. Add a note about release as well.

sridhargaddam:
Thanks @aswinsuryan for addressing the comments.

/lgtm

cadawson:
Hi @aswinsuryan and @sridhargaddam. Made some minor editing changes. Please make sure that I didn't inadvertently change anything. 

Also, is this 2.5-specific? 

Thanks!

aswinsuryan:
> Hi @aswinsuryan and @sridhargaddam. Made some minor editing changes. Please make sure that I didn't inadvertently change anything.
> 
> Also, is this 2.5-specific?
> 
> Thanks!

yes this is 2.5 specific

PR #4087, opened 2022-09-14
Clarify Submariner steps 2.5

oafischer:
https://github.com/stolostron/backlog/issues/25666

Dev approval: https://github.com/stolostron/backlog/issues/25666#issuecomment-1248305475

### DIFF
--- a/add-ons/submariner/submariner_deploy_mcaddon_api.adoc
+++ b/add-ons/submariner/submariner_deploy_mcaddon_api.adoc
@@ -68,7 +68,9 @@ After the `ManagedClusterAddOn` is created, the `submariner-addon` deploys Subma
 +
 *Note:* The name of `ManagedClusterAddOn` must be `submariner`.
 
-. Repeat the steps from xref:../submariner/submariner_prepare_hosts.adoc#preparing-selected-hosts-to-deploy-submariner[Preparing selected hosts to deploy Submariner] for your hosting environment.
+. Repeat steps three and four for all of the managed clusters that you want to enable Submariner on.
++ 
+*Note:* Make sure your clusters are already prepared before repeating steps three and four.
 
 . After Submariner is deployed on the managed cluster, you can verify the Submariner deployment status by checking the status of submariner `ManagedClusterAddOn` by entering the following command: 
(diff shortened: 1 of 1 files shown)


dfarrell07 (changes requested):
I don't think this is correct, see https://github.com/stolostron/backlog/issues/25666#issuecomment-1248260857

oafischer:
Thanks @dfarrell07, I saw your comment. I will leave this open until we get confirmation on what clarification we actually need. Thanks again for bringing this up with your team!

tpantelis (approved):

add-ons/submariner/submariner_deploy_mcaddon_api.adoc
```diff
@@ -69,6 +69,8 @@ After the `ManagedClusterAddOn` is created, the `submariner-addon` deploys Subma
 *Note:* The name of `ManagedClusterAddOn` must be `submariner`.
 
 . Repeat steps three and four for all of the managed clusters that you want to enable Submariner on.
++ 
+*Note:* Make sure your clusters are already prepared before repeating steps three and four.
```
dfarrell07: Maybe this should be at the top of `1.1.4.2.2. Deploy Submariner with the ManagedClusterAddOn API`? In this case the user hit issues at this step, but in general starting the deploy before preparing the clusters is bug-prone.

> *Note*: Make sure you complete 1.1.4.2.1. Preparing selected hosts to deploy Submariner(link) before starting this section.

oafischer: If we move it there it might even make sense to just create a prereq section and list 1.1.4.2.1 as a requirement. Those are also harder to skip/miss than a special tag. Let me implement and then review with my team.

oafischer: @dfarrell07 I just injected the information into the intro since that worked better. Let me know what you think.

dfarrell07 (approved):
Yes, that looks great thank you @oafischer!

tpantelis (approved):

PR #4105, opened 2022-09-16
Add GitHub issue point guide

oafischer:
https://github.com/stolostron/backlog/issues/25736

DRAFT

Still requires more content and input from team members. Please fill out some examples in the higher point ranges because I haven't had many (or any) issues with points that high up yet.

The rich text isn't displayed properly so `checkout` the branch and run `bundle exec jekyll serve` to see a preview of the site.

If you need Jekyll, see https://github.com/stolostron/rhacm-docs/blob/gh-pages/README.md#building-with-jekyll

### DIFF
--- a/_pages/arch_points.md
+++ b/_pages/arch_points.md
@@ -46,7 +46,7 @@ The most basic type of issue usually consisting of fixing typos and adding or re
 
 ## `2` points
 
-A simple issue that requires mostly copy and paste. These are often straightforward new known issues being requested by developers, complete with a full and mostly edited draft. At most you might need to remind yourself of a style item or rearrange some sentences, but the information is complete and ready to be implemented in your solution.
+A simple issue that requires mostly copy and paste. These are often straightforward new known issues being requested by developers, complete with a full and mostly edited draft. You might need to remind yourself of a style item or rearrange some sentences, but the information is complete and ready to be implemented in your solution.
 
 | **Rubric** | | **`2` points description** |
 | ----------- | | ----------- |
--- a/_pages/arch_points.md
+++ b/_pages/arch_points.md
@@ -166,7 +166,7 @@ An extremely complex issue that requires massive work on all fronts and input fr
 | Difficulty | | Complex authoring, editing, and decision-making on content that might not exist yet. You need mastery of the doc architecture to make the right decisions. | 
 | Research | | Extensive research on all fronts, from style to technical concepts and customer needs. Information comes from a multitude of sources. Recurring meetings with product managers, SMEs, developers, and team leaders are required. | 
 | Length | | Up to a whole book in size (massive input in hundreds of files). | 
-| Review | | Multiple peer, developer, team leader, and proudct manager reviews through meetings and via GitHub. |
+| Review | | Multiple peer, developer, team leader, and product manager reviews through meetings and via GitHub. |
 
 &nbsp;
 
--- a/_pages/arch_points.md
+++ b/_pages/arch_points.md
@@ -1,0 +1,315 @@
+---
+layout: page
+title: GitHub issue points
+permalink: arch_points
+horizontal: false
+---
+
+Not all issues are created equal. To help plan our anticipated workloads for each sprint and distribute issues efficiently, we assign points to GitHub issues that provide an estimate for how difficult and time-intensive each issue is. While assigning points is a subjective process, there are still some guidelines you can follow.
+
+## General rules
+
+- The range of points conists of integers from `1` to `10`, with `1` being "I can knock this out in 5 minutes" and `10` being "this will give me grey hairs." Do not use fractions or integers above 10.
+- Estimate points when triaging issues, but feel free to change them any time during your work process, including after you have already closed the issue.
+
+## Rubrics
+
+Consider the following rubrics when assinging points to GitHub issues:
+
+| **Rubric** | | **Description** |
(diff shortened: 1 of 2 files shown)


_pages/arch_points.md
```diff
@@ -0,0 +1,223 @@
...
+- The range of points conists of integers from `1` to `10`, with `1` being "I can knock this out in 5 minutes" and `10` being "this will give me grey hairs." Do not use fractions or integers above 10.
+- Estimate points when triaging issues, but feel free to change them any time during your work process, including after you have already closed the issue.
+
+## Rubrics
+
+Consider the following rubrics when assinging points to GitHub issues:
+
+| **Rubric** | | **Description** |
+| ----------- | | ----------- |
+| Time | | How long it takes to solve an issue. |
```
cadawson, line 22: We don't want to base the point system on time because different people might require different amounts of time to complete a similar task. Instead we want to base it on the complexity of the task. I understand that sometimes time and complexity can be pretty closely related, though.

_pages/arch_points.md
```diff
@@ -0,0 +1,223 @@
...
+&nbsp;
+
+## `9` points
+
+Draft
+
+| **Rubric** | | **`9` points description** |
+| ----------- | | ----------- |
+| Time | | Requires multiple sprints. |
+| Difficulty | | Draft |
```
cadawson: Maybe: Complex authoring of new content that is still being decided. It will likely require more effort to try to work these issues because you have to deal with frequent changes, You need to have strong awareness of the doc architecture to make the right decisions.

_pages/arch_points.md
```diff
@@ -0,0 +1,223 @@
...
+
+## `9` points
+
+Draft
+
+| **Rubric** | | **`9` points description** |
+| ----------- | | ----------- |
+| Time | | Requires multiple sprints. |
+| Difficulty | | Draft | 
+| Research | | Draft |
```
cadawson: Requires lengthy research to understand the content and explain it. Often requires putting together information from multiple sources.

cadawson (commented):
Thanks for all of the work that you did on this, Oliver. It looks good. I like the descriptions that provide guidance for assigning points consistently. I wonder whether it is too specific, or if we can make sure that we stress that this is guidance. I know you mention that it is ultimately up to you to make the decision, but I don't want someone to feel like they can't assign it if it doesn't meet all of the criteria. Just a thought.

_pages/arch_points.md
oafischer: Thanks for bringing that up, it's a good point. I definitely understand that some issues can vary drastically, often due to things out of our control. I'm interested to hear what everyone else thinks about this as well (@swopebe @dockerymick). I think from a board perspective it can be helpful to have some form of time frame to work with. For example, if an issue is being worked on for several sprints but is 2 points, then it's easy to spot a potential process problem/inefficiency. Even if it's as simple as "the person isn't responding." (Think DDF) Might also help prevent stale bots.

I definitely hear what you are saying though. My suggestion based on your feedback would be to keep it as a rubric, but to make it even vaguer. So less of an indicator of individual work pace and more of a complexity indicator. So something like "quick" for a 1 point issue or "relatively quick" for a 2 point issue and then maybe move to sprint-centric timing (less than a sprint, more than a sprint, and full release, the latter of which also varies every time).

_pages/arch_points.md
oafischer: That's good insight, thanks! Frequent changes as things were still being decided hadn't crossed my mind.

oafischer:
Finished adding examples, I want to merge this soon but I want everyone on the team to have seen this since it's a collaborative effort. Also want more input on the time rubric. Should we keep it with the updates I made based on feedback from Chris? Or still remove entirely? I'm fine with either.

_pages/arch_points.md
swopebe: I do think we should have time _mentioned_ because honestly a known issue that has a complete draft should not take more than an hour. An issue to redo a whole section will take numerous sprints. 

Those can be the baseline of the time piece, though it is not the only thing to consider. This will help when sharing work, as well.

I have done story point training on other teams and we did use time as a guideline, but we based it off **what information is given to the team**, which I think is fair. I have spoken to senior managers about time/efficiency issues and while we don't want to pressure folks to do a known issue within an hour, we do want to _identify_ things that are preventing efficiency with easy, short issues. It is realistic to strive for efficiency in time/effort, I think.

For instance, if you have a lot of 1, 2 point issues that are getting done but no 6,7 issues, it can be because you have too many meetings and no time blocked (4 hours, say) to research and address those issues. If you are constantly interrupted and not able to work a 2 point issue, what are the interruptions and how can we eliminate them? (Some examples.)

_pages/arch_points.md
swopebe: This is great...thanks so much. I left a comment about "time" and only have one more comment:

Even with 1, 2 copy/paste issues, we all should stop and consider the concept of the content: Why are we writing this.

The reasons?

We catch potential for architectural problems (we can identify that this may not be in the right spot, we consider user questions, we organize according to those factors)

We ensure we know more about the product with each update we are making

swopebe (approved):

dockerymick (approved):
Great job Oliver! i appreciate you doing this and identifying examples for each point. Well done!

PR #4110, opened 2022-09-21
Adding Application known issue

dockerymick:
https://github.com/stolostron/backlog/issues/25467

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -815,6 +815,19 @@ The `allow` and `deny` list feature does not work in Object storage application 
 
 The `ApplicationSet` topology status icon spins continuously if an `ApplicationSet` application is deployed, but has no associated Argo applications.
 
+[#unsupported-{ocp-short}-versions-listed-after-upgrade]
+=== Unsupported {ocp-short} versions listed after hub cluster upgrade
+// 2.6:25467
+
+After you upgrade your hub cluster from versions before {previous-version} to {product-version}, some unsupported {ocp-short} versions are listed on the _Cluster_ page from the console.
+
+//i assume orphan is similar to 'stale'. What are your thoughts about using stale instead of orphan?
+The orphan `clusterImageSet` resources deployed by earlier versions before {previous-version} subscription controller are not deleted after the upgrade. To resolve this, manually delete the `clusterImageSet` resources that have unsupported {ocp-short} versions. For example, run the following command to delete the `img4.7.0-x86-64-appsub` `clusterImageSet`:
+
+----
+oc delete clusterimageset img4.7.0-x86-64-appsub
+----
+
 [#governance-known-issues]
 == Governance known issues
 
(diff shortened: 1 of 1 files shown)


xiangjingli (approved):
It looks good to me. And using stale instead of orphan is better.

release_notes/known_issues.adoc
```diff
@@ -815,6 +815,18 @@ The `allow` and `deny` list feature does not work in Object storage application
 
 The `ApplicationSet` topology status icon spins continuously if an `ApplicationSet` application is deployed, but has no associated Argo applications.
 
+[#unsupported-{ocp-short}-versions-listed-after-upgrade]
```
cadawson: I don't think we use conrefs in the anchors. I know we have been successful using them in the titles, but I have not seen it in an anchor. I worry that if we start to do that, we might link to it with a hard-coded link, and it might not resolve. Also, I am not sure whether the conrefs are processed before or after the links are created. I would recommend removing the conref from the anchor.

release_notes/known_issues.adoc
```diff
@@ -815,6 +815,18 @@ The `allow` and `deny` list feature does not work in Object storage application
...
+=== Unsupported {ocp-short} versions listed after hub cluster upgrade
+// 2.6:25467
+
+After you upgrade your hub cluster from versions before {previous-version} to {product-version}, some unsupported {ocp-short} versions are listed on the _Cluster_ page from the console.
```
cadawson: in the console? 

Also, just a note about using the {previous-version} and {product-version} here: Keep in mind that if this known issue continues to the next version (which we know happens more than sometimes), those versions will be bumped to 2.6 and 2.7, respectively. I would guess that we might not want that to happen, in this case. It looks like this would be specific to versions before 2.5 being upgraded to 2.6. If it is a consistent issue that applies to all versions (which calls into question whether it should be a known issue), then we are ok. If it applies specifically to these versions, then I might go with the hardcoded 2.5 and 2.6.

We had a similar situation with a bare metal that was an issue in 2.1 or something like that. It was not removed, and someone asked why we couldn't use it in 2.5. The problem was specific to 2.1, but it just kept getting updated as we moved because of the conref.

release_notes/known_issues.adoc
```diff
@@ -815,6 +815,18 @@ The `allow` and `deny` list feature does not work in Object storage application
...
+
+The stale `clusterImageSet` resources deployed by earlier versions before {previous-version} subscription controller are not deleted after the upgrade. To resolve this, manually delete the `clusterImageSet` resources that have unsupported {ocp-short} versions. For example, run the following command to delete the `img4.7.0-x86-64-appsub` `clusterImageSet`:
```
cadawson: The stale `clusterImageSet` resources that are/were deployed by versions earlier than {previous-version} subscription controller are/were not deleted after the upgrade.

cadawson (changes requested):
Looks good, but added a couple of considerations. I don't need to see it again. Thanks!

release_notes/known_issues.adoc
dockerymick: Thanks Chris, you are correct about how the statement is displayed w/the conref. I will hardcode the versions for now.

dockerymick:
@cadawson @xiangjingli Thanks to you both for your reviews and approval. Moving forward with this merge

PR #4111, opened 2022-09-21
Updating clone command

dockerymick:
https://github.com/stolostron/backlog/issues/25546


dockerymick:
Since this is a small change and only required command replacement, I will move forward with merging this PR.

PR #4118, opened 2022-09-22
No issue, scrum discussion

swopebe:
Adding info to release note about support.

### DIFF
--- a/release_notes/release_notes.adoc
+++ b/release_notes/release_notes.adoc
@@ -3,7 +3,7 @@
 
 Learn about the current release. 
 
-*Note:* The 2.2 and earlier versions of {product-title-short} are _removed_ and no longer supported. Documentation for versions 2.2 and earlier is not updated. The documentation might remain available, but is deprecated without any Errata or other updates available. 
+*Note:* The 2.2 and earlier versions of {product-title-short} are _removed_ from service, and are no longer supported. Documentation for versions 2.2 and earlier is not updated. The documentation might remain available, but is deprecated without any Errata or other updates available. 
 
 * xref:../release_notes/whats_new.adoc#whats-new[What's new in {product-title}]
 * xref:../release_notes/errata.adoc#errata-updates[Errata updates]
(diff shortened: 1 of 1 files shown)


release_notes/release_notes.adoc
```diff
@@ -1,11 +1,17 @@
...
+*Note:* The 2.2 and earlier versions of {product-title-short} are _removed_ from service, and are no longer supported. Documentation for versions 2.2 and earlier is not updated. The documentation might remain available, but is deprecated without any Errata or other updates available. 
 
 * xref:../release_notes/whats_new.adoc#whats-new[What's new in {product-title}]
 * xref:../release_notes/errata.adoc#errata-updates[Errata updates]
 * xref:../release_notes/known_issues.adoc#known-issues[Known issues and limitations]
 * xref:../release_notes/deprecate_remove.adoc#deprecations-removals[Deprecations and removals]
 * xref:../release_notes/gdpr_readiness.adoc#red-hat-advanced-cluster-management-for-kubernetes-platform-considerations-for-gdpr-readiness[{product-title} considerations for GDPR readiness]
 * xref:../release_notes/fips_readiness.adoc#fips-readiness[FIPS readiness]
+
+To report issues with one of the currently supported releases or its documentation, open a https://bugzilla.redhat.com/enter_bug.cgi?product=Red%20Hat%20Advanced%20Cluster%20Management%20for%20Kubernetes[support ticket].
```
cadawson: The original link only allowed me to complete an issue against the Customer Portal. I hope that is not a permissions issue that I don't have, but others will also have problems if it is. I replaced with a link to BZ. We might have to change it to Jira at some point. 

I am asking Josh Carter if he knows about any problems with that.

Any thoughts about putting it in a level 2 heading titled "Support" or "Contacting support"?

cadawson (changes requested):
Updated with comments. Let me know if you have questions. Thank you for working on this!

release_notes/release_notes.adoc
swopebe: @cadawson the link was to the support docs, we don't want to link just to BZ when there are other avenues... that is why I chose that link. I recommend we link them to the higher support authority. We would not put contacting support here, but that can go in the Troubleshooting spot.

swopebe: I will change this to the higher support contact.

swopebe: We don't use possessives in tech doc as a rule.

swopebe: Please see changes...we don't want to bypass all questions to just a BZ...the user can get to the case open tab from that spot but also has access to other documentation that may help them first.

cadawson (approved):
Looks good! Thanks!

PR #4123, opened 2022-09-23
removed incorrect conref

dockerymick:
https://github.com/stolostron/backlog/issues/25467


dockerymick:
quick change after running build

PR #4137, opened 2022-09-26
Updating credentials topic 

dockerymick:
[Create an Azure service principal with the Azure CLI](https://github.com/stolostron/backlog/issues/25519)


Randy424 (approved):
/lgtm

cadawson (changes requested):
Looks good. I suggest changing the link from "Microsoft Azure documentation" to the topic name. If you are concerned about going to that depth of the docs because they might move it (which is not unrealistic), you can list the topic name and just link to the highest level of the Azure docs and let them find it. If you go that route, leave the link where it is, but change it to the overall docs. No need to send it back. Thanks!

dockerymick:
Thanks for the reviews and feedback @Randy424 @cadawson . I have made changes according to chris' suggestion. Merging this

PR #4139, opened 2022-09-26
Enable assisted on AWS in 2.5

oafischer:
https://github.com/stolostron/backlog/issues/25693

### DIFF
--- a/clusters/create_infra_env.adoc
+++ b/clusters/create_infra_env.adoc
@@ -5,6 +5,7 @@ You can use the {product-title} console to create an infrastructure environment 
 
 * <<infra-env-prerequisites,Prerequisites>>
 * <<enable-cim,Enable Central Infrastructure Management service>>
+** <<enable-cim,Enable Central Infrastructure Management on Amazon Web Services>>
 * <<creating-your-infra-env-with-the-console,Creating your infrastructure environment with the console>>
 * <<accessing-your-infra-env,Accessing your infrastructure environment>>
 
--- a/clusters/create_infra_env.adoc
+++ b/clusters/create_infra_env.adoc
@@ -173,7 +173,7 @@ oc get routes --all-namespaces | grep assisted-image-service
 Your domain might resemble the following example:
 `assisted-image-service-multicluster-engine.apps.<yourdomain>.com`
 
-. Add <yourdomain> to the `domain` parameter in `IngressController` by replacing `<domain>` in `nlb-apps.<domain>.com` with `<yourdomain`.
+. Add <yourdomain> to the `domain` parameter in `IngressController` by replacing `<domain>` in `nlb-apps.<domain>.com` with `<yourdomain>`.
 
 . Run the following command to edit the `assisted-image-service` route to use the `nlb-apps` location:
 +
--- a/clusters/create_infra_env.adoc
+++ b/clusters/create_infra_env.adoc
@@ -190,7 +190,7 @@ oc edit route assisted-image-service -n <namespace>
     router-type: nlb
 ----
 
-. In the `assisted-image-service` route, find the URL value of `spec.hosts`. It might resemble the following example:
+. In the `assisted-image-service` route, find the URL value of `spec.hosts`. The URL might resemble the following example:
 +
 `assisted-image-service-multicluster-engine.apps.<yourdomain>.com`
 
--- a/clusters/create_infra_env.adoc
+++ b/clusters/create_infra_env.adoc
@@ -135,6 +135,62 @@ agentserviceconfig.agent-install.openshift.io/agent created
 
 Your CIM service is configured. You can verify that it is healthy by checking the `assisted-service` and `assisted-image-service` deployments and ensuring that their pods are ready and running. Continue with xref:../clusters/create_infra_env.adoc#creating-your-infra-env-with-the-console[Creating your infrastructure environment with the console].
 
+[#enable-cim-aws]
+=== Enabling CIM on Amazon Web Services
+
+If you are running CIM on Amazon Web Services, complete the following additional steps after  <<enable-cim,Enabling CIM>>:
+
+. Create a new `IngressController` with a unique domain using the `NLB` `type` parameter:
++
+[source,yaml]
+----
+apiVersion: operator.openshift.io/v1
+kind: IngressController
+metadata:
+  name: ingress-controller-with-nlb
+  namespace: openshift-ingress-operator
+spec:
+  domain: nlb-apps.<domain>.com
+  routeSelector:
+      matchLabels:
+        router-type: nlb
+  endpointPublishingStrategy:
+    type: LoadBalancerService
+    loadBalancer:
+      scope: External
+      providerParameters:
+        type: AWS
+        aws:
+          type: NLB
+----
+
+. Find the unique domain configured on the assisted-image-service by running the following command:
++
+----
+oc get routes --all-namespaces | grep assisted-image-service
+----
++
+Your domain might resemble the following example:
+`assisted-image-service-multicluster-engine.apps.<yourdomain>.com`
+
+. Add <yourdomain> to the `domain` parameter in `IngressController` by replacing `<domain>` in `nlb-apps.<domain>.com` with `<yourdomain`.
+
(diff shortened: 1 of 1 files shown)


clusters/create_infra_env.adoc
```diff
@@ -135,6 +136,62 @@ agentserviceconfig.agent-install.openshift.io/agent created
...
+
+. Add the following lines to the metadata in the `assisted-image-service` route:
++
+[source,yaml]
+----
+  labels:
+    router-type: nlb
+----
+
+. In the `assisted-image-service` route, replace `spec` with `hosts` and replace `apps` with `nlb-apps`.
```
jelkosz: As it is written, it sounds a bit confusing to me. It sounds like you need to replace spec and hosts somehow. 

Id suggest to write something like:
In the assisted-image-service route, find the URL value of "spec.hosts". It will look a something like this:
assisted-image-service-multicluster-engine.apps.<yourdomain>.com
In this URL replace the "apps" by "nlb-apps" to match the domain configured in the IngressController above.

oafischer:
Implemented the feedback, waiting for writing review as well.

jelkosz:
lgtm

cadawson (approved):
Looks good. Thanks!

clusters/create_infra_env.adoc
```diff
@@ -135,6 +136,66 @@ agentserviceconfig.agent-install.openshift.io/agent created
 
 Your CIM service is configured. You can verify that it is healthy by checking the `assisted-service` and `assisted-image-service` deployments and ensuring that their pods are ready and running. Continue with xref:../clusters/create_infra_env.adoc#creating-your-infra-env-with-the-console[Creating your infrastructure environment with the console].
 
+[#enable-cim-aws]
+=== Enabling Central Infrastructure Management on Amazon Web Services
+
+If you are running CIM on Amazon Web Services, complete the following additional steps after  <<enable-cim,Enabling CIM>>:
```
berenss, line 142: The customer does not really know that they are running CIM. I mean yeah the section above describes CIM but really what we're meaning here is 'if you're running hub on AWS'

clusters/create_infra_env.adoc
```diff
@@ -135,6 +136,66 @@ agentserviceconfig.agent-install.openshift.io/agent created
...
+
+. Create a new `IngressController` with a unique domain using the `NLB` `type` parameter. See the following example:
++
```
berenss, line 145: when I reviewed this step I immediately thought: ok so right away I need to create this object from yaml. am I supposed to oc apply this yaml or do something else?  So I need to login to the hub first and prepare a yaml file. Then as I read the next steps, you don't really have all of the information to make this YAML file ready until like step 4 when you've got the `<domain>` sorted out. 
So I think move this step lower and ask the user to ensure they are logged in at the hub. create the file, then oc apply -f the-file

clusters/create_infra_env.adoc
```diff
@@ -135,6 +136,66 @@ agentserviceconfig.agent-install.openshift.io/agent created
...
+Your domain might resemble the following example:
+`assisted-image-service-multicluster-engine.apps.<yourdomain>.com`
+
+. Add `<yourdomain>` to the `domain` parameter in `IngressController` by replacing `<domain>` in `nlb-apps.<domain>.com` with `<yourdomain>`.
+
+. Run the following command to edit the `assisted-image-service` route to use the `nlb-apps` location:
++
+----
+oc edit route assisted-image-service -n <namespace>
+----
```
berenss, line 183: we can give a hint here that the namespace is where they installed the multicluster engine, or multicluster-engine as default.

clusters/create_infra_env.adoc
```diff
@@ -135,6 +136,66 @@ agentserviceconfig.agent-install.openshift.io/agent created
...
+. Add the following lines to the metadata in the `assisted-image-service` route:
++
+[source,yaml]
+----
+  labels:
+    router-type: nlb
+----
+
+. In the `assisted-image-service` route, find the URL value of `spec.hosts`. The URL might resemble the following example:
++
```
berenss, line 194: for me it was `spec.host` not `spec.hosts`

clusters/create_infra_env.adoc
```diff
@@ -135,6 +136,66 @@ agentserviceconfig.agent-install.openshift.io/agent created
...
+oc edit route assisted-image-service -n <namespace>
+----
+
+. Add the following lines to the metadata in the `assisted-image-service` route:
++
+[source,yaml]
+----
+  labels:
+    router-type: nlb
+----
```
berenss, line 191: for a n00b like me this made me nervous. There are not any existing labels in the `assisted-image-service` so I wasn't exactly sure where to add it.

clusters/create_infra_env.adoc
```diff
@@ -135,6 +136,66 @@ agentserviceconfig.agent-install.openshift.io/agent created
...
+
+. In the `assisted-image-service` route, find the URL value of `spec.hosts`. The URL might resemble the following example:
++
+`assisted-image-service-multicluster-engine.apps.<yourdomain>.com`
+
+. Replace `apps` in the URL with `nlb-apps` to match the domain configured in the new `IngressController`.
+
```
berenss, line 198: anything I should do here at the end to check/verify? I looked at the pods and saw the 2, 
- assisted-image-service-0  1/1  
- assisted-service-blah-blah  2/2

come up to running after about 2 mins.

oafischer:
Thanks for your comments @berenss! Since this is already merged, would you mind creating a new issue? No need to spend much time copy-pasting your comments, just link to this PR in the description. That way, I can also pull in @jelkosz again and we can work on getting this section up to scratch. Thanks again!

clusters/create_infra_env.adoc
oafischer: Does keeping the current topic title still make sense? Or should we change that too?

clusters/create_infra_env.adoc
oafischer: @jelkosz Any insight here?

clusters/create_infra_env.adoc
oafischer: Pinging @jelkosz again for extra insight.

jelkosz: What I typically do is to create a new infra env and make sure the download url is using this new URL.
@filanov anything better you can think of?

clusters/create_infra_env.adoc
jelkosz: it needs to go to metadata. This is a little larger snippet:
```
...
metadata:
  labels:
    router-type: nlb
  name: assisted-image-service
...
```

oafischer: Thanks @jelkosz. Since metadata seems to be a parameter, I'll monospace it and hopefully it should be clearer from the previous step. Or I just use the example above. Which makes more sense to you @berenss?

berenss:
opened! https://github.com/stolostron/backlog/issues/26095

clusters/create_infra_env.adoc
berenss: I guess it does make sense as the section title - we're introducing them about CIM as 'an extra thing to turn on'
I think it's just the way that first sentence is written, it should reinforce that what we're really pointing to is if their hub is running on AWS.
"If you are running your hub on Amazon Web Services and wish to enable the Central Infrastructure Management service," ...

clusters/create_infra_env.adoc
berenss: Using the new example snippet from Tomas makes sense to me; this provides more context about the new entry and where it should go. thx!

clusters/create_infra_env.adoc
berenss: Yes- lets advise them to do both things
check the pods are up and healthy,

> oc get pods -n multicluster-engine |grep assist
> assisted-image-service-0                                1/1     Running   0             115s
> assisted-service-6647dbfdc4-p5nqh                       2/2     Running   0             116s

And, then create a new infrastructure environment and ensure the new download URL is reflected with the `nlb-apps` in it.

PR #4164, opened 2022-10-04
Add kubeconfig troubleshooting in 2.6

oafischer:
https://github.com/stolostron/backlog/issues/25922


troubleshooting/main.adoc
```diff
@@ -8,6 +8,7 @@ include::trouble_install_status.adoc[leveloffset=+2]
 include::trouble_reinstall.adoc[leveloffset=+2]
 include::trouble_cluster_offline.adoc[leveloffset=+2]
 include::trouble_cluster_import_fails.adoc[leveloffset=+2]
```
cadawson, line 10: This might need to go into the mce docs, rather than/in addition to RHACM. @skeeey is out until this weekend, so he can let us know where it makes sense. 

We try to avoid including the product name in the title, as it is pretty long. If there is a reason for not using something like "Reimporting cluster fails with unknown authority error," then it is ok. If not, try to avoid the product title. If this belongs in MCE, then we need to make sure that those are updated. Thanks!

troubleshooting/trouble_cluster_reimport_fails.adoc
```diff
@@ -0,0 +1,81 @@
...
++
+----
+oc get ns
+----
+
+If you receive an error that resembles the following message, your cluster API server ceritificates have been changed and your `kubeconfig` file is invalid.
+
+`Unable to connect to the server: x509: certificate signed by unknown authority`
+
+[#resolving-the-problem-openstack-fails]
```
cadawson: This might be a copy-not-changed ID. If it is, it will probably result in an error when you build. I'm guessing you intended something like: #resolving-the-problem-cluster-reimport-kubeconfig?

troubleshooting/trouble_cluster_reimport_fails.adoc
```diff
@@ -0,0 +1,81 @@
...
+== Resolving the problem: Reimporting cluster fails with unknown authority error
+
+The managed cluster admin must create a new valid `kubeconfig` file for your managed cluster.
```
cadawson, line 61: Administrator? If you are referring to a role, it's good.

cadawson (changes requested):

troubleshooting/trouble_cluster_reimport_fails.adoc
oafischer: Indeed! I actually already changed it after noticing, it may have been showing you an old version for some reason.

troubleshooting/trouble_cluster_reimport_fails.adoc
oafischer: @skeeey Can you clarify?

troubleshooting/trouble_cluster_reimport_fails.adoc
cadawson: It looks like there might still be one that did not get updated. It is still showing up when I view the changed file, and is being flagged by the linter. Please take a look. Thanks!

oafischer: Hopefully that did the trick!

troubleshooting/trouble_cluster_reimport_fails.adoc
skeeey: yeah, here should be Administrator, not referring to a role

troubleshooting/main.adoc
skeeey: yeah, this should be in the mce, because this is about cluster importing

skeeey: @cadawson @oafischer 

And I think it's better to add this to acm 2.5 and 2.4, because this problem may also occur in these versions

oafischer: Thanks for clarifying @skeeey, I'll switch 2.6 over to MCE and add it to 2.5 and 2.4 as well.

oafischer:
I'll close this and start over, seems easier.

PR #4166, opened 2022-10-05
Add port 8081 in 2.6

oafischer:
https://github.com/stolostron/backlog/issues/26128

### DIFF
--- a/add-ons/submariner/submariner.adoc
+++ b/add-ons/submariner/submariner.adoc
@@ -10,6 +10,7 @@ See the link:https://access.redhat.com/articles/6968787[{product-title-short} Su
 * <<submariner-prereq,Prerequisites>>
 * <<submariner-subctl,subctl command utility>>
 * <<submariner-globalnet,Globalnet>>
+
 * xref:../submariner/submariner_deploy_over.adoc#submariner-deploy-over[Deploying Submariner]
 ** xref:../submariner/submariner_deploy_console.adoc#submariner-deploy-console[Deploying Submariner with the console]
 ** xref:../submariner/submariner_deploy_manual.adoc#deploying-submariner-manually[Deploying Submariner manually]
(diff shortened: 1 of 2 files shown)


add-ons/submariner/submariner.adoc
```diff
@@ -54,6 +54,10 @@ Ensure that you have the following prerequisites before using Submariner:
 | Submariner metrics port
 | 8080/TCP
 | No
+
+| Globalnet port
```
vthapar:
```suggestion
| Globalnet Metrics port
```

add-ons/submariner/submariner.adoc
```diff
@@ -54,6 +54,10 @@ Ensure that you have the following prerequisites before using Submariner:
...
+| 8081/TCP
+| No
```
vthapar, line 67: Can we add an asterix or something? Need to indicate that this is only applicable if Globalnet is enabled.

vthapar (changes requested):

add-ons/submariner/submariner.adoc
oafischer: I think with some restructuring and an additional column we can make it work, let me make some changes real quick.

swopebe: Please remove what is causing bold in the source (likely an asterisk or code snippet?) 
The source should match the production, that is why we use asterisks and such sparingly, it can change the source somewhere before the table.

add-ons/submariner/submariner.adoc
```diff
@@ -39,21 +39,29 @@ Ensure that you have the following prerequisites before using Submariner:
 - Start with an alphanumeric character
 - End with an alphanumeric character
 
-.Submariner required ports
+.Submariner ports
```
swopebe: I am not sure that this is the way we introduce a table, and that code would be causing the bold source. I see many things wrong with how this topic is built but will send an issue to @cadawson and raise the issues in Slack.

swopebe: If you can fix the code on this snippet, I will approve.

swopebe (approved):

oafischer:
Merging since the one dev comment that was left has been addressed.

swopebe:
Reminder for myself to open an issue to 1. look at table titles and intros and 2. rebuild if source is still showing bold and find out why.

PR #4213, opened 2022-10-18
Update VXLAN best practice in 2.7

oafischer:
https://github.com/stolostron/backlog/issues/26430

Also let me know if this needs to be backported (this best practice exists in 2.4, 2.5, and 2.6). Thanks!

### DIFF
--- a/add-ons/submariner/submariner_customizations.adoc
+++ b/add-ons/submariner/submariner_customizations.adoc
@@ -154,4 +154,4 @@ spec:
    
 ----
 
-*Best practice:* Only use unencrypted VXLAN connections in on-premises enviroments.
+*Best practice:* Do not use the unencrypted VXLAN cable driver on public networks. Only use VXLAN to avoid unnecessary double encryption on private networks. For example, some on-premise environments handle the tunnel's encryption with a dedicated hardware device.
--- a/add-ons/submariner/submariner_customizations.adoc
+++ b/add-ons/submariner/submariner_customizations.adoc
@@ -154,4 +154,4 @@ spec:
    
 ----
 
-*Best practice:* Do not use the unencrypted VXLAN cable driver on public networks. Only use VXLAN to avoid unnecessary double encryption on private networks. For example, some on-premise environments handle the tunnel's encryption with a dedicated hardware device.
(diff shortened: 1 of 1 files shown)


add-ons/submariner/submariner_customizations.adoc
```diff
@@ -154,4 +154,4 @@ spec:
    
 ----
 
-*Best practice:* Only use unencrypted VXLAN connections in on-premises enviroments.
+*Best practice:* Do not use the unencrypted VXLAN cable driver on public networks. Only use VXLAN to avoid unnecessary double encryption on private networks. For example, some on-premise environments might handle the tunnel's encryption with a dedicated hardware device.
```
dfarrell07: This is okay, but I wonder if we can be a bit stricter.

For example, a user might read "the unencrypted VXLAN cable driver" and think "is there an encrypted VXLAN?" or "if I use this other config is VXLAN encrypted?". We know that's not true, but that's basically what users have said before. My suggestion ("The VXLAN cable drive is unencrypted") doesn't leave any ambiguity.

oafischer: Good call, especially if you've heard that from users before. I'll go ahead and change it.

cadawson (approved):
Looks good. Thanks!

add-ons/submariner/submariner_customizations.adoc
dfarrell07: One other way we should be strict to mitigate possible user misinterpretation:

It's normal for application-level encryption (L7) to be nested in IP-level encryption (L3). That doesn't count as double encryption. So we could say "with a dedicated line-level hardware device" and/or add a sentence like "Note that it's normal and desirable for application-level L7 encryption to run nested in encrypted L3 tunnels".

dfarrell07: (sorry to come back with yet another iteration! you're just so fast lol)

oafischer: No worries! We love getting feedback and input, it helps us build better docs which in turn is better for the user. Always happy to hear suggestions at any stage. We don't want too many notes though, so I'll go with your more specific line-level hardware suggestion. Thanks!

dfarrell07 (approved):
Looks great!

PR #4220, opened 2022-10-20
Updates for TOC in Backup and restore

dockerymick:
https://github.com/stolostron/backlog/issues/25895

After reviewing the stage environment, I noticed the table of contents (TOC) needed to be updated. 
Screen Shot 2022-10-20 at 12 00 49 PM
Screen Shot 2022-10-20 at 12 00 58 PM

Moving the topic "Managing backup and restore operator". The TOC is expected to be displayed in the following order:

1. Backup and restore
2. Managing backup and restore operator (@birsanv should this be renamed to backup and restore component? I remember we changed some references in the last live review)

### DIFF
--- a/backup_restore/main.adoc
+++ b/backup_restore/main.adoc
@@ -3,4 +3,4 @@ include::modules/common-attributes.adoc[]
 = Backup and restore
 
 include::backup_intro.adoc[leveloffset=+1]
-include::manage_backup_restore.adoc[leveloffset=+2]
+include::manage_backup_restore.adoc[leveloffset=+1]


birsanv (approved):
lgtm

birsanv:
@dockerymick we can change the `operator` to `component` but it should be done in a few other places

dockerymick:
@birsanv Thanks for your suggestion! As you explained in Slack, the project name is `cluster-backup-operator`, so it is fine to use 'operator'. Let's keep the name as is since the terms can be used interchangeably

dockerymick:
After discussion in the doc architecture call. We agreed not to move forward with this change. Thanks for your review @birsanv 

Closing this PR

PR #4161, opened 2022-10-04
Backup and restore updates for 2.6

dockerymick:
https://github.com/stolostron/backlog/issues/25895

PR #4229, opened 2022-10-24
Reverting change that was applicable for 2.7 only; adding note back for 2.6

dockerymick:
[Related PR](https://github.com/stolostron/rhacm-docs/pull/4161/files)
[Related issue for 2.7](https://github.com/stolostron/backlog/issues/25978)
[Related issue where I made the incorrect commit](https://github.com/stolostron/backlog/issues/25895)

@birsanv will you confirm that the following notes should be listed for 2.6:

* The OADP Operator 1.0 has disabled building multi-arch builds and only produces `x86_64` builds for the official release. This means that if you are using an architecture other than `x86_64`, the OADP Operator installed by the backup component must be replaced with the correct version. In this case, uninstall the OADP Operator and find the operator matching your architecture, then install it.

* If you have previously installed and used the OADP Operator on the hub cluster, in a namespace different from the backup component namespace, uninstall this version since the backup component works now with OADP installed in the component namespace. Use the same storage location for the link:https://github.com/openshift/oadp-operator/blob/master/docs/install_olm.md#create-the-dataprotectionapplication-custom-resource[`DataProtectionApplication`] resource owned by the OADP Operator installed with the backup component; it accesses the same backup data as the previous operator. Velero backup resources are now loaded within the new OADP Operator namespace on this hub cluster.

### DIFF
--- a/backup_restore/manage_backup_restore.adoc
+++ b/backup_restore/manage_backup_restore.adoc
@@ -3,7 +3,11 @@
 
 The cluster backup and restore operator is not installed automatically. Enable the backup component by setting the `cluster-backup` parameter to `true`, in the `MultiClusterHub` resource. The OADP Operator is automatically installed, in the same namespace as the backup resource, when you install the cluster backup and restore operator.
 
-*Note*: If you have previously installed and used the OADP Operator on the hub cluster, in a namespace different from the backup component namespace, uninstall this version since the backup component works now with OADP installed in the component namespace. Use the same storage location for the link:https://github.com/openshift/oadp-operator/blob/master/docs/install_olm.md#create-the-dataprotectionapplication-custom-resource[`DataProtectionApplication`] resource owned by the OADP Operator installed with the backup component; it accesses the same backup data as the previous operator. Velero backup resources are now loaded within the new OADP Operator namespace on this hub cluster.
+*Notes*: 
+
+- The OADP Operator 1.0 has disabled building multi-arch builds and only produces `x86_64` builds for the official release. This means that if you are using an architecture other than `x86_64`, the OADP Operator installed by the backup component must be replaced with the correct version. In this case, uninstall the OADP Operator and find the operator matching your architecture, then install it.
+
+- If you have previously installed and used the OADP Operator on the hub cluster, in a namespace different from the backup component namespace, uninstall this version since the backup component works now with OADP installed in the component namespace. Use the same storage location for the link:https://github.com/openshift/oadp-operator/blob/master/docs/install_olm.md#create-the-dataprotectionapplication-custom-resource[`DataProtectionApplication`] resource owned by the OADP Operator installed with the backup component; it accesses the same backup data as the previous operator. Velero backup resources are now loaded within the new OADP Operator namespace on this hub cluster.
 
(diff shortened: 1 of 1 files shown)


backup_restore/manage_backup_restore.adoc
```diff
@@ -3,7 +3,11 @@
...
 
-*Note*: If you have previously installed and used the OADP Operator on the hub cluster, in a namespace different from the backup component namespace, uninstall this version since the backup component works now with OADP installed in the component namespace. Use the same storage location for the link:https://github.com/openshift/oadp-operator/blob/master/docs/install_olm.md#create-the-dataprotectionapplication-custom-resource[`DataProtectionApplication`] resource owned by the OADP Operator installed with the backup component; it accesses the same backup data as the previous operator. Velero backup resources are now loaded within the new OADP Operator namespace on this hub cluster.
+*Notes*: 
+
+- The OADP Operator 1.0 has disabled building multi-arch builds and only produces `x86_64` builds for the official release. This means that if you are using an architecture other than `x86_64`, the OADP Operator installed by the backup component must be replaced with the correct version. In this case, uninstall the OADP Operator and find the operator matching your architecture, then install it.
```
oafischer: Is full present tense in the first sentence OK? So:

The OADP Operator 1.0 disables building multi-arch builds and only produces `x86_64` builds for the official release.

oafischer: I think we can probably leave out  the phrase "This means that". Or maybe replace it with "As a result"? I'll let you decide. Just some thoughts.

oafischer: Instead of "In this case", maybe "To replace the version, uninstall the OADP Operator and find the operator matching your architecture, then install it." Or something along those lines that's a bit more specific.

backup_restore/manage_backup_restore.adoc
```diff
@@ -3,7 +3,11 @@
...
+
+- If you have previously installed and used the OADP Operator on the hub cluster, in a namespace different from the backup component namespace, uninstall this version since the backup component works now with OADP installed in the component namespace. Use the same storage location for the link:https://github.com/openshift/oadp-operator/blob/master/docs/install_olm.md#create-the-dataprotectionapplication-custom-resource[`DataProtectionApplication`] resource owned by the OADP Operator installed with the backup component; it accesses the same backup data as the previous operator. Velero backup resources are now loaded within the new OADP Operator namespace on this hub cluster.
```
oafischer, line 10: The first sentence is a little clunky. Maybe we can break it up or reword it a little. A simple change might be:

If you have previously installed and used the OADP Operator on the hub cluster, in a namespace different from the backup component namespace, uninstall this version. The backup component now works with OADP installed in the component namespace.

backup_restore/manage_backup_restore.adoc
dockerymick: @oafischer Thanks for your suggestion! How about:

The OADP Operator 1.0 no longer supports multi-arch builds and only produces `x86_64` builds for official release. As a result,  if you are using an architecture other than.....

Thoughts?

oafischer: That's perfect!

birsanv (approved):
lgtm

dockerymick:
Thanks so much for your approval Valentina! @birsanv

oafischer (approved):

dockerymick:
@oafischer Thanks Oliver!

PR #4230, opened 2022-10-24
Attempting to resolve BZ for 2.4

dockerymick:
https://github.com/stolostron/backlog/issues/25916

### DIFF
--- a/observability/observe_environments.adoc
+++ b/observability/observe_environments.adoc
@@ -46,6 +46,10 @@ Learn from the {ocp-short} documentation what types of metrics are collected and
 
 [#observability-pod-capacity-requests]
 == Observability pod capacity requests
+//add sizing column, request from issue 25916; Observability requirements give us the details for each observability workload expected resources but does not give the sizing for the nodes.
+
+//The sizing documentation is not clear.
+//The documentation mention capacity if installed on worker nodes but no in infra nodes. We are still not capable to estimate the CPU/RAM for ACM components in the infra nodes. MJ, 10/24
 
 Observability components require 2701mCPU and 11972Mi memory to install the observability service. The following table is a list of the pod capacity requests for five managed clusters with `observability-addons` enabled:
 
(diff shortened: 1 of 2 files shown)


observability/observe_environments.adoc
```diff
@@ -46,6 +46,10 @@ Learn from the {ocp-short} documentation what types of metrics are collected and
 
 [#observability-pod-capacity-requests]
 == Observability pod capacity requests
+//add sizing column, request from issue 25916; Observability requirements give us the details for each observability workload expected resources but does not give the sizing for the nodes.
```
swopebe: So we have an issue requesting that we add a sizing column? If that is the case, consider removing the comment since the request is recorded in the issue already and we will address it as either release work or tech debt? Thank you.

swopebe: Same with the next comment.

dockerymick: @swopebe I have not seen a specific issue to add a sizing column. It was a comment mentioned in the BZ. @jlpadilla Would it be beneficial to consider adding a sizing column? This is the comment from the author: 

["Observability requirements give us the details for each observability workload expected resources but does not give the sizing for the nodes."](https://github.com/stolostron/backlog/issues/25916#issue-1382188008). 

If this should be doc'ed we can create a new issue to add a sizing column

jlpadilla: @subbarao-meduri This one is for the metrics-collector sizing. Can you help answer?

swopebe:
poke

dockerymick:
Based on the progress of the [issue](https://github.com/stolostron/backlog/issues/25916) that is attached to this PR and response, I removed my hidden comment from the `observe_environments.adoc` file. Sizing is still a topic that's discussed. Changes related to sizing can be addressed in 2.7. If there is a customer request to update earlier versions, it will be updated. Moving forward with this merge.

dockerymick:
From the issue, you can see that previous PRs with similar updates were approved by dev and a writer

PR #4234, opened 2022-10-24
Document MSA feature in 2.7 (Fixed version)

oafischer:
https://github.com/stolostron/backlog/issues/26324

I have 2 concerns:

The limitations section seems like it should belong into known issues. Does it make sense to move it there? I left it in here for the moment.

This new section is rather long and quite convoluted. I think the entire file needs its architecture reevaluated. Some of these topics should probably go into their own files. Not something for now, but I'd like to discuss that in the future.


backup_restore/manage_backup_restore.adoc
```diff
@@ -462,6 +463,102 @@ To reconnect imported clusters on the new hub cluster, manually create the `auto
 
 Create the `auto-import-secret` resource in the managed cluster namespace for each cluster in `Pending Import` state. Use a `kubeconfig` or token with enough permissions for the import component to start the automatic import on the new hub cluster. You must have access for each managed cluster by using a token to connect with the managed cluster. The token must have a `klusterlet` role binding or a role with the same permissions.
 
+[#auto-connect-clusters-managed-service-account-tech-pre]
+==== Automatically connecting clusters by using a Managed Service Account (Technology Preview)
```
dockerymick, line 470: Be sure to add the childlink to the TOC at the beginning of the doc, I believe I also forgot to add a link to the "Restoring imported managed clusters" section. Can you add that to the list as well?

dockerymick: does Managed Service Account need to the capitalized?

backup_restore/manage_backup_restore.adoc
```diff
@@ -462,6 +463,102 @@ To reconnect imported clusters on the new hub cluster, manually create the `auto
...
+
+The backup controller implements a solution to automatically connect imported clusters on the new hub cluster. Automatic connection uses the link:../https://github.com/open-cluster-management-io/managed-serviceaccount[Managed Service Account] component on the primary hub cluster to create a token for each imported cluster. The token is backed up in each managed cluster namespace and uses a `klusterlet-bootstrap-kubeconfig` `ClusterRole` binding, which allows an auto import operation to use the token. The `klusterlet-bootstrap-kubeconfig` `ClusterRole` can only get or update the `bootstrap-hub-kubeconfig` secret.
```
dockerymick: Throughout the doc, we are using **backup component** and **backup operator** interchangeably. I would reconsider using "controller". Just want to be consistent. 

The link reference needs to be updated:

link:https://github.com/open-cluster-management-io/managed-serviceaccount[Managed Service Account]

I suggest the following change:

Use the Managed Service Account component on the primary hub cluster to automatically connect imported clusters to the new hub cluster.  You must create a token for each imported cluster, which is backed up in each managed cluster namespace. Update the `klusterlet-bootstrap-kubeconfig` `ClusterRole` to enable access to the token during the import operation. The `klusterlet-bootstrap-kubeconfig` `ClusterRole` can only get or update the `bootstrap-hub-kubeconfig` secret.

I recommend these changes based on the [Anthropomorphism guidance in the style guide](https://www.ibm.com/docs/en/ibm-style?topic=grammar-anthropomorphism)

backup_restore/manage_backup_restore.adoc
```diff
@@ -462,6 +463,102 @@ To reconnect imported clusters on the new hub cluster, manually create the `auto
...
+
+When the activation data is restored on the new hub cluster, the restore controller runs a post restore operation and looks for all managed clusters in the `Pending Import` state. If there is a valid token generated by the Managed Service Account, the controller creates an `auto-import-secret` using the token. As a result, the import component tries to reconnect the managed cluster. If the cluster is accessible, the operation is successful.
```
dockerymick: Does the user have to run the restore operation manually or does it run automatically? 

(if manual) I recommend the following updates: 

When the activation data is restored on the new hub cluster, use the restore operator to run a post restore operation and view all managed clusters in the `Pending Import` state. If there is a valid token generated by the Managed Service Account, the controller creates an `auto-import-secret` by using the token. As a result, the import component tries to reconnect with the managed cluster. The token is only created if the managed cluster is accessible when you create the Managed Service Account, otherwise it is created later once the the managed cluster becomes available.

backup_restore/manage_backup_restore.adoc
```diff
@@ -462,6 +463,102 @@ To reconnect imported clusters on the new hub cluster, manually create the `auto
...
+
+[#enabling-auto-import]
+===== Enabling automatic import
+
```
dockerymick, line 478: Be sure to add this section to the TOC at the beginning of this doc

backup_restore/manage_backup_restore.adoc
```diff
@@ -462,6 +463,102 @@ To reconnect imported clusters on the new hub cluster, manually create the `auto
...
+kind: MultiClusterEngine
+metadata:
+  name: multiclusterhub
+spec:
+  overrides:
+    components:
+      - enabled: true
+        name: managedserviceaccount-preview
+----
+
```
dockerymick, line 495: I recommend the following update:

. Enable the Managed Service Account component by listing `managedserviceaccount-preview` in the `MultiClusterEngine` resource. See the following example:


How will the user view this resource? Will they need to run a specific command to access this resource? If yes, we should doc the command that they need to run

backup_restore/manage_backup_restore.adoc
```diff
@@ -462,6 +463,102 @@ To reconnect imported clusters on the new hub cluster, manually create the `auto
...
+. Enable the automatic import feature for the `BackupSchedule.cluster.open-cluster-management.io` resource. See the following example:
```
dockerymick: I recommend the following update: 

Enable the automatic import feature for the `BackupSchedule.cluster.open-cluster-management.io` resource by setting the `useManagedServiceAccount` parameter to `true`. See the following example:

backup_restore/manage_backup_restore.adoc
```diff
@@ -462,6 +463,102 @@ To reconnect imported clusters on the new hub cluster, manually create the `auto
...
+metadata:
+  name: schedule-acm-msa
+spec:
+  veleroSchedule:
+  veleroTtl: 120h
+  useManagedServiceAccount: true
+  managedServiceAccountTTL: 2h
+----
+
+After you set `useManagedServiceAccount` to `true`, the backup controller starts processing imported managed clusters. The backup controller creates the following for each imported managed cluster:
```
dockerymick: I recommend the following update:

After you enable the automatic import feature, the backup component starts processing imported managed clusters.

backup_restore/manage_backup_restore.adoc
```diff
@@ -462,6 +463,102 @@ To reconnect imported clusters on the new hub cluster, manually create the `auto
...
+
+- A `ManagedServiceAddon` named `managed-serviceaccount`.
+- A `ManagedServiceAccount` named `auto-import-account`.
+- A `ManifestWork` for each `ManagedServiceAccount` to set up a `klusterlet-bootstrap-kubeconfig` `RoleBinding` for the `ManagedServiceAccount` token on the managed cluster.
+
+The `ManagedServiceAccount` resource creates a token on the managed cluster that is also called `ManagedServiceAccount`. The token is then pushed to the hub cluster in the managed cluster namespace. This hub cluster secret is backed up. The token is only created if the managed cluster is accessible when you create the Managed Service Account, otherwise it is created later once the the managed cluster becomes available.
```
dockerymick: I recommend removing the following statement since something similar is mentioned earlier: 

"The token is only created if the managed cluster is accessible when you create the Managed Service Account, otherwise it is created later once the the managed cluster becomes available."

backup_restore/manage_backup_restore.adoc
```diff
@@ -462,6 +463,102 @@ To reconnect imported clusters on the new hub cluster, manually create the `auto
...
+kind: BackupSchedule
+metadata:
+  name: schedule-acm-msa
+spec:
+  veleroSchedule:
+  veleroTtl: 120h
+  useManagedServiceAccount: false
+----
+
+The default value is `false`. After setting the value to `false`, the backup controller removes all created resources, including `ManagedServiceAddon`, `ManagedServiceAccount`, and `ManifestWork`. Removing the resources deletes the auto import token on the hub cluster and managed cluster.
```
dockerymick: Removing the resources deletes the automatic import token on the hub cluster and managed cluster.

backup_restore/manage_backup_restore.adoc
```diff
@@ -462,6 +463,102 @@ To reconnect imported clusters on the new hub cluster, manually create the `auto
...
+
+[#limitations-auto-import]
+===== Automatic import limitations
+
```
dockerymick: I recommend that you move this section to the [Known issue and limitations doc. I would add it as a subsection to the [Backup and restore section](https://github.com/stolostron/rhacm-docs/blob/2.7_stage/release_notes/known_issues.adoc#backup-known-issues)

dockerymick (changes requested):

backup_restore/manage_backup_restore.adoc
oafischer: It's the name of the component and it's capitalized in its own docs so I just followed that. They refer to it as "a" as well and proceed to capitalize it still. Though the writing in that doc wasn't great so let me know if you think it would be better to not capitalize.

dockerymick: Gotchu, sounds good to me

backup_restore/manage_backup_restore.adoc
oafischer: I checked again, it's not clear if it's manual or not. Maybe @birsanv can provide some insight?

backup_restore/manage_backup_restore.adoc
oafischer: @birsanv Does the user need to run a specific command here?

oafischer:
Thanks for the detailed feedback @dockerymick! I've addressed what I could and pulled in @birsanv where I need help.

backup_restore/manage_backup_restore.adoc
```diff
@@ -21,6 +21,9 @@ For a list of supported storage providers for Velero, see link:https://velero.io
 ** <<restore-passive-resources,Restoring passive resources>>
 ** <<restore-all-resources,Restoring all resources>>
 ** <<restore-imported-managed-clusters,Restoring imported managed clusters>>
+*** <<auto-connect-clusters-managed-service-account-tech-pre,Automatically connecting clusters by using a Managed Service Account (Technology Preview)>>
```
dockerymick: I would reconsider shortening the anchor. Maybe just use `auto-connect-clusters-managed-service-account` or `auto-connect-clusters-msa`. If you do decide to update the anchor, remember to update the section too

dockerymick (approved):
Made one more comment, however I am good with your changes here.

backup_restore/manage_backup_restore.adoc
oafischer: `auto-connect-clusters-msa` is a good one, I'll use that. Thanks!

oafischer:
Thanks @dockerymick! I'll wait for @birsanv to approve before merging.

backup_restore/manage_backup_restore.adoc
```diff
@@ -462,6 +466,89 @@ To reconnect imported clusters on the new hub cluster, manually create the `auto
 
 Create the `auto-import-secret` resource in the managed cluster namespace for each cluster in `Pending Import` state. Use a `kubeconfig` or token with enough permissions for the import component to start the automatic import on the new hub cluster. You must have access for each managed cluster by using a token to connect with the managed cluster. The token must have a `klusterlet` role binding or a role with the same permissions.
 
+[#auto-connect-clusters-msa]
+==== Automatically connecting clusters by using a Managed Service Account (Technology Preview)
+
+Use the link:https://github.com/open-cluster-management-io/managed-serviceaccount[Managed Service Account] component on the primary hub cluster to automatically connect imported clusters to the new hub cluster.  You must create a token for each imported cluster, which is backed up in each managed cluster namespace. Update the `klusterlet-bootstrap-kubeconfig` `ClusterRole` to enable access to the token during the import operation. The `klusterlet-bootstrap-kubeconfig` `ClusterRole` can only get or update the `bootstrap-hub-kubeconfig` secret.
```
birsanv: @dockerymick I don't really agree with the way this information is presented .. 
It is as if the user must do all this things, which is not the case - this is the whole point, the backup component does all this for the user , automatically ..

By saying `You must create a token..` we lost the automatic , done for you by the controller story. It should say `The backup controller implements a solution to automatically connect imported clusters on the new hub.` 

Not sure how to better rephrase this, other then pointing to the readme :) 
https://github.com/stolostron/cluster-backup-operator#automatically-connecting-clusters-using-managedserviceaccount---tech-preview

_The backup controller implements a solution to automatically connect imported_ ...

backup_restore/manage_backup_restore.adoc
birsanv: Restore is invoked by the user, by creating a `Restore` resource.
So I'd say the restore is still automatic .. 

I find this confusing : 
When the activation data is restored on the new hub cluster, use the restore operator to run a post restore operation and view all managed clusters in the Pending Import state.

`use the restore operator` - you are already using it since you said `When the activation data is restored`

backup_restore/manage_backup_restore.adoc
birsanv: @oafischer no specific command, just regular update of a resource, can be done through OCP ui or command line, nothing to document here.

@dockerymick why do we want to say `Enable the Managed Service Account component by listing` ? that option is already there, we just switch the `enabled` value to true or false

birsanv (changes requested):

backup_restore/manage_backup_restore.adoc
oafischer: Hmm I think we can leave out "implements a solution" and just directly say:

"The backup controller automatically connects imported clusters on the new hub cluster."

Does that work @birsanv @dockerymick?

birsanv: @oafischer that would work for me

birsanv: @oafischer I mean I assume this would be something like 

"The backup controller automatically connects imported clusters on the new hub cluster using the link:https://github.com/open-cluster-management-io/managed-serviceaccount[Managed Service Account] component... etc

birsanv: What about 

"The backup controller automatically connects imported clusters on the new hub cluster using the link:[https://github.com/open-cluster-management-io/managed-serviceaccount[Managed](https://github.com/open-cluster-management-io/managed-serviceaccount%5BManaged) Service Account] component.
The MaangedServiceAccount creates a token for each imported cluster in each managed cluster namespace., which is backed up. This token is set to use a klusterlet-bootstrap-kubeconfig ClusterRole binding, which allows the token to be used by an auto import operation. The klusterlet-bootstrap-kubeconfig ClusterRole can only get or update the bootstrap-hub-kubeconfig secret.
When the activation data is restored on the new hub, the restore controller runs a post restore operation and looks for all managed clusters in Pending Import state. For these managed clusters, it checks if there is a valid token generated by the ManagedServiceAccount and if found, it creates an auto-import-secret using this token. As a result, the import component will try to reconnect the managed cluster and if the cluster is accessible the operation should be successful.

dockerymick: @birsanv Thanks for clarifying ! From. my perspective, I wanted to put emphasis on the requirement. i agree that writing "You must.." suggests that the user needs to do some manual work, which is not the case for this. 

I like your suggestion. I recommend that we prompt the user for the link reference. The link might distract the reader. i recommend that we add the reference link for Managed Service Account be placed at the end of the paragraph.

Let's be sure to use "automatic import" instead of _auto import_

oafischer: I implemented most of your suggestion, let me know if it looks good now. Thanks!

backup_restore/manage_backup_restore.adoc
```diff
@@ -462,6 +466,89 @@ To reconnect imported clusters on the new hub cluster, manually create the `auto
...
 
+[#auto-connect-clusters-msa]
+==== Automatically connecting clusters by using a Managed Service Account (Technology Preview)
+
+The backup controller automatically connects imported clusters to the new hub cluster by using the link:https://github.com/open-cluster-management-io/managed-serviceaccount[Managed Service Account] component. The Managed Service Account creates a token that is backed up for each imported cluster in each managed cluster namespace. The token uses a `klusterlet-bootstrap-kubeconfig` `ClusterRole` binding, which allows the token to be used by an automatic import operation. The `klusterlet-bootstrap-kubeconfig` `ClusterRole` can only get or update the `bootstrap-hub-kubeconfig` secret.
+
+When the activation data is restored on the new hub cluster, the restore controller runs a post restore operation and looks for all managed clusters in the `Pending Import` state. If there is a valid token generated by the Managed Service Account is found, the controller creates an `auto-import-secret` using the token. As a result, the import component tries to reconnect the managed cluster. If the cluster is accessible, the operation is successful.
```
birsanv: @oafischer  small typo 

If there is a valid token generated by the Managed Service Account is found

should be 

If a valid token generated by the Managed Service Account is found

backup_restore/manage_backup_restore.adoc
dockerymick: Understood. I thought the user needed to update the resource to enable it, Thanks for clarifying. I recommend this instead:

Enable the Managed Service Account component by setting the `enabled` parameter to `true`.

birsanv: @oafischer this is still not addressed 

why do we want to say Enable the Managed Service Account component by listing ? that option is already there, we just switch the enabled value to true or false

I suggest to replace 

Enable the Managed Service Account component by listing

with 

Enable the Managed Service Account component by setting the `managedserviceaccount-preview` component's enable option to `true`

release_notes/known_issues.adoc
```diff
@@ -1054,6 +1054,20 @@ When you restore the backup of the changed or rotated certificate of authority (
...
+=== Automatic import limitations
+//2.7:26324
+
+The following situations can prevent the managed cluster from being automatically imported when moving to a new hub cluster:
+
+- When running a hub backup without a `ManagedServiceAccount` token, for example when you create the `ManagedServiceAccount` resource while the managed cluster is not accessible, the backup does not contain a token to auto import the managed cluster.
+
+- When you import a managed cluster right after a lookup is completed, the managed cluster is not processed until the next call. Any backups created before the new lookup is processed do not contain an auto import token for newly imported managed clusters.
+
+- The auto import operation fails when you run the restore operation while the `auto-import-account` token is valid and backed up but the backup with the token has expired. The `restore.cluster.open-cluster-management.io` resource reports invalid token issues for each managed cluster.
```
birsanv: This is confusing @oafischer ( The auto import operation fails when you run the restore operation while the `auto-import-account` token is valid and backed up but the backup with the token has expired. )

I want to say : 
"The auto import operation fails - If the auto-import-account secret token is valid at the time when is backed up 
 but the restore operation is run at a time when the token available with the backup has already expired"

birsanv (changes requested):

release_notes/known_issues.adoc
```diff
@@ -1064,9 +1064,9 @@ The following situations can prevent the managed cluster from being automaticall
 
 - When you import a managed cluster right after a lookup is completed, the managed cluster is not processed until the next call. Any backups created before the new lookup is processed do not contain an auto import token for newly imported managed clusters.
 
-- The auto import operation fails when you run the restore operation while the `auto-import-account` token is valid and backed up but the backup with the token has expired. The `restore.cluster.open-cluster-management.io` resource reports invalid token issues for each managed cluster.
+- The auto import operation fails the `auto-import-account` secret token is valid and is backed up but the restore operation is run when the token available with the backup has already expired. The `restore.cluster.open-cluster-management.io` resource reports invalid token issues for each managed cluster.
```
birsanv: @oafischer small typo :) great otherwise !

This 
The auto import operation fails the ...

Should be 

The auto import operation fails IF the ...

birsanv (changes requested):

birsanv (approved):
lgtm

release_notes/known_issues.adoc
```diff
@@ -1081,7 +1081,7 @@ spec:
       url: <apiserver>
 ----
 +
-When a cluster is imported on the hub cluster, the `apiserver` is only set up automatically on {ocp} clusters You must set the `apiserver` manually on other types of managed clusters, such as EKS clusters, otherwise the automatic import feature ignores the clusters. As a result, they remain in the `Pending Import` state when you moved them to the import hub cluster.
+When a cluster is imported on the hub cluster, the `apiserver` is only set up automatically on {ocp} clusters You must set the `apiserver` manually on other types of managed clusters, such as EKS clusters, otherwise the automatic import feature ignores the clusters. As a result, the clusters remain in the `Pending Import` state when you moved them to the import hub cluster.
```
birsanv: Typo : 
You must

. You must

oafischer:
Thanks @dockerymick and @birsanv for your help with this PR! I've also implemented the suggestions from here while this is open: https://github.com/stolostron/backlog/issues/26593

I think I addressed all the comments and typos as well. If either of you wants to take one last quick look at everything I'd appreciate it, otherwise I plan to merge tomorrow morning. Thanks again!

release_notes/known_issues.adoc
birsanv: As a result, the clusters remain in the `Pending Import` state when you moved them to the import hub cluster.

should be 

As a result, the clusters remain in the `Pending Import` state when you moved them to the restore hub cluster.

birsanv (changes requested):

birsanv (approved):
lgtm

PR #4242, opened 2022-10-25
Update clean up in 2.7

oafischer:
https://github.com/stolostron/backlog/issues/26291

I checked for instances where CleanupAll is referenced or clean up before applying restore is mentioned. I couldn't find anything additional other than in the known issues, but I think we will address those in the cross squad issues, so I left those for now.

### DIFF
--- a/backup_restore/manage_backup_restore.adoc
+++ b/backup_restore/manage_backup_restore.adoc
@@ -16,7 +16,7 @@ For a list of supported storage providers for Velero, see link:https://velero.io
 * <<schedule-backup,Scheduling a cluster backup>>
 * <<restore-backup,Restoring a backup>>
 ** <<prepare-new-hub,Preparing the new hub cluster>>
-** <<clean-hub-restore,Cleaning the hub cluster before restore>>
+** <<clean-hub-restore,Cleaning the hub cluster after restore>>
 ** <<restore-passive-resources-check-backups,Restoring passive resources while checking for backups>>
 ** <<restore-passive-resources,Restoring passive resources>>
 ** <<restore-all-resources,Restoring all resources>>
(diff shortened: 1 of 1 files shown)


backup_restore/manage_backup_restore.adoc
```diff
@@ -332,19 +332,18 @@ Before running the restore operation on a new hub cluster, you need to manually
...
-* `CleanupRestored`: Clean up all resources created by a previous {product-title-short} restore. It is recommended to use this property because it is less intrusive than the `CleanupAll` property.
-* `CleanupAll`: Clean up all resources on the hub cluster, which can be part of an {product-title-short} backup, even if the resources are not created as a result of a restore operation. This is to be used when extra content has been created on the hub cluster, which requires clean up. Use this option with caution because this option cleans up resources on the hub cluster created by the user, not by a previous backup. It is strongly recommended to use the `CleanupRestored` option, and to refrain from manually updating hub cluster content when the hub cluster is designated as a passive cluster for a disaster scenario. Use the `CleanupAll` option as a last alternative.
+- `None`: No clean up necessary, just begin Velero restore. This is to be used on a brand new hub cluster.
+- `CleanupRestored`: Clean up all resources created by a previous {product-title-short} restore that are not part of the currently restored backup.
```
birsanv, line 344: sorry, @oafischer , I decided to add the `CleanupAll` option; it was used by a customer so I wanted to be backward compatible

As a result , add this content as a third bullet : 
- CleanupAll : clean up all resources on the hub which could be part of a {product-title-short} backup, even if they were not created as a result of a restore operation. This is to be used when content has been created on this hub before the restore operation is executed. Use this option with extreme caution as this will also cleanup resources on the hub created by the user, not just by a previously restored backup. It is strongly recommended to use the CleanupRestored option instead and to refrain from manually updating hub content when the hub is designated as a passive candidate for a disaster scenario. Use a clean hub as a passive cluster. Avoid situations where you have to swipe the cluster using the CleanupAll option; this is given as a last alternative.

birsanv (changes requested):

backup_restore/manage_backup_restore.adoc
oafischer: No worries! I'll add it back. Thanks for letting me know.

backup_restore/manage_backup_restore.adoc
```diff
@@ -332,19 +332,21 @@ Before running the restore operation on a new hub cluster, you need to manually
...
 
 [#clean-hub-restore]
-=== Cleaning the hub cluster before restore
+=== Cleaning the hub cluster after restore
 
-Velero currently skips existing backed up resources on the hub cluster. This limits the scenarios that can be used when you restore hub cluster data on a new hub cluster. If the new hub cluster is used and the restore is applied more than once, the hub cluster is not recommended to use as a passive configuration unless the data is cleaned before restore is ran. The data on the new hub cluster is not reflective of the data available with the restored resources.
+Velero updates existing resources if they have changed with the currently restored backup. Velero does not clean up delta resources, which are resources created by a previous restore and not part of the currently restored backup. This limits the scenarios you can use when restoring hub data on a new hub cluster. Unless the restore is applied only once, you cannot reliably use the new hub cluster as a passive configuration. The data on the hub cluster does not reflect the data available with the restored resources.
```
dockerymick: Is 'delta' a resource kind? Should this be monospaced?

Also I suggest the following update:

This limits the scenarios you can use when restoring hub cluster data on a new hub cluster.

backup_restore/manage_backup_restore.adoc
```diff
@@ -332,19 +332,21 @@ Before running the restore operation on a new hub cluster, you need to manually
...
+- `None`: No clean up necessary, just begin Velero restore. Use `None` on a brand new hub cluster.
+- `CleanupRestored`: Clean up all resources created by a previous {product-title-short} restore that are not part of the currently restored backup.
+- `CleanupAll`: Clean up all resources on the hub cluster that might be part of a {product-title-short} backup, even if they were not created as a result of a restore operation. This is to be used when extra content has been created on a hub cluster before the restore operation starts.
```
dockerymick: I suggest the following update: 

This is to be used when extra content is created on a hub cluster before the restore operation starts.

dockerymick (approved):
Made a couple of comments. Thanks for making the changes !

backup_restore/manage_backup_restore.adoc
oafischer: Couldn't find any references to delta resources in our or other docs for verification. @birsanv Can you chime in?

birsanv: Delta in not a resource;  delta resources = extra resources created by a previous restore and not part of the currently restored backup.
For example : app1, app2 is on the restore hub. The new restored backup only contains an updated version of app2; there is no app1 in the new backup. So app1 is an extra resource ( delta from this new backup ) and should be removed

oafischer:
Addressed all the comments, does it look good now @birsanv?

birsanv (approved):
lgtm

PR #4261, opened 2022-10-28
adding info about replace upgrades and inplace upgrades per osdocs-4180

lahinson:
Related OCP doc Jira: https://issues.redhat.com/browse/OSDOCS-4180

This PR adds information about replace upgrades and in-place upgrades for hosted control planes.

I requested review from Yu Qi Zhang in the related Jira. I did not find him in GitHub to add as a reviewer.

### DIFF
--- a/multicluster_engine/hosted_control_planes/hosted_control_planes_configure.adoc
+++ b/multicluster_engine/hosted_control_planes/hosted_control_planes_configure.adoc
@@ -348,9 +348,15 @@ spec:
           - <zone-2>
 ----
 
-Replace `zone-1` with the name of the zone in the region.
+Replace `zone-1` with the name of the zone in the region. Replace `zone-2` with the name of the zone in the region. 
 
-Replace `zone-2` with the name of the zone in the region. 
+In the previous example, the `upgradeType` is `Replace`. To upgrade a hosted cluster, you can use a replace upgrade or an in-place upgrade. 
+
+A _replace_ upgrade creates instances in the new version while it removes old instances from the previous version. This upgrade type is a good choice in cloud environments where this level of immutability is cost effective. 
+
+An _in-place_ upgrade directly updates the operating systems of the instances. This type is a good choice for environments where the infrastructure constraints are higher, such as bare metal.
+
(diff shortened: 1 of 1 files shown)


yuqi-zhang (commented):
Generally looks fine, thanks!

oafischer (approved):
Looks good, thanks!

multicluster_engine/hosted_control_planes/hosted_control_planes_configure.adoc
```diff
@@ -348,9 +348,11 @@ spec:
           - <zone-2>
 ----
 
-Replace `zone-1` with the name of the zone in the region.
+Replace `zone-1` with the name of the zone in the region. Replace `zone-2` with the name of the zone in the region. 
 
-Replace `zone-2` with the name of the zone in the region. 
+In the previous example, the `upgradeType` is `Replace`. To upgrade a hosted cluster, you can use a replace upgrade or an in-place upgrade. A replace upgrade creates instances in the new version while it removes old instances from the previous version. This upgrade type is a good choice in cloud environments where this level of immutability is cost effective. In-place upgrades directly update the operating systems of the instances. This type is a good choice for environments where the infrastructure constraints are higher, such as bare metal.
```
swopebe: For clarity, we may want to break out the upgrade types so user can easily find def. and term?

A _replace upgrade_ creates instances in the new version while it removes old instances from the previous version. This upgrade type is a good choice in cloud environments where this level of immutability is cost effective. 

An _in-place_ upgrade directly updates the operating systems of the instances. This type is a good choice for environments where the infrastructure constraints are higher, such as bare metal.

(parallel and off set in two paragraphs to draw user to the term. We could also do a DL but I think with just these two, not necessary.)

thoughts?

lahinson: Great idea. I can make those changes. Thanks!

lahinson:
@swopebe I just committed the changes that you suggested. Thanks!

swopebe:
With Olivers approval, the merge is ready for you when you are. 

@oafischer since you want to update the readme, this may be a good tidbit:

_Link to your Jira or GitHub issue in the PR. You can merge after your PR is reviewed by a peer and a developer._ (though it looks like Laura has the dev. discussion in the JIRA so that works, too).

lahinson:
Thanks @swopebe and @oafischer!

PR #4275, opened 2022-11-02
Adding infrastructure node info

dockerymick:
https://github.com/stolostron/backlog/issues/25916

### DIFF
--- a/install/cluster_size.adoc
+++ b/install/cluster_size.adoc
@@ -34,21 +34,7 @@ Each {product-title} cluster is unique and the following guidelines provide samp
 | 6
 |===
 
-In addition to {product-title-short}, the {ocp-short} cluster runs additional services to support cluster features. The recommended node size is three nodes. 
-
-
-// should we point the users to this topic? (install_connected.adoc)
-//#installing-on-infra-node] 
-//== Installing the {product-title-short} hub cluster on infrastructure nodes 
-//
-//
-//
-//Or to this topic (create_infra_env.adoc)
-//#creating-an-infrastructure-environment
-//Creating an infrastructure environment
-//
-//OR point them to both topics
-//
+In addition to {product-title-short}, the {ocp-short} cluster runs additional services to support cluster features. The recommended node size is three nodes. See xref:../install/install_connected.adoc#installing-on-infra-node[Installing the {product-title-short} hub cluster on infrastructure nodes] for more details.
 
 [#openshift-cluster-on-additional-services]
 === {ocp-short} on additional services
--- a/install/cluster_size.adoc
+++ b/install/cluster_size.adoc
@@ -27,7 +27,7 @@ Each {product-title} cluster is unique and the following guidelines provide samp
 | Per {ocp-short} sizing guidelines
 | Per {ocp-short} sizing guidelines
 
-| Worker
+| Worker or infrastructure
 | 3
 | 1
 | 12 GB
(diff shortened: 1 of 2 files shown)


jnpacker:
So I don't mean to keep asking questions, but I took a look at the whole page.
1. @xrajesh I need someone from search to confirm they are okay with the Infrastructure to work node comparison.
2. @dockerymick it should be either you use Worker Nodes or Infrastructure Nodes, but the table makes it seem like you would use both.  So strictly speaking the ACM configurations are:  Control Plane Nodes + Worker nodes or Control Plane Nodes + Infrastructure Nodes. There is technically a third scenario with Control Plane Nodes + InfrastructureNodes + Worker Nodes (but NO ACM on the worker nodes, and we don't need to call this one out).

swopebe:
This topic was restructured in later releases. Please see the install team about this update, which I am not sure is going in the right spot. Please look at 2.6 and 2.5.

Also consider that install is more for the hub and managed cluster is in cluster lifecycle. 2.6 and later, that is found in MCE doc. 

This is the corrected and updated sizing topic: https://github.com/stolostron/rhacm-docs/blob/2.7_stage/install/cluster_size.adoc

dockerymick:
@swopebe thanks for your suggestion and information Brandi! @JakobGray can you provide some feedback here when you have an opportunity?

install/perform_scale.adoc
```diff
@@ -35,16 +35,7 @@ The following table shows the configuration information for the clusters on the
...
-| m5.2xlarge
-| 8
-| 32
-| gp2
-| 100
-| 3 or 5 nodes
-| us-east-1
-
-| Infrastructure
+| Worker or Infrastructure
```
jnpacker, line 38: 👍

xrajesh:
@jnpacker We did not test specifically , Search components running on Infrastructure Nodes or Worker Nodes , part of search V1 . I am okay with claiming `Infrastructure or Worker Nodes` . We will qualify on specific Nodes for search V-2 . @bjoydeep FYI.

dockerymick:
@swopebe @jnpacker @eemurphy i made a couple of changes and added some comments. Let me know your thoughts when you have an opportunity

dockerymick:
Received confirmation from @JakobGray 
Screen Shot 2022-11-17 at 2 05 55 PM

@swopebe cc

JakobGray (approved):

install/cluster_size.adoc
```diff
@@ -36,7 +36,21 @@ Each {product-title} cluster is unique and the following guidelines provide samp
 
 In addition to {product-title-short}, the {ocp-short} cluster runs additional services to support cluster features. The recommended node size is three nodes.
 
-//is the second statement in line 37 neccessary since we have the Best practice notice o line 13? 
+//is the second statement in line 37 neccessary since we have the Best practice notice o line 13?
```
swopebe: Any chance we can remove these comments>

oafischer (approved):
Looks good, just make sure to sign the comments if you want to keep them.

install/cluster_size.adoc
dockerymick: Comments removed

dockerymick:
@swopebe @oafischer Thanks for the reminders and reviews. Thanks @jnpacker @JakobGray for your reviews as well! Moving forward with merging this

PR #4295, opened 2022-11-07
Remove owners file

oafischer:
https://coreos.slack.com/archives/GVDCLR7A5/p1667599929541279?thread_ts=1667589925.338969&cid=GVDCLR7A5


swopebe (approved):

cadawson:
The owner's file seems like it is a good list of who owns the files on our open repo. It is a common place for people to look for contacts in an open repo. I agree that it probably doesn't have the same importance now that we have a different repo for the community now. We can remove it for now, but we might also want to see if a `codeowners` file might be helpful for us now that we have more PRs coming in. Thanks.

cadawson (approved):

PR #4296, opened 2022-11-07
Deprecate HypershiftDeployment CRD in 2.7

oafischer:
https://github.com/stolostron/backlog/issues/25594


swopebe (changes requested):
we might need to slow down and visit deprecation and and removal process, as they are not recorded in what's new, but in the deprecation topic and then labeled in the topic of interest for the user...this system is to keep SEO up, user expectation up, and make it easier for us to follow internally.

PR #4331, opened 2022-11-15
stolostron/backlog#26868 small changes.

swopebe:
Offset folder names to remove bold and highlight, adding updating bc I was confused what maintaining meant at first


oafischer (approved):
Looks good!

PR #4332, opened 2022-11-15
Adv config and name changes stolostron/backlog#26868

oafischer:
https://github.com/stolostron/backlog/issues/26868

Adv config changes we discussed AND some renaming after learning that the old naming scheme is an obsolete technical limitation that no longer applies.

The Google script was removed because it doesn't work.


swopebe (approved):
perfection!

PR #4341, opened 2022-11-16
Adding best practice for policy generator 2.4

dockerymick:
https://github.com/stolostron/backlog/issues/25740#issuecomment-1279102380

### DIFF
--- a/governance/policy_generator.adoc
+++ b/governance/policy_generator.adoc
@@ -144,7 +144,6 @@ spec:
   name: openshift-gitops-operator
   source: redhat-operators
   sourceNamespace: openshift-marketplace
-  startingCSV: openshift-gitops-operator.v<version>
 ----
 
 To pin to a specific version of the operator, you can set the `spec.startingCSV` value to `openshift-gitops-operator.v<version>`. Replace `<version>` with your preferred version.
--- a/governance/policy_generator.adoc
+++ b/governance/policy_generator.adoc
@@ -142,9 +142,10 @@ spec:
   name: openshift-gitops-operator
   source: redhat-operators
   sourceNamespace: openshift-marketplace
+  startingCSV: openshift-gitops-operator.v<version>
 ----
 
-To pin to a specific version of the operator, you can set the `spec.startingCSV` value to `openshift-gitops-operator.v1.2.1` (replace `v1.2.1` with your preferred version).
+To pin to a specific version of the operator, you can set the `spec.startingCSV` value to `openshift-gitops-operator.v<version>`. *Best practice*: Use policy generator v1.2.3.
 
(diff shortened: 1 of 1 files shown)


oafischer (approved):
Would it be better to have the tag in a separate line?

mprahl:
@dockerymick that's not quite right because where you made the change is on the version of the example operator that we are generating a policy for.

I would probably just add a sentence under the `2.4.2. Policy generator` section that states the bundled/included Policy Generator version.

dockerymick:
@mprahl thanks for your review, i just made an update. When you have an opportunity, will you rereview?

governance/policy_generator.adoc
```diff
@@ -3,6 +3,8 @@
 
 The policy generator is a part of the {product-title} application lifecycle subscription GitOps workflow that generates {product-title} policies using Kustomize. The policy generator builds {product-title} policies from Kubernetes manifest YAML files, which are provided through a `PolicyGenerator` manifest YAML file that is used to configure it. The policy generator is implemented as a Kustomize generator plugin. For more information on Kustomize, see the link:https://kustomize.io/[Kustomize documentation]. 
 
+The latest version of the policy generator is included in the installation. The policy generator version used is v1.4.1.
```
mprahl:
```suggestion
The Policy Generator version bundled in this version of RHACM is v1.4.1. 
```

dockerymick:
@mprahl thanks for your suggestion, made an update. If you are ok with that update, i will merge and apply similar changes to the other versions

governance/policy_generator.adoc
```diff
@@ -142,9 +144,10 @@ spec:
   name: openshift-gitops-operator
   source: redhat-operators
   sourceNamespace: openshift-marketplace
+  startingCSV: openshift-gitops-operator.v<version>
```
mprahl: I would like the example to be valid without user alteration.
```suggestion
```

mprahl (approved):

governance/policy_generator.adoc
```diff
@@ -144,7 +146,7 @@ spec:
   sourceNamespace: openshift-marketplace
 ----
 
-To pin to a specific version of the operator, you can set the `spec.startingCSV` value to `openshift-gitops-operator.v1.2.1` (replace `v1.2.1` with your preferred version).
+To pin to a specific version of the operator, you can set the `spec.startingCSV` value to `openshift-gitops-operator.v<version>`. Replace `<version>` with your preferred version.
```
dockerymick: @mprahl since you removed the parameter, is this statement accurate or should it be removed?

mprahl: Yes, I think it still applies. Perhaps you can change the word `set` to `add` instead?

dockerymick:
@oafischer @mprahl thanks so much for your reviews. I applied the suggestions and ready to merge. Also making similar updates to the other versions

PR #4350, opened 2022-11-17
Deprecate HypershiftDepl CRD 2.7 stolostron/backlog#25594

oafischer:
https://github.com/stolostron/backlog/issues/25594

Let me know if there is a **Deprecated** tag too many, I added more just in case to cover all instances. Easy to remove if I wrongly added them to a topic.


swopebe (approved):
colons where needed, moved the parentheses behind title

PR #4371, opened 2022-11-22
Specify min size for assisted installer database

jhernand:
This patch documents that the minium size required by the assisted installer is 1G.

Related: https://issues.redhat.com/browse/MGMT-11700
Related: https://github.com/openshift/assisted-service/pull/4463


jhernand:
Note also that the minimum size for the filesystem, as validated by the software, is 1G, not the 100G specified here. And for the image storage is 10G, not 50G. We should probably synchronize these documentation and the software, but not sure in what direction.

swopebe (approved):

PR #4399, opened 2022-12-01
installer: deprecate ingress.sslCiphers field

ray-harris:
Signed-off-by: Ray Harris <raharris@redhat.com>


swopebe:
There is no other action the user should take???

Also, we have to note this in the spot for the doc: where?

https://github.com/stolostron/rhacm-docs/blob/2.7_stage/install/adv_config_install.adoc

I promise I will end up with a bug at some point asking what the user should do instead or asking to note the deprecation where the user comes across the component...so for deprecations, we need a bit more info.

swopebe:
According to scrum, no new action to take as it is deprecated with the sunset of UI.

PR #4428, opened 2022-12-08
Fix YAML formatting 2.7 stolostron/backlog#26993

oafischer:
https://github.com/stolostron/backlog/issues/26993

### DIFF
--- a/add-ons/volsync/volsync_replicate.adoc
+++ b/add-ons/volsync/volsync_replicate.adoc
@@ -368,6 +368,7 @@ By sharing the same S3 bucket, each `ReplicationSource` has write access to the 
 . Configure your backup policy by creating a `ReplicationSource` object that resembles the following YAML content: 
 +
 [source,yaml]
+\
 ----
 apiVersion: volsync.backube/v1alpha1
 kind: ReplicationSource
@@ -376,7 +377,7 @@ metadata:
 spec:
   sourcePVC: <source>
   trigger:
-    schedule: \*/30 * * * *"
+    schedule: "*/30 * * * *"
   restic:
     pruneIntervalDays: 14
     repository: <restic-config>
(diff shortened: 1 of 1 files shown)


add-ons/volsync/volsync_replicate.adoc
```diff
@@ -371,25 +369,24 @@ By sharing the same S3 bucket, each `ReplicationSource` has write access to the
...
-  pruneIntervalDays: 14
-  repository: <restic-config>
-  retain:
-    hourly: 6
-    daily: 5
-    weekly: 4
-    monthly: 2
-    yearly: 1
-  copyMethod: Clone
+    schedule: "*/30 * * * *" #\*
```
tesshuflower, line 379: I'm not sure what the #\* is for, it's some comment for formatting?  The previous formatting for this line appeared ok to me.

tesshuflower (approved):
/lgtm

add-ons/volsync/volsync_replicate.adoc
oafischer: Yeah I will discuss with Brandi as well and I won't merge until then, but this is what it looked like before:

image

It's an AsciiDoctor/GitHub bug.

oafischer: Adding that comment fixes it but it is definitely a workaround/hack:

image

oafischer: @tesshuflower Brandi asks if we can just remove the `*` from `schedule` since it is causing formatting issues? I'm guessing not since it's part of the configuration but just wanted to ask to be sure.

tesshuflower: @oafischer no we can't remove the `*` characters from the schedule as they are exactly what the user would specify.  Note that in the final docs these did seem to be formatted ok as far as I could tell:
Screenshot 2022-12-09 at 8 23 35 AM

swopebe (approved):

PR #4432, opened 2022-12-12
stolostron/backlog#25988 --proxy (hub--Chris tagged)

swopebe:
@cadawson will need help with this one, see issue comments and see that this is related to the hub but now that it is in MCE, that needs to change.

### DIFF
--- a/multicluster_engine/cluster_lifecycle/cluster_proxy_addon.adoc
+++ b/multicluster_engine/cluster_lifecycle/cluster_proxy_addon.adoc
@@ -6,6 +6,7 @@ In some environments, a managed cluster is behind a firewall and cannot be acces
 *Required access*: Editor
 
 To configure a cluster proxy add-on for a hub cluster and a managed cluster, complete the following steps:
+//this references hub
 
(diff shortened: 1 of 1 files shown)


xuezhaojun:
@swopebe  Hello, I'm a little lost here. Because the cluster-proxy-addon is managed by MCE operator from the code-level. 
And I checked the original issue and found it's mainly talks about another concept of proxy -- the cluster-wide proxy.

xuezhaojun:
I found these 2 concepts look very similar but actually 2 different thing:
* The cluster-proxy-addon doc(only include one addon): https://github.com/stolostron/rhacm-docs/blob/2.6_stage/multicluster_engine/cluster_lifecycle/cluster_proxy_addon.adoc
* The cluster-wide proxy configuration doc(involved all addons): https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/clusters/managing-your-clusters#enable-cluster-wide-proxy-addon

To me the second one is more related to the issue #25988

PR #4434, opened 2022-12-12
customCAConfigmap deprecation

eemurphy:
Signed-off-by: Erin Murphy <erinmurp@redhat.com>


swopebe (approved):
I thought we did this but OK, approved. Anything that is mentioned in the doc has the deprecated tag on it.

PR #4442, opened 2022-12-13
stolostron/backlog#25868

swopebe:
--election leader GA doc WIP


applications/config_leader_election.adoc
```diff
@@ -0,0 +1,26 @@
+[#config_leader_election]
+= Configuring leader election
+
+You can reduce your etcd load by configuring `leaseDuration`, `renewDeadline`, or `retryPeriod` of controllers.
+
```
swopebe, line 5: Would like maybe another line or two about why user wants to do this.

applications/config_leader_election.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+[#controller-change]
+== Editing the controller
+
```
swopebe, line 18: Was not sure what to call this procedure...adding the flag (though I see no flag)? The topic is already titled configuring so we cannot repeat it.

applications/config_leader_election.adoc
swopebe: I need any not-obvious prereqs and the role is usually: Editor, Cluster adniminstrator, etc.... that we try to add to the docs. (RBAC requirement)

applications/config_leader_election.adoc
```diff
@@ -0,0 +1,26 @@
...
+- leaseDuration:  137 seconds
+- renewDeadline: 107 seconds
+- retryPeriod: 26 seconds
+
+See the following steps to set the `multicluster-operators-application`, `multicluster-operators-channel`, `multicluster-operators-standalone-subscription`, or `multicluster-operators-hub-subscription` controllers:
+
+. Run `oc get csv -n open-cluster-management` to get and edit the {product-title-short} `csv` file. Use the `advanced-cluster-management` name, for example. The command resembles `oc edit csv -n open-cluster-management advanced-cluster-management.v2.7.0`.
+
+. Locate and specify a controller and insert a `- command:`  for a flag with a specified time in seconds. Search `name: <controller>`, for example: `name: multicluster-operators-hub-subscription`
```
swopebe: I didn't really understand this part.

swopebe: and example would be good (with the seconds)

applications/config_leader_election.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+. Save the file. The controller restarts to apply the flag.
```
swopebe: we don't need to run an apply command?

applications/config_leader_election.adoc
```diff
@@ -0,0 +1,26 @@
...
+
+. Run `oc get csv -n open-cluster-management` to get and edit the {product-title-short} `csv` file. Use the `advanced-cluster-management` name, for example. The command resembles `oc edit csv -n open-cluster-management advanced-cluster-management.v2.7.0`.
+
+. Locate and specify a controller and insert a `- command:`  for a flag with a specified time in seconds. Search `name: <controller>`, for example: `name: multicluster-operators-hub-subscription`
+
+. Save the file. The controller restarts to apply the flag. 
+
+. Repeat this procedure for each controller that you want to change.
+
+See the following example output of a successful flag:
```
swopebe: example would be great.

applications/config_leader_election.adoc
lennysgarage: Regarding why a user may want to do this is not just for reducing etcd load, that was just the reasoning for choosing the default values. A user may want to increase/decrease the time a controller takes to acquire the LeaderElection lock to. This changes how often the controllers will make requests to choose a new leader in case of failure. A user may want to decrease the default time so in the case of a failure a leader is chosen quicker with the tradeoff being a decrease in overall system performance.

The required access for this would be, `Cluster administrator`

applications/config_leader_election.adoc
lennysgarage: Since what this is doing is editing the controller's flag options I think editing the controller is fine or possibly adding controller flags?

applications/config_leader_election.adoc
lennysgarage: Since we're editing the csv directly when we save and exit, it will automatically restart the controller without needing to `oc apply`

applications/config_leader_election.adoc
lennysgarage: An example of the `multicluster-operators-hub-subscription` controller with all flags default time doubled:
```
containers:
- command:                                                                                                                                                                 
  - /usr/local/bin/multicluster-operators-subscription
  - --sync-interval=60
  - --retry-period=52
  - --leader-election-lease-duration=274
  - --renew-deadline-114
```

applications/config_leader_election.adoc
lennysgarage: Okay so we're not actually inserting  `- comamnd`, but a command under this section.
Under each of the above containers they already have a `- comamnd:` section and so we're simply adding new commands under this section. Any combination of `leaseDuration`, `renewDeadline`, `retryPeriod` can be added. So an example change to the `retryPeriod` of the `multicluster-operators-hub-subscription` controller would be first locating the controller:

0. Run the edit command to enter the csv
1. Locate controller we want to modify, searching `name: <controller>`
2. Head down to the `containers` section under the controller and insert a flag under `- command`
3. Inserting the `retryPeriod` flag itself doubling the default time it takes to retry for acquiring leaderElection.:
```
containers:
- command:                                                                                                                                                                 
  - /usr/local/bin/multicluster-operators-subscription
  - --sync-interval=60
  - --retry-period=52
```

applications/config_leader_election.adoc
swopebe: Ok, so I am not following and the user will not follow. We don't write "under" because that alienates users who cannot see. I will try to rewrite a bit.

applications/config_leader_election.adoc
swopebe: those changes are in

applications/config_leader_election.adoc
swopebe: changed

applications/config_leader_election.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+See the following steps to change the `multicluster-operators-application`, `multicluster-operators-channel`, `multicluster-operators-standalone-subscription`, or `multicluster-operators-hub-subscription` controllers:
+
+. Run `oc get csv -n open-cluster-management` to locate the {product-title-short} `csv` file.
+
+. Edit the `csv` file by adding the `advanced-cluster-management` name to the `oc edit` command. For example, the command resembles `oc edit csv -n open-cluster-management advanced-cluster-management.v2.7.0`.
+
+. Locate controller by searching for `<controller>`.
+
+. From the containers section in the controller, insert a `- command` flag. For instance, instert `RetryPeriod`.
```
swopebe: @lennysgarage we need to list these out in steps and this part is a bit confusing to me, so where does searching controller fit in? See my edits with what you gave me yesterday.

lennysgarage: Okay these steps are good, the searching of the controller is just for locating where exactly to insert the desired commands. It's the first step after opening the csv and it will only appear once in the csv so by heading to `- name: <controller>` they will be in the correct place.

swopebe: Ok, thanks @lennysgarage -- if you are good just approve the PR for me, then @dockerymick can approve as peer review.

applications/config_leader_election.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+To configure `LeaderElection`, you change the following default values:
+
+- `leader-election-lease-duration: 137 seconds`
+- `renew-deadline: 107 seconds`
+- `retry-period: 26 seconds`
+
+See the following steps to change the `multicluster-operators-application`, `multicluster-operators-channel`, `multicluster-operators-standalone-subscription`, or `multicluster-operators-hub-subscription` controllers:
+
+. Run `oc get csv -n open-cluster-management` to locate the {product-title-short} `csv` file.
```
dockerymick: I recommend the following update:

Run the following command to locate the {product-title-short} `csv` file:

```
oc get csv -n open-cluster-management
```

or for consistency:

Locate the {product-title-short} `csv` file with the following command:

applications/config_leader_election.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+. Edit the `csv` file by adding the `advanced-cluster-management` name to the `oc edit` command. For example, the command resembles `oc edit csv -n open-cluster-management advanced-cluster-management.v2.7.0`.
```
dockerymick: I recommend that you change: 

For example, the command resembles `oc edit csv -n open-cluster-management advanced-cluster-management.v2.7.0`.

to

Your command might resemble the following example command:

 ```
oc edit csv -n open-cluster-management advanced-cluster-management.v2.7.0
```

File: applications/config_leader_election.adoc [hunk 9]
```diff
@@ -0,0 +1,41 @@
...
+
+. Locate controller by searching for `- name: <controller>`.
+
+. From the containers section in the controller, insert a `- command` flag. For instance, insert `RetryPeriod`.
```
dockerymick, line 36: is _containers_ a parameter name? if yes, I recommend that you monospace it

applications/config_leader_election.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+. Save the file. The controller automatically restarts to apply the flag. 
+
+. Repeat this procedure for each controller that you want to change.
+
+See the following example output of a successful edit to the `-command`, where the `retryPeriod` flag doubles the previously mentioned default time to `52`, which is allotted to retry acquiring `leaderElection`:
+containers:
```
dockerymick: Remove `containers:` on line 34

dockerymick (changes requested):

applications/config_leader_election.adoc
swopebe: I don't like command in the same line twice, but will break it out for ya.

applications/config_leader_election.adoc
swopebe: not sure...it is a section, we could get away with both I think.

applications/config_leader_election.adoc
swopebe: Sometimes we embed command in the step, but I will break it out. It is more UX friendly.

applications/config_leader_election.adoc
swopebe: I am going to keep as is bc I don't have the example or demo -- it wasn't on the playback. The developer called it a section.

applications/config_leader_election.adoc
swopebe: changed

applications/config_leader_election.adoc
swopebe: changed

swopebe:
I think I got all the discussion and changes addressed. Dev approved. Waiting on peer review follow up.

dockerymick (approved):
Looks good to me! Thanks Brandi !

PR #4449, opened 2022-12-15
VolSync least priviliged 2.7 stolostron/backlog#26896

oafischer:
https://github.com/stolostron/backlog/issues/26896

### DIFF
--- a/add-ons/volsync/volsync_replicate.adoc
+++ b/add-ons/volsync/volsync_replicate.adoc
@@ -413,7 +413,7 @@ For additional information about the backup options, see https://volsync.readthe
 oc annotate namespace <namespace> volsync.backube/privileged-movers=true
 ----
 
-Replace `namespace` with the name of your namespace.
+Replace `<namespace>` with the name of your namespace.
 
 [#restore-restic-backup]
 === Restoring a restic backup
@@ -559,4 +559,4 @@ Replace `my-vsc` with the name of the `VolumeSnapshotClass` to use, if you speci
 oc annotate namespace <namespace> volsync.backube/privileged-movers=true
 ----
 
-Replace `namespace` with the name of your namespace.
+Replace `<namespace>` with the name of your namespace.
--- a/add-ons/volsync/volsync_replicate.adoc
+++ b/add-ons/volsync/volsync_replicate.adoc
@@ -407,6 +407,14 @@ Best practice: Use `Clone` for the value of `CopyMethod` to ensure that a point-
 +
 For additional information about the backup options, see https://volsync.readthedocs.io/en/latest/usage/restic/index.html#backup-options[Backup options] in the VolSync documentation.
 
+*Note:* restic movers run without root permissions by default. If you want to run restic movers as root, run the following command to add the elevated permissions annotation to your namespace.
++
+----
+oc annotate namespace <namespace> volsync.backube/privileged-movers=true
+----
++
+Replace `namespace` with the name of your namespace.
+
 [#restore-restic-backup]
 === Restoring a restic backup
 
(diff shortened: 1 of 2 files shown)


add-ons/volsync/volsync_replicate.adoc
```diff
@@ -407,6 +407,14 @@ Best practice: Use `Clone` for the value of `CopyMethod` to ensure that a point-
 +
 For additional information about the backup options, see https://volsync.readthedocs.io/en/latest/usage/restic/index.html#backup-options[Backup options] in the VolSync documentation.
 
+*Note:* restic movers run without root permissions by default. If you want to run restic movers as root, run the following command to add the elevated permissions annotation to your namespace.
```
swopebe: Let's check "restic" here, meaning it is either capitalized or in monospace after a note. Last time I checked per IBM style it is colon: Caps (I think English rules say otherwise).

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -407,6 +407,14 @@ Best practice: Use `Clone` for the value of `CopyMethod` to ensure that a point-
...
+
+----
+oc annotate namespace <namespace> volsync.backube/privileged-movers=true
+----
+
+Replace `namespace` with the name of your namespace.
```
swopebe: Be clear which one to replace here...there are two, use <> to define the variable:


Replace `<namespace>` with your namespace.

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -544,3 +552,11 @@ The `capacity` specifies the size of the destination volume, which must be large
...
 +
 Replace `my-vsc` with the name of the `VolumeSnapshotClass` to use, if you specified `Snapshot` as your `copyMethod`. This is not required for other types of `copyMethod`. If not included, the system default `VolumeSnapshotClass` is used. 
+
+*Note:* Rclone movers run without root permissions by default. If you want to run Rclone movers as root, run the following command to add the elevated permissions annotation to your namespace.
+
+----
+oc annotate namespace <namespace> volsync.backube/privileged-movers=true
+----
+
+Replace `namespace` with the name of your namespace.
```
swopebe: same change here

swopebe: Do we already know what Rclone is?

If the user has not been introduced to it:

**Note:** Rclone is the command-line program that is used to manage files. By default, Rclone movers run without root permissions. To run Rclone movers as root, run the following command to add the elevated permissions annotation to your namespace:

swopebe (approved):
I wanted to be able to peer review, I am approving but with these comments and recommended changes before merge.

add-ons/volsync/volsync_replicate.adoc
oafischer: It's in the rclone topic at the very end so I assume the user would know by this point?

add-ons/volsync/volsync_replicate.adoc
oafischer: @tesshuflower 

It it like macOS where the first letter is always lower case?

tesshuflower: @oafischer I think we can use `Restic` capitalized - I've seen it both ways in their docs, but they do capitalize it on their main start page: https://restic.readthedocs.io/en/latest/

oafischer: Awesome thanks for the quick response, I'll update it!

tesshuflower (changes requested):
Could we have some sort of summary about the changes in addition to the note that says they'll run as root by default?

Something like:

The Restic and Rclone movers have been enhanced to run with reduced permissions. Previously the movers would run with a service account that used an elevated SCC in order to run as root with enhanced capabilities. The new movers will run as non-root by default and drop all SE Linux capabilities.

This means the mover pods for Restic and Rclone will not require elevating permissions on the namespace if Pod Security Standards require that pods in the namespace run with restricted permissions.

oafischer:
We can add that info to the What's new section, how does that sound? We should try not to clutter procedures with concept information, unless the user needs to know it to complete the task. If that doesn't work then I can see if maybe we can create a space for more concept information somewhere.

tesshuflower:
> We can add that info to the What's new section, how does that sound? We should try not to clutter procedures with concept information, unless the user needs to know it to complete the task. If that doesn't work then I can see if maybe we can create a space for more concept information somewhere.

Sure, that sounds good

oafischer:
Added info to what's new.

tesshuflower (approved):
/lgtm

PR #4452, opened 2022-12-15
https://github.com/stolostron/backlog/issues/26607, 2.7 remove (peer/tech edit)

swopebe:
…ased on a 2.6 announcement and alternative from dev team, @cadawson FYI/


cadawson (approved):
Looks good. Thanks.

cadawson:
This was confirmed for removal by @qiujian16 in the scrum meeting. It was intended to be a procedure that internal Red Hat teams use, rather than customers.

cadawson:
Hi @qiujian16. Will you please take a look at this change for removing the "Deploying workloads using ManifestWork" topic? We talked about it during the scrum meeting. Is there a different procedure that customers should use for this that we can link to from the removal and deprecations files? Thanks!

swopebe:
In that case, I will remove from 2.7 and simply add it to the removal table for 2.6. But I still need info about other mentions of manifest work. Do we remove all mentions of it? (you can search to see it)

PR #4464, opened 2022-12-19
Peer review stolostron/backlog#25897

swopebe:


install/adv_config_install.adoc
```diff
@@ -1,7 +1,7 @@
 [#advanced-config-hub]
 = MultiClusterHub advanced configuration 
 
-{product-title} is installed using an operator that deploys all of the required components. {product-title-short} can be further configured during or after installation by adding one or more of the following attributes to the `MultiClusterHub` custom resource:
+{product-title} is installed by using an operator that deploys all of the required components. {product-title-short} can be further configured during or after installation by adding one or more of the following attributes to the `MultiClusterHub` custom resource:
 
 [#cosole-config]
```
oafischer: Not part of your changes but missing n in console. If you fix it, make sure it doesn't break any links.

install/adv_config_install.adoc
```diff
@@ -22,13 +22,13 @@ spec:
       enabled: true
 ----
 
-Alternatively, you can run the following command. Replace `namespace` with the name of your project and the value.
+Alternatively, you can run the following command. Replace `namespace` with the name of your project and the value:
```
oafischer, line 25: Should the "Replace..." sentence go after the command? That's how we seem to do it across most of the doc.

oafischer (approved):
Couple comments but I'll approve in advance. Feel free to address them as you see fit. Thanks!

install/adv_config_install.adoc
swopebe: Actually, this is the second PR so that is my change....thanks for catching that. No links, yet.

install/adv_config_install.adoc
swopebe: @oafischer that actually drives me nuts that we do that. I never have.... my philosophy is the user needs to know before they copy the command that they need to replace something. Unless the list of replacements is long, I do them first for UX...I have copied and ran commands without glancing under them for info. This is not my preference...it's a trend from one person's preference on the team and I asked about it...it was pretty prevalent so we kept it.

oafischer: I've been doing it that way for months since it's what I saw in the docs. Should we agree to switch to this format? Maybe we can go through the doc and change it everywhere after GA. And let's make sure to document that in our internal docs as well if we move that direction.

swopebe: @oafischer I am fine with people doing what fits best here...if it is one change, I recommend bringing that up to the top. If multiple, a list after is fine.

Run _____ command to ___, changing `<cluster-namespace>` to the namespace for your cluster: 

That is a simple example where having one change after the command is silly, IMO. I merge those.

oafischer: That's good guidance, thanks!

PR #4467, opened 2022-12-20
Deprecated resources--upgrade known issue

swopebe:


release_notes/known_issues.adoc
```diff
@@ -53,6 +53,27 @@ If this happens, you can either find the specified section manually or complete
 [#installation-known-issues]
 == Installation known issues
 
+[#upgrade-remaining-resource]
+=== Deprecated resources remain after upgrade to Errata versions
```
cadawson: "Errata versions" makes me think of 2.4.5 > 2.4.6, for example. I know technically 2.4-2.5 is a y-stream errata, but I don't think we refer to them now as Errata versions. Would it be clearer to say "later versions" or similar? If there is a reason for it, it might be good for us to know what we mean by errata versions.

release_notes/known_issues.adoc
```diff
@@ -53,6 +53,27 @@ If this happens, you can either find the specified section manually or complete
...
+//2.6X: 26987
+
+After you upgrade from 2.4.x to 2.5.x, and then to 2.6.x, the following deprecated resources in the managed cluster namespace might remain. You need to manually delete these deprecated resources if version 2.6.x was upgraded from 2.4.x:
```
cadawson: Suggestion: Change "..the following deprecated resources in the managed cluster namespace might remain." to "..some deprecated resources might remain in the managed cluster namespace." "The following" implies to me that it is directly introducing the list.

release_notes/known_issues.adoc
```diff
@@ -53,6 +53,27 @@ If this happens, you can either find the specified section manually or complete
...
+-klusterlet-addon-appmgr
+-klusterlet-addon-certpolicyctrl
+-klusterlet-addon-crds
+-klusterlet-addon-iampolicyctrl
+-klusterlet-addon-operator
+-klusterlet-addon-policyctrl
+-klusterlet-addon-workmgr
+----
+
+**Best practice:** Wait 30 minutes or more before you upgrade from version 2.5.x to version 2.6.x.
```
cadawson: Is this a best practice, or an important note? It sounded like it might be a real issue if it is not done, but maybe it doesn't apply to every case.

cadawson (changes requested):
Looks good! A few questions/suggestions, but nothing show-stopping. Approved after your consideration. Thanks!

release_notes/known_issues.adoc
swopebe: That is mentioned in the content, headers need to be simple. Is your feedback on the header or the content?

Simple is best for headers and SEO.

cadawson: My comment was on the header. Thinking "Deprecated resources remain after upgrade to later versions" might be simpler. I don't know that we refer to them as "Errata versions" anywhere. Maybe we are, and I missed it.

release_notes/known_issues.adoc
zhiweiyin318: I think this is an important note. the user may meet this issue if the waiting time is too short.

swopebe:
Please see the other changes are implemented. I changed to Errata release -- instead of versions?

cadawson (approved):
I think the important note was removed from the end. Was that intentional? It might be good to add that back, unless you had a different opinion from someone. This line: 
*Important:* Wait 30 minutes or more before you upgrade from version 2.5.x to version 2.6.x.
Otherwise, it looks good. Thanks!

PR #4473, opened 2022-12-21
Updating the RBAC topic

dockerymick:
https://github.com/stolostron/backlog/issues/26917


access_control/rbac_implementation.adoc
```diff
@@ -0,0 +1,524 @@
...
+** <<console-api-rbac-gov,Console and API RBAC table for governance lifecycle>>
+* <<observability-RBAC,Observability RBAC>>
+** <<console-api-rbac-obs,Console and API RBAC table for observability lifecycle>>
+
+[#cluster-lifecycle-RBAC]
+== Cluster lifecycle RBAC
+
+View the following cluster lifecycle RBAC operations:
+
+- Create and administer cluster role bindings for all managed clusters. For example, create a cluster role binding to the cluster role `open-cluster-management:cluster-manager-admin` by entering the following command:
```
oafischer, line 22: I would try playing around with formatting here to fix the bold issues.

dockerymick: Thanks Oliver!

oafischer (approved):
lgtm

access_control/rbac.adoc
```diff
@@ -64,533 +46,4 @@ Some product resources are cluster-wide and some are namespace-scoped. You must
 
 * If a user does not have role access to a cluster, the cluster name is not visible. The cluster name is displayed with the following symbol: `-`.
```
ckandag: The cluster name is displayed with the following symbol: `-`.   

This is not always correct, its mostly just not shown/displayed. e.g. when the user goes to view main Clusters overview - Only the clusters, user has access to, are displayed in the page.   The ones that are not accessible are simply Not shown at all.

There may be one or two UI pages where its displayed with '_'.

I would suggest changing to indicate something like ,    if the cluster name is showing as "-" then user may not have to it

ckandag (changes requested):

dockerymick:
@ckandag thanks for your review. i have made updates based on your comment. Merging this PR

PR #4476, opened 2023-01-04
no issue - add mce rel not to main file

cadawson:


cadawson:
Hi @swopebe. It looks like the release notes for MCE content is not being pulled in to the build. I think that is because it is not in the MCE main file. It was in a release notes-specific main file, but that is not pulled in because the release notes is not in its own book. This PR should resolve the issue. I am hoping that we can get away with 2 folders named release_notes. I think we will be ok because they are in separate books. If this wasn't where you planned on having them, feel free to move them. 
You might have been planning to get to this, but I don't want a group of errors building up while these changes are being made to the cluster docs. Thanks!

swopebe (approved):

PR #4477, opened 2023-01-05
Add broker known issue 2.7 stolostron/backlog#27008

oafischer:
https://github.com/stolostron/backlog/issues/27008

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -996,3 +996,9 @@ Globalnet is not supported with Red Hat OpenShift Data Foundation disaster recov
 //2.7:26928
 
 The length of cluster names for Azure deployments cannot exceed 13 characters.
+
+[#certificates-prevent-connection-broker]
+=== Self-signed certificates prevent connection to broker
+//2.7:27008
+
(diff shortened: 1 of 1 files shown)


cadawson (approved):
Looks good. A couple of quick questions: 
1. Does this always happen, or is it occasional? In the title it sounds like always, but in the text, it says "might". Just provides someone with expectations, if it does or doesn't always happen. 
2. Will they know where to set the InsecureBrokerConnection to true?

oafischer:
It happens sometimes so I'll change the title to reflect that!

Second question might be better suited for @skitt

skitt:
Regarding the second question, perhaps we can add a link to the Submariner customizations section, and add a mention there — the result would be

```
apiVersion: submarineraddon.open-cluster-management.io/v1alpha1
kind: SubmarinerConfig
metadata:
   name: submariner
   namespace: <managed-cluster-namespace>
spec:
   insecureBrokerConnection: true
```

oafischer:
Thanks for providing the sample @skitt! I think it's better to keep everything together, so I'll just add that sample to the known issue. Let me know if the changes look good.

skitt (approved):

PR #4486, opened 2023-01-06
Adding known issue

dockerymick:

After reading the statement that you provided, it seemed more appropriate to add this as a known issue

### DIFF
--- a/governance/custom_template.adoc
+++ b/governance/custom_template.adoc
@@ -166,7 +166,7 @@ spec:
 [#lookup-func]
 === lookup function
 
-The `lookup` function returns the Kubernetes resource as a JSON compatible map. Note that if the requested resource does not exist, an empty map is returned. If the resource does not exist and the value is provided to another template function, you might get the following error, `invalid value; expected string`.
+The `lookup` function returns the Kubernetes resource as a JSON compatible map. If the requested resource does not exist, an empty map is returned. If the resource does not exist and the value is provided to another template function, you might get the following error: `invalid value; expected string`.
 
 View the following syntax for the function:
 
--- a/governance/custom_template.adoc
+++ b/governance/custom_template.adoc
@@ -166,7 +166,7 @@ spec:
 [#lookup-func]
 === lookup function
 
-The `lookup` function returns the Kubernetes resource as a JSON compatible map. Note that if the requested resource does not exist, an empty map is returned.
+The `lookup` function returns the Kubernetes resource as a JSON compatible map. Note that if the requested resource does not exist, an empty map is returned. If the resource does not exist and the value is added to another template function, you might get the following error, `invalid value; expected string`.
 
 View the following syntax for the function:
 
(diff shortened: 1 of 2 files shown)


governance/custom_template.adoc
```diff
@@ -166,7 +166,7 @@ spec:
 [#lookup-func]
 === lookup function
 
-The `lookup` function returns the Kubernetes resource as a JSON compatible map. Note that if the requested resource does not exist, an empty map is returned.
+The `lookup` function returns the Kubernetes resource as a JSON compatible map. Note that if the requested resource does not exist, an empty map is returned. If the resource does not exist and the value is added to another template function, you might get the following error, `invalid value; expected string`.
```
mprahl:
```suggestion
The `lookup` function returns the Kubernetes resource as a JSON compatible map. Note that if the requested resource does not exist, an empty map is returned. If the resource does not exist and the value is provided to another template function, you might get the following error, `invalid value; expected string`.
```

governance/custom_template.adoc
```diff
@@ -166,7 +166,9 @@ spec:
 [#lookup-func]
 === lookup function
 
-The `lookup` function returns the Kubernetes resource as a JSON compatible map. Note that if the requested resource does not exist, an empty map is returned. View the following syntax for the function:
+The `lookup` function returns the Kubernetes resource as a JSON compatible map. Note that if the requested resource does not exist, an empty map is returned. If the resource does not exist and the value is provided to another template function, you might get the following error, `invalid value; expected string`.
```
oafischer: I think we can leave out the "Note that" unless we want to break it out into a special tag.

oafischer: Should there be a colon instead of comma after error?

oafischer (approved):
Couple of comments to consider but I'll leave the decisions up to you, approving.

dockerymick:
Thanks for your review and approval! I made updates based on your review. Merging this PR

release_notes/known_issues.adoc
```diff
@@ -926,6 +926,11 @@ While installation into an ARM environment is supported, operators that are depl
 * link:https://github.com/stolostron/policy-collection/blob/main/stable/SI-System-and-Information-Integrity/policy-imagemanifestvuln.yaml[{product-title-short} policy for the Quay Container Security Operator]
 * link:https://github.com/stolostron/policy-collection/blob/main/stable/CA-Security-Assessment-and-Authorization/policy-compliance-operator-install.yaml[{product-title-short} policy for the Compliance Operator] 
 
+[#lookup-template-error]
```
mprahl, line 929: @dockerymick, I wouldn't consider this a known issue. This is just warning the user of valid behavior that they may not expect.

governance/custom_template.adoc
dockerymick: I think either option is fine, however I agree that a colon is better

release_notes/known_issues.adoc
dockerymick: Spoke through Slack and received clarification. Agreed to remove this as a known issue

PR #4495, opened 2023-01-11
Removing the known issue for the lookup function

dockerymick:
https://github.com/stolostron/rhacm-docs/pull/4486/files#r1064964836


dockerymick:
Reverting change from previous [PR](https://github.com/stolostron/rhacm-docs/pull/4486). Simple removal that was discussed during the doc call on 01/09 and confirmed by Matt

PR #4497, opened 2023-01-11
Warning about lookup template 2.6

dockerymick:

### DIFF
--- a/governance/custom_template.adoc
+++ b/governance/custom_template.adoc
@@ -169,7 +169,7 @@ spec:
 
 The `lookup` function returns the Kubernetes resource as a JSON compatible map. If the requested resource does not exist, an empty map is returned. If the resource does not exist and the value is provided to another template function, you might get the following error: `invalid value; expected string`.
 
-*Best practice:* Use the `default` template function, so the correct type is provided to later template functions.
+*Note:* Use the `default` template function, so the correct type is provided to later template functions. See the _Open source community functions_ section.
 
 View the following syntax for the function:
 
--- a/governance/custom_template.adoc
+++ b/governance/custom_template.adoc
@@ -167,7 +167,11 @@ spec:
 [#lookup-func]
 === lookup function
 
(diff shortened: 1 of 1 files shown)


governance/custom_template.adoc
```diff
@@ -167,7 +167,11 @@ spec:
 [#lookup-func]
 === lookup function
 
-The `lookup` function returns the Kubernetes resource as a JSON compatible map. Note that if the requested resource does not exist, an empty map is returned. View the following syntax for the function:
+The `lookup` function returns the Kubernetes resource as a JSON compatible map. If the requested resource does not exist, an empty map is returned. If the resource does not exist and the value is provided to another template function, you might get the following error: `invalid value; expected string`.
```
cadawson, line 170: ...you might see the following error message: ...

governance/custom_template.adoc
```diff
@@ -167,7 +167,11 @@ spec:
...
+
+*Best practice:* Use the `default` template function, so the correct type is provided to later template functions.
```
cadawson: This might be my ignorance about the topic, but are you using the `default` template function instead of the `lookup` function? Maybe you use the default template function earlier in the process earlier so the lookup function does not have a problem later? This might be clear, but I am not sure.

cadawson (approved):
Looks good. A couple of thoughts added. Thanks!

governance/custom_template.adoc
dockerymick: updated

governance/custom_template.adoc
dockerymick: @cadawson Good point, will get more clarity

mprahl (approved):
Looks good to me!

governance/custom_template.adoc
dockerymick: Gained clarification. The `default` function is mentioned later in the Open source community functions section: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.6/html-single/governance/index#open-source-community-functions

PR #4544, opened 2023-01-17
stolostron/backlog#27130 Update upgrade_hub.adoc

swopebe:


install/upgrade_hub.adoc
```diff
@@ -46,7 +46,10 @@ While it is upgrading, the `Status` field shows `Updating`. After upgrading is c
 
 . Now that the _Upgrade Status_ is *Up to date*, click the value in the _Channel_ field to edit it.  
 
-. Select the channel for the next available feature release. *Deprecated:* `release-2.4` and `release-2.3` channels do not receive updates, you must use the `stable-2.0` channel. To import, you must use the `stable-2.0` channel of the klusterlet operator for {product-version}. You cannot skip channels when upgrading. For example, you cannot skip versions 2.2.z through 2.4. 
+. Select the channel for the next available feature release. *Deprecated:* `release-2.4` and `release-2.3` channels do not receive updates, you must use the `stable-2.0` channel. To import, you must use the `stable-2.0` channel of the klusterlet operator for {product-version}.
```
swopebe: asking for a review of this section from the install team.

cadawson: Suggestion: *Deprecated:* The `release-2.4` and `release-2.3` channels do not receive updates. You must use the `stable-2.0` channel to receive updates. (or similar). 

My thought is to limit the deprecated sentence to the deprecated content. That is all I have. Thanks!

cadawson (approved):
One small suggestion. Looks good! Thanks!

install/upgrade_hub.adoc
JakobGray: (Copying comment over)
The emphasis on not skipping versions looks good to me. The whole comment following deprecated seems out of place though. Looking at the history it is coming from a doc issue about Managed OSD Klusterlet Operator. That isn't clear in this context and I think it would be better off removed entirely.

PR #4545, opened 2023-01-17
stolostron/backlog#27130

swopebe:


install/upgrade_hub.adoc
```diff
@@ -46,7 +46,10 @@ While it is upgrading, the `Status` field shows `Updating`. After upgrading is c
 
 . Now that the _Upgrade Status_ is *Up to date*, click the value in the _Channel_ field to edit it.  
 
-. Select the channel for the next available feature release. *Deprecated:* `release-2.4` and `release-2.3` channels do not receive updates, you must use the `stable-2.0` channel. To import, you must use the `stable-2.0` channel of the klusterlet operator for {product-version}. You cannot skip channels when upgrading. For example, you cannot skip versions 2.2.z through 2.4. 
+. Select the channel for the next available feature release. *Deprecated:* `release-2.4` and `release-2.3` channels do not receive updates, you must use the `stable-2.0` channel. To import, you must use the `stable-2.0` channel of the klusterlet operator for {product-version}.
```
swopebe, line 49: asking for a review of this section from the install team.

cadawson (approved):
Only suggestion is the same as the other one: *Deprecated:* `release-2.4` and `release-2.3` channels do not receive updates. You must use the `stable-2.0` channel to receive updates. (or similar). 
Thanks!

swopebe:
@cadawson -- the deprecated tag is on that from previous release (@JakobGray can provide insight if that should be removed.)

That specific line is related to the UI process: **Select the channel for the next available feature release.** Are you saying that is the spot that is causing the issue?

install/upgrade_hub.adoc
JakobGray: The emphasis on not skipping versions looks good to me. The whole comment following **deprecated** seems out of place though. Looking at the history it is coming from a doc issue about Managed OSD Klusterlet Operator. That isn't clear in this context and I think it would be better off removed entirely.

swopebe: Ok, so I am getting that we should just remove the Deprecated stuff and that the simple change to the skip channel guidance is fine.

I wonder if something in the operator hub can call this out if it remains something unknown to the user.

Gonna merge and build.

PR #4581, opened 2023-01-20
stolostron/backlog#26941

swopebe:
-- main file changes


swopebe:
used to get rid of main.adoc files that are not necessary and potentially complicating build

PR #4600, opened 2023-01-23
stolostron/backlog#26801 architecture RHACM only (not MCE)

swopebe:


about/architecture.adoc
```diff
@@ -21,7 +19,9 @@ The _hub_ cluster is the common term that is used to define the central controll
 
 Additionally, you can enable _observability_ on your hub cluster to monitor metrics from your managed clusters across your cloud providers.
 
-The hub cluster uses the `MultiClusterHub` operator to manage, upgrade, and install hub cluster components. The hub cluster aggregates information from multiple clusters by using an asynchronous work request model and search collectors. With a graph database, the hub cluster maintains the state of clusters and applications that run on it. The _local cluster_ is the term used to define a hub cluster that is also a managed cluster, discussed in the next section.
+The {product-title-short} hub cluster uses the `MultiClusterHub` operator to manage, upgrade, and install hub cluster components and runs in the `open-cluster-management` namespace. The hub cluster aggregates information from multiple clusters by using an asynchronous work request model and search collectors. With a graph database, the hub cluster maintains the state of clusters and applications that run on it. 
+
+The _local cluster_ is the term used to define a hub cluster that is also a managed cluster, discussed in the following sections.
```
oafischer, line 24: Thanks for highlighting this in its own paragraph! Very useful to know for the hosted control planes section.

oafischer (approved):
lgtm

about/architecture.adoc
```diff
@@ -53,14 +53,14 @@ See link:../applications/app_management_overview.adoc#managing-applications[Mana
...
 
 [#observability-arch]
 == Observability
 
-The _Observability_ component collects and reports the status and health of the {ocp-short} version 4.x, or later, managed clusters to the hub cluster. You can create custom alerts to inform you of problems with your fleet of managed clusters. Because it requires configured persistent storage, observability must be enabled after the {product-title-short} installation. You also have access to the Grafana dashboard. 
+The _Observability_ component collects and reports the status and health of the {ocp-short} version 4.x, or later, managed clusters to the hub cluster. You can create custom alerts to inform you of problems with your managed clusters. Because it requires configured persistent storage, Observability must be enabled after the {product-title-short} installation. You also have access to the Grafana dashboard.
```
dockerymick: Observability should be lower-cased. It is fine for now. I will create an issue to verify consistent references

swopebe: @dockerymick another time in that paragraph I think it wasn't. As long as we are consistent....we are doing the same with app and other areas. We will discuss for 2.8.

(I think we don't need an issue but we need to add it to the arch call mid-Feb because we are not all doing the same thing. Also klusterlet was supposed to be lower case and people are writing it uppercase and now it is both. lol)

dockerymick: Sounds good! Thanks for maintaining consistency throughout the doc!

about/architecture.adoc
jnpacker: I would take this out. `With a graph database,`

about/architecture.adoc
jnpacker: `You also have access to the Grafana dashboard. ` Is this a statement?  We could just add to the first sentence,  `, managed clusters to the hub cluster,  visible via Grafana dashboards.`

Also @subbarao-meduri please comment, but we have OCP 3.11 metrics dashboards which implies we collect health info on them (is this true)?

PR #4620, opened 2023-01-26
Updating certificate management ingress

dockerymick:
https://github.com/stolostron/backlog/issues/27009


release_notes/deprecate_remove.adoc
```diff
@@ -225,6 +225,12 @@ A _removed_ item is typically function that was deprecated in previous releases
 |===
 |Product or category | Affected item | Version | Recommended action | More details and links
 
+| Governance
+| Replacing certificate management ingress is no longer supported.
```
gparvin: Can you reword to "Replacing the management ingress certificate is no longer supported."?
Thanks

dockerymick: @gparvin  how about 

Certificate management ingress is no longer supported.

gparvin: That makes it sound like we have something called the "Certificate management ingress" which is not the case.

Here are 2 different ways of describing what we did in 2.7 -- not intended for use in the doc -- more for context to help guide what you think we should say.
1. In ACM 2.7 the management ingress available in previous releases has been removed. One impact to removing the management ingress is you can no longer customize the certificate used by the route associated with that ingress.  If you customized that certificate, you no longer need to manage that custom certificate and it should be removed.

2. The management ingress component of ACM has been removed because the ACM console has been fully integrated into the OCP console.  You can no longer access the ACM console separately through the management ingress.  This means the only certificate customization is done in OCP and the custom certificate for the ACM management ingress is no longer needed.  

Maybe these descriptions make it more clear what we did.

dockerymick:
@gparvin thanks for your suggestion. When you have an opportunity, will you re-review this PR?

release_notes/deprecate_remove.adoc
```diff
@@ -243,6 +249,11 @@ A _removed_ item is typically function that was deprecated in previous releases
 | Use the integrated web console.
 | See link:../console/console_access.adoc#accessing-your-console[Accessing your console] for more information.
 
+| Governance
+| Integrity shield (Technology Preview)
+| 2.7
+| You can continue to use Integrity shield as a community-provided signing solution. For more details, see the Integrity Shield documentation, link:https://github.com/stolostron/integrity-shield/blob/master/docs/ACM/README_GETTING_STARTED.md[Getting Started documentation].
```
gparvin, line 255: Does this Integrity Shield entry show up for you?  To me it looks like it should have another column.

The management ingress details look good.

gparvin: When I view it I see "Clusters" show up as the last column -- which should be the first column of the next row.

dockerymick: Great catch! Updated

gparvin (approved):

swopebe (approved):

dockerymick:
@swopebe @gparvin thanks for the reviews and approvals! Merging this PR

PR #4621, opened 2023-01-26
Add missing parameter for cert policy controller

dockerymick:
https://github.com/stolostron/backlog/issues/27033

### DIFF
--- a/governance/cert_policy_ctrl.adoc
+++ b/governance/cert_policy_ctrl.adoc
@@ -38,6 +38,9 @@ spec:
     exclude: []
     matchExpressions: []
     matchLabels: {}
+  labelSelector:
+    include:
+    exclude:
   remediationAction:
   severity: 
   minimumDuration:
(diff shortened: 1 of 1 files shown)


governance/cert_policy_ctrl.adoc
```diff
@@ -38,6 +38,9 @@ spec:
     exclude: []
     matchExpressions: []
     matchLabels: {}
+  labelSelector:
+    include:
+    exclude:
```
gparvin: The label selector doesn't contain an include and exclude, instead it contains a key and a value: example:

```
labelSelector:
  myLabelKey: myLabelValue
```

dockerymick: Thanks ! @gparvin updated

governance/cert_policy_ctrl.adoc
```diff
@@ -78,6 +80,10 @@ If there is a different value for the `category` key in your certificate policy,
 
 *Note:* If the `namespaceSelector` for the certificate policy controller does not match any namespace, the policy is considered compliant.
 
+| `spec.labelSelector`
+| Optional
+| Specify identifying attributes of objects. See the link:https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/[Kubernetes labels and selectors] documentation.
```
cadawson: If this is a single value, then "specifies" is better. It looks like it might allow multiple labels, so it might be correct as it is.

cadawson (approved):
One thought, but it might be ok. Approving. Thanks!

dockerymick:
@cadawson Thanks for the review, I updated the PR based on your comment.

gparvin (approved):

PR #4654, opened 2023-01-31
stolostron/backlog#27154

oafischer:
Also for 2.6


cadawson (approved):
Guessing no workaround for that one, huh? Looks good. Thanks!

Is this a long-term issue that should be in the docs, or just a known issue? Just a tought.

oafischer:
This will be fixed in a z stream (planned in 2.7.1 so pretty soon) so no workaround and suitable for a known issue.

PR #4661, opened 2023-01-31
Firefox scrolling known issue stolostron/backlog#27107

oafischer:
stolostron/backlog#27107


chenz4027 (approved):
/lgtm

release_notes/known_issues.adoc
```diff
@@ -157,6 +157,12 @@ When you edit namespace bindings for a cluster set with the `admin` role or `bin
 
 To resolve the issue, make sure you also have permission to create or delete a `ManagedClusterSetBinding` resource in the namespace you want to bind. The role bindings only allow you to bind the cluster set to the namespace.
 
+[#scrolling-hosted]
+=== Horizontal scrolling does not work after provisioning hosted control plane cluster
+//2.7:27107
+
+After provisioning a hosted control plane cluster, you might not be able to scroll horizontally in the cluster overview of the {product-title-short} console if the `ClusterVersionUpgradeable` parameter is too long. You are not able to view the hidden data as a result. To work around the issue, zoom out by using your browser zoom controls, increase your {product-title-short} console window size, or copy and paste the text to a different location.
```
swopebe: You are not able to
changes to 
You cannot

Move the workaround down to its own paragraph...always think: Can you use find the task quickly...then structure the content as such. 

We trypically use "resolve" because workaround can be a verb or noun...better for comprehension.

swopebe (changes requested):
small stuff that can make a big diff mentioned

release_notes/known_issues.adoc
oafischer: I'll update it, thanks! In this case the workaround does not resolve the issue though so I'd rather stick with "work around" if that's ok.

swopebe (approved):

PR #4663, opened 2023-02-01
Adding note about new behavior when cleaning resources

dockerymick:
https://github.com/stolostron/backlog/issues/27025

### DIFF
--- a/governance/create_policy.adoc
+++ b/governance/create_policy.adoc
@@ -311,6 +311,6 @@ You can set the value directly in the YAML as you create a policy from the CLI. 
 
 *Notes:*
 
-- If a policy that installs an operator uses the `pruneObjectBehavior` parameter defined, then additional clean up is needed to complete the operator uninstall. Additional clean up might include deleting the operator `ClusterServiceVersion` object.
+- If a policy that installs an operator has the `pruneObjectBehavior` parameter defined, then additional clean up is needed to complete the operator uninstall. You might need to delete the operator `ClusterServiceVersion` object as part of this cleanup.
 
-- As you disable the `config-policy-addon` and the Kubernetes configuration policy custom resource is deleted on the managed cluster, you must first remove the resources from the managed cluster. 
+- As you disable the `config-policy-addon` resource on the managed cluster, the `pruneObjbectBehavior` is ignored. To automatically clean up the related resources on the policies, you must remove the policies from the managed cluster before the add-on is disabled.
--- a/governance/create_policy.adoc
+++ b/governance/create_policy.adoc
@@ -313,4 +313,4 @@ You can set the value directly in the YAML as you create a policy from the CLI. 
 
 - If a policy that installs an operator uses the `pruneObjectBehavior` parameter defined, then additional clean up is needed to complete the operator uninstall. Additional clean up might include deleting the operator `ClusterServiceVersion` object.
 
-- As you disable the `ManagedClusterAddOn` and the Kubernetes configuration policy custom resource is deleted on the managed cluster, you must first remove the resources from the managed cluster. 
+- As you disable the `config-policy-addon` and the Kubernetes configuration policy custom resource is deleted on the managed cluster, you must first remove the resources from the managed cluster. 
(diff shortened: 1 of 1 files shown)


governance/create_policy.adoc
```diff
@@ -309,4 +309,8 @@ Use the `pruneObjectBehavior` parameter in a configuration policy to clean up re
 
 You can set the value directly in the YAML as you create a policy from the CLI. From the console, you can select the value in the _Prune Object Behavior_ section of the the _Policy templates_ step.
 
-*Note:* If a policy that installs an operator uses the `pruneObjectBehavior` parameter defined, then additional clean up is needed to complete the operator uninstall. Additional clean up might include deleting the operator `ClusterServiceVersion` object.
+*Notes:*
+
+- If a policy that installs an operator uses the `pruneObjectBehavior` parameter defined, then additional clean up is needed to complete the operator uninstall. Additional clean up might include deleting the operator `ClusterServiceVersion` object.
+
+- As you disable the `config-policy-addon` and the Kubernetes configuration policy custom resource is deleted on the managed cluster, you must first remove the resources from the managed cluster.
```
JustinKuli: This makes it sound like you need to remove the resources before the addon can be disabled, which isn't quite right.

When the addon is disabled, the `pruneObjectBehavior` is ignored, and related resources are not cleaned up automatically. If the user wanted the resources deleted, they could either remove the policies from the cluster first (which would trigger the cleanup), or they could manually remove the resources after the addon is disabled.

I think we would suggest the first option here, something like:

```suggestion
- As you disable the `config-policy-addon` on the managed cluster, the `pruneOjbectBehavior` is ignored. To automatically clean up the related resources on the policies, the policies should be removed from the managed cluster before the addon is disabled.
```

JustinKuli (changes requested):

governance/create_policy.adoc
dockerymick: @JustinKuli Thanks for your suggestion! What are your thoughts about this update:

As you disable the `config-policy-addon` on the managed cluster, the `pruneOjbectBehavior` is ignored. To automatically clean up the related resources on the policies, you must remove the policies from the managed cluster before the add-on is disabled.

governance/create_policy.adoc
```diff
@@ -309,4 +309,8 @@ Use the `pruneObjectBehavior` parameter in a configuration policy to clean up re
 
 You can set the value directly in the YAML as you create a policy from the CLI. From the console, you can select the value in the _Prune Object Behavior_ section of the the _Policy templates_ step.
 
-*Note:* If a policy that installs an operator uses the `pruneObjectBehavior` parameter defined, then additional clean up is needed to complete the operator uninstall. Additional clean up might include deleting the operator `ClusterServiceVersion` object.
+*Notes:*
+
+- If a policy that installs an operator uses the `pruneObjectBehavior` parameter defined, then additional clean up is needed to complete the operator uninstall. Additional clean up might include deleting the operator `ClusterServiceVersion` object.
```
cadawson: This reads a little strangely. Alternate suggestion: If a policy that installs an operator has the `pruneObjectBehavior` parameter defined, additional cleanup is needed to complete the operator uninstallation. You might need to delete the operator `ClusterServiceVersion` object as part of this cleanup.

governance/create_policy.adoc
cadawson: Typo in pruneObjectBehavior

cadawson (approved):
Suggestion added, otherwise looks good.

governance/create_policy.adoc
JustinKuli: Typo is my fault! Yes, that sounds better

governance/create_policy.adoc
JustinKuli: I like that

governance/create_policy.adoc
dockerymick: Updated

dockerymick:
@cadawson @JustinKuli thanks for the reviews and suggestions. I have updated the PR. When you have an opportunity, will you re-review?

JustinKuli (approved):
LGTM

dockerymick:
Thanks to you both! Merging this PR

PR #4670, opened 2023-02-01
Update custom_template.adoc

dockerymick:
backport to 2.6
https://github.com/stolostron/backlog/issues/27141


dockerymick:
Similar changes to 2.6. Developer confirmed that this was added for 2.7 and backported to 2.6. No need to update What's new doc

PR #4671, opened 2023-02-02
Known issue stolostron/backlog#27156

oafischer:
stolostron/backlog#27156

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -874,7 +874,7 @@ To resolve the issue, disable the policy and compare the differences between `ob
 
 - The `objectDefinition` in the `ConfigurationPolicy` does not match because Kubernetes processing the object when the policy is applied.
 +
-To resolve the issue, remove the keys from the `objectDefinition` in the policy. If the keys are different or missing, Kubernetes might have processed the keys before applying them to the object. For `PodSecurityPolicy`, for example, Kubernetes removes keys with values set to `false`, which you can see in the resulting object on the managed cluster. 
+To resolve the issue, disable the policy and compare the differences between `objectDefinition` in the policy and the object on the managed cluster. If the keys are different or missing, Kubernetes might have processed the keys before applying them to the object. For `PodSecurityPolicy`, for example, Kubernetes removes keys with values set to `false`, which you can see in the resulting object on the managed cluster. In this case, remove these keys from the `objectDefinition` in the policy.
 
 [#backup-known-issues]
 == Backup and restore known issues
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -872,7 +872,7 @@ If a policy is set to `remediationAction: enforce` and is repeatedly updated, th
 
 - The `objectDefinition` in the `ConfigurationPolicy` does not match because Kubernetes is processing the object on apply.
 
-The resolve th issue, disable the policy and compare the differences between `objectDefinition` in the policy and the object on the managed cluster. If the values are different, another controller or process might be updating them.
+To resolve the issue, disable the policy and compare the differences between `objectDefinition` in the policy and the object on the managed cluster. If the values are different, another controller or process might be updating them.
 
 You can also look for process leftovers in the object, such as `metadata`. If the keys are different or missing, Kubernetes might have processed the keys before applying them to the object.
 
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -863,6 +863,21 @@ The configuration policy resources are automatically removed from the cluster an
 
 When modifying an existing configuration policy, `DeleteAll` or `DeleteIfCreated` in the `pruneObjectBehavior` feature does not clean up old resources that were created before modifying. Only new resources from policy creations and policy updates are tracked and deleted when you delete the configuration policy.
 
+[#policy-status-repeated-updates]
+=== Policy status shows repeated updates when enforced
+
+If a policy is set to `remediationAction: enforce` and is repeatedly updated, the {product-title-short} console shows repeated violations with successful updates. This might happen in the following cases:
+
+- Another controller or process is also updating the object with different values.
+
+- The `objectDefinition` in the `ConfigurationPolicy` does not match because Kubernetes is processing the object on apply.
+
+The resolve th issue, disable the policy and compare the differences between `objectDefinition` in the policy and the object on the managed cluster. If the values are different, another controller or process might be updating them.
+
+You can also look for process leftovers in the object, such as `metadata`. If the keys are different or missing, Kubernetes might have processed the keys before applying them to the object.
+
+For PodSecurityPolicy, for example, Kubernetes removes keys with values set to `false`, which you can see in the resulting object on the managed cluster. Removing the keys from the `objectDefinition` in the policy resolves the issue.
+
 [#backup-known-issues]
 == Backup and restore known issues
 
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -870,7 +870,7 @@ If a policy is set to `remediationAction: enforce` and is repeatedly updated, th
 
 - Another controller or process is also updating the object with different values.
 +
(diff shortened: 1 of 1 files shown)


release_notes/known_issues.adoc
```diff
@@ -863,6 +863,21 @@ The configuration policy resources are automatically removed from the cluster an
...
+[#policy-status-repeated-updates]
+=== Policy status shows repeated updates when enforced
+
+If a policy is set to `remediationAction: enforce` and is repeatedly updated, the {product-title-short} console shows repeated violations with successful updates. This might happen in the following cases:
+
+- Another controller or process is also updating the object with different values.
+
+- The `objectDefinition` in the `ConfigurationPolicy` does not match because Kubernetes is processing the object on apply.
+
+The resolve th issue, disable the policy and compare the differences between `objectDefinition` in the policy and the object on the managed cluster. If the values are different, another controller or process might be updating them.
```
dhaiducek:
```suggestion
To resolve the issue, disable the policy and compare the differences between `objectDefinition` in the policy and the object on the managed cluster. If the values are different, another controller or process might be updating them.
```

release_notes/known_issues.adoc
```diff
@@ -863,6 +863,21 @@ The configuration policy resources are automatically removed from the cluster an
...
+
+You can also look for process leftovers in the object, such as `metadata`. If the keys are different or missing, Kubernetes might have processed the keys before applying them to the object.
+
+For PodSecurityPolicy, for example, Kubernetes removes keys with values set to `false`, which you can see in the resulting object on the managed cluster. Removing the keys from the `objectDefinition` in the policy resolves the issue.
```
dhaiducek:
```suggestion
If the keys are different or missing, Kubernetes might have processed the keys before applying them to the object. For PodSecurityPolicy, for example, Kubernetes removes keys with values set to `false`, which you can see in the resulting object on the managed cluster. Removing the keys from the `objectDefinition` in the policy resolves the issue.
```

release_notes/known_issues.adoc
```diff
@@ -863,6 +863,21 @@ The configuration policy resources are automatically removed from the cluster an
...
+
+If a policy is set to `remediationAction: enforce` and is repeatedly updated, the {product-title-short} console shows repeated violations with successful updates. This might happen in the following cases:
+
+- Another controller or process is also updating the object with different values.
+
+- The `objectDefinition` in the `ConfigurationPolicy` does not match because Kubernetes is processing the object on apply.
+
+The resolve th issue, disable the policy and compare the differences between `objectDefinition` in the policy and the object on the managed cluster. If the values are different, another controller or process might be updating them.
+
+You can also look for process leftovers in the object, such as `metadata`. If the keys are different or missing, Kubernetes might have processed the keys before applying them to the object.
```
dhaiducek: These are two separate points:
- The `metadata` reference refers to clues left by another controller or process. These controllers sometimes leave labels or annotations to signal that they've updated the object.
- The keys being different or missing goes with the PodSecurityPolicy point in the next paragraph.

dhaiducek (changes requested):

release_notes/known_issues.adoc
oafischer: Thanks for clarifying, might be clearer then to structure the scenario and resolution together. Let me update it.

release_notes/known_issues.adoc
```diff
@@ -863,6 +863,19 @@ The configuration policy resources are automatically removed from the cluster an
...
 When modifying an existing configuration policy, `DeleteAll` or `DeleteIfCreated` in the `pruneObjectBehavior` feature does not clean up old resources that were created before modifying. Only new resources from policy creations and policy updates are tracked and deleted when you delete the configuration policy.
 
+[#policy-status-repeated-updates]
+=== Policy status shows repeated updates when enforced
+
+If a policy is set to `remediationAction: enforce` and is repeatedly updated, the {product-title-short} console shows repeated violations with successful updates. This might happen in the following cases:
+
+- Another controller or process is also updating the object with different values.
+
+- The `objectDefinition` in the `ConfigurationPolicy` does not match because Kubernetes is processing the object on apply.
```
dockerymick: "on apply" sounds interesting to me 

What about "processing the object when the policy is applied."

release_notes/known_issues.adoc
```diff
@@ -863,6 +863,19 @@ The configuration policy resources are automatically removed from the cluster an
...
+
+To resolve the issue, disable the policy and compare the differences between `objectDefinition` in the policy and the object on the managed cluster. If the values are different, another controller or process might be updating them.
+
+If the keys are different or missing, Kubernetes might have processed the keys before applying them to the object. For PodSecurityPolicy, for example, Kubernetes removes keys with values set to `false`, which you can see in the resulting object on the managed cluster. Removing the keys from the `objectDefinition` in the policy resolves the issue.
```
dockerymick: monospace PodSecurityPolicy

dockerymick (changes requested):

oafischer:
Thanks for the comments @dhaiducek and @dockerymick, made some updates to hopefully make it clearer.

release_notes/known_issues.adoc
```diff
@@ -863,6 +863,19 @@ The configuration policy resources are automatically removed from the cluster an
...
 When modifying an existing configuration policy, `DeleteAll` or `DeleteIfCreated` in the `pruneObjectBehavior` feature does not clean up old resources that were created before modifying. Only new resources from policy creations and policy updates are tracked and deleted when you delete the configuration policy.
 
+[#policy-status-repeated-updates]
+=== Policy status shows repeated updates when enforced
+
+If a policy is set to `remediationAction: enforce` and is repeatedly updated, the {product-title-short} console shows repeated violations with successful updates. This might happen in the following two cases:
+
+- Another controller or process is also updating the object with different values.
++
+To resolve the issue, disable the policy and compare the differences between `objectDefinition` in the policy and the object on the managed cluster. If the values are different, another controller or process might be updating them.
```
dhaiducek: Adding something like might help users troubleshoot:

```suggestion
To resolve the issue, disable the policy and compare the differences between `objectDefinition` in the policy and the object on the managed cluster. If the values are different, another controller or process might be updating them. If this is the case, the controller or process may have left clues in the `metadata` of the object, such as a label or annotation, that can help identify what is making the changes.
```

release_notes/known_issues.adoc
```diff
@@ -863,6 +863,19 @@ The configuration policy resources are automatically removed from the cluster an
...
+
+- The `objectDefinition` in the `ConfigurationPolicy` does not match because Kubernetes processing the object when the policy is applied.
++
+To resolve the issue, remove the keys from the `objectDefinition` in the policy. If the keys are different or missing, Kubernetes might have processed the keys before applying them to the object. For `PodSecurityPolicy`, for example, Kubernetes removes keys with values set to `false`, which you can see in the resulting object on the managed cluster.
```
dhaiducek: Removing keys might not be the only solution, so the first part should be more generic.

```suggestion
To resolve the issue, disable the policy and compare the differences between `objectDefinition` in the policy and the object on the managed cluster. If the keys are different or missing, Kubernetes might have processed the keys before applying them to the object. For `PodSecurityPolicy`, for example, Kubernetes removes keys with values set to `false`, which you can see in the resulting object on the managed cluster. In this case, remove these keys from the `objectDefinition` in the policy.
```

dhaiducek (changes requested):
Thanks, @oafischer! It's looking good--I have a few more refinements to consider.

release_notes/known_issues.adoc
oafischer: Not a huge fan of mentioning clues since it's a little vague. We mention two examples (label/annotation). Can we list more specific things like that to look out for?

dhaiducek: That's fair. Unfortunately what a controller leaves on an object isn't consistent nor guaranteed. So we can leave it off if that's best practice for documentation.

oafischer: @dockerymick What are your thoughts on this?

dockerymick: I recommend stating: 

Check the `metadata` of the object to help identify why the values are different.

dhaiducek:
Thanks, @oafischer!

PR #4688, opened 2023-02-06
Known issue AAP Operator 2.7 stolostron/backlog#27113

oafischer:
Wondering if I should add examples of integrations (such as automation templates) or if integrations by itself is clear enough?

### DIFF
--- a/release_notes/known_issues.adoc
+++ b/release_notes/known_issues.adoc
@@ -169,7 +169,7 @@ To work around the issue, zoom out by using your browser zoom controls, increase
 === Error when using integrations with the {aap} Operator
 //2.7:27113
 
-If you use integrations that depend on the {aap-short} Operator and you do not have permissions to view installed Operators on the {ocp} cluster, you might see an error message that resembles the following:
+If you use integrations that depend on the {aap-short} Operator and do not have permissions to view installed Operators on the {ocp} cluster, you might see an error message that resembles the following:
 
 `The Ansible Automation Platform Operator is required to use automation templates. Version 2.2.1 or greater is required to use workflow job templates in automation templates.`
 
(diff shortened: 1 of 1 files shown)


KevinFCormier (approved):
Looks good to me. If you want to give an example of Ansible integrations for which this message applies, you could mention automation templates for cluster install/upgrade and configuring automations for a policy.

release_notes/known_issues.adoc
```diff
@@ -165,6 +165,16 @@ After provisioning a hosted control plane cluster, you might not be able to scro
 
 To work around the issue, zoom out by using your browser zoom controls, increase your {product-title-short} console window size, or copy and paste the text to a different location.
 
+[#error-integrations-aap]
+=== Error when using integrations with {aap} Operator
```
cadawson:
```suggestion
=== Error when using integrations with the {aap} Operator
```

release_notes/known_issues.adoc
```diff
@@ -165,6 +165,16 @@ After provisioning a hosted control plane cluster, you might not be able to scro
...
+//2.7:27113
+
+If you use integrations that depend on the {aap-short} Operator and you do not have permissions to view installed Operators on the {ocp} cluster, you might see an error message that resembles the following:
```
cadawson:
```suggestion
If you use integrations that depend on the {aap-short} Operator and do not have permissions to view installed Operators on the {ocp} cluster, you might see an error message that resembles the following:
```

release_notes/known_issues.adoc
```diff
@@ -165,6 +165,16 @@ After provisioning a hosted control plane cluster, you might not be able to scro
...
+
+`The Ansible Automation Platform Operator is required to use automation templates. Version 2.2.1 or greater is required to use workflow job templates in automation templates.`
+
+You can safely ignore the warning if you confirm with your system administrator that the Operator is installed.
```
cadawson: Should we stay consistent here with `message` or `error message`, rather than `warning`? Might not be a big deal, but `warnings` are often used for specific instances where you might be about to lose some data.

cadawson (approved):
A couple of small suggestions, but not required. Looks good!

release_notes/known_issues.adoc
oafischer: Good point, I agree and I'll update it

PR #4691, opened 2023-02-06
removing link reference to Quick start guide

dockerymick:
Screen Shot 2023-02-06 at 12 07 43 PM


dockerymick:
@swopebe are there thought about us adding a link to What's new since we removed the reference to Quick start?

swopebe (approved):
I think I already have this in a PR for cluster,

PR #4705, opened 2023-02-07
broken links for back up and restore

swopebe:


dockerymick (approved):
Looks good to me! There were only 2 links that are broken and I know it's bc we are pointing to private info (support matrix)

oafischer (approved):

PR #4712, opened 2023-02-07
updating links for gov book (which is green)

dockerymick:
https://github.com/stolostron/backlog/issues/26793

Because the anchors match, the links are working from Pantheon. When I run the link check tool, the link appears as broken bc the file names don't match


cadawson (approved):
Looks like they will work to me! Thanks!

PR #4736, opened 2023-02-13
ACM-3419 Clarify volume mode 2.7

oafischer:
https://issues.redhat.com/browse/ACM-3419

### DIFF
--- a/add-ons/volsync/volsync.adoc
+++ b/add-ons/volsync/volsync.adoc
@@ -3,7 +3,7 @@
 
 VolSync is a Kubernetes operator that enables asynchronous replication of persistent volumes within a cluster, or across clusters with storage types that are not otherwise compatible for replication. It uses the Container Storage Interface (CSI) to overcome the compatibility limitation. After deploying the VolSync operator in your environment, you can leverage it to create and maintain copies of your persistent data. VolSync can only replicate persistent volume claims on {ocp} clusters that are at version 4.8 or later.
 
-*Important:* VolSync only supports the volume mode of Filesystem. If you do not select the volume mode, it defaults to Filesystem.
+*Important:* VolSync only supports replicatin persistent volume claims with the volume mode of Filesystem. If you do not select the volume mode, it defaults to Filesystem.
 
 * xref:../volsync/volsync_replicate.adoc#volsync-rep[Replicating persistent volumes with VolSync]
 ** xref:../volsync/volsync_replicate.adoc#volsync-install-clusters[Installing VolSync on the managed clusters]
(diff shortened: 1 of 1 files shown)


add-ons/volsync/volsync.adoc
```diff
@@ -1,7 +1,9 @@
...
 
-VolSync is a Kubernetes operator that enables asynchronous replication of persistent volumes within a cluster, or across clusters with storage types that are not otherwise compatible for replication. It uses the Container Storage Interface (CSI) to overcome the compatibility limitation. After deploying the VolSync operator in your environment, you can leverage it to create and maintain copies of your persistent data. VolSync can only replicate persistent volume claims on {ocp} clusters that are at version 4.8 or later. 
+VolSync is a Kubernetes operator that enables asynchronous replication of persistent volumes within a cluster, or across clusters with storage types that are not otherwise compatible for replication. It uses the Container Storage Interface (CSI) to overcome the compatibility limitation. After deploying the VolSync operator in your environment, you can leverage it to create and maintain copies of your persistent data. VolSync can only replicate persistent volume claims on {ocp} clusters that are at version 4.8 or later.
+
+*Important:* VolSync only supports the volume mode of Filesystem. If you do not select the volume mode, it defaults to Filesystem.
```
JohnStrunk:
```suggestion
*Important:* VolSync only supports a PV or PVC `volumeMode` of `Filesystem`. If you do not select the volume mode, it defaults to Filesystem.
```

Do you want to be more specific about the PV/PVC and the "volumeMode" name?

Otherwise, looks fine

tesshuflower (changes requested):
Quick suggestion - can we mention persistent volume claims specifially, something like:

*Important:* VolSync only supports replicating persistent volume claims with volume mode of Filesystem. If you do not select the volume mode, it defaults to Filesystem.

add-ons/volsync/volsync.adoc
oafischer: What does PV/PVC stand for? If it's important for the context or for the user to know we can add it, otherwise I don't think we need to.

oafischer:
I'll try to combine both suggestions, thanks @JohnStrunk and @tesshuflower

oafischer:
Made the updates, ready for another review.

add-ons/volsync/volsync.adoc
cadawson: Another option: VolSync only supports a persistent volume or persistent volume claim with a `volumeMode` value of `Filesystem`. If you do not specify a value for `volumeMode`, it defaults to `Filesystem`. 

If you want to spell out PV and PVC.

cadawson (approved):
Another option based on John's suggestion. Looks good! Thanks!

tesshuflower (approved):
/lgtm

Thanks Oliver and Chris!

JohnStrunk (approved):
/lgtm

Thanks!

PR #4738, opened 2023-02-13
2023 Q1 internal docs update

oafischer:
More updates to follow. Please add and update content as you see fit.


_pages/wr_special_tag_guidance.md
```diff
@@ -5,29 +5,19 @@ permalink: /wr_special_tag_guidance
...
-**Important:** A little stronger than a note.
+| **Special tag** | | **How and when to use** |
+| ----------- | | ----------- |
+| **Note:** | |  Use for extra information that is helpful. |
+| **Important:** | |  A little stronger than a note. |
+| **Optional:**  | | For any step not required, but user may want to take or needs to take in certain cases. |
+| **Best practice:** | | Use with caution. Avoid when possible by writing more descriptive text to guide the user. |
+| **(Technology preview)** | | Place at the highest level of the component that is tech preview on the title (to get this in the navigation). Include a special tag on the What's new entry. |
 
-**Tip:** Use for a shortcut, for instance.
```
oafischer, line 14: Tip and note are so similar IMO, I think we can reduce it to just using note. I've never seen tip used except for that YAML view thing.

_pages/wr_special_tag_guidance.md
```diff
@@ -5,29 +5,19 @@ permalink: /wr_special_tag_guidance
...
-**Note:** Use for extra information that is helpful.
+# Table of special tags
 
-**Important:** A little stronger than a note.
+| **Special tag** | | **How and when to use** |
+| ----------- | | ----------- |
+| **Note:** | |  Use for extra information that is helpful. |
+| **Important:** | |  A little stronger than a note. |
+| **Optional:**  | | For any step not required, but user may want to take or needs to take in certain cases. |
+| **Best practice:** | | Use with caution. Avoid when possible by writing more descriptive text to guide the user. |
```
dockerymick, line 17: Do we want to use Best practice rather than recommendation? -Brandi 

Continue using Best practice, as recommended, with caution. Try to rewrite with more content -Oliver

Legal concerns when stating _recommends_. Continue verifying if the content is best practice -Laura

Will revisit the option to remove Best practice

PR #4741, opened 2023-02-14
Delete .DS_Store (Please try to get this ignored :-) )

swopebe:


cadawson:
What a pain. I think I have already done it with mine. If it is mine that is still causing it, I will try something else.

PR #4747, opened 2023-02-15
ACM-3386 move lcl-clst ACM 2 mce

cadawson:


cadawson:
https://issues.redhat.com/browse/ACM-3386

Hi @ray-harris. Can you review the draft text for this issue in this PR? Thanks!

cadawson:
@JakobGray: Will you take a look at this draft? Thanks!

JakobGray (changes requested):
Now that hub-self-management is treated as a component of MCE I think the section can more closely resemble those below it, like the Hypershift add-on section. In that vein it could go something like this:

> By default, the cluster running {mce-short} will manage itself. To install with this feature disabled, include the following in the `MultiClusterEngine` values in `spec.overrides`:
(yaml content but set enabled to false)

Everything after "If the setting is not included in the YAML" can be removed, because that is describing the old way to do it by manually creating the resource. The last line "A hub cluster that is managed by itself..." can stay.

clusters/install_upgrade/adv_config_install.adoc
```diff
@@ -6,34 +6,25 @@ The {mce-short} is installed using an operator that deploys all of the required
...
 
-. Run the following command to apply the file:
-+
-----
-oc apply -f import-hub.yaml
-----
+* The `name` value identifies the hub cluster as a `local-cluster`.
+* The `enabled` setting specifies whether the feature is enabled or disabled. When the value is `true`, the hub cluster manages itself. When the value is `false`, the hub cluster does not manage itself.
+
+If the setting is not included in the YAML file that defines the custom resource, you need to add it. The hub cluster can only be managed with this option.
```
JakobGray: This line: "If the setting is not included in the YAML file that defines the custom resource, you need to add it. The hub cluster can only be managed with this option." can be deleted, and then I think this is good to go.

cadawson: Made that update @JakobGray. Thanks! Does that need to be removed from the ACM version, as well, or is it ok there?

oafischer (approved):
lgtm

PR #4764, opened 2023-02-21
fix the hypershift config doc

rokej:
A couple of corrections


oafischer (approved):
Thanks for the updates!

PR #4767, opened 2023-02-21
Update issue points internal docs

oafischer:
We can slowly update the examples as we receive more issues in Jira.

One thing I noticed is that the integer labels are used by other projects/teams/issues. Not sure what they are officially for, I think some others might be using them as a points system as well but not sure.


cadawson (approved):
Very nice. Thanks!

PR #4787, opened 2023-03-03
[2.7_stage] Update policy generator link to upstream

willkutler:


dockerymick:
https://issues.redhat.com/browse/ACM-2598

dockerymick:
@willkutler will you confirm if all instances of `stolostron` need to be updated to `open-cluster-management-io`?

dockerymick:
Merging this before @swopebe create the 2.8 branch. It's a concern that there is no doc issue for this change. It is pretty straight forward tho. If all instances need to change, then we need to create a doc issue for the link updates for 2.8 @willkutler

dockerymick (approved):
Approving before 2.8_stage branch is created

willkutler:
@dockerymick by all instances do you mean the other Github repos we link to? The only link that needs to be updated is the one to policy-generator-plugin - I think the change I made for this PR is the only time it is used. I can create an issue if needed.

dockerymick:
Thanks for clarifying. All instances in the rhacm-docs repo only.

PR #4799, opened 2023-03-07
PM discovery, no issue, missing in the TOC

swopebe:


swopebe:
@cadawson anything to add here? Thanks.

swopebe:
@cadawson you wanted to double check on this...did you have anything to return in favor or against adding this?

If we add, needs to be added to 2.8, updated in any intro file/container topic.

elgnay:
@swopebe Before adding this topic back, we need to update the content of https://github.com/stolostron/rhacm-docs/blob/2.7_stage/clusters/cluster_lifecycle/upgrade_cluster_disconn.adoc to remove the dependency on policy add-on. That effort is traced by https://issues.redhat.com/browse/ACM-4471

cadawson:
After discussion, this will be added back to the ACM install and upgrade book. Policies are not provided with MCE, so replacing this into the clusters doc in its current form will result in issues for those using only MCE. 

A revised version that does not use policies is being worked for placement in the cluster docs. Closing this PR and will open a new one for the new locations.

PR #4800, opened 2023-03-07
fix ACM-4051 disabling the hypershift feature

rokej:
Fixing https://issues.redhat.com/browse/ACM-4051. The doc is missing a step to disable the hosted control plane feature.


oafischer (approved):
Made some style edits, feel free to merge if it looks good. Thanks!

PR #4811, opened 2023-03-14
https://issues.redhat.com/browse/ACM-4060 Add Rsync-TLS docs 2.8

oafischer:
https://issues.redhat.com/browse/ACM-4060

Needs updated example outputs. Check examples for accuracy. Check if any of the process/flow for Rsync-TLS should be different from Rsync.


add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,227 @@ oc apply -f volsync-mcao.yaml
...
++
+----
+ADDRESS=`oc get replicationdestination <destination> -n <destination-ns> --template={{.status.rsyncTLS.address}}`
+echo $ADDRESS
+----
++
+The output appears similar to the following, which is for an Amazon Web Services environment:
++
+----
+a831264645yhrjrjyer6f9e4a02eb2-5592c0b3d94dd376.elb.us-east-1.amazonaws.com
```
oafischer, line 189: Is this output still a good example?

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,227 @@ oc apply -f volsync-mcao.yaml
...
+----
+
+.. Run the following command to copy the name of the secret and the contents of the secret that are provided as the value of `.status.rsyncTLS.keySecret`.
++
+----
+KEYSECRET=`oc get replicationdestination <destination> -n <destination-ns> --template={{.status.rsyncTLS.keySecret}}`
```
oafischer, line 195: Is this still applicable to Rsync-TLS? I looked over the upstream docs and from what I can tell, the key secret works a little differently from the SSH key, so I'm not sure.

add-ons/volsync/volsync_replicate.adoc
tesshuflower: This still looks good 👍

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,227 @@ oc apply -f volsync-mcao.yaml
...
+apiVersion: volsync.backube/v1alpha1
+kind: ReplicationSource
+metadata:
+  name: <source>
+  namespace: <source-ns>
+spec:
+  sourcePVC: <persistent_volume_claim>
+  trigger:
+    schedule: "*/3 * * * *" #/*
+  rsync:
```
tesshuflower: This needs to be `rsyncTLS` rather than `rsync`

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,227 @@ oc apply -f volsync-mcao.yaml
...
+----
++
+Replace `destination` with the name of your replication destination custom resource.
++
+Replace `destination-ns` with the name of the namespace where your destination is located.
++
+You will have to enter it on the source cluster when you configure the source. The output should be the name of your SSH keys secret file, which might resemble the following name:
++
+----
+volsync-rsync-dst-src-destination-name
```
tesshuflower: Here's a better example output from the `echo $KEYSECRET` cmd:

`volsync-rsync-tls-destination-name`

add-ons/volsync/volsync_replicate.adoc
tesshuflower: This individual command is good, but it seems like we're missing some steps here - this command will essentially just get the name of the keySecret secret.   I'll put some more comments a bit below ..

add-ons/volsync/volsync_replicate.adoc
tesshuflower: Note that after this step all we've done is echo the secret name (i.e. `volsync-rsync-tls-destination-name` is the name of the secret that VolSync created.  It actually looks like the rsync docs have this same problem.  They don't cover the steps to actually copy the secret over (although line 191 above mentions the steps are how to find the secret and copy its contents).

To get the contents users would need to do something along the lines of:

```
$ oc get secret $KEYSECRET -n <destination-ns> -o yaml > /tmp/keysecret.yaml

# Now remove unnecessary fields:
$ vi /tmp/keysecret.yaml
$ ^^^ change the namespace to the source namespace where the replicationsource will be created
$ ^^^ remove unnecessary fields from .metadata (i.e. fields that are not name or namespace)

# Create the secret on the source cluster
$ oc create -f /tmp/keysecret.yaml


Note essentially the same steps are needed with rsync as well - I thought this was previously documented, but maybe I'm just imagining it?  Even though the contents of the secrets are different between rsync and rsync-tls, just copying over the secret is the same for each (other than the names themselves and how to get it from the status).

File: add-ons/volsync/volsync_replicate.adoc [hunk 5]
```diff
@@ -107,10 +107,227 @@ oc apply -f volsync-mcao.yaml
...
+
+The following example shows how to configure by using the Rsync-TLS method. For additional information about Rsync-TLS, see https://volsync.readthedocs.io/en/latest/usage/index.html[Usage] in the VolSync documentation.
+ 
+[#volsync-rsync-tls-clusters]
+=== Configuring Rsync-TLS replication across managed clusters
+
+For Rsync-TLS-based replication, configure custom resources on the source and destination clusters. The custom resources use the `address` value to connect the source to the destination, and a TLS-protected tunnel provided by stunnel to ensure that the transferred data is secure.
+
+*Note:* The source and destination must have access to the shared key for the TLS connection. You can find the key location in the `keySecret` field. If you do not provide a secret name in `.spec.rsyncTLS.keySecret`, the secret name is automatically generated and added to `.status.rsyncTLS.keySecret`.
```
tesshuflower: If we're going to mention that the secret name can be provided, we probably need to document the format of the secret if the user wants to create it themselves.

You can find upstream documentation for the rsync-tls secret here:  https://volsync.readthedocs.io/en/latest/usage/rsync-tls/index.html

Look in the "Rsync-specific considerations > TLS authentication" section.  There is info on creating your own secret.

Note that the steps following this note show how to create the replicationdestination and let VolSync create the secret for you.

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,227 @@ oc apply -f volsync-mcao.yaml
...
++
+Replace `destination-ns` with the name of the namespace where your destination is located.
++
+You will have to enter it on the source cluster when you configure the source. The output should be the name of your SSH keys secret file, which might resemble the following name:
++
+----
+volsync-rsync-dst-src-destination-name
+----
+
+. Identify the source persistent volume claim that you want to replicate.
```
tesshuflower, line 234: Note that these steps may be taking place on another cluster. I.e. we have a destination cluster where we created the replicationdestination and now we have the source cluster where the pvc that we want to replicate lives.  Maybe this is self-explanatory - line 214 does mention creating the replication_source on the "source cluster".

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,227 @@ oc apply -f volsync-mcao.yaml
...
+    volumeSnapshotClassName: gp2-csi
+----
++
+Replace `source` with the name for your replication source custom resource. See step _3-vi_ of this procedure for instructions on how to replace this automatically.
++
+Replace `source-ns` with the namespace of the persistent volume claim where your source is located. See step _3-vi_ of this procedure for instructions on how to replace this automatically. 
++
+Replace `persistent_volume_claim` with the name of your source persistent volume claim.
++
+Replace `mykeysecret` with the keys that you copied from the `.status.rsyncTLS.keySecret` field of the `ReplicationDestination` when you configured it.
```
tesshuflower: This should be replace `mykeysecret` with the name of the secret that was copied over from the destination cluster to the source cluster.  this will be the value of $KEYSECRET if they followed the steps above

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,227 @@ oc apply -f volsync-mcao.yaml
...
++
+Replace `source` with the name for your replication source custom resource. See step _3-vi_ of this procedure for instructions on how to replace this automatically.
++
+Replace `source-ns` with the namespace of the persistent volume claim where your source is located. See step _3-vi_ of this procedure for instructions on how to replace this automatically. 
++
+Replace `persistent_volume_claim` with the name of your source persistent volume claim.
++
+Replace `mykeysecret` with the keys that you copied from the `.status.rsyncTLS.keySecret` field of the `ReplicationDestination` when you configured it. 
++
+Replace `my.host.com` with the host address that you copied from the `.status.rsyncTLS.address` field of the `ReplicationDestination` when you configured it.
```
tesshuflower: Should we mention this will be the value of $ADDRESS if they ran the cmd on line 181 above?

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,227 @@ oc apply -f volsync-mcao.yaml
...
++
+Replace `my.host.com` with the host address that you copied from the `.status.rsyncTLS.address` field of the `ReplicationDestination` when you configured it. 
++
+If your storage driver supports cloning, using `Clone` as the value for `copyMethod` might be a more streamlined process for the replication.
++ 
+*Optional:* Specify the values of the `storageClassName` and `volumeSnapshotClassName` parameters if you are using a storage class and volume snapshot class name that are different than the default values for your environment. 
++
+You can now set up the synchronization method of the persistent volume.
+
+.. Copy the key secret from the destination cluster by entering the following command against the destination cluster:
```
tesshuflower: Oh, my bad!  This is where the copying of the secret is documented!  Sorry - Uh...I think maybe it just makes sense to move this higher up rather than after the section about creating the replicationsource?  I think they'd want to copy over the secret, then create their replicationsource.

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,227 @@ oc apply -f volsync-mcao.yaml
...
+* Change the namespace to the namespace of your source cluster. For this example, it is `source-ns`.
+* Remove the owner references (`.metadata.ownerReferences`).
+
+.. On the source cluster, create the secret file by entering the following command on the source cluster:
++
+----
+oc create -f /tmp/secret.yaml
+----
+
+.. On the source cluster, modify the `replication_source.yaml` file by replacing the value of the `address` and `keySecret` in the `ReplicationSource` object with the values that you noted from the destination cluster by entering the following commands:
```
tesshuflower, line 277: Maybe redundant since we told them to do this already above?  I guess the sed cmds below might be useful though

add-ons/volsync/volsync_replicate.adoc
tesshuflower: Update: Sorry there actually are steps below about copying the secret over...maybe it makes sense to move them here instead.

add-ons/volsync/volsync_replicate.adoc
oafischer: Yes that's a good idea, stuff like this is why I wanted to essentially restructure and rewrite large parts of the VolSync doc. Chris actually ended up taking that issue, so he'll probably arrange a meeting to go over the existing docs and find all the places that need improvement.

tesshuflower (changes requested):
Looking good so far!  I added some comments, please let me know if you have any questions or if anything I wrote doesn't make sense.

add-ons/volsync/volsync_replicate.adoc
oafischer: It got really messy so I rearranged some stuff and hopefully the flow makes more sense now. I made the secret creation an optional step, given the content after it, as you said, shows you how to let the secret be created for you. Let me know if this works.

add-ons/volsync/volsync_replicate.adoc
oafischer: I moved those steps further up as suggested. The whole topic is still a little convoluted and hard to follow, but that is something Chris will hopefully address with the rewrite soon-

oafischer: Fixed on both rsync topics, let me know if the order makes more sense now.

add-ons/volsync/volsync_replicate.adoc
oafischer: If it helps the user or might otherwise be easy to miss/confusing, then let's add it.

add-ons/volsync/volsync_replicate.adoc
oafischer: Is it worth keeping the sed commands? It's better to simplify the topic if the user won't need them. If an advanced user might be interested, we can move the commands into an "Advanced config" topic and declutter the main procedure that way.

oafischer:
Made a round of edits, thanks for the review @tesshuflower. I think this needs a what's new entry as well so I'll add that soon. Left a couple of questions that we can address.

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,251 @@ oc apply -f volsync-mcao.yaml
...
+volsync-rsync-tls-destination-name
+----
+
+.. Copy the key secret from the destination cluster by entering the following command against the destination cluster:
++
+----
+oc get secret -n <destination-ns> $KEYSECRET -o yaml > /tmp/secret.yaml
+----
++
+Replace `destination-ns` with the namespace of the persistent volume claim where your destination is located.
```
tesshuflower: Can we update to "Replace `destination-ns` with the namespace where your replication destination is located."

add-ons/volsync/volsync_replicate.adoc
tesshuflower: Maybe we can just mention that there is an example of using sed commands below - i.e. what we tell them to do on line 191 we show how to do later on via cmds

add-ons/volsync/volsync_replicate.adoc
tesshuflower: maybe we keep the cmds, as it's not really an advanced item, it's more like specific cmds to help users who are not as advanced.

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,251 @@ oc apply -f volsync-mcao.yaml
...
+echo $ADDRESS
+----
++
+The output appears similar to the following, which is for an Amazon Web Services environment:
++
+----
+a831264645yhrjrjyer6f9e4a02eb2-5592c0b3d94dd376.elb.us-east-1.amazonaws.com
+----
+
+.. Run the following command to copy the name of the secret and the contents of the secret that are provided as the value of `.status.rsyncTLS.keySecret`.
```
tesshuflower: Since we now have the optional section above on how to create their own secret - we may need to mention here that if they did that optional step, they just need to make sure the secret exists on both the source and destination clusters.

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,251 @@ oc apply -f volsync-mcao.yaml
...
+
+. Configure your destination cluster.
+
+.. Run the following command on the destination cluster to create the namespace:
++
+----
+oc create ns <destination-ns>
+----
++
+Replace `destination-ns` with a name for the namespace that contains your destination persistent volume claim.
```
tesshuflower: Can we change this to be consistent with other lines:

"Replace `destination-ns` with the namespace where your replication destination is located."

add-ons/volsync/volsync_replicate.adoc
tesshuflower: I think this looks good - I just added a single comment about how we may need to reference this later on - as the steps later are assuming you didn't create the secret yourself

oafischer:
Thanks @tesshuflower, I think I addressed everything and added a what's new as well. Let me know if it looks good now.

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,250 @@ oc apply -f volsync-mcao.yaml
...
+  name: tls-key-secret
+type: Opaque
+----
+
+[#volsync-rsync-tls-clusters]
+=== Configuring Rsync-TLS replication across managed clusters
+
+For Rsync-TLS-based replication, configure custom resources on the source and destination clusters. The custom resources use the `address` value to connect the source to the destination, and a TLS-protected tunnel provided by stunnel to ensure that the transferred data is secure.
+
+See the following and examples to configure an Rsync-TLS replication from a persistent volume claim on the `source` cluster in the `source-ns` namespace to a persistent volume claim on a `destination` cluster in the `destination-ns` namespace. Replace the values where necessary:
```
cadawson:
```suggestion
See the following information and examples to configure an Rsync-TLS replication from a persistent volume claim on the `source` cluster in the `source-ns` namespace to a persistent volume claim on a `destination` cluster in the `destination-ns` namespace. Replace the values where necessary:
```

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,250 @@ oc apply -f volsync-mcao.yaml
...
++
+Replace `destination-ns` with the name of the namespace where your destination is located.
++
+After the `replicationdestination` resource is created, the following parameters and values are added to the resource: 
++
+|==========
+| Parameter | Value
+
+| `.status.rsyncTLS.address` | IP address of the destination cluster that is used to enable the source and destination clusters to communicate.
+| `.status.rsyncTLS.keySecret` | Name of the secret containing the TLS key that authenticates the connection with the source cluster.
```
cadawson:
```suggestion
| `.status.rsyncTLS.address` 
| IP address of the destination cluster that is used to enable the source and destination clusters to communicate.

| `.status.rsyncTLS.keySecret` 
| Name of the secret containing the TLS key that authenticates the connection with the source cluster.
```

cadawson (approved):
Looks good. A couple of suggestions. Thanks!

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,253 @@ oc apply -f volsync-mcao.yaml
...
+    volumeSnapshotClassName: gp2-csi
+----
++
+Replace `source` with the name for your replication source custom resource. See step _3-vi_ of this procedure for instructions on how to replace this automatically.
++
+Replace `source-ns` with the namespace of the persistent volume claim where your source is located. See step _3-vi_ of this procedure for instructions on how to replace this automatically. 
++
+Replace `persistent_volume_claim` with the name of your source persistent volume claim.
++
+Replace `mykeysecret` with the the name of the secret that you copied from the destination cluster to the source cluster (the value of `$KEYSECRET`.
```
tesshuflower: Missing closing bracket )

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -107,10 +107,253 @@ oc apply -f volsync-mcao.yaml
...
+echo $ADDRESS
+----
++
+The output appears similar to the following, which is for an Amazon Web Services environment:
++
+----
+a831264645yhrjrjyer6f9e4a02eb2-5592c0b3d94dd376.elb.us-east-1.amazonaws.com
+----
+
+.. Run the following command to copy the name of the secret and the contents of the secret that are provided as the value of `.status.rsyncTLS.keySecret`. If you created your own secret, make sure that the secret exists on both the source and destination clusters.
```
tesshuflower: I think this is still a bit confusing - We say "Run the following command to copy the name of the secret **and the contents of the secret** ..." But this is not true, the command only copies the name of the secret.

It's only later on down below where we have steps to export the secret as yaml (i.e. the contents of the secret) and save to /tmp/secret, edit it etc.

One more thing  is with the part about creating the secret on your own.  Something is still a bit awkward about this - If you create your own secret as we mention above then these steps are required at all - but instead, they need to set spec.rsyncRLS.keySecret in the replicationdestination spec instead.  It's maybe weird as we have steps to export the ADDRESS and KEYSECRET (name of the secret) to env vars, and then mention generally what to replace in the spec, and then later after that have specific steps on how to replace ADDRESS and KEYSECRET in the spec.

^ Not sure if that is clear, let me know if you'd rather discuss directly.

tesshuflower (changes requested):
sorry, just a couple more nitpicks...

add-ons/volsync/volsync_replicate.adoc
oafischer: I think the main problem we are running into is that we want to cater to multiple solutions at the same time. Might be worth it to streamline by telling the user to either create their own secret, or telling them that the secret will be created for them and to just explain the rest of the procedure under that assumption. I assume it is preferred to have the secret be created for you? Would a typical user want to create their own secret? Or is that more of an advanced config thing?

tesshuflower: I agree completely here - I think the easiest is to leave the instructions with letting VolSync create the secret for you.  And I do think it would be good to point to information if they want to create the secret themselves (but this could be afterwards or in some extra info at the bottom).  We can even link to the upstream docs for the user-created secret if that would be easier.

oafischer: Sounds like a plan, we are currently overhauling parts of our doc to remove inline links and move them to an "Additional resources" topic at the end of every file. Sounds like it would be a good spot for a link to upstream or a separate topic if we create one ourselves. You can see an example here: https://github.com/stolostron/rhacm-docs/blob/2.8_stage/governance/config_policy_ctrl.adoc

Chris would probably follow that structure when working on the planned VolSync docs overhaul, but in the meantime, I'll implement the "let VolSync create it for you" option as discussed.

oafischer:
Made more updates @tesshuflower, let me know if it looks good now.

tesshuflower:
> Made more updates @tesshuflower, let me know if it looks good now.

Great stuff - thanks a lot @oafischer !

tesshuflower (approved):
/lgtm

add-ons/volsync/volsync_replicate.adoc
```diff
@@ -784,6 +784,6 @@ Replace `<namespace>` with the name of your namespace.
 
 See the following topics for more information:
 
-- See xref:../volsync/adv_config_volsync.adoc##volsync-rsync-tls-clusters[Creating a secret for Rsync-TLS replication ] to learn how to create your own secret for an Rsync-TLS replication.
+- See xref:../volsync/adv_config_volsync.adoc##volsync-rsync-tls-clusters[Creating a secret for Rsync-TLS replication] to learn how to create your own secret for an Rsync-TLS replication.
 - For additional information about Rsync, see https://volsync.readthedocs.io/en/latest/usage/index.html[Usage] in the VolSync documentation.
-- For additional information about the backup options, see https://volsync.readthedocs.io/en/latest/usage/restic/index.html#backup-options[Backup options] in the VolSync documentation.
\ No newline at end of file
+- For additional information about the backup options, see https://volsync.readthedocs.io/en/latest/usage/restic/index.html#backup-options[Backup options] in the VolSync documentation.
```
tesshuflower: sorry, I didn't notice this one before - should we say "For additional information about restic options" .... as it's pointing to a restic section of the docs

oafischer: I'll update it, also waiting for a QE review (new process we are trying to include in 2.8 for all new feature items in our docs).

tesshuflower (approved):
/lgtm

Thanks a lot Oliver!

oafischer:
Merging now as this is blocking work, QE can still look at this PR for their check and recommend changes in comments that we can address with a new PR.

PR #4817, opened 2023-03-15
[ACM-4111]: Updates to creating a hosted cluster on bare metal

lahinson:
Jira: https://issues.redhat.com/browse/ACM-4111

@rokej @avishayt PTAL at my comment in the doc about creating a hosted cluster on bare metal. I think we can be clearer about what the user is supposed to do. I'd appreciate your feedback. Thanks!

### DIFF
--- a/clusters/hosted_control_planes/managing_hosted_bm.adoc
+++ b/clusters/hosted_control_planes/managing_hosted_bm.adoc
@@ -24,8 +24,8 @@ You must configure hosted control planes for bare metal before you can create ho
 
 Verify that you have a default storage class configured for your cluster. Otherwise, you might end up with pending PVCs.
 
-// What exactly should users do in this step? Do they enter the following commands? Or is the following code the output of another step?
-
+. In a shell, enter the following commands:
++
 ----
 export CLUSTERS_NAMESPACE="clusters"
 export HOSTED_CLUSTER_NAME="example"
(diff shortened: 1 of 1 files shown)


clusters/hosted_control_planes/managing_hosted_bm.adoc
```diff
@@ -24,6 +24,8 @@ You must configure hosted control planes for bare metal before you can create ho
 
 Verify that you have a default storage class configured for your cluster. Otherwise, you might end up with pending PVCs.
 
+// What exactly should users do in this step? Do they enter the following commands? Or is the following code the output of another step?
```
avishayt: Good call - these are commands to run in a shell.  Same goes for the "oc -n ..." command on line 51.

lahinson: @avishayt Thanks for the feedback! I made a couple of updates to the file. When you can, PTAL and let me know what you think.

avishayt:
Looks great, thanks!

clusters/hosted_control_planes/managing_hosted_bm.adoc
```diff
@@ -24,6 +24,8 @@ You must configure hosted control planes for bare metal before you can create ho
 
 Verify that you have a default storage class configured for your cluster. Otherwise, you might end up with pending PVCs.
 
+. In a shell, enter the following commands:
```
oafischer: You can leave out the shell here, I don't think we use that.

lahinson: Done. Thanks @oafischer!

oafischer (approved):

PR #4833, opened 2023-03-18
Removing asterisks clusters folder

dockerymick:
https://issues.redhat.com/browse/ACM-3392


dockerymick:
Merging, I only fixed the asterisks in the API docs. Confirmed that all of the APIs are updated

PR #4859, opened 2023-03-23
Updates for multiline YAML strings in templates

dockerymick:
https://issues.redhat.com/browse/ACM-3116

### DIFF
--- a/governance/config_policy_ctrl.adoc
+++ b/governance/config_policy_ctrl.adoc
@@ -50,7 +50,7 @@ spec:
           name: pod-name
           ports:
           - containerPort: 80
-  object-templates-raw:
+  object-templates-raw: |
   - complianceType: musthave
     objectDefinition:
     apiVersion: v1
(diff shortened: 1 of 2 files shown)


governance/config_policy_ctrl.adoc
```diff
@@ -50,6 +50,17 @@ spec:
           name: pod-name
           ports:
           - containerPort: 80
+  object-templates-raw:
```
willkutler: This would need a `|` character at the end of the line so that the YAML after is read as a string rather than an object. So it would be `object-templates-raw: |`

governance/config_policy_ctrl.adoc
```diff
@@ -113,6 +124,10 @@ spec:
 | Required
 | A Kubernetes object (either fully defined or containing a subset of fields) for the controller to compare with objects on the managed cluster.
 
+| `spec.object-templates-raw`
+| Optional
+| Used to add multiple YAML strings for policy templates. *Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, at least one of the two parameter fields must be set.
```
willkutler: I think this could be a little confusing because the field is only one string - maybe something like "used to set object templates with a raw YAML string"? Also, could you move this so it's next to object-templates in the table? Thanks!

dockerymick:
@willkutler thanks for your patience and feedback. When you have an opportunity, will you take another look

swopebe (approved):

dockerymick:
@willkutler @swopebe thanks for the review and approval!

PR #4874, opened 2023-03-30
[ACM-4338]: Adding docs about distributing hosted cluster workloads

lahinson:
Related Jira task: https://issues.redhat.com/browse/ACM-4338

This PR adds a new topic about distributing hosted cluster workloads. This is an admin task that applies to all platforms.


lahinson:
@enxebre When you're back from PTO, please take a look at this PR. It adds information about distributing hosted cluster workloads to the product docs. Thanks!

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -0,0 +1,41 @@
+[#hosted-cluster-workload-distributing]
+= Distributing hosted cluster workloads (Technology Preview)
+
+If you are the administrator of a management cluster, you can colocate and isolate pods for hosted clusters.
+//Do we need to say more about why this is needed, or will the intended audience understand the importance of colocating and isolating pods?
```
enxebre: This might be useful to prevent noisy neighbour issues or to allocate workload with specific needs.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -0,0 +1,41 @@
...
++
+----
+  spec:
+    nodeSelector:
+      role.kubernetes.io/infra: "" 
+----
+
+[#hosted-cluster-workload-distributing-topology]
+== Topology
+
```
enxebre: this is not correct. They don't "work best".
The behaviour is as described here https://hypershift-docs.netlify.app/how-to/distribute-hosted-cluster-workloads/ based on toleration and preference at scheduling time.

lahinson: @enxebre Thanks for the clarification. I saw the wording "...prefer to be..." in the upstream docs, and I assumed it meant that the pods work the best in those situations. I'll update the wording. Unfortunately, I can't use the phrase "prefer" because that suggests that pods have human characteristics (i.e., they have preferences the same way that people do). Is the following revision accurate?

>Pods for a hosted cluster can work with taints for `control-plane` and `cluster`, and can be scheduled into `control-plane` nodes and their own `cluster` nodes.

lahinson: Seth pointed me to a resource that helped be better understand what the doc meant by "preferred". I updated the file again to use wording that is hopefully clear and accurate. Can you take a look and let me know?

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -0,0 +1,42 @@
...
+. If the `ControllerAvailabilityPolicy` is `HighlyAvailable`, schedule pods for each deployment within a hosted cluster across different failure domains. To do so, set `topology.kubernetes.io/zone` as the topology key.
+
+. To enable a hosted cluster to require its pods to be scheduled into particular nodes, set `HostedCluster.spec.nodeSelector`, as shown in the following example:
++
+----
+  spec:
+    nodeSelector:
+      role.kubernetes.io/infra: "" 
+----
```
zanetworker: This setting is actually recommended. This is to avoid charging additional subscriptions for control-planes hosted on the nodes. 

The above example shows how to provision control-planes on infra nodes, this should be the recommendation

lahinson: Thanks for the feedback, @zanetworker! I updated the text as follows:

> To enable a hosted cluster to require its pods to be scheduled into infra nodes, set `HostedCluster.spec.nodeSelector`, as shown in the following example. This setting avoids charging additional subscriptions for control planes that are hosted on the nodes:

Is that revision accurate?

zanetworker: The below statement is more accurate, I'd say.

This way, hosted control planes for each tenant are considered eligible infrastructure node workloads, and no need to entitle the underlying OpenShift nodes.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -0,0 +1,42 @@
+[#hosted-cluster-workload-distributing]
+= Distributing hosted cluster workloads (Technology Preview)
+
+If you are the administrator of a management cluster, you can colocate and isolate pods for hosted clusters. For example, you might need to allocate workload for specific needs or prevent a user from allocating load on resources in such a way that other users are negatively affected.  
+
+* To distribute hosted cluster workloads, you can use the following node labels and taints:
```
enxebre: >To distribute hosted cluster workloads, you can use the following node labels and taints:
>//Is this enough information? Where does the admin use the node labels and taints?

We could say:

"As a management cluster admin you can use the following labels and taints in the management cluster Nodes to influence the scheduling of the control plane workload:
** `hypershift.openshift.io/control-plane: true`
** `hypershift.openshift.io/cluster: ${HostedControlPlane Namespace}`

Pods for a hosted cluster have tolerations and are scheduled according to the affinity rules. Based on this, pods for a hosted cluster function as follows:
* They tolerate taints for `control-plane` and their own `cluster`.
* The scheduler will prefer to schedule them into `control-plane` nodes.
* The scheduler will prefer to schedule them into their own `cluster` nodes.
* The scheduler will prefer to schedule them into the same node.
"


FWIW this is the kube core definition for PreferredDuringSchedulingIgnoredDuringExecution
```
	// The scheduler will prefer to schedule pods to nodes that satisfy
	// the affinity expressions specified by this field, but it may choose
	// a node that violates one or more of the expressions. The node that is
	// most preferred is the one with the greatest sum of weights, i.e.
	// for each node that meets all of the scheduling requirements (resource
	// request, requiredDuringScheduling affinity expressions, etc.),
	// compute a sum by iterating through the elements of this field and adding
	// "weight" to the sum if the node has pods which matches the corresponding podAffinityTerm; the
	// node(s) with the highest sum are the most preferred.
	// +optional
	PreferredDuringSchedulingIgnoredDuringExecution
```
I'd avoid mentioning PreferredDuringSchedulingIgnoredDuringExecution explicitely since it's an impl detail subject to change

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -0,0 +1,42 @@
...
+* They work with taints for `control-plane` and `cluster`.
+* They can be scheduled into `control-plane` nodes.
+* They can be scheduled into their own `cluster` nodes.
+* They can be scheduled into the same node.
+
+[#hosted-cluster-workload-distributing-priority]
+== Priority classes
+
+Hosted control planes use `PriorityClass` resources to drive priority and preemption of managed pods. It installs four priority classes in a management cluster in the following order of priority from highest to lowest:
+// In the previous sentence, what does "It" refer to? Does it mean "Hosted control planes installs four priority classes..."? Or does it mean "The `PriorityClass` resources install four priority classes..."?
```
enxebre: this is something the dev cli installs for you, we could say. It's not clear if will provide oot automation for creating the priority clases in the management cluster. And admin can always create them. For now, wWe could say:

"There are four built-in priority classes that influences priority an preemption of the hosted cluster pods. They can be created in the management cluster with the following order of priority from highest to lowest:
* `hypershift-operator`: HyperShift Operator pods.
* `hypershift-etcd`: Pods for etcd.
* `hypershift-api-critical`: Pods that are required for API calls and resource admission to succeed. These pods include pods such as `kube-apiserver`, aggregated API servers, and web hooks.
* `hypershift-control-plane`: Pods in the control plane that are not API-critical but still need elevated priority, such as the cluster version Operator.
"

enxebre:
Thanks! I dropped some more comments. Feel free to address here or follow up if they make sense

/lgtm

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+If you are the administrator of a management cluster, you can colocate and isolate pods for hosted clusters. For example, you might need to allocate workload for specific needs or prevent a user from allocating load on resources in such a way that other users are negatively affected.  
+
+* To distribute hosted cluster workloads, you can use the following node labels and taints:
+
+** `hypershift.openshift.io/control-plane: true`
+** `hypershift.openshift.io/cluster: ${HostedControlPlane Namespace}`
+//Is this enough information? Where does the admin use the node labels and taints?
+
+* If the `ControllerAvailabilityPolicy` is `HighlyAvailable`, schedule pods for each deployment within a hosted cluster across different failure domains. To do so, set `topology.kubernetes.io/zone` as the topology key.
```
zanetworker: We should either communicate this as the default or highlight that the recommended option is `HighleyAvailable`

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -0,0 +1,42 @@
...
+
+* They work with taints for `control-plane` and `cluster`.
+* They can be scheduled into `control-plane` nodes.
+* They can be scheduled into their own `cluster` nodes.
+* They can be scheduled into the same node.
+
+[#hosted-cluster-workload-distributing-priority]
+== Priority classes
+
+Hosted control planes use `PriorityClass` resources to drive priority and preemption of managed pods. It installs four priority classes in a management cluster in the following order of priority from highest to lowest:
```
zanetworker: "uses" given that Hosted control planes is a single product.

lahinson: Good catch!

lahinson:
@zanetworker @enxebre Thanks so much for your feedback! I made a few more updates based on your latest comments. When you can, PTAL.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -13,7 +13,7 @@ Pods for a hosted cluster have tolerations and are scheduled according to affini
 * The scheduler prefers to schedule them into their own `cluster` nodes.
 * The scheduler prefers to schedule them into the same node.
 
-If the `ControllerAvailabilityPolicy` is `HighlyAvailable`, schedule pods for each deployment within a hosted cluster across different failure domains. To do so, set `topology.kubernetes.io/zone` as the topology key.
+The default `ControllerAvailabilityPolicy` option is `HighlyAvailable`. If you use that default option, schedule pods for each deployment within a hosted cluster across different failure domains by setting `topology.kubernetes.io/zone` as the topology key.
```
zanetworker: I don't think it's the default is `HA` but its definitely the recommended option, right @enxebre? 

That said, It would make sense to make HA the default option in the future.

lahinson: If we want people to use HA, should I adjust the wording to state that explicitly? For example:

> For the `ControllerAvailabilityPolicy` option, use `HighlyAvailable`. When you use that option, you can schedule pods for each deployment within a hosted cluster across different failure domains by setting `topology.kubernetes.io/zone` as the topology key.

zanetworker: Yes, we can also explicitly state that non-HA is only good for development and testing.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -0,0 +1,35 @@
...
+* `hypershift.openshift.io/control-plane: true`
+* `hypershift.openshift.io/cluster: ${HostedControlPlane Namespace}`
+
+Pods for a hosted cluster have tolerations and are scheduled according to affinity rules. As a result, pods for a hosted cluster function as follows: 
+
+* They tolerate taints for `control-plane` and their own `cluster`.
+* The scheduler prefers to schedule them into `control-plane` nodes.
+* The scheduler prefers to schedule them into their own `cluster` nodes.
+* The scheduler prefers to schedule them into the same node.
+
```
swopebe, line 10: Usually ask that we try to avoid anthropomorphism and pronouns: https://www.ibm.com/docs/en/ibm-style?topic=grammar-anthropomorphism.

I tried to edit but I am not exactly sure what this is saying--happy to have a 5 minute discussion to brainstorm for clarity and to remove some of the possessive and such.

Do we need code block around all of these?

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -0,0 +1,35 @@
...
+For the `ControllerAvailabilityPolicy` option, use `HighlyAvailable`. When you use that option, you can schedule pods for each deployment within a hosted cluster across different failure domains by setting `topology.kubernetes.io/zone` as the topology key.
+
+To enable a hosted cluster to require its pods to be scheduled into infra nodes, set `HostedCluster.spec.nodeSelector`, as shown in the following example. This way, hosted control planes for each tenant are eligible infrastructure node workloads, and you do not need to entitle the underlying {ocp-short} nodes.
++
+----
+  spec:
+    nodeSelector:
+      role.kubernetes.io/infra: "" 
+----
```
swopebe, line 20: The grayed out content here is a sign that the formatting in the PR may be off - usually due to an asterisks. I will play with it a little.

lahinson: Thank you!

swopebe: I cannot fix easily, it seems there is something off in this YAML and it could be a copy/paste issue. I find at times if I go back and insert it myself, this issue can be resolved. Just an thought.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
swopebe: Wonder if something like:

The scheduler default preference setting schedules pods to `control-plane` nodes.

swopebe: The `Preferred` setting is...

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -0,0 +1,38 @@
+[#hosted-cluster-workload-distributing]
+= Distributing hosted cluster workloads (Technology Preview)
+
+As a management cluster administrator, you can use the following labels and taints in the management cluster nodes to schedule the control plane workload:
+
+* `hypershift.openshift.io/control-plane: true`
+* `hypershift.openshift.io/cluster: ${HostedControlPlane Namespace}`
+
```
lahinson, line 8: @enxebre @zanetworker When you can, PTAL at lines 9 - 14. During peer review, the reviewer pointed out that the wording in that list wasn't following our [style guidelines](https://www.ibm.com/docs/en/ibm-style?topic=grammar-anthropomorphism), which warn against anthropomorphism (i.e., assigning human characteristics to inanimate objects).

I adjusted the wording so that it follows the style guidance, and I think it still retains the original meaning, but I'd like you to confirm that. In the revision the phrase "...the `Preferred` setting..." refers to the setting where the admin specifies preferences for scheduling.

swopebe (approved):

PR #4883, opened 2023-04-03
no issue, clearing old content

swopebe:


dockerymick (approved):
Thanks for removing the comments and keeping the content! The updates look good to me

PR #4886, opened 2023-04-03
https://issues.redhat.com/browse/ACM-2113--adding 2.8 changes for con…

swopebe:
…sole

### DIFF
--- a/console/console.adoc
+++ b/console/console.adoc
@@ -47,8 +47,6 @@ For more information about managing clusters, see link:../clusters/cluster_lifec
 
 Additionally, see specific information on these cluster types at link:../clusters/cluster_lifecycle/ansible_config_cluster.adoc#ansible-config-cluster[Configuring {aap-short} tasks to run on managed clusters].
 
-You can 
-
 [#applications-nav]
 == Applications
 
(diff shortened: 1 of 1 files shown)


jeswanke (approved):

console/console.adoc
```diff
@@ -19,23 +19,19 @@ image:../images/nav-2.7.png[Navigation]
 [#home-page]
 == Home
 
-From the {product-title} _Home_ page, you get more information about the product and you can access header features, as well as the pages for the major components of the product. 
+From the {product-title} _Home_ page, you can access more infomation and you can search across the product. 
 
-- Access the _Welcome_ page and the _Overview_, which gives you visibility into your clusters.
+Click *Welcome* for introductory information about each product function. Click *Overview* to see updates, such as _Application_ status, or to access clickable _Cluster_ percentage values for policy violations and more.
```
cadawson, line 24: Not sure if this works much better, but take a look.

cadawson (approved):
Looks good. One thought, but not required to change. Thank you!

console/console.adoc
swopebe: no we should not be using like @cadawson -- I always get rid of it.

PR #4889, opened 2023-04-04
Fix hosted link

oafischer:
image


dockerymick (approved):
Ran the link check tool for both files. All links work according to the tool. Thanks for updating this!

PR #4895, opened 2023-04-04
@swopebe https://issues.redhat.com/browse/ACM-2113--adding what's new

swopebe:


jeswanke (approved):

release_notes/whats_new.adoc
```diff
@@ -15,6 +15,8 @@
 [#web-console-whats-new]
 == Web console
 
+* You can get information about your cluster add-ons from the _Overview_ page. Click *Cluster add-ons* for information about availability and health.
```
cadawson, line 18: Overview page in the other PR was not in italics. It is here. The style guide says use quotes with sentence capitalization, while using none with headline capitalization. I think it used to be italic, so I see where that came from. With one word, it is not evident which capitalization format is being used. I left it as "no highlighting" in the other one, but one of them should probably change to be consistent.

cadawson (commented):
Looks good other than a highlighting inconsistency with the other PR. Thanks!

release_notes/whats_new.adoc
swopebe: @cadawson remember that we used italics to describe a page or a window. This is pretty consistent within the doc, and bold for clickable items and never quotes unless in code. I will make them both italics, thanks.

PR #4922, opened 2023-04-11
2.6.5 docs errata

oafischer:
Might still change according to release manager.

GA on May 2nd 2023.


dockerymick (approved):
Looks good from my perspective, I would add the `link:` syntax, but I understand that this change is for 2.6 so we don't need to update the syntax. Syntax consistency effort done in 2.8

PR #4923, opened 2023-04-11
https://issues.redhat.com/browse/ACM-4815 Update credentials CA

oafischer:
While working on this I noticed that the console has changed a decent amount since these docs were written. Some steps are in the wrong order and most are not necessary anymore in my opinion since the console is clear enough. Might be another tech debt issue to go through all credentials and update the console steps.

### DIFF
--- a/clusters/credentials/credential_openstack.adoc
+++ b/clusters/credentials/credential_openstack.adoc
@@ -23,33 +23,15 @@ See https://access.redhat.com/documentation/en-us/openshift_container_platform/4
 
 To create a credential from the {mce-short} console, complete the steps in the console. 
 
-Start at the navigation menu. Click *Credentials* to choose from existing credential options. *Tip:* Create a namespace specifically to host your credentials, for both convenience and added security.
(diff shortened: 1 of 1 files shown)


dockerymick (changes requested):
Will you also add an additional resources section so that you can add inline links at the end? As a reminder or consideration an inline link might be needed in the Prereqs

oafischer:
Excellent reminder, I can convert the inline links, but that means we need to commit to rewriting the credentials docs because it might be a bit weird to have only one topic be written according to our new standards while the others remain untouched.

dockerymick:
@oafischer good point, seems like tech debt to improve the credentials doc. Approving the PR

dockerymick (approved):
Links were checked as well . thanks for updating this

oafischer:
Thanks for approving @dockerymick, gonna merge so we can address the reported issue for now. I've already talked a bit with Scott to lay out a plan for rewriting credentials. I will create a tech debt issue for this soon.

PR #4930, opened 2023-04-13
Update whats_new.adoc

dockerymick:
Updating the link 
https://issues.redhat.com/browse/ACM-4822


dockerymick:
Quick update and verified that the link was correct by using the link check tool, and selecting the link that I replaced. Confirmed that is the correct link. Merging

PR #4905, opened 2023-04-06
acm-4374 update import instructions 2.8

cadawson:

PR #4947, opened 2023-04-18
acm4374 update import doc 2.7

cadawson:


cadawson:
This is a backport that was approved in: https://github.com/stolostron/rhacm-docs/pull/4905

PR #4953, opened 2023-04-19
https://issues.redhat.com/browse/ACM-formatting issue

swopebe:


cadawson (changes requested):
From Slack:
Two thoughts here. 1. I agree that when it goes out for GA, we don’t want errata before what’s new. 2. I can see the benefit of having the errata before what’s new when we have some errata (when the first z-stream goes out). 
In some earlier releases, we removed the errata topic from the initial GA release. I think we ran out of time to do that last release, but I don’t think we need to include the errata at GA if it is empty. We can add it when we need it. 
I guess whether the errata goes before the What’s new after some errata is included is another conversation. If you still feel that way, I am not adamant about it, but it might be a good discussion. Thanks!

swopebe:
Hmmm not sure what that means? The first topic should be the What's new, we don't even have an Errata when 2.8 comes out? I didn't mean to derail progress...this really is a simple change and matches the intro.

cadawson (approved):
Talked about this one. All good. Thanks!

swopebe:
added clarification for 2.8

PR #4956, opened 2023-04-20
add addon to apis

haoqing0110:
@haoqing0110 for a change like this, we definitely need an issue because it takes time to edit this...we also need to see the TRAIN label on this issue so we can relate it to the train and prioritize it as such. Thank you so much!

Related Jira https://issues.redhat.com/browse/ACM-5055


haoqing0110:
Thanks @dockerymick and cc @cadawson , plz hold this PR and we are waiting for response in https://issues.redhat.com/browse/ACM-2200 whether we are going to release this feature in 2.8.

apis/addondeploymentconfig.json.adoc
```diff
@@ -0,0 +1,285 @@
...
+[[_rhacm-docs_apis_addondeploymentconfig_jsonaddondeploymentconfig_nodePlacement]]
+*nodePlacement*
+
+[options="header", cols=".^2a,.^3a,.^4a"]
+|===
+|Name|Description|Schema
+|*nodeSelector* +
+__optional__|Defines which nodes the pods are scheduled to be enabled. If the selector is an empty list, the selector matches all nodes.|map[string]string
+|*tolerations* +
+__optional__|Connected by pods and used to list any taint that matches the triple <key,value,effect> using the matching operator <operator>.|[]corev1.Toleration
```
dockerymick: @haoqing0110 I made an update here, can you confirm that my change is accurate

haoqing0110: @dockerymick It seems to have changed a little bit of meaning. 
For the node selector, it defines which Nodes the Pods will be scheduled and running on, using "enabled" is not accurate. 
For the tolerations, it is to tolerate any taint that matches the triple <key,value,effect> using the matching operator <operator>.

dockerymick: "For the node selector, it defines which Nodes the Pods will be scheduled and **running on**" seems incomplete. where do the pods and nodes run?  Is it running on the hub cluster? managed cluster? all clusters? 

Pod exceptions for any taint that matches the triple <key,value,effect> using the matching operator <operator>.|[]corev1.Toleration

haoqing0110: The pod will be running on the nodes. 

The nodeSelector defined in this API will finally be applied to pods. The meaning of node selector is the same as k8s pods' [nodeselector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector). 

For the toleration, the toleration defined in this API will also finally be applied to pods, and the meaning of toleration is the same as k8s [taint and toleration](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/). 

Hope the above info can help better understand this field.

dockerymick: @haoqing0110 thanks for the sources, I misunderstood at first. Making updates now

dockerymick: @haoqing0110 I updated `nodeSelector` to read: Define which nodes the pods are scheduled to run on. If the selector has an empty list, the selector matches all nodes.

When the selector matches all nodes, is that done automatically and how is it matched? When you say selector, are we referring to the user or to the parameter name (`nodeSelector`). Based on the reference that you provided, the pods are scheduled to nodes that have `nodeSelector` labels.

for tolerations: Applied to pods and used to schedule pods to any taint that matches the `<key,value,effect>` toleration using the matching operator (`<operator>`).

haoqing0110: > When the selector matches all nodes, is that done automatically and how is it matched? 

selector refer to the `nodeSelector`. when the `nodeSelector` is empty, so no change will make to the pods, it's by default selecting all the nodes.

haoqing0110: @dockerymick The update looks good to me. Thx!

apis/clustermanagementaddon.json.adoc
```diff
@@ -0,0 +1,299 @@
...
+
+[[_rhacm-docs_apis_clustermanagementaddon_jsonqueryclustermanagementaddon]]
+=== Query a single ClusterManagementAddOn
+....
+GET /addon.open-cluster-management.io/v1alpha1/clustermanagementaddons/{clustermanagementaddon_name}
+....
+
+
+==== Description
+Query a single ClusterManagementAddOn for more details.
```
oafischer, line 158: Is the "for more details" supposed to be there?

dockerymick: Yea, the user will receive more details about the add-on. @haoqing0110 will you confirm?

apis/managedclusteraddon.json.adoc
```diff
@@ -0,0 +1,274 @@
+[#managedclusteraddon-api]
+= ManagedClusterAddOn API (v1alpha1)
+
+[[_rhacm-docs_apis_managedclusteraddon_jsonoverview]]
+== Overview
+This documentation is for the ManagedClusterAddOn resource for {product-title}. The ManagedClusterAddOn resource has four possible requests: create, query, delete, and update.
```
oafischer, line 6: We should avoid saying "this documentation". Maybe something like "see the following overview for..." instead.

apis/clustermanagementaddon.json.adoc
```diff
@@ -0,0 +1,299 @@
+[#clustermanagementaddon-api]
+= ClusterManagementAddOn API (v1alpha1)
+
+[[_rhacm-docs_apis_clustermanagementaddon_jsonoverview]]
+== Overview
+This documentation is for the ClusterManagementAddOn resource for {product-title}. The ClusterManagementAddOn resource has four possible requests: create, query, delete, and update.
```
oafischer, line 6: Same here regarding "this documentation".

apis/addondeploymentconfig.json.adoc
```diff
@@ -0,0 +1,285 @@
+[#addondeploymentconfig-api]
+= AddOnDeploymentConfig API (v1alpha1)
+
+[[_rhacm-docs_apis_addondeploymentconfig_jsonoverview]]
+== Overview
+This documentation is for the AddOnDeploymentConfig resource for {product-title}. The AddOnDeploymentConfig resource has four possible requests: create, query, delete, and update.
```
oafischer, line 6: Same here regarding "this documentation".

oafischer (approved):
One suggestion but otherwise looks good, thanks!

apis/clustermanagementaddon.json.adoc
haoqing0110: yes, correct.

apis/managedclusteraddon.json.adoc
haoqing0110: "This documentation" is referring to other API docs in https://github.com/stolostron/rhacm-docs/tree/2.8_stage/apis, to keep consistency, personally not recommended to modify it here. Maybe another PR to modify the expression.

dockerymick: @haoqing0110 I agree with your suggestion. @oafischer I also agree with yours. Therefore, let's create a separate do issue to update that line in all of the PRs. Could be a practice issue for Faisa, I'll leave that decision for you Oliver. Let me know what your thoughts

dockerymick:
@oafischer thanks for your review and feedback. I agree with your suggestions. As discussed in Slack, we will make the statement updates in a different issue and PR. Merging this

PR #4957, opened 2023-04-20
https://issues.redhat.com/browse/ACM-4062 dr4hub arch

oafischer:
Need to fix dr4hub links in known issues still.


backup_restore/backup_activation.adoc
```diff
@@ -0,0 +1,45 @@
+[#managed-cluster-activation-data]
+= Managed cluster activation data
+
```
swopebe, line 3: If this is a task, and it seems it is, we may want to indicate that with a gerund as we are trying to do throughout the content.

backup_restore/backup_hub_config.adoc
```diff
@@ -0,0 +1,86 @@
+[#dr4hub-config]
+= Hub cluster configuration
```
swopebe: This title could be a title for any feature within RHACM or OCP so I would try a more specific approach here:
Configuring active or passive hub cluster (remember the gerund)

backup_restore/backup_install.adoc
```diff
@@ -0,0 +1,176 @@
+[#install-backup-and-restore]
+= Installing the backup and restore operator
```
swopebe, line 2: like this ^^^

backup_restore/backup_msa.adoc
```diff
@@ -0,0 +1,120 @@
+[#auto-connect-clusters-msa]
+= Automatically connecting clusters by using a Managed Service Account
+
```
swopebe, line 3: Connecting clusters automatically

backup_restore/backup_schedule.adoc
```diff
@@ -0,0 +1,222 @@
+[#using-backup-restore]
+= Scheduling and restoring backups
+
```
swopebe, line 3: see these are gerunds and for the most part on titles, sections, we want to try to be parallel unless we are dealing with a concept or reference.

swopebe (changes requested):
Couple changes to titles--I didn't review content. 

I see you have an intro but didn't see it here....that will link to all this I assume?

Thanks!

oafischer:
Implemented the suggestions, thanks for the quick review. Yes, the intro topic has all the new structure with links. I think you need to click "load diff" to see it.

swopebe (approved):

birsanv (approved):
lgtm

PR #4968, opened 2023-04-20
https://issues.redhat.com/browse/ACM-3650-- editing

swopebe:
1. Industry standard change, like reducing modals.
2. Some words removed for conciseness, like extra to-be phrases
3. Imagine replaced with a more concrete line for the example
4. DNS prefix, what the asterisks symbolizes, replaced.


swopebe:
@lahinson this is ready for feedback, also node pool is represented a few ways in this doc. I found this from GKE: https://cloud.google.com/kubernetes-engine/docs/concepts/node-pools#:~:text=A%20node%20pool%20is%20a,pool's%20name%20as%20its%20value.

Do we want to go with "node pool" throughout?

lahinson (approved):
lgtm

PR #5009, opened 2023-04-28
Add note about upstream links

oafischer:
Per recent discussion.

### DIFF
--- a/_pages/basic_content_standards.md
+++ b/_pages/basic_content_standards.md
@@ -88,7 +88,7 @@ From the Developing Quality Technical Information handbook (IBM Press), do not o
 * Name the link so that the user sees value in the title and can better decide whether to click or come back later. Don't name the link [here].
 * URLs to GitHub issues are usually long and uninformative. Take the extra step to define the link with a good link title. 
 * Never link to internal materials like internal GitHub issues, files, other internal-only docs like scripts. Have the developer replace a script with steps. Some initial drafts can require this removal. Ensure linked materials are from public spaces.
-* Avoid linking to upstream sources unless necessary.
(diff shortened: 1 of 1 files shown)


_pages/basic_content_standards.md
```diff
@@ -88,6 +88,7 @@ From the Developing Quality Technical Information handbook (IBM Press), do not o
 * Name the link so that the user sees value in the title and can better decide whether to click or come back later. Don't name the link [here].
 * URLs to GitHub issues are usually long and uninformative. Take the extra step to define the link with a good link title. 
 * Never link to internal materials like internal GitHub issues, files, other internal-only docs like scripts. Have the developer replace a script with steps. Some initial drafts can require this removal. Ensure linked materials are from public spaces.
+* Avoid linking to upstream sources unless necessary.
```
dockerymick: Is it appropriate for us to specify what upstream is, as in what repository is upstream? Or maybe provide an example of what an upstream source might be. For example, community folders. Policy-collection repo is upstream, however we want to avoid linking them to [`community` policies](https://github.com/open-cluster-management-io/policy-collection/tree/main/community)

dockerymick (changes requested):

_pages/basic_content_standards.md
oafischer: Good idea, we should definitely specify more. Maybe something like:

* Avoid linking to upstream documentation and community policies unless necessary.

Does that work?

dockerymick: @oafischer Yea, that works.

dockerymick (approved):

PR #5011, opened 2023-05-01
https://issues.redhat.com/browse/ACM-3383--move topics, redo descript…

swopebe:
…ions where needed, move folders and remvoe items not needed


about/docinfo.xml
```diff
@@ -1,8 +1,8 @@
 <productname>{product-title}</productname>
 <productnumber>{product-version}</productnumber>
-<subtitle>Read more about the product including the architecture, components, and quick start guide.</subtitle>
+<subtitle>Read more about the product including the architecture and components.</subtitle>
 <abstract>
-    <para>Read more about the product including the architecture, components, and quick start guide.</para>
+    <para>Learn about architecure and other specific information about {product-title}.</para>
```
dockerymick, line 5: In the later `.xml` file the subtitle and paragraph description are the same. Is there a reason why the paragraph isn't the same as the subtitle?

dockerymick (changes requested):
Left a comment. Did you update link references? For example, changing a link reference from backup_restore/backup_activation.adoc to business_continuity/backup_restore/backup_activation.adoc. I did a search and received 33 results
Screenshot 2023-05-01 at 1 25 11 PM

about/docinfo.xml
swopebe: fixed, subtitles changed.

dockerymick (approved):
thanks for the changes! this looks good to me !

PR #5013, opened 2023-05-01
https://issues.redhat.com/browse/ACM-5097 2.5.8 Errata

oafischer:
GA May 4 2023.

### DIFF
--- a/release_notes/errata.adoc
+++ b/release_notes/errata.adoc
@@ -11,7 +11,7 @@ FIPS notice: If you do not specify your own ciphers in `spec.ingress.sslCiphers`
 
 * The `must-gather` command now collects the {ocp} version number. (https://issues.redhat.com/browse/ACM-2857[ACM-2857])
 
-* Fixes an issue that caused the `max_item_size` setting in the `MEMCACHED` index to not propogate changes to all `MEMCACHED` clients. (link:https://issues.redhat.com/browse/ACM-4683[ACM-4683])
+* Fixes an issue that caused the `max_item_size` setting in the `MEMCACHED` index to not propogate changes to all `MEMCACHED` clients. (https://issues.redhat.com/browse/ACM-4683[ACM-4683])
 
 * The policy status of policies with dots in their names now updates faster. (https://issues.redhat.com/browse/ACM-4736[ACM-4736])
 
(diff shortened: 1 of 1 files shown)


release_notes/errata.adoc
```diff
@@ -7,6 +7,14 @@ By default, Errata updates are automatically applied when released. See link:../
 
 FIPS notice: If you do not specify your own ciphers in `spec.ingress.sslCiphers`, then the `multiclusterhub-operator` provides a default list of ciphers. For 2.4, this list includes two ciphers that are _not_ FIPS approved. If you upgrade from a version 2.4.x or earlier and want FIPS compliance, remove the following two ciphers from the `multiclusterhub` resource: `ECDHE-ECDSA-CHACHA20-POLY1305` and `ECDHE-RSA-CHACHA20-POLY1305`.
 
+== Errata 2.5.8
+
+* The `must-gather` command now collects the {ocp} version number. (https://issues.redhat.com/browse/ACM-2857[ACM-2857])
+
+* Fixes an issue that caused the `max_item_size` setting in the `MEMCACHED` index to not propogate changes to all `MEMCACHED` clients. (link:https://issues.redhat.com/browse/ACM-4683[ACM-4683])
```
cadawson:
```suggestion
* Fixes an issue that caused the `max_item_size` setting in the `MEMCACHED` index to not propogate changes to all `MEMCACHED` clients. (https://issues.redhat.com/browse/ACM-4683[ACM-4683])
```

cadawson (approved):
One comment. Looks good! Thanks!

release_notes/errata.adoc
cadawson: Just a consistency note here (does not affect function): The link in the second bullet item leads with "link:" while the other two do not. At one time, we said that we would include "link:" in external links, but it looks like one or more of them might not in the errata. The reason is probably because it is 2.5, and it is not worth changing all instances now. Unless there is a reason for them to be different, can we make them consistent across 2.5.8? Thanks.

oafischer: Didn't notice, thanks for catching that. We want to use link: starting in 2.7 I believe. I'll stick with not using it on older versions.

PR #5018, opened 2023-05-01
https://issues.redhat.com/browse/ACM-4774

oafischer:
Update troubleshooting server foundation

### DIFF
--- a/troubleshooting/trouble_cluster_offline_avail.adoc
+++ b/troubleshooting/trouble_cluster_offline_avail.adoc
@@ -20,4 +20,4 @@ oc edit managedcluster <cluster-name>
 ----
 Replace _cluster-name_ with the name of your managed cluster.
 
-. Increase the value of `leaseDurationSeconds` in your `ManagedCluster` specification. The default value is 5 minutes, but that might not be enough time to maintain the connection with the network issues. Specify a greater amount of time for the lease. For example, you can raise the setting to 20 minutes.
+. Increase the value of `leaseDurationSeconds` in your `ManagedCluster` specification. The default value is 60 seconds. If the `ManagedCluster` lease is not renewed in 300 seconds, the cluster status changes from available to offline. The default value might not be enough time to maintain the connection with the network issues. Specify a greater amount of time for the lease. For example, you can raise the setting to 240 seconds.
--- a/troubleshooting/trouble_cluster_offline_avail.adoc
+++ b/troubleshooting/trouble_cluster_offline_avail.adoc
@@ -20,4 +20,4 @@ oc edit managedcluster <cluster-name>
 ----
 Replace _cluster-name_ with the name of your managed cluster.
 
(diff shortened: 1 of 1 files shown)


troubleshooting/trouble_cluster_offline_avail.adoc
```diff
@@ -20,4 +20,4 @@ oc edit managedcluster <cluster-name>
 ----
 Replace _cluster-name_ with the name of your managed cluster.
 
-. Increase the value of `leaseDurationSeconds` in your `ManagedCluster` specification. The default value is 5 minutes, but that might not be enough time to maintain the connection with the network issues. Specify a greater amount of time for the lease. For example, you can raise the setting to 20 minutes.
+. Increase the value of `leaseDurationSeconds` in your `ManagedCluster` specification. The default value is 60 seconds. If the `ManagedCluster` lease is not renewed in 300 seconds, the cluster status changes from available to offline. The default value might not be enough time to maintain the connection with the network issues. Specify a greater amount of time for the lease. For example, you can raise the setting to 240 seconds.
```
elgnay: we have to explain the relationship between  the toleration time (300 seconds) and `leaseDurationSeconds` (60 seconds).  So it might be better to replace
> If the `ManagedCluster` lease is not renewed in 300 seconds, the cluster status changes from available to offline.

with

> If the `ManagedCluster` lease is not renewed in 300 seconds (5 times of `leaseDurationSeconds`), the cluster status changes from available to offline.

oafischer: If we want to highlight that there is a connection, I would recommend stating it directly. How about the following?

> . Increase the value of `leaseDurationSeconds` in your `ManagedCluster` specification. The default value is 60 seconds.  The toleration time is the value of `leaseDurationSeconds` multiplied by 5....

elgnay: Then we need to explain the `toleration time`. How about this one?
> Increase the value of `leaseDurationSeconds` in your `ManagedCluster` specification. The default value is 60 seconds. The toleration time is the value of `leaseDurationSeconds` multiplied by 5, which dictates how long the `ManagedCluster` will stay available after the last time the `ManagedCluster` lease is renewed.

oafischer: That works, I'll update it. Thanks!

troubleshooting/trouble_cluster_offline_avail.adoc
```diff
@@ -20,4 +20,4 @@ oc edit managedcluster <cluster-name>
 ----
 Replace _cluster-name_ with the name of your managed cluster.
 
-. Increase the value of `leaseDurationSeconds` in your `ManagedCluster` specification. The default value is 5 minutes, but that might not be enough time to maintain the connection with the network issues. Specify a greater amount of time for the lease. For example, you can raise the setting to 20 minutes.
+. Increase the value of `leaseDurationSeconds` in your `ManagedCluster` specification. The default value is 60 seconds. The toleration time is the value of `leaseDurationSeconds` multiplied by 5, which controls how long the `ManagedCluster` remains available after the last time the `ManagedCluster` lease is renewed. If the `ManagedCluster` lease is not renewed in 300 seconds, the cluster status changes from available to offline. The default value might not be enough time to maintain the connection with the network issues. Specify a greater amount of time for the lease. For example, you can raise the setting to 240 seconds.
```
swopebe: Looks great, one small suggestion is to break up in to two paragraphs for easier reading.

swopebe (approved):

PR #5020, opened 2023-05-02
2.6 errata refresh

oafischer:
2.6.5 refresh (shipped today).


dockerymick (approved):
recommend to add `link:` but it's not required. In efforts to move forward, you can ignore my suggestion.

PR #5040, opened 2023-05-05
https://issues.redhat.com/browse/ACM-5024 Gatekeeper operator updates

dockerymick:
https://issues.redhat.com/browse/ACM-5024


gparvin:
@dockerymick Hey Mikela,  At this time we do not want to make this change.  We are still finalizing what gatekeeper will look like for 2.8.  We can discuss more in our sync call next week, but we are not planning on pointing customers to Helm as the solution forward for installing gatekeeper.

The current plan is to leave the current operator deprecated, and the doc update I think for deprecation should indicate that a proper migration path to a supported way of installing gatekeeper will be provided (and we may be able to provide it separately from the ACM release process so it gets to customers faster than having to wait for an ACM release).

release_notes/deprecate_remove.adoc
```diff
@@ -225,6 +219,12 @@ A _removed_ item is typically function that was deprecated in previous releases
...
 |Product or category | Affected item | Version | Recommended action | More details and links
 
+| Governance 
+| Gatekeeper operator
+| 2.8
+| Install Gatekeeper by using the gatekeeper image with Helm.
+| See link:../governance/create_gatekeeper.adoc#gatekeeper-policy-with-helm[Installing Gatekeeper by using Helm]
+
 | Governance
 | The management ingress used in previous releases is is removed.
```
gparvin:
```suggestion
| The management ingress used in previous releases is removed.
```

release_notes/deprecate_remove.adoc
```diff
@@ -225,6 +219,12 @@ A _removed_ item is typically function that was deprecated in previous releases
 |===
 |Product or category | Affected item | Version | Recommended action | More details and links
 
+| Governance 
+| Gatekeeper operator
+| 2.8
```
gparvin:
```suggestion
| 2.6
```

gparvin:
```suggestion
| 2.6
```

release_notes/deprecate_remove.adoc
```diff
@@ -225,6 +219,12 @@ A _removed_ item is typically function that was deprecated in previous releases
...
+| Install Gatekeeper by using the gatekeeper image with Helm.
```
gparvin: Here is where we want to encourage continued use of the deprecated operator.
The Gatekeeper Operator can continue to be used while it is deprecated.  A migration path to an updated  solution will be provided.  Gatekeeper continues to be supported while changes to the operator are being finalized.

Is that confusing?  But that's kind of the idea.

release_notes/deprecate_remove.adoc
```diff
@@ -225,6 +219,12 @@ A _removed_ item is typically function that was deprecated in previous releases
...
+| See link:../governance/create_gatekeeper.adoc#gatekeeper-policy-with-helm[Installing Gatekeeper by using Helm]
```
gparvin: Instead of pointing to helm, the previous text I think is mostly ok -- maybe change it slightly:

Continue using the deprecated operator to install Gatekeeper.  See [Managing gatekeeper operator](https://github.com/stolostron/rhacm-docs/blob/2.8_stage/governance/create_gatekeeper.adoc#managing-gatekeeper-operator-policies) policies.

gparvin (commented):
Added some quick comments to the deprecation section that I think updates based on our recent meeting.

release_notes/deprecate_remove.adoc
dockerymick: Thanks Gus! Updated

release_notes/deprecate_remove.adoc
dockerymick: Updated

release_notes/deprecate_remove.adoc
```diff
@@ -225,8 +219,14 @@ A _removed_ item is typically function that was deprecated in previous releases
 |===
 |Product or category | Affected item | Version | Recommended action | More details and links
 
+| Governance 
+| Gatekeeper operator
+| 2.6
+| Install the Gatekeeper operator by using the migration path instructions.. You can continue to use the Gatekeeper operator while it is deprecated.
```
dockerymick:
```suggestion
| Install the Gatekeeper operator by using the migration path instructions. You can continue to use the Gatekeeper operator while it is deprecated.
```

release_notes/deprecate_remove.adoc
```diff
@@ -225,8 +219,14 @@ A _removed_ item is typically function that was deprecated in previous releases
 |===
 |Product or category | Affected item | Version | Recommended action | More details and links
 
+| Governance 
+| Gatekeeper operator
+| 2.6
+| Install the Gatekeeper operator by using the migration path instructions. You can continue to use the Gatekeeper operator while it is deprecated.
+| See link:../governance/create_gatekeeper.adoc#managing-gatekeeper-operator-policies[Managing Gatekeeper operator policies] for more details.
```
dockerymick: @mprahl @gparvin are we adding the migration instructions to the Managing Gatekeeper operator policies topic?

gparvin: @dockerymick We don't have migration instructions yet because we haven't created the gatekeeper solution we want to migrate to.  Instead of pointing to migrations instructions, can we reword the content so it assures the customer we will provide migrations instructions before the current Gatekeeper operator is removed?

Something like:
You can continue to use the Gatekeeper operator while it is deprecated. Before the Gatekeeper operator is removed a new solution with migration details will be provided.

dockerymick: From my perspective, I wouldn't mention the migration until it is documented, seems like we are pre-announcing. I'll remove the following statement: "Install the Gatekeeper operator by using the migration path instructions."

governance/create_gatekeeper.adoc
```diff
@@ -1,42 +1,43 @@
...
-* <<install-gatekeeper-operator-policy,Installing gatekeeper using a gatekeeper operator policy>>
-* <<creating-a-gatekeeper-policy-from-the-console,Creating a gatekeeper policy from the console>>
-** <<gatekeeper-operator-sample,Gatekeeper operator CR>>
-* <<upgrading-gatekeeper-gatekeeper-operator,Upgrading gatekeeper and the gatekeeper operator>>
-* <<updating-gatekeeper-operator-policy,Updating gatekeeper operator policy>>
-** <<viewing-gatekeeper-operator-policy-from-the-console,Viewing gatekeeper operator policy from the console>>
-** <<disabling-gatekeeper-operator-policy,Disabling gatekeeper operator policy>>
-* <<deleting-gatekeeper-operator-policy,Deleting gatekeeper operator policy>>
-* <<uninstalling-gatekeeper,Uninstalling gatekeeper policy, gatekeeper, and gatekeeper operator policy>>
+* <<gatekeeper-policy-with-helm,Installing Gatekeeper by using Helm>>
```
mprahl: Should this still be here?

release_notes/deprecate_remove.adoc
mprahl: @dockerymick I think it's important to reassure customers that a replacement will come though. Right now, it just sounds like we will remove the operator eventually without an alternative.

dockerymick: @mprahl thanks for your feedback. From my perspective, maybe we should move the Gatekeeper operator out of the removal table and into deprecations. I'm most concerned about "promising" a solution bc what happens if the migration solution is not provided? Also, future tense tone does not align with the IBM Style Guide

governance/create_gatekeeper.adoc
dockerymick: no, great catch

release_notes/deprecate_remove.adoc
mprahl: @dockerymick I agree that it should be in the deprecations section and not the removals section. I think saying something along the lines of "The operator will not be removed unless a replacement is provided" might be a good compromise?

dockerymick: @mprahl Your suggested statement uses future tense. My interpretation is that we are guaranteeing a replacement, which is not accurate bc that decision can change. The statement before the deprecation table is sufficient:

"A _deprecated_ component, feature, or service is supported, but no longer recommended for use and might become obsolete in future releases. Consider the alternative actions in the _Recommended action_ and details that are provided in the following table:"

It's fair that we explicitly tell the user that Gatekeeper can still be used:
 
"You can continue to use the Gatekeeper operator while it is deprecated."

We can add the migration or workaround in the _Recommended action_ column when it becomes available.

cadawson: Hi all. Mikela asked for my thoughts on this one, so I am including them here for convenience. It looks like you are correct that this is still a deprecation, and not a removal. A removal suggests that an alternative procedure is required to continue using the function (unless the function is altogether removed without replacement). A deprecation suggests that you can still use it until it is removed. No alternative needs to be provided because they can still use the deprecated function. It seems like that is what we are trying to say here. 

We should avoid specifying what we will do or provide in the future, unless it is complete. Plans can change beyond our control, there could be legal issues and reputation hits that we can experience if a customer based their decisions on something that we said would happen and didn't. Hope this helps.

dockerymick:
@mprahl @cadawson @gparvin thanks for yalls review and feedback. I have made updates, will you all re-review when you have an opportunity?

gparvin:
Thanks @dockerymick , it looks good to me!

release_notes/deprecate_remove.adoc
```diff
@@ -185,6 +179,17 @@ A _deprecated_ component, feature, or service is supported, but no longer recomm
 | None
 | See link:../install/adv_config_install.adoc[Advanced Configuration] for configuring install.
 
+| Governance 
+| Gatekeeper operator
+| 2.6
+| You can continue to use the Gatekeeper operator while it is deprecated. {product-title-short} governance team is currently investigating an alternative.
```
mprahl:
```suggestion
| You can continue to use the Gatekeeper operator while it is deprecated. The {product-title-short} Governance team is currently investigating an alternative.
```

mprahl (approved):

nelsonjean:
Hi @dockerymick , @cadawson,  I discussed this with @mprahl and the concern is we had previously announced deprecation of the current Operator, based on a certain strategy to use Helm charts instead, and we've now changed the strategy back to using an Operator, but one where the design approach is different than the current operator. 

So the action we have in front of us is how best to communicate to the customer that they can continue to use the deprecated operator, and that they will have a solution to move to in the future, so that they feel comfortable in using it and won't end up in a situation where they have nothing supported. If we don't provide that messaging, customers will likely think they should not use Gatekeeper because the message they've received over a number of releases already is that the Operator is going away (i.e. deprecated).

dockerymick:
@nelsonjean the team agreed with the following [statement](https://github.com/stolostron/rhacm-docs/pull/5040/files#diff-772a6a9cd62fa839ae27c43982f84bc23ba2d3f0c3f31d3148d76c828d560929R185): 

You can continue to use the Gatekeeper operator while it is deprecated. The {product-title-short} Governance team is currently investigating an alternative.

https://github.com/stolostron/rhacm-docs/pull/5040#discussion_r1200645802
https://github.com/stolostron/rhacm-docs/pull/5040#discussion_r1202511035

nelsonjean:
That looks like it strikes the delicate balance. Thanks @dockerymick!

PR #5072, opened 2023-05-15
`get routes` issue (also needed in 2.8)

swopebe:
…ed, also removing content that is no longer relevant

### DIFF
--- a/install/install_connected.adoc
+++ b/install/install_connected.adoc
@@ -240,6 +240,7 @@ oc get mch -o=jsonpath='{.items[0].status.phase}'
 ----
 oc get routes
 ----
+// no longer correct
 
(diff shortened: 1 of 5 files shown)


install/install_connected.adoc
```diff
@@ -240,6 +240,7 @@ oc get mch -o=jsonpath='{.items[0].status.phase}'
 ----
 oc get routes
 ----
+// no longer correct
```
joeg-pro: So I guess this final step (step 10) was added to have the user get the route to the ACM console. (though the wording "to find your routes" was pretty vague).  With the ACM stuff integrated into the OCP console as plugins, there is no need to find the console route.   And since this was the last step, the results were not needed by a subsequent step.  So I think it is fine just to delete the entire step.

The lead in words of the to-be-deleted step include a nugget of info, telling the user how to know when the install is done. (when status was running).  The preceding step talks about checking status, so I suggest we add the following sentence to the end of step 9 (which will now be the last).

"When the MulticlusterHub status is Running, the installation is complete."

PR #5077, opened 2023-05-17
Add symlink to fix build

oafischer:


business_continuity/backup_restore/images
```diff
@@ -0,0 +1 @@
+../../images
```
dockerymick, line 1: Is this a symlink? the images are still broken from the files

swopebe: what she said ^^^ :-)

Thanks for going through all of this!

oafischer: It is a symlink, just checked and you are right, guess this isn't a solution either then. I'll keep looking. Thanks!

PR #5082, opened 2023-05-18
https://issues.redhat.com/browse/ACM-5272 updating custom search limit

dockerymick:
https://issues.redhat.com/browse/ACM-5272

### DIFF
--- a/observability/manage_search.adoc
+++ b/observability/manage_search.adoc
@@ -65,7 +65,7 @@ The previous ConfigMap example allows `services` and `pods` to be collected from
 //i have concerns about adding the info here bc it mentions mce. It seems like this would be best in the clusters folder. I noticed that there wasn't a console folder within clusters, should there be? 
 You can customize the search result limit. Update the `console-mce-config` in the `multicluster-engine` namespace. These settings apply to all users and might affect performance. View the following performance parameter descriptions:
 
-- `SAVED_SEARCH_LIMIT` - The maximum amount of saved searches per user. By default, there is a limit of ten saved searches for each user. The default value is `10`. To update the limit, add the following key value to the `console-config` ConfigMap: `key:value: SAVED_SEARCH_LIMIT: x`.
+- `SAVED_SEARCH_LIMIT` - The maximum amount of saved searches per user. By default, there is a limit of ten saved searches for each user. The default value is `10`. To update the limit, add the following key value to the `console-config` ConfigMap: `SAVED_SEARCH_LIMIT: x`.
 - `SEARCH_RESULT_LIMIT` - The maximum amount of search results displayed in the console. Default value is `1000`. To remove this limit set to `-1`.
 - `SEARCH_AUTOCOMPLETE_LIMIT` - The maximum number of suggestions retrieved for the search bar typeahead. Default value is `10,000`. To remove this limit set to `-1`.
 
--- a/observability/manage_search.adoc
+++ b/observability/manage_search.adoc
@@ -9,6 +9,7 @@ Continue reading the following topics:
 
 - <<creating-search-configurable-collection,Creating search configurable collection>>
 - <<updating-saved-search-limit,Updating saved search limit>>
+- <<customizing-search-console,Customizing the search console>>
 - <<querying-in-the-console,Querying in the console>>
 ** <<querying-argo-apps,Querying ArgoCD applications>>
 - <<updating-klusterlet-addons-managed,Updating klusterlet-addon-search deployments on managed clusters>>
@@ -64,6 +65,15 @@ The previous ConfigMap example allows `services` and `pods` to be collected from
 == Updating saved search limit
 
 By default, there is a limit of ten saved searches for each user. To update the limit, add the following key value to the `console-config` ConfigMap, `key:value: SAVED_SEARCH_LIMIT: x`.
+
+[#customizing-search-console]
+== Customizing the search console
+//i have concerns about adding the info here bc it mentions mce. It seems like this would be best in the clusters folder. I noticed that there wasn't a console folder within clusters, should there be? 
+You can customize the search results of the search console. Update the `console-mce-config` in the `multicluster-engine` namespace. These settings apply to all users and might affect performance. View the following performance parameter descriptions:
+
+- `SAVED_SEARCH_LIMIT` - The maximum amount of saved searches per user. Default value is `10`.
+- `SEARCH_RESULT_LIMIT` - The maximum amount of results displayed in the console. Default value is `1000`.
+- `SEARCH_AUTOCOMPLETE_LIMIT` - The maximum number of suggestions retrieved for the search bar typeahead. Default value is `10,000`.
 
 [#querying-in-the-console]
 == Querying in the console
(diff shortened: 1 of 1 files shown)


observability/manage_search.adoc
```diff
@@ -63,7 +64,21 @@ The previous ConfigMap example allows `services` and `pods` to be collected from
...
+By default, there is a limit of ten saved searches for each user. To update the limit, add the following key value to the `console-config` ConfigMap: `key:value: SAVED_SEARCH_LIMIT: x`.
+
+[#customizing-search-console]
+=== Customizing the search console
+//i have concerns about adding the info here bc it mentions mce. It seems like this would be best in the clusters folder. I noticed that there wasn't a console folder within clusters, should there be? 
+You can customize the search result limit. Update the `console-mce-config` in the `multicluster-engine` namespace. These settings apply to all users and might affect performance. View the following performance parameter descriptions:
+
+- `SAVED_SEARCH_LIMIT` - The maximum amount of saved searches per user. Default value is `10`.
+- `SEARCH_RESULT_LIMIT` - The maximum amount of results displayed in the console. Default value is `1000`.
+- `SEARCH_AUTOCOMPLETE_LIMIT` - The maximum number of suggestions retrieved for the search bar typeahead. Default value is `10,000`.
```
jlpadilla:
```suggestion
- `SEARCH_RESULT_LIMIT` - The maximum amount of search results displayed in the console. Default value is `1000`. To remove this limit set to `-1`.
- `SEARCH_AUTOCOMPLETE_LIMIT` - The maximum number of suggestions retrieved for the search bar typeahead. Default value is `10,000`. To remove this limit set to `-1`.
```

observability/manage_search.adoc
```diff
@@ -63,7 +64,21 @@ The previous ConfigMap example allows `services` and `pods` to be collected from
 [#updating-saved-search-limit]
 == Updating saved search limit
 
-By default, there is a limit of ten saved searches for each user. To update the limit, add the following key value to the `console-config` ConfigMap, `key:value: SAVED_SEARCH_LIMIT: x`.
+By default, there is a limit of ten saved searches for each user. To update the limit, add the following key value to the `console-config` ConfigMap: `key:value: SAVED_SEARCH_LIMIT: x`.
+
+[#customizing-search-console]
+=== Customizing the search console
+//i have concerns about adding the info here bc it mentions mce. It seems like this would be best in the clusters folder. I noticed that there wasn't a console folder within clusters, should there be?
```
jlpadilla: Here is the best place because the customization is specific to search.  It would be harder to find in another topic, but agree with the concern of mentioning mce.

observability/manage_search.adoc
```diff
@@ -63,7 +64,21 @@ The previous ConfigMap example allows `services` and `pods` to be collected from
 [#updating-saved-search-limit]
 == Updating saved search limit
 
-By default, there is a limit of ten saved searches for each user. To update the limit, add the following key value to the `console-config` ConfigMap, `key:value: SAVED_SEARCH_LIMIT: x`.
+By default, there is a limit of ten saved searches for each user. To update the limit, add the following key value to the `console-config` ConfigMap: `key:value: SAVED_SEARCH_LIMIT: x`.
```
jlpadilla:
```suggestion
By default, there is a limit of ten saved searches for each user. To update the limit, add the following key value to the `console-config` ConfigMap: `SAVED_SEARCH_LIMIT: x`.
```

jlpadilla: Actually, should we remove this now that it's covered by the next topic?

observability/manage_search.adoc
```diff
@@ -9,6 +9,7 @@ Continue reading the following topics:
 
 - <<creating-search-configurable-collection,Creating search configurable collection>>
 - <<updating-saved-search-limit,Updating saved search limit>>
+- <<customizing-search-console,Customizing the search console>>
```
jlpadilla: Should we consolidate these 2 into a single topic?

dockerymick: @jlpadilla yea good point. I imagine Updating saved search to be a subtopic of Customizing the search console? Can the user update the search limit from the console or can the task only be completed from the CLI?

observability/manage_search.adoc
dockerymick: Wondering if there should be a search folder added to the [clusters folder](https://github.com/stolostron/rhacm-docs/tree/2.8_stage/clusters). Alternatively, we can be more specific in the title and description. For example:

"You can customize the search result limit for mce."

@swopebe will you provide your thoughts when you can?

observability/manage_search.adoc
dockerymick: @jlpadilla thoughts? ^

observability/manage_search.adoc
dockerymick: @swopebe thoughts?

dockerymick: @jlpadilla @swopebe ^

swopebe: @jlpadilla we cannot place MCE-specific only content in RHACM, pretend MCE and RHACM are not even in the same doc collection and the user doesn't know much about MCE because it is under-the-covers.

`You can customize the search result limit. Update the `console-mce-config` in the `multicluster-engine` namespace.`

If you feel strongly about this, I would like to discuss with PM, as they have indicated the need for the separation.

dockerymick: @swopebe thank you for your input! Pulling in @berenss for his thoughts

swopebe: It could be that this resource was named MCE without understanding that MCE for the rhacm user is really just "cluster".

berenss:
No, this should NOT be in the Clusters folder. This is NOT a clusters topic. This is a configuration specific to Observability/Search.
The fact that it must be edited in a resource related to MCE is what it is.
ACM always includes MCE as prerequisite. So it will be fine to instruct the user to do this knowing that MCE will already be there.

observability/manage_search.adoc
jlpadilla: @dockerymick Sorry I missed this.  
- Yes, I'd suggest a general topic "Customizing the search console" and then a subtopic for each of the 3 things we allow to customize.
- The task can be completed from either OCP console (not ACM) or `oc` CLI.  The `oc patch ...` command in this PR is the simplest 1 line sample.  Another alternative is `oc edit configmap console-mce-config -n multicluster-engine` and then show a sample of the ConfigMap.  I prefer the 1 line patch that the user can just copy/paste, but either will work.

swopebe:
@berenss remember we don't really go into MCE much for the RHACM user, very little in the doc. Since we obviously have `mce` in the resource names, we are as a team often having to ask. Sometimes it is related only to MCE users and sometimes it is not.

observability/manage_search.adoc
dockerymick: @jlpadilla thanks for clarifying that I've made updates based on your feedback

dockerymick:
@jlpadilla @swopebe @berenss i've made some updates based on the feedback and info. When you all have an opportunity, will you provide another review?

berenss (approved):
looks great to me! I know Amadeus will really appreciate to see it

dockerymick:
@berenss Thanks for your approval!

jlpadilla (approved):
/lgtm

swopebe (approved):

dockerymick:
Thanks for the approvals everyone! @swopebe @jlpadilla @berenss Merging this PR

PR #5089, opened 2023-05-19
https://issues.redhat.com/browse/ACM-5559 yaml callout

faaden123:
This may be an uneeded callout but nevertheless, still practice.


oafischer (approved):
lgtm

PR #5104, opened 2023-05-23
internal: Add ts for new book no main file

cadawson:


cadawson:
Hi Brandi. I added a troubleshooting procedure in here that I think fixed the book. You should be able to test it on 2.8 to see if it will work. If it does, please approve this PR and it will be forever in our procedure for creating a guide. Thanks!

swopebe (approved):

PR #5108, opened 2023-05-24
[ACM-5636]: Fixing bug in bare metal docs

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-5636

This PR adds information to the "Handling Ingress" section of the bare metal docs to fix an issue where the system could not access the web console.

After the wording is approved and this PR is merged, I'll backport the change to 2.7.

### DIFF
--- a/clusters/hosted_control_planes/managing_hosted_bm.adoc
+++ b/clusters/hosted_control_planes/managing_hosted_bm.adoc
@@ -487,6 +487,7 @@ EOF
 
 . In the `<hosted-cluster-name>.kubeconfig` file, make the following changes so that the Ingress controller is exposed to the L2 network:
 +
+[source,yaml]
 ----
 apiVersion: metallb.io/v1beta1
 kind: L2Advertisement
--- a/clusters/hosted_control_planes/managing_hosted_bm.adoc
+++ b/clusters/hosted_control_planes/managing_hosted_bm.adoc
@@ -485,11 +485,9 @@ metadata:
 EOF
 ----
 
-. Apply the following configuration so that the Ingress controller is exposed to the L2 network:
+. In the `<hosted-cluster-name>.kubeconfig` file, make the following changes so that the Ingress controller is exposed to the L2 network:
 +
 ----
-cat <<"EOF" | oc --kubeconfig ${HOSTED_CLUSTER_NAME}.kubeconfig apply
----
 apiVersion: metallb.io/v1beta1
 kind: L2Advertisement
 metadata:
@@ -498,7 +496,12 @@ metadata:
 spec:
   ipAddressPools:
     - ingress-public-ip
-EOF
+----
+
+. Enter the following command to apply the changes to your `<hosted-cluster-name>.kubeconfig` file:
++
+----
+oc --kubeconfig <hosted-cluster-name>.kubeconfig apply
 ----
 +
 The {ocp-short} node starts to detect the IP from the host network, and the system becomes operational.
--- a/clusters/hosted_control_planes/managing_hosted_bm.adoc
+++ b/clusters/hosted_control_planes/managing_hosted_bm.adoc
@@ -106,6 +106,7 @@ spec:
   pullSecretRef:
     name: pull-secret
   sshAuthorizedKey: ${SSH_PUB_KEY}
+  EOF
 ----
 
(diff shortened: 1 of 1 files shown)


ikke-t:
Hi, I'm the same guy as who created the Jira.

Otherwise great, but why do you remove the EOF lines? They are needed there, they shouldn't be removed. The cat command is reading until it hits line with EOF, and now if ypu remove EOF, the user needs to know to type it.

So please add them back, and I'm happy with the change. Thanks!

ikke-t:
Looks good.
/approve

clusters/hosted_control_planes/managing_hosted_bm.adoc
```diff
@@ -485,6 +485,24 @@ metadata:
...
+---
+apiVersion: metallb.io/v1beta1
+kind: L2Advertisement
+metadata:
+  namespace: metallb
+  name: ingress-public-advertisement
+spec:
+  ipAddressPools:
+    - ingress-public-ip
+EOF
```
dockerymick:
```suggestion
```

clusters/hosted_control_planes/managing_hosted_bm.adoc
```diff
@@ -485,6 +485,24 @@ metadata:
 EOF
 ----
 
+. Apply the following configuration so that the Ingress controller is exposed to the L2 network:
++
+----
+cat <<"EOF" | oc --kubeconfig ${HOSTED_CLUSTER_NAME}.kubeconfig apply
```
dockerymick:
```suggestion
oc --kubeconfig ${HOSTED_CLUSTER_NAME}.kubeconfig apply
```

dockerymick (approved):
Made some suggestions. thanks!

clusters/hosted_control_planes/managing_hosted_bm.adoc
swopebe: Our documentation always removed the EOF and cat or extra code within the commands we are given. 

#1. Not everyone sends them and our official doc (not open doc) sets standards where the UX and standards are the same, so they are not in all the commands or samples we get.
#2. It is used more in the open doc and not official doc.

Do these work without the `cat <<"EOF"`. Understand as professional writers, we are trying to keep the documentation within "one voice" standards.

swopebe: @ikke-t happy to discuss the standards of the portal doc in a meeting.

swopebe (approved):
Thanks @lahinson for the Slack discussion and implementing the standards here.

clusters/hosted_control_planes/managing_hosted_bm.adoc
ikke-t: This is confusing to me @swopede. Now that the lines got removed, the instruction doesn't work. It shouldn't be about guidelines and beauty over functionality. The instruction needs to be written differently if you follow such guide, as this is just plain wrong. You either put the text into file, and feed it to oc apply -f <file>, or you do it exactly as it was, which is handier.

Now the doc is just wrong. For the reference, here is a guide how here-docs work, the one used in here:

https://linuxize.com/post/bash-heredoc/

The original way is just perfect and commonly used practice. With that one can just copy paste the command to terminal. It can not be left as it is now.

swopebe: It's not about beauty, sorry, it is about combining steps in one command without explanation. We normally give the file and then the apply command separately for doc. If it is a two step command, please give both steps and not assume the user knows that with this single command.

ikke-t: then it would be like this for copy paste:

```
cat <<"EOF" > l2advertisement.yaml
---
kind: L2Advertisement
metadata:
  namespace: metallb
  name: ingress-public-advertisement
spec:
  ipAddressPools:
    - ingress-public-ip
EOF
oc --kubeconfig ${HOSTED_CLUSTER_NAME}.kubeconfig apply -f l2advertisement.yaml
```

The same would be needed to do for all the places EOF got removed from. I don't see how it's better though, but if that's wanted this would be the copy paste way. Changes oneliner to twoliner.

ikke-t: the original is nicer in my opinion, and doesn't leave a temp file laying around. It's one oc command that applies the given input.

PR #5113, opened 2023-05-24
[ACM-5637]: Updating autoscaling docs for bare metal

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-5637

This PR adds additional information to the bare metal docs about autoscaling.

### DIFF
--- a/clusters/hosted_control_planes/managing_hosted_bm.adoc
+++ b/clusters/hosted_control_planes/managing_hosted_bm.adoc
@@ -561,7 +561,7 @@ clusteroperator.config.openshift.io/storage                                    4
 
 When you need more capacity in your hosted cluster and spare agents are available, you can enable auto-scaling to install new agents. 
 
-. To enable auto-scaling, enter the following command. In this case, the minimum number of nodes is 2, and the maximum number is 5.
+. To enable auto-scaling, enter the following command. In this case, the minimum number of nodes is 2, and the maximum number is 5. The maximum number of nodes that you can add is bound by the number of available agents.
 +
 ----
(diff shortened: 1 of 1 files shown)


eranco74 (approved):
/lgtm

swopebe:
Will come back to check, seems the source is off a bit.

swopebe (commented):
I have a few questions:

In OCP doc, do you guys keep #comments in the commands and the commands all together like they are in this file? 

There is a step that says: Wait.... and then what looks like an output after. I think if the step is to wait for the following output, that can be clearer. 

Because of the large command blocks, the PR source seems to be off a little, not sure exactly where, but it could be a simple as an issue.

Also, in OCP, do you guys build really long docs like this or make them into sections depending on the function of the doc?

We are doing a little more modularity, I think? This file is getting really long, that is why I ask.

I will approve as to not hold people up but if we can address this together, that would be helpful.

lahinson:
Thanks for your comments, @swopebe! A few responses:

> In OCP doc, do you guys keep #comments in the commands and the commands all together like they are in this file?

Nope. I can clean that up.

>There is a step that says: Wait.... and then what looks like an output after. I think if the step is to wait for the following output, that can be clearer.

Agree. I can address that.

>Because of the large command blocks, the PR source seems to be off a little, not sure exactly where, but it could be a simple as an issue.

I'll take a look.

>Also, in OCP, do you guys build really long docs like this or make them into sections depending on the function of the doc?

We actually typically build much smaller files for OCP. You're right -- this one is getting really long. I can add a to-do for myself to restructure it.

lahinson:
@swopebe I plan to address your questions re: comments in the code, file length, etc. in a tech-debt Jira. The purpose of this PR is mainly to add the line from Eran about the number of nodes that you can add being bound by the number of agents available. With that in mind, would it be okay if we went ahead and merged this one and then addressed the other issues in a different PR?

swopebe:
Ok, I think it's fair just to address the long topics from this point forward and only when there is time we can modularize more. I just notice these docs are growing and growing and wondered what OCP practice is. It sounds like we all have the same standards, just do what we can within the time frame we have.

swopebe (approved):

PR #5141, opened 2023-05-30
https://issues.redhat.com/browse/ACM-5510 Updating object-templates-raw

dockerymick:
https://issues.redhat.com/browse/ACM-5510

### DIFF
--- a/governance/adv_template_process.adoc
+++ b/governance/adv_template_process.adoc
@@ -22,7 +22,7 @@ If you need to manually initiate an update, use the special annotation, `policy.
 [#raw-object-template-processing]
 == Object template processing
 
-Set object templates with a raw YAML string. The `object-template-raw` parameter is an optional parameter that support if-else statements. You can specify conditions `object-templates` specification. The following example is defined to add the `species-category: mammal` label  to any ConfigMap in the `default` namespace that has a `name` key equal to `Sea Otter`:
+Set object templates with a YAML string representation. The `object-template-raw` parameter is an optional parameter that support if-else statements. You can specify conditions in the `object-templates` specification. The following example is defined to add the `species-category: mammal` label  to any ConfigMap in the `default` namespace that has a `name` key equal to `Sea Otter`:
 
 [source,yaml]
 ----
--- a/governance/adv_template_process.adoc
+++ b/governance/adv_template_process.adoc
@@ -26,7 +26,6 @@ Set object templates with a raw YAML string. The `object-template-raw` parameter
 
 [source,yaml]
 ----
-object-templates:
 object-templates-raw: |
 {{- range (lookup "v1" "ConfigMap" "default" "").items }}
     {{- if eq .data.name "Sea Otter" }}
--- a/governance/adv_template_process.adoc
+++ b/governance/adv_template_process.adoc
@@ -42,7 +42,7 @@ object-templates-raw: |
   {{- end }}
 ----
 
-*Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, at least one of the two parameter fields must be set. 
+*Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, one of the two parameter fields must be set.
  
 [#bypass-template-processing]
 == Bypass template processing
--- a/governance/adv_template_process.adoc
+++ b/governance/adv_template_process.adoc
@@ -8,6 +8,7 @@ Use both managed cluster and hub cluster templates to reduce the need to create 
 Continue reading for advanced template use-cases:
 
 * <<special-annotation-processing,Special annotation for reprocessing>>
+* <<raw-object-template-processing,Object template processing>>
 * <<bypass-template-processing,Bypass template processing>>
 * <<additional-resources-hub-temp,Additional resources>>
 
@@ -18,6 +19,32 @@ Hub cluster templates are resolved to the data in the referenced resources durin
 
 If you need to manually initiate an update, use the special annotation, `policy.open-cluster-management.io/trigger-update`, to indicate changes for the data referenced by the templates. Any change to the special annotation value automatically initiates template processing. Additionally, the latest contents of the referenced resource are read and updated in the policy definition that is propagated for processing on managed clusters. A way to use this annotation is to increment the value by one each time.
 
+[#raw-object-template-processing]
+== Object template processing
+
+Set object templates with a raw YAML string. The `object-template-raw` parameter is an optional parameter that support if-else statements. You can specify conditions `object-templates` specification. The following example is defined to add the `species-category: mammal` label  to any ConfigMap in the `default` namespace that has a `name` key equal to `Sea Otter`:
+
+[source,yaml]
+----
+object-templates:
+object-templates-raw: |
+{{- range (lookup "v1" "ConfigMap" "default" "").items }}
+    {{- if eq .data.name "Sea Otter" }}
+    - complianceType: musthave
+      objectDefinition:
+        kind: ConfigMap
+        apiVersion: v1
+        metadata:
+          name: {{ .metadata.name }}
+          namespace: {{ .metadata.namespace }}
+          labels:
+            species-category: mammal
+    {{- end }}
+    {{- end }}
+----
+
+*Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, at least one of the two parameter fields must be set. 
+ 
 [#bypass-template-processing]
 == Bypass template processing
 
@@ -30,6 +57,7 @@ Alternatively, you can add the following annotation in the `ConfigurationPolicy`
 [#additional-resources-hub-temp]
 == Additional resources
 
-* xref:../governance/template_support_intro.adoc#template-processing[Template processing]
-* xref:../governance/template_functions.adoc#template-functions[Template functions]
+* See xref:../governance/template_functions.adoc#template-functions[Template functions] for more details.
+* Return to xref:../governance/template_support_intro.adoc#template-processing[Template processing].
+* See xref:../governance/config_policy_ctrl.adockubernetes-configuration-policy-controller[Kubernetes configuration policy controller] for more details.
 * Also refer to the link:https://docs.openshift.com/container-platform/4.11/security/encrypting-etcd.html[{ocp} etcd encryption documentation].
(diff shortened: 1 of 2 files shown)


governance/config_policy_ctrl.adoc
```diff
@@ -51,6 +51,19 @@ spec:
           ports:
           - containerPort: 80
   object-templates-raw: |
```
mprahl: @dockerymick we can't have both `object-templates` and `object-templates-raw` defined in the same `ConfigurationPolicy`. This is why a separate example is required.

dockerymick: @mprahl understood, well i can remove the info out of object-templates-raw section in the YAML. I will only add the info in the table. I recommend that we keep the single, semi-empty sample and not add a separate example for object-templates-raw.

dhaiducek: @dockerymick That makes sense for this YAML sample. I think more examples of `object-templates-raw` are still desired in the documentation, so would that mean we should create a new section in one of these to flesh out more examples?
- https://github.com/stolostron/rhacm-docs/blob/2.8_stage/governance/template_support_intro.adoc
- https://github.com/stolostron/rhacm-docs/blob/2.8_stage/governance/template_functions.adoc
- https://github.com/stolostron/rhacm-docs/blob/2.8_stage/governance/adv_template_process.adoc

governance/config_policy_ctrl.adoc
```diff
@@ -118,9 +131,9 @@ spec:
 
 `{{- if eq .metadata.name "policy-grc-your-meta-data-name" }}
                   replicas: 2
-                  {{- else }}
+ {{- else }}
                   replicas: 1
-                  {{- end }}`
+ {{- end }}`
```
dhaiducek, line 121: There are a couple updates here:
- The note in the `spec.object-templates`/`spec.object-templates-raw` sections should not have the phrase "at least" since one or the other is required but can't have both.
- Remove the `spec.object-templates-raw.spec` row entirely since that field doesn't exist. `spec.object-templates-raw` is a string, so all doc about it should be there, but I think removing all of its content and moving the `object-templates-raw` example from the YAML sample to the table would work for now.

dhaiducek (changes requested):
Thanks for the updates! I have a few more updates for this PR.

governance/config_policy_ctrl.adoc
dockerymick: Thanks for the suggestions Dale. From my perspective, I would add the example and short description to the [Advanced template processing in configuration policies](https://github.com/stolostron/rhacm-docs/blob/2.8_stage/governance/adv_template_process.adoc) doc

dockerymick: Updated this resolving the conversation

governance/config_policy_ctrl.adoc
dockerymick: Updated

dockerymick:
thanks for the suggestions and feedback. @mprahl @dhaiducek will either of you provide another review please?

oafischer (approved):
lgtm

dockerymick:
@oafischer thanks for your approval Oliver!

governance/adv_template_process.adoc
```diff
@@ -18,6 +19,32 @@ Hub cluster templates are resolved to the data in the referenced resources durin
 
 If you need to manually initiate an update, use the special annotation, `policy.open-cluster-management.io/trigger-update`, to indicate changes for the data referenced by the templates. Any change to the special annotation value automatically initiates template processing. Additionally, the latest contents of the referenced resource are read and updated in the policy definition that is propagated for processing on managed clusters. A way to use this annotation is to increment the value by one each time.
 
+[#raw-object-template-processing]
+== Object template processing
+
+Set object templates with a raw YAML string. The `object-template-raw` parameter is an optional parameter that support if-else statements. You can specify conditions `object-templates` specification. The following example is defined to add the `species-category: mammal` label  to any ConfigMap in the `default` namespace that has a `name` key equal to `Sea Otter`:
```
mprahl:
```suggestion
Set object templates with a YAML string representation. The `object-template-raw` parameter is an optional parameter that support if-else statements. You can specify conditions in the `object-templates` specification. The following example is defined to add the `species-category: mammal` label  to any ConfigMap in the `default` namespace that has a `name` key equal to `Sea Otter`:
```

governance/adv_template_process.adoc
```diff
@@ -18,6 +19,32 @@ Hub cluster templates are resolved to the data in the referenced resources durin
...
+    - complianceType: musthave
+      objectDefinition:
+        kind: ConfigMap
+        apiVersion: v1
+        metadata:
+          name: {{ .metadata.name }}
+          namespace: {{ .metadata.namespace }}
+          labels:
+            species-category: mammal
+    {{- end }}
```
mprahl: This should all be indented.

governance/adv_template_process.adoc
```diff
@@ -18,6 +19,32 @@ Hub cluster templates are resolved to the data in the referenced resources durin
...
 If you need to manually initiate an update, use the special annotation, `policy.open-cluster-management.io/trigger-update`, to indicate changes for the data referenced by the templates. Any change to the special annotation value automatically initiates template processing. Additionally, the latest contents of the referenced resource are read and updated in the policy definition that is propagated for processing on managed clusters. A way to use this annotation is to increment the value by one each time.
 
+[#raw-object-template-processing]
+== Object template processing
+
+Set object templates with a raw YAML string. The `object-template-raw` parameter is an optional parameter that support if-else statements. You can specify conditions `object-templates` specification. The following example is defined to add the `species-category: mammal` label  to any ConfigMap in the `default` namespace that has a `name` key equal to `Sea Otter`:
+
+[source,yaml]
+----
+object-templates:
```
mprahl: You can't specify both object-templates and object-templates-raw.
```suggestion
```

governance/config_policy_ctrl.adoc
mprahl: If this is the baseline sample, I think it should use `object-templates` like it used to. `object-templates-raw` is for advanced cases only.

dockerymick: @mprahl thanks for your feedback. So should I remove the entire `object-templates-raw` parameter and value from the sample and table?

governance/adv_template_process.adoc
dockerymick: Updated based on the example that is provided in the issue

governance/config_policy_ctrl.adoc
dhaiducek: @dockerymick Don't remove it from the table--just from the sample.

governance/adv_template_process.adoc
```diff
@@ -18,6 +19,31 @@ Hub cluster templates are resolved to the data in the referenced resources durin
 
 If you need to manually initiate an update, use the special annotation, `policy.open-cluster-management.io/trigger-update`, to indicate changes for the data referenced by the templates. Any change to the special annotation value automatically initiates template processing. Additionally, the latest contents of the referenced resource are read and updated in the policy definition that is propagated for processing on managed clusters. A way to use this annotation is to increment the value by one each time.
 
+[#raw-object-template-processing]
+== Object template processing
+
+Set object templates with a YAML string representation. The `object-template-raw` parameter is an optional parameter that support if-else statements. You can specify conditions in the `object-templates` specification. The following example is defined to add the `species-category: mammal` label  to any ConfigMap in the `default` namespace that has a `name` key equal to `Sea Otter`:
```
dhaiducek: if-else is one example of many functions, so I think mentioning `range` and implying there are more would be good.
```suggestion
Set object templates with a YAML string representation. The `object-template-raw` parameter is an optional parameter that supports advanced templating use-cases, such as if-else and the `range` function. The following example is defined to add the `species-category: mammal` label  to any ConfigMap in the `default` namespace that has a `name` key equal to `Sea Otter`:
```

governance/config_policy_ctrl.adoc
```diff
@@ -105,22 +118,20 @@ spec:
...
+| The array of Kubernetes objects (either fully defined or containing a subset of fields) for the controller to compare with objects on the managed cluster. *Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, one of the two parameter fields must be set.
 
 | `spec.object-templates-raw`
 | Optional
-| Used to set object templates with a raw YAML string. *Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, at least one of the two parameter fields must be set. 
-
-|`spec.object-templates-raw.spec`
-| Optional
-| Used to specify conditions for the object templates. If else statements are supported values. For example, add the following value to avoid duplication in your `object-templates` definition:
+| Used to set object templates with a raw YAML string. Specify conditions for the object templates, where if-else statements are supported values. For example, add the following value to avoid duplication in your `object-templates` definition:
```
dhaiducek:
```suggestion
| Used to set object templates with a raw YAML string. Specify conditions for the object templates, where advanced functions like if-else statements and the `range` function are supported values. For example, add the following value to avoid duplication in your `object-templates` definition:
```

dockerymick:
@dhaiducek thanks for your feedback. I removed the parameter from the sample. When you have an opportunity, will you re-review?

governance/adv_template_process.adoc
```diff
@@ -18,6 +19,31 @@ Hub cluster templates are resolved to the data in the referenced resources durin
...
+      metadata:
+        name: {{ .metadata.name }}
+        namespace: {{ .metadata.namespace }}
+        labels:
+          species-category: mammal
+  {{- end }}
+  {{- end }}
+----
+
+*Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, at least one of the two parameter fields must be set.
```
dhaiducek:
```suggestion
*Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, one of the two parameter fields must be set.
```

dhaiducek (changes requested):
Looking good! One minor update to match the other notes from the table.

dockerymick:
@dhaiducek thank you again Dale! I accepted your suggestion, are you ok with us merging?

mprahl (approved):

PR #5150, opened 2023-06-01
[ACM-5636]: Update to L2 file and command info in HyperShift bare metal docs

lahinson:
Continuing work from #5108

### DIFF
--- a/clusters/hosted_control_planes/managing_hosted_bm.adoc
+++ b/clusters/hosted_control_planes/managing_hosted_bm.adoc
@@ -482,30 +482,11 @@ kind: L2Advertisement
 metadata:
   name: ingress-public-ip
   namespace: metallb
-EOF
-----
-
-. Expose the Ingress controller to the L2 network by creating a .yaml file named `l2advertisement`:
-+
-[source,yaml]
-----
-apiVersion: metallb.io/v1beta1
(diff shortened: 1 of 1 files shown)


lahinson:
Hi @swopebe @ikke-t - This PR is a continuation from #5108. When you can, PTAL at the new changes and let me know what you think. Thank you both!

ikke-t:
Now that I read the whole doc again, I see the addition is already partly there. See the previous step has:

```
---
kind: L2Advertisement
metadata:
  name: ingress-public-ip
  namespace: metallb
EOF
```

We actually need to add only some lines to that L2Advertisement section so it becomes like:

```
---
kind: L2Advertisement
metadata:
  name: ingress-public-ip
  namespace: metallb
spec:
  ipAddressPools:
    - ingress-public-ip
EOF
```

The Spec part is missing from the original. Then you can remove the paragraphs we added here.

lahinson:
Thanks, @ikke-t! I added the spec part and removed the other paragraphs that we had added last week. Let me know what you think when you can.

lahinson:
Just need a final approval from someone on the docs team, and then this one can be merged.

swopebe (approved):
Thanks so so much!

PR #5151, opened 2023-06-01
https://issues.redhat.com/browse/ACM-4743 Adenf4743

faaden123:
Went ahead and changed this clarification on upgrades to match the one for RHACM. This was from a mock draft to kind of utilize what was already written by Brandi and adding some more information from the jira issue. 

Not sure if the " see upgrading for details" is the same for RHACM. This is information I may need further clarification on.

### DIFF
--- a/clusters/release_notes/whats_new.adoc
+++ b/clusters/release_notes/whats_new.adoc
@@ -18,7 +18,7 @@ Learn more about what is new this release:
 
 If you installed {ocp-short} or {product-title-short}, you automatically receive {mce-short}. If any new features exist for {mce-short} install only, you can view them in this section.
 
-You can now rely on the {olm} `operatorcondition` resource to enforce an efficient {mce-short} upgrade procedure, which will prevent errors from attempts to skip releases. See link:../install_upgrade/upgrade_cluster.adoc#upgrading[Upgrading] for details.
+You can now use the {olm} `operatorcondition` resource to enforce the {mce-short} upgrade procedure, which prevent errors from attempts to skip releases. See link:../install_upgrade/upgrade_cluster.adoc#upgrading[Upgrading] for details.
 
 [#cluster-lifecycle]
 == Cluster lifecycle
--- a/clusters/release_notes/whats_new.adoc
+++ b/clusters/release_notes/whats_new.adoc
@@ -18,7 +18,7 @@ Learn more about what is new this release:
 
 If you installed {ocp-short} or {product-title-short}, you automatically receive {mce-short}. If any new features exist for {mce-short} install only, you can view them in this section.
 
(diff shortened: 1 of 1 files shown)


clusters/release_notes/whats_new.adoc
```diff
@@ -18,6 +18,8 @@ Learn more about what is new this release:
 
 If you installed {ocp-short} or {product-title-short}, you automatically receive {mce-short}. If any new features exist for {mce-short} install only, you can view them in this section.
 
+You can now rely on the {olm} `operatorcondition` resource to enforce an efficient {mce-short} upgrade procedure, which will prevent errors from attempts to skip releases. See link:../install_upgrade/upgrade_cluster.adoc#upgrading[Upgrading] for details.
```
dockerymick: Remember to use [active tone](https://www.ibm.com/docs/en/ibm-style?topic=grammar-verbs#tense). When I read "rely", i interpreted the procedure to be reliable as if it was a person, which is an example of Anthropomorphism: https://www.ibm.com/docs/en/ibm-style?topic=grammar-anthropomorphism

```suggestion
You can now use the {olm} `operatorcondition` resource to enforce the {mce-short} upgrade procedure, which prevent errors from attempts to skip releases. See link:../install_upgrade/upgrade_cluster.adoc#upgrading[Upgrading] for details.
```

dockerymick (changes requested):

clusters/release_notes/whats_new.adoc
faaden123: Thank you for your review!

JakobGray (commented):
I didn't see any more details in `install_upgrade/upgrade_cluster.adoc` and that looks to be about updating managed clusters rather than MCE version so I don't think that sentence is necessary. 

Based on the above line:
> If any new features exist for {mce-short} install only, you can view them in this section.

It sounds like a release note isn't necessary for MCE at all because the operatorcondition isn't MCE only.

dockerymick (approved):

faaden123:
> I didn't see any more details in `install_upgrade/upgrade_cluster.adoc` and that looks to be about updating managed clusters rather than MCE version so I don't think that sentence is necessary.
> 
> Based on the above line:
> 
> > If any new features exist for {mce-short} install only, you can view them in this section.
> 
> It sounds like a release note isn't necessary for MCE at all because the operatorcondition isn't MCE only.

Ah- I see the sentence prior to this added note. 

> I didn't see any more details in `install_upgrade/upgrade_cluster.adoc` and that looks to be about updating managed clusters rather than MCE version so I don't think that sentence is necessary.
> 
> Based on the above line:
> 
> > If any new features exist for {mce-short} install only, you can view them in this section.
> 
> It sounds like a release note isn't necessary for MCE at all because the operatorcondition isn't MCE only.

I've taken the link out, but are you saying this note is not needed entirely?

JakobGray:
I would say nothing more needs to be added since the operatorcondition isn't user controlled anyway

PR #5156, opened 2023-06-05
[ACM-4770]: removing link to upstream HyperShift docs

lahinson:
The [Configuring the hosting cluster on AWS](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.7/html/clusters/cluster_mce_overview#hosted-enable-private-link) topic links to the upstream HyperShift docs for more information about AWS PrivateLink. I ported those upstream docs to `clusters/hosted_control_planes/deploying_aws_private_clusters.adoc` as part of an effort to replace any links to upstream docs since upstream docs are not officially supported.

The original reference to the upstream docs said "For details about the AWS secret, see Deploying AWS private clusters". However, neither the upstream docs nor the docs that I ported over from upstream to the new file really mention more details about the AWS secret, so I don't think that link is helpful or necessary.

### DIFF
--- a/clusters/hosted_control_planes/configure_hosted_aws.adoc
+++ b/clusters/hosted_control_planes/configure_hosted_aws.adoc
@@ -356,7 +356,7 @@ If you plan to provision hosted control plane clusters on the AWS platform with 
 | Required
 |===
 +
-See _Deploying AWS private clusters_ in the HyperShift documentation for more information. The following example shows the sample `hypershift-operator-private-link-credentials` secret template:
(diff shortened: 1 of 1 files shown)


oafischer (approved):
Looks great, thanks! The link you removed is from the original draft/content for 2.7 Hypershift doc updates, so I think it's safe to remove given how we have restructured in 2.8.

lahinson:
Thanks @oafischer!

PR #5174, opened 2023-06-08
https://issues.redhat.com/browse/ACM-5813 2.8 AI KI

oafischer:


clusters/release_notes/known_issues.adoc
```diff
@@ -400,4 +400,27 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
 
 When you use the console to add a new node pool to an existing hosted cluster, the same version of {ocp-short} might appear more than once in the list of options. You can select any instance in the list for the version that you want. 
 
+[#delpoy-sno-managed-fails]
```
chadcrum: #delpoy should be #deploy

clusters/release_notes/known_issues.adoc
```diff
@@ -400,4 +400,27 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
...
+=== Deploying a single node OpenShift managed cluster fails
+//2.8:ACM-5813
 
+Deploying a single node OpenShift (SNO) managed cluster might fail when the Assisted Installer agent starts slowly. You can work around the issue by disabling converged flow. Complete the following steps:
+
+. Create the following ConfigMap:
```
chadcrum: Create the following ConfigMap on the Hub Cluster

clusters/release_notes/known_issues.adoc
oafischer: Fixed!

clusters/release_notes/known_issues.adoc
oafischer: Fixed!

cadawson (approved):
Looks good, Thanks!

clusters/release_notes/known_issues.adoc
```diff
@@ -400,4 +400,27 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
 
 When you use the console to add a new node pool to an existing hosted cluster, the same version of {ocp-short} might appear more than once in the list of options. You can select any instance in the list for the version that you want. 
 
+[#deploy-sno-managed-fails]
+=== Deploying a single node OpenShift managed cluster fails
+//2.8:ACM-5813
 
+Deploying a single node OpenShift (SNO) managed cluster might fail when the Assisted Installer agent starts slowly. You can work around the issue by disabling converged flow. Complete the following steps:
```
carbonin: I'm not sure that the deployment actually "fails" in this case.

IIRC what happens is that the agent resource is never created so the install never even begins.

carbonin: This is also not specific to SNO.

Any host could fail to start our agent in this way regardless of the type of cluster the user intents to create.

Maybe it would be better as "Agent is never created after host boots" or something?

oafischer: What would the user see? If it ends in the cluster not being created when they expect it to, I would assume it still appears as a failure to them, even if it's not technically one.

Either way, we need to describe the symptom or what the user might see. If they expect to see the agent resource, but it does not appear, we can just say "The agent resource might not appear". (or "does not appear" if we can verify that is happens 100% of the time)

I'll remove the SNO portion, thanks for clarifying that.

oafischer: I'd also be interested in knowing what causes the Assisted Installer agent to start slowly? I think that would be good context to provide.

carbonin: >What would the user see?

The user would see the cluster in some kind of pending or insufficient state indefinitely.
The would also see no agent resources.

> If they expect to see the agent resource, but it does not appear, we can just say "The agent resource might not appear". (or "does not appear" if we can verify that is happens 100% of the time)

If the issue is what we're describing here then the agent resource will not appear.

> I'd also be interested in knowing what causes the Assisted Installer agent to start slowly?

Unfortunately we don't know. We haven't been able to find the full root cause for why this happens in one environment and not others.

oafischer: So it sounds like when this happens, the user will see (in the console) a cluster stuck in the `Pending` state AND expected agent resources will be missing from the console as well? If that is the case I will update the text to reflect that. Thanks!

clusters/release_notes/known_issues.adoc
```diff
@@ -400,4 +400,27 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
 
 When you use the console to add a new node pool to an existing hosted cluster, the same version of {ocp-short} might appear more than once in the list of options. You can select any instance in the list for the version that you want. 
 
+[#deploy-managed-stuck-pending]
+=== Managed cluster stuck in _Pending_ state after deployment
+//2.8:ACM-5813
 
+If the Assisted Installer agent starts slowy and you delpoy a managed cluster, the managed cluster might become stuck in the `Pending` state and not have any agent resources. You can work around the issue by disabling converged flow. Complete the following steps:
```
carbonin, line 407: I think technically it's in the "Insufficient" state, but maybe @dgonyier would be able to comment about what it looks like in the UI.

oafischer: I've used `Pending` before for UI, so I assume this might be similar. But I'll wait for @dgonyier to confirm.

oafischer: Not to rush, but if we want this to publish along with 2.8, it needs to be merged within the next hour. Otherwise it will not be picked up for GA tomorrow.

carbonin: Go with pending then if you've used it before.

It gets the idea across well enough.

carbonin (approved):

PR #5182, opened 2023-06-12
Doc fix for https://issues.redhat.com/browse/ACM-4277

rokej:
Doc fix for https://issues.redhat.com/browse/ACM-4277

### DIFF
--- a/clusters/hosted_control_planes/managing_hosted_bm.adoc
+++ b/clusters/hosted_control_planes/managing_hosted_bm.adoc
@@ -458,7 +458,10 @@ metadata:
 EOF
 ----
 
-. Configure the MetalLB Operator by creating an `IPAddressPool` with a single IP address and `BGPAdvertisement` to advertise the LoadBalancer IPs provided by the IPAddressPool via BGP protocol. This IP address must be on the same subnet as the network that the cluster nodes use.
+. Configure the MetalLB Operator by creating two resources:
++
+** An `IPAddressPool` resource with a single IP address. This IP address must be on the same subnet as the network that the cluster nodes use.
+** A `BGPAdvertisement` resource to advertise the load balancer IP addresses that the `IPAddressPool` resource provides through the BGP protocol.
 +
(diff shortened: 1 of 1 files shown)


clusters/hosted_control_planes/managing_hosted_bm.adoc
```diff
@@ -458,7 +458,7 @@ metadata:
 EOF
 ----
 
-. Configure the MetalLB Operator by creating an `IPAddressPool` with a single IP address. This IP address must be on the same subnet as the network that the cluster nodes use.
+. Configure the MetalLB Operator by creating an `IPAddressPool` with a single IP address and `BGPAdvertisement` to advertise the LoadBalancer IPs provided by the IPAddressPool via BGP protocol. This IP address must be on the same subnet as the network that the cluster nodes use.
```
lahinson:
```suggestion
. Configure the MetalLB Operator by creating two resources:
+
** An `IPAddressPool` resource with a single IP address. This IP address must be on the same subnet as the network that the cluster nodes use.
** A `BGPAdvertisement` resource to advertise the load balancer IP addresses that the `IPAddressPool` resource provides through the BGP protocol.
```
- In general, it's a best practice to include a noun after the name of a resource or variable. In my revision, I included "resource" after `IPAddressPool` and `BGPAdvertisement`. If "resource" is not the right term to use, let me know.
- The first sentence was becoming a little unwieldy with the addition of the new information, so I restructured the content to use a list for easier readability. Also, I assumed that the second sentence ("This IP address must be on the same subnet...") was referring to the IP address for the `IPAddressPool` resource.

lahinson (changes requested):
Hi @rokej - Thanks for pulling me in for this review! I suggested a revision and included a couple of comments. Let me know what you think when you can.

rokej:
@lahinson Thank you for the comments. I committed your suggested changes.

eranco74 (commented):
/lgtm

lahinson (approved):
Thanks, @rokej!

PR #5188, opened 2023-06-12
https://issues.redhat.com/browse/ACM-5754

faaden123:


release_notes/known_issues_governance.adoc
```diff
@@ -97,4 +97,9 @@ The support of pod security policies is removed from {ocp-short} 4.12 and later,
...
 ----
 violation - couldn't find mapping resource with kind PodSecurityPolicy, please check if you have CRD deployed
-----
\ No newline at end of file
+----
+
+[#duplicate-policy-name-in-the-same-policy]
+== Duplicate policy names in the same policy
+
+Creating a policy with duplicate policy template names is not detected and will lead to inconsistent results. This might happen when defining a policy with multiple configuration policies named `create-pod`. To resolve this issue, it is required to not use duplicate names for policy templates.
```
faaden123: This sounds repetitive-- The second line is alos just an example but it was included in the issue. I don't think it is needed.

release_notes/known_issues_governance.adoc
```diff
@@ -97,4 +97,9 @@ The support of pod security policies is removed from {ocp-short} 4.12 and later,
 
 ----
 violation - couldn't find mapping resource with kind PodSecurityPolicy, please check if you have CRD deployed
-----
\ No newline at end of file
+----
+
+[#duplicate-policy-name-in-the-same-policy]
+== Duplicate policy names in the same policy
+
```
cadawson: Don't forget to add the issue number: //2.8:5754

release_notes/known_issues_governance.adoc
```diff
@@ -97,4 +97,9 @@ The support of pod security policies is removed from {ocp-short} 4.12 and later,
...
+Creating a policy with duplicate policy template names is not detected and will lead to inconsistent results. To resolve this issue, it is required to not use duplicate names for policy templates.
```
cadawson:
```suggestion
Creating a policy that contains two instances of the same policy template name leads to inconsistent results. To resolve this issue, do not use duplicate names for policy templates.
```

cadawson: Are we saying that:
We create a policy that contains policy templates. 
`Mypolicy` contains
    `policy template name A`
    `Policy template name B`
We cannot use `policy template name A` for both of these. 

Or are we saying that: 
We cannot create a new policy named `Mypolicy` because there is already one with that name? 

See if the suggestion is clearer, or if I am misunderstanding. Thanks!

cadawson (dismissed):
See the comments to see if we need to make it a little clearer. Let me know if you have questions. Thanks!

release_notes/known_issues_governance.adoc
faaden123: Thank you, just did!

release_notes/known_issues_governance.adoc
faaden123: In the example given by Mathew- or the reporter of the issue, it seems to be the earlier suggestion( the policy template names). I have went and asked Mathew for further clarification for confirmation but do believe it is that first one. Thank you for your suggestions and being able to a great help on this.

release_notes/known_issues_governance.adoc
swopebe: _This might happen when_ (not the most concise way to write this, I think the style guide will give you some examples of concise writing, so please spend some time reviewing the manual)

I read this and did not understand what you were trying to say. Also, please always make things active if you can. 

Consider if this is better?

_When you create a policy with duplicate policy template names, the policy is not detected and leads to inconsistent results._ 

_Defining a policy with multiple configuration policies named `create-pod` might cause this issue._ 

Or

_When you define a policy ..... , errors can occur._ 


To resolve this issue, it is required to not use duplicate names for policy templates. (_subject, predicate disagreement)_

swopebe (changes requested):
let a few comments and revision options

release_notes/known_issues_governance.adoc
cadawson: Another option, though Brandi's are good:
Creating a policy that contains multiple policy templates with the same name is not reported as an error, but causes inconsistent results. To resolve this issue, do not use duplicate names for policy templates within the same policy.

release_notes/known_issues_governance.adoc
```diff
@@ -97,4 +97,10 @@ The support of pod security policies is removed from {ocp-short} 4.12 and later,
...
 violation - couldn't find mapping resource with kind PodSecurityPolicy, please check if you have CRD deployed
-----
\ No newline at end of file
+----
+
+[#duplicate-policy-template-name-in-the-same-policy]
+== Duplicate policy template name in the same policy
+//2.8:5754
+
+When you create duplicate policy template names within a single policy, the policy is not detected and leads to inconsistent results. Defining a policy with multiple configuration policies named `create-pod` might cause this issue. To resolve the issue, avoid using duplicate names for policy templates.
```
cadawson: When I read the issue, I understood it as:
- You are creating a policy that includes 2 (or more) policy templates within it. 
- If the 2 policy templates have the same name, there will be no error or notification that there is a problem, but there is a problem. 
- The problem that is caused (though you are not alerted of it) is that there are inconsistent results. I'm not sure what this means. Maybe one of the policy templates is skipped? 

Thoughts:
- I like the example you included. I think it helps. This can be difficult to describe. 
- Is the policy not detected, or is the error of having duplicate names not detected?  
- I might be misunderstanding, but I see it with a modification of the first sentence. Example: You must use unique names for each policy template that you include in a policy. When you create a policy that contains multiple policy templates that have the same name, it results in inconsistent results. For example, defining a a policy with multiple configuration policies named `create-pod` might cause this issue. To resolve the issue, avoid using duplicate names for policy templates in the same policy.

release_notes/known_issues_governance.adoc
```diff
@@ -103,4 +103,4 @@ violation - couldn't find mapping resource with kind PodSecurityPolicy, please c
 == Duplicate policy template name in the same policy
 //2.8:5754
 
-When you create a policy with identical policy template names, it results in inconsistent outcomes and is not detected. This means you are not made aware of the issue it can cause. For example, defining a policy with multiple configuration policies named `create-pod` leads to this problem. To resolve the issue,  avoid using duplicate names for policy templates. 
+When you create a policy with identical policy template names, it results in inconsistent outcomes and is not detected which means you are not made aware of the issue it can cause. For example, defining a policy with multiple configuration policies named `create-pod` leads to this problem. To resolve the issue, avoid using duplicate names for policy templates.
```
swopebe: Struggling with this sentence a bit still. Some things to keep in mind:

1. "which means that you" is a bit wordy, and there should be a comma before which
2. We want to try to avoid using "it" (what is "it"?)
3. No subject/predicate agreement here.

swopebe: When you create a policy with identical policy template names, you (the subject is understood you) receive inconsistent results that are not detected, so you might not know what is causing the results.

swopebe: Something like that, less wordy, all surrounding the subject, and not using it as much.

swopebe (changes requested):
see comments

release_notes/known_issues_governance.adoc
faaden123: @swopebe  ah I see it. I knew this was very long- and "which means" I knew would be an error. I definitely need to pay more attention to subject predicate agreement

swopebe (approved):

swopebe:
resolved conflict, discussed with writer. She is free to merge when engineer gives review.

mprahl (approved):

PR #5198, opened 2023-06-13
https://github.com/stolostron/rhacm-docs/pull/4945- adding reqs 

swopebe:


clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -19,6 +19,13 @@ Hosted control plane clusters offer several advantages:
 
 * Supports turn-key deployments or fully customized {ocp-short} provisioning
 
+See the following hosted control plane requirements, which were tested with {ocp-short} version 4.12.9 and later:
```
oafischer: Looks good, I personally prefer tables for minimum requirements but I understand it's preference. Approving.

oafischer (approved):

clusters/hosted_control_planes/hosted_intro.adoc
rokej: See the following highly available hosted control plane requirements, which were tested with {ocp-short} version 4.12.9 and later:

rokej: OR ... See the following hosted control plane requirements with HighlyAvailable option, which were tested with {ocp-short} version 4.12.9 and later:

swopebe: @oafischer I never do tables for these but we can review that later so we are all doing the same things. I actually think we need a requirements topic for cluster where this would go but for now, it can stay here. Thanks for that suggestion.

swopebe: @roke, I didn't realize we were talking HA, I added that.

rokej (approved):

PR #5211, opened 2023-06-20
[ACM-5665]: Restructure bare metal docs and clarify procedures and commands

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-5665

Several changes in this PR:
- Moved bare metal content from a few very long files to multiple, shorter files. The biggest changes were to `configure_hosted_bm` and `managing_hosted_bm`. The `configure_hosted_bm` file is now structured similarly to the `hosted_intro` file. Much of the content that was in the `managing_hosted_bm` and `configure_hosted_bm` files was moved to individual files.
- Pulled comments out of code blocks and into numbered call-outs after the code blocks.
- Reordered a few steps based on feedback from an SA who tested the docs. He found that he needed to create the InfraEnv first and then create the hosted cluster.
- Revised code that used heredoc syntax to be clearer and align with doc standards for both RHACM and OpenShift.

**Note:** I added comments, which begin with `// lahinson - June 2023` to sections where I either have a question or especially want SME review for accuracy.


lahinson:
@rokej When you can, PTAL at this PR, which makes several changes to the bare metal docs. I summarized the changes in the first comment in the PR. If you have any questions, please feel free to reach out to me. 

@swopebe FYI - for your awareness

rokej (approved):

lahinson:
@avishayt @eranco74 When you can, I would really appreciate your feedback on these changes to the bare metal docs. A summary of the changes is in the [first comment](https://github.com/stolostron/rhacm-docs/pull/5211#issue-1766036266) in this PR. Please let me know if you have any questions for me. Thank you!

clusters/hosted_control_planes/hosted_bare_metal_access_cluster.adoc
```diff
@@ -0,0 +1,24 @@
+[#access-hosted-cluster-bm]
+= Accessing the hosted cluster on bare metal (Technology Preview)
```
avishayt: There shouldn't be anything specific to bare metal in this file.  It should say that to run "oc" or "kubectl" commands on the hosted cluster then they need to fetch the kubeconfig file.

clusters/hosted_control_planes/hosted_bare_metal_access_cluster.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+The hosted control plane is running and the agents are ready to join the hosted cluster. Before the agents join the hosted cluster, you need to access the hosted cluster.
```
avishayt: These sentences are not needed.  The second one about needing access to the hosted cluster for agents to join is actually not correct.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
+[#hosted-bare-metal-adding-agents]
+= Adding agents to hosted control planes on bare metal (Technology Preview)
+
+You can add agents by manually configuring the machine to start with the live ISO or by using Metal3.
+
+* To manually add agents, follow these steps:
+
+. Download the live ISO and use it to start a node (bare metal or VM). At startup, the node communicates with the Assisted Service and registers as an agent in the same namespace as the `InfraEnv` resource.
```
avishayt: 1. We should specify that the URL for the live ISO can be found in the `InfraEnv` resource, under status.isoDownloadURL
2. We usually refer to a "node" as a host that is part of an OpenShift cluster.  I believe we should refer to it as "host" here.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+
+. After each agent is created, you can optionally set its `installation_disk_id` and `hostname` in the spec. Then, approve it to indicate that the agent is ready for use.
+//lahinson - June 2023 - We need to better clarify what the following command does
```
avishayt: The "get agents" command lists the agents with some properties.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+----
++
+.Example output
+----
+NAME                                   CLUSTER   APPROVED   ROLE          STAGE
+86f7ac75-4fc4-4b36-8130-40fa12602218                        auto-assign
+e57a637f-745b-496e-971d-1abbf03341ba                        auto-assign
+----
++
+// lahinson - June 2023 - We need to better clarify what the following commands do
```
avishayt: Here we are setting the installation disk and hostname, as well as approving the agent.  This is what's mentioned above on line 10.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+----
+oc -n ${HOSTED_CONTROL_PLANE_NAMESPACE} patch agent 86f7ac75-4fc4-4b36-8130-40fa12602218 -p '{"spec":{"installation_disk_id":"/dev/sda","approved":true,"hostname":"worker-0.example.krnl.es"}}' --type merge
+
+oc -n ${HOSTED_CONTROL_PLANE_NAMESPACE} patch agent 23d0c614-2caa-43f5-b7d3-0b3564688baa -p '{"spec":{"installation_disk_id":"/dev/sda","approved":true,"hostname":"worker-1.example.krnl.es"}}' --type merge
+----
++
+----
+oc -n ${HOSTED_CONTROL_PLANE_NAMESPACE} get agents
```
avishayt, line 40: Now after the "patch" command above, we can see in the output below that the agents are approved for use.

avishayt (commented):
Started a review, will continue later

clusters/hosted_control_planes/configure_hosted_bm.adoc
```diff
@@ -15,7 +15,7 @@ On {product-title-short} 2.7, you can use the managed hub cluster, also known as
 
 - Run the hub cluster and workers on the same platform for hosted control planes.
 
-- To provision hosted control planes on bare metal, you can use the Agent platform. The Agent platform uses the Central Infrastructure Management (CIM) service to add worker nodes to a hosted cluster. 
+- To provision hosted control planes on bare metal, you can use the Agent platform. The Agent platform uses the Central Infrastructure Management (CIM) service to add worker nodes to a hosted cluster. For an introduction to the CIM service, see link:https://github.com/openshift/assisted-service/blob/master/docs/hive-integration/kube-api-getting-started.md[Kube API - Getting Started Guide].
```
avishayt, line 20: Some comments on the text above - I can't add a comment on the necessary line.
1. On line 4, it says "you need a _hosting_ cluster and a _hosted_ cluster".  This is not accurate because we are creating the hosted cluster - at this point, the user needs the hosting cluster only.
2. Are we settled on the term "hosting cluster"?  I thought it was called "management cluster".
3. The term "hosting cluster" is defined on line 10 - it should be closer to line 4 where it was first used.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+.Example output
+----
+NAME                                   CLUSTER   APPROVED   ROLE          STAGE
+86f7ac75-4fc4-4b36-8130-40fa12602218             true       auto-assign
+e57a637f-745b-496e-971d-1abbf03341ba             true       auto-assign
+----
+
+* To add agents by using Metal3, follow these instructions:
+
+. Use the Assisted Service to create the custom ISO and the Baremetal Operator to perform the installation.
```
avishayt: This line is inaccurate and can be removed.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
++
+**Important:** Because the `BaremetalHost` objects are created outside the baremetal-operator namespace, you must configure the Operator to watch all namespaces.
++
+//lahinson - June 2023 - Would it be clearer to add "by entering the following command:" to the end of line 44?
```
avishayt: Yes, that would be clearer.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+----
+oc patch provisioning provisioning-configuration --type merge -p '{"spec":{"watchAllNamespaces": true }}'
+----
++
+The `metal3` pod is restarted in the `openshift-machine-api` namespace.
+
+. Wait until the `metal3` pod is ready again:
+// lahinson - June 2023 - We need to better clarify this step. It says to "Wait", but then it provides a command. Is the user supposed to enter the command, or is the following text output?
```
avishayt: The following line is a command that continuously checks the pod's status and returns once it is ready (or once it has reached a 10 second timeout).

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+oc -n ${HOSTED_CONTROL_PLANE_NAMESPACE} get agent
+----
++
+.Example output
+----
+NAME                                   CLUSTER   APPROVED   ROLE          STAGE
+4dac1ab2-7dd5-4894-a220-6a3473b67ee6             true       auto-assign   
+d9198891-39f4-4930-a679-65fb142b108b             true       auto-assign 
+da503cf1-a347-44f2-875c-4960ddb04091             true       auto-assign 
+----
```
avishayt: Which "other two nodes"?  Maybe this text should say "Repeat for all other hosts"?

clusters/hosted_control_planes/configure_hosted_bm.adoc
```diff
@@ -37,7 +33,6 @@ You must have the following prerequisites to configure a hosting cluster:
 * You need the {mce} 2.2 and later installed on an {ocp-short} cluster. The {mce-short} is automatically installed when you install {product-title-short}. You can also install {mce-short} without {product-title-short} as an Operator from the {ocp-short} OperatorHub.
 
 * You need the {mce-short} must have at least one managed {ocp-short} cluster. The `local-cluster` is automatically imported in {mce-short} 2.2 and later. See xref:../install_upgrade/adv_config_install.adoc#advanced-config-engine[Advanced configuration] for more information about the `local-cluster`. You can check the status of your hub cluster by running the following command:
-
```
avishayt, line 40: They also need to enable Central Infrastructure Management (CIM), as described here:
https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html-single/clusters/index#enable-cim

clusters/hosted_control_planes/hosted_bare_metal_infraenv.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+.. Save the file as `infraenv-config.yaml`.
+
+.. Apply the configuration by entering the following command:
++
+----
+oc apply -f infraenv-config.yaml
+----
+
+. To generate a live ISO that allows virtual machines or bare metal machines to join as agents, enter the following command:
```
avishayt: This command doesn't generate the live ISO, but rather it fetches the URL to download it from the InfraEnv resource.

clusters/hosted_control_planes/hosted_bare_metal_verifying_cluster.adoc
```diff
@@ -0,0 +1,129 @@
+[#verifying-cluster-creation-bm]
+= Verifying hosted cluster creation on bare metal (Technology Preview)
```
avishayt: There is nothing specific to bare metal in this file - consider sharing it across platforms.

clusters/hosted_control_planes/configure_hosted_bm.adoc
```diff
@@ -47,29 +42,11 @@ oc get managedclusters local-cluster
...
+To continue with the configuration of hosted control planes on bare metal, take the following steps:
 
-* You can now deploy the SR-IOV Operator. See link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.13/html/networking/hardware-networks#sriov-operator-hosted-control-planes_configuring-sriov-operator[Deploying the SR-IOV Operator for hosted control planes] to learn more about deploying the SR-IOV Operator.
+. xref:../hosted_control_planes/hosted_bare_metal_dns.adoc#configure-dns-bm[Configure DNS on bare metal]
+. xref:../hosted_control_planes/hosted_bare_metal_infraenv.adoc#hosted-control-planes-create-infraenv[Create an InfraEnv resource for hosted control planes on bare metal]
+. xref:../hosted_control_planes/hosted_create_bare_metal.adoc#creating-a-hosted-cluster-bm[Create a hosted cluster on bare metal]
+. xref:../hosted_control_planes/hosted_bare_metal_agents.adoc#hosted-bare-metal-adding-agents[Add agents to hosted control planes on bare metal]
```
eranco74: 1. This is actually just adding Agents to the infraEnv not to the hosted control plane.
2. This should be performed prior to creating a hosted control plane.
While creating a hosted control plane the user will configure nodepool (set the nuber or workers) as well, in order to have available agents to add there they need to create the agents first.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
+[#hosted-bare-metal-adding-agents]
+= Adding agents to hosted control planes on bare metal (Technology Preview)
```
eranco74: This explains how adds agents to the infraEnv not to the hosted control plane.
There should be already documented for CIM

avishayt: From what I saw, the CIM docs are only UI, and here is API.  Though we have UI here as well... @lahinson should we be documenting UI or console or both?

clusters/hosted_control_planes/hosted_bare_metal_access_cluster.adoc
lahinson: @avishayt So do the steps here work for any platform (i.e., AWS, bare metal, OpenShift Virtualization)? I ask because we have [separate docs on how to access the cluster on AWS](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/clusters/cluster_mce_overview#hosting-service-cluster-access-aws), so now I'm wondering whether those are necessary.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
lahinson: @avishayt Is line 42 inaccurate, or line 44? Just want to double-check.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
lahinson: @avishayt If we expect that users will use either the UI or console, I think it's best to document both.

clusters/hosted_control_planes/configure_hosted_bm.adoc
lahinson: @avishayt I recently just asked Adel about "management cluster" vs. "hosting cluster". I'd prefer if we simply picked one term, but apparently both are used synonymously. For OCP 4.14, we're publishing a glossary of terms for hosted control planes that explains that the management cluster and hosting cluster are synonyms. I'll update the text in the intro of this file to incorporate your feedback here. Thank you!

clusters/hosted_control_planes/hosted_bare_metal_access_cluster.adoc
rokej: https://github.com/stolostron/rhacm-docs/blob/lahinson-acm-5665-restructure-bm-docs/clusters/hosted_control_planes/hosted_bare_metal_access_cluster.adoc works for any platform. https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/clusters/cluster_mce_overview#hosting-service-cluster-access-aws works for any platform as well and I think the two documents should be combined as a single doc section with two different options. 

https://github.com/stolostron/rhacm-docs/blob/lahinson-acm-5665-restructure-bm-docs/clusters/hosted_control_planes/hosted_bare_metal_access_cluster.adoc talks about one option where users use `hypershift` CLI to generate a kubeconfig file and access.

https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/clusters/cluster_mce_overview#hosting-service-cluster-access-aws talks about another option of getting the kubeconfig file and kubeadmin credential directly from resources.

lahinson: Thanks @rokej!

> https://github.com/stolostron/rhacm-docs/blob/lahinson-acm-5665-restructure-bm-docs/clusters/hosted_control_planes/hosted_bare_metal_access_cluster.adoc works for any platform. https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/clusters/cluster_mce_overview#hosting-service-cluster-access-aws works for any platform as well and I think the two documents should be combined as a single doc section with two different options.

Will do

>https://github.com/stolostron/rhacm-docs/blob/lahinson-acm-5665-restructure-bm-docs/clusters/hosted_control_planes/hosted_bare_metal_access_cluster.adoc talks about one option where users use hypershift CLI to generate a kubeconfig file and access.
https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/clusters/cluster_mce_overview#hosting-service-cluster-access-aws talks about another option of getting the kubeconfig file and kubeadmin credential directly from resources.

That is helpful to know. I can add some wording to better differentiate those docs.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
avishayt: Just 44

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
avishayt: I guess that can be a separate JIRA and PR

clusters/hosted_control_planes/hosted_bare_metal_access_cluster.adoc
lahinson: @rokej I combined the content from the two files here into `clusters/hosted_control_planes/managing_hosted_aws.adoc`. You can see the changes in the latest commit. I had a couple of questions, which start with `\\lahinson`. PTAL when you can. Thanks!

clusters/hosted_control_planes/managing_hosted_aws.adoc
```diff
@@ -124,12 +124,40 @@ oc get hostedclusters -n <hypershift-hosting-service-cluster>
...
 [#hosting-service-cluster-access-aws]
-== Accessing a hosting cluster on AWS
+== Accessing the hosted cluster
 
-The access secrets for hosted control plane clusters are stored in the `hypershift-management-cluster` namespace. Learn about the following secret name formats:
+You can access the hosted cluster by either getting the `kubeconfig` file and `kubeadmin` credential directly from resources or by using the `hypershift` CLI to generate a `kubeconfig` file.
 
-- `kubeconfig` secret: `<hostingNamespace>-<name>-admin-kubeconfig` (clusters-hypershift-demo-admin-kubeconfig)
-- `kubeadmin` password secret: `<hostingNamespace>-<name>-kubeadmin-password` (clusters-hypershift-demo-kubeadmin-password)
+* To access the hosted cluster by getting the `kubeconfig` file and credentials directly from resources, you need to be familiar with the access secrets for hosted control plane clusters. The secrets are stored in the `hypershift-management-cluster` namespace. Learn about the following secret name formats:
```
rokej: The secrets are stored in the `hypershift-management-cluster` namespace. ---> The secrets are stored in the hosted cluster (hosting) namespace. Note: The hosted cluster (hosting) namespace contains hosted cluster resources and the hosted control plane namespace is where the actual hosted control plane runs.

clusters/hosted_control_planes/managing_hosted_aws.adoc
```diff
@@ -124,12 +124,40 @@ oc get hostedclusters -n <hypershift-hosting-service-cluster>
...
+
+** `kubeconfig` secret: `<hostingNamespace>-<name>-admin-kubeconfig` (clusters-hypershift-demo-admin-kubeconfig)
+** `kubeadmin` password secret: `<hostingNamespace>-<name>-kubeadmin-password` (clusters-hypershift-demo-kubeadmin-password)
+
+//lahinson - June 2023 - Does the user need to take any steps with the secrets to access the hosted cluster? The latter section seems incomplete.
```
rokej: The secret contains  base64 encoded `kubeconfig` field. This can be base64 decoded and saved into a file to be used with `oc --kubeconfig .. ` command to get access to the hosted cluster. This is equivalent to the next section about generating the `kubeconfig` file using `hypershift create kubeconfig` command.

clusters/hosted_control_planes/managing_hosted_aws.adoc
```diff
@@ -124,12 +124,40 @@ oc get hostedclusters -n <hypershift-hosting-service-cluster>
...
+
+* To access the hosted cluster by using the `hypershift` CLI to generate the `kubeconfig` file, take the following steps:
+
+. Generate the `kubeconfig` file by entering the following command:
++
+----
+hypershift create kubeconfig --namespace ${CLUSTERS_NAMESPACE} --name ${HOSTED_CLUSTER_NAME} > ${HOSTED_CLUSTER_NAME}.kubeconfig
+----
++
+. If you access the cluster by entering the following command, the output shows that you do not have any nodes and that the ClusterVersion is trying to reconcile the {ocp} release:
```
rokej: Replace the whole sentence with...

Once the `kubeconfig` file is saved, you can access the hosted cluster by entering the following example command.

clusters/hosted_control_planes/managing_hosted_aws.adoc
```diff
@@ -124,12 +124,40 @@ oc get hostedclusters -n <hypershift-hosting-service-cluster>
...
++
+----
+oc --kubeconfig ${HOSTED_CLUSTER_NAME}.kubeconfig get clusterversion,nodes
```
rokej: oc --kubeconfig ${HOSTED_CLUSTER_NAME}.kubeconfig get nodes

clusters/hosted_control_planes/managing_hosted_aws.adoc
```diff
@@ -124,12 +124,40 @@ oc get hostedclusters -n <hypershift-hosting-service-cluster>
...
+----
```
rokej, line 171: Remove lines 150 ~ 160. This is not about accessing the cluster.

clusters/hosted_control_planes/managing_hosted_aws.adoc
```diff
@@ -124,12 +124,40 @@ oc get hostedclusters -n <hypershift-hosting-service-cluster>
 ----
 
 [#hosting-service-cluster-access-aws]
-== Accessing a hosting cluster on AWS
+== Accessing the hosted cluster
```
rokej, line 135: @lahinson This section should be on its own and referenced by managing_hosted_aws, managing_hosted_bm, managing_hosted_kubevirt (where is this doc? :) ). In each doc after "creating a cluster" section.

clusters/hosted_control_planes/managing_hosted_aws.adoc
lahinson: @rokej Thanks for the additional information! I'm learning a lot :)

Would it make sense if I added the following text after line 134?

```
The secret contains a Base64-encoded `kubeconfig` field. To access the hosted cluster, you can decode the field and save it into a file to be used with the following command:

----
oc --kubeconfig ${HOSTED_CLUSTER_NAME}.kubeconfig get nodes
----
```

rokej: @lahinson Yes, that sounds good. Sorry but we should say the same for the kubeadmin password that the secret's `data.password` field is base64-encoded and should be decoded.

lahinson: @rokej Does this sound correct?

```
The `kubeconfig` secret contains a Base64-encoded `kubeconfig` field, and the `kubeadmin` password secret contains a Base64-encoded `data.password` field. To access the hosted cluster, you can decode the fields and save them into a file to be used with the following command:

----
oc --kubeconfig ${HOSTED_CLUSTER_NAME}.kubeconfig get nodes
----
```

rokej: The `kubeconfig` can be saved into a file and used with `oc --kubeconfig ...` command. The decoded password can be used to log into the hosted cluster's console or API server. Thank you Laura.

clusters/hosted_control_planes/managing_hosted_aws.adoc
lahinson: @rokej Good catch.

For bare metal, I added references to this section from the following places:
- configure_hosted_bm
- hosted_create_bare_metal

The Kubevirt docs already contain a section about accessing the hosted cluster (https://github.com/stolostron/rhacm-docs/blob/2.8_stage/clusters/hosted_control_planes/managing_hosted_kubevirt.adoc#accessing-the-hosted-cluster). I can update the kubevirt docs to point to this section, but I think I'd prefer to do that in a separate PR.

clusters/hosted_control_planes/managing_hosted_aws.adoc
```diff
@@ -124,12 +124,48 @@ oc get hostedclusters -n <hypershift-hosting-service-cluster>
...
+----
+hypershift create kubeconfig --namespace ${CLUSTERS_NAMESPACE} --name ${HOSTED_CLUSTER_NAME} > ${HOSTED_CLUSTER_NAME}.kubeconfig
+----
++
+. After you save the `kubeconfig` file, you can access the hosted cluster by entering the following example command:
++
+----
+oc --kubeconfig ${HOSTED_CLUSTER_NAME}.kubeconfig get nodes
+----
++
```
rokej: We can remove lines 160 ~ 166 which answers your question in line 168. The example command above was to show user how to use the `kubeconfig` to run commands agains the hosted cluster, not for doing anything with the nodes.

lahinson:
@rokej Thanks again for your review! I just made the updates that we discussed.

rokej (approved):
Thank you @lahinson

lahinson:
@swopebe When you're back from PTO, I would appreciate your peer review on these changes. Thank you!

clusters/hosted_control_planes/configure_hosted_bm.adoc
```diff
@@ -1,21 +1,21 @@
 [#configuring-hosting-service-cluster-configure-bm]
 = Configuring the hosting cluster on bare metal (Technology Preview)
 
-To configure hosted control planes, you need a _hosting_ cluster and a _hosted_ cluster. You can enable a managed cluster to be a hosting cluster by using the `hypershift` add-on to deploy the HyperShift Operator on that cluster. Then, you can start to create the hosted cluster. 
+You can deploy hosted control planes by configuring a cluster to function as a hosting cluster. The hosting cluster is the {ocp-short} cluster where the control planes are hosted. The hosting cluster is also known as the _management_ cluster.
```
swopebe, line 5: This is gonna be so confusing because we have a "managed cluster" that the hub manages. I wonder if we can do this:

*Note:* The management cluster is not the same as _managed_ cluster. A managed cluster is a cluster that the hub cluster manages. 

or something...willing to brainstorm.

clusters/hosted_control_planes/configure_hosted_bm.adoc
```diff
@@ -1,21 +1,21 @@
 [#configuring-hosting-service-cluster-configure-bm]
 = Configuring the hosting cluster on bare metal (Technology Preview)
 
-To configure hosted control planes, you need a _hosting_ cluster and a _hosted_ cluster. You can enable a managed cluster to be a hosting cluster by using the `hypershift` add-on to deploy the HyperShift Operator on that cluster. Then, you can start to create the hosted cluster. 
+You can deploy hosted control planes by configuring a cluster to function as a hosting cluster. The hosting cluster is the {ocp-short} cluster where the control planes are hosted. The hosting cluster is also known as the _management_ cluster. 
 
-The {mce-short} 2.2 supports only the default `local-cluster` and the hub cluster as the hosting cluster.
+You can enable a managed cluster to be a hosting cluster by using the `hypershift` add-on to deploy the HyperShift Operator on that cluster. Then, you can start to create the hosted cluster.
```
swopebe, line 8: Here is where we switch to _managed_.

clusters/hosted_control_planes/configure_hosted_bm.adoc
```diff
@@ -1,21 +1,21 @@
...
 = Configuring the hosting cluster on bare metal (Technology Preview)
 
-To configure hosted control planes, you need a _hosting_ cluster and a _hosted_ cluster. You can enable a managed cluster to be a hosting cluster by using the `hypershift` add-on to deploy the HyperShift Operator on that cluster. Then, you can start to create the hosted cluster. 
+You can deploy hosted control planes by configuring a cluster to function as a hosting cluster. The hosting cluster is the {ocp-short} cluster where the control planes are hosted. The hosting cluster is also known as the _management_ cluster. 
 
-The {mce-short} 2.2 supports only the default `local-cluster` and the hub cluster as the hosting cluster.
+You can enable a managed cluster to be a hosting cluster by using the `hypershift` add-on to deploy the HyperShift Operator on that cluster. Then, you can start to create the hosted cluster. 
 
-Hosted control planes is a Technology Preview feature, so the related components are disabled by default.
+The {mce-short} 2.2 supports only the default `local-cluster` and the hub cluster as the hosting cluster.
```
swopebe, line 11: Does the user at this point know what the `local cluster` is, if not, maybe something I have added in a couple other spots: `local-cluster`, which is a hub cluster that is managed... 

Up to you if you think the user was introduced to that by now.

clusters/hosted_control_planes/configure_hosted_bm.adoc
```diff
@@ -37,39 +33,34 @@ You must have the following prerequisites to configure a hosting cluster:
...
-[#additional-resources-config-bm]
-== Additional resources
+* DNS: The API and Ingress endpoints must be routable.
 
-* For an introduction to the CIM service, see link:https://github.com/openshift/assisted-service/blob/master/docs/hive-integration/kube-api-getting-started.md[Kube API - Getting Started Guide]. 
+[#configuring-hcp-bare-metal]
+== Configuring hosted control planes on bare metal
 
-* See link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.13/html/installing/installing-on-any-platform#installation-load-balancing-user-infra_installing-platform-agnostic[Load balancing requirements for user-provisioned infrastructure] for more information about load balancers.
+After you meet the prerequisites, take the following steps to configure hosted control planes on bare metal:
```
swopebe: take I think would be better written as "complete," as I think take could be taken many ways in different languages

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+
+. Download the live ISO and use it to start a host (bare metal or VM). The URL for the live ISO can be found in the `InfraEnv` resource, in the `status.isoDownloadURL` field. At startup, the host communicates with the Assisted Service and registers as an agent in the same namespace as the `InfraEnv` resource.
+
+. To list the agents and some of their properties, enter the following command:
++
+----
+oc -n ${HOSTED_CONTROL_PLANE_NAMESPACE} get agents
+----
++
+.Example output
```
swopebe: I have been changing this when I see it to"

See the following example output: 

(if not too much trouble)

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+----
+NAME                                   CLUSTER   APPROVED   ROLE          STAGE
+86f7ac75-4fc4-4b36-8130-40fa12602218                        auto-assign
+e57a637f-745b-496e-971d-1abbf03341ba                        auto-assign
+----
+
+. After each agent is created, you can optionally set its `installation_disk_id` and `hostname` in the spec and approve the agent by entering the following commands: 
++
```
swopebe, line 29: we usually write specification, avoiding the colloquial developer speak.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+----
+oc -n ${HOSTED_CONTROL_PLANE_NAMESPACE} patch agent 86f7ac75-4fc4-4b36-8130-40fa12602218 -p '{"spec":{"installation_disk_id":"/dev/sda","approved":true,"hostname":"worker-0.example.krnl.es"}}' --type merge
+
+oc -n ${HOSTED_CONTROL_PLANE_NAMESPACE} patch agent 23d0c614-2caa-43f5-b7d3-0b3564688baa -p '{"spec":{"installation_disk_id":"/dev/sda","approved":true,"hostname":"worker-1.example.krnl.es"}}' --type merge
+----
+
```
swopebe, line 35: might be a format issue here, as the codeblock is usually grayed out. I will play with it a bit.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+. To verify that the agents are approved for use, enter the following command and check the output:
++
+----
+oc -n ${HOSTED_CONTROL_PLANE_NAMESPACE} get agents
+----
++
+.Example output
```
swopebe: same comment here on line 37

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+* To add agents by using Metal3, follow these instructions:
++
+**Important:** Because the `BaremetalHost` objects are created outside the baremetal-operator namespace, you must configure the Operator to watch all namespaces.
+
+. Enter the following command: 
++
+----
+oc patch provisioning provisioning-configuration --type merge -p '{"spec":{"watchAllNamespaces": true }}'
+----
++
```
swopebe, line 64: will look at the format here, as well.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+The `metal3` pod is restarted in the `openshift-machine-api` namespace.
+
+. Enter the following command, which continuously checks the status of the pod and returns the status when it is ready:
++
+----
+until oc wait -n openshift-machine-api $(oc get pods -n openshift-machine-api -l baremetal.openshift.io/cluster-baremetal-operator=metal3-state -o name) --for condition=containersready --timeout 10s >/dev/null 2>&1 ; do sleep 1 ; done
+----
```
swopebe, line 72: and here

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+export BMC_USERNAME=$(echo -n "root" | base64 -w0) <1>
+export BMC_PASSWORD=$(echo -n "calvin" | base64 -w0) <2>
+export BMC_IP="192.168.124.228" <3>
+export WORKER_NAME="ocp-worker-0" <4>
+export BOOT_MAC_ADDRESS="aa:bb:cc:dd:ee:ff" <5>
+export UUID="1" <6>
+export REDFISH_SCHEME="redfish-virtualmedia" <7>
+export REDFISH="${REDFISH_SCHEME}://${BMC_IP}/redfish/v1/Systems/${UUID}" <8>
+----
```
swopebe, line 86: and here

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+<5> The MAC address of the NIC that is connected to the machine network.
+<6> The Redfish UUID. This value is usually `1`. If you are using sushy-tools, this value is a long UUID. If you are using iDrac, this value is `System.Embedded.1`. You might need to check with the vendor.
+<7> The Redfish provider to use. If you are using hardware that uses a standard Redfish implementation, you can set this value to `redfish-virtualmedia`. If you use iDrac, this value is `idrac-virtualmedia`. If you use iLO5, this value is `ilo5-virtualmedia`. You might need to check with the vendor.
+<8> The Redfish connection endpoint.
+
+. Create the BareMetalHost object by following these steps:
+
+.. Create the BMC Secret:
+//lahinson -June 2023 - We need to replace this heredoc syntax with a substep to create the yaml file with the BMC secret info and another substep to enter the command to create the secret.
++
```
swopebe, line 101: thank you for line 87

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+  namespace: ${HOSTED_CONTROL_PLANE_NAMESPACE}
+type: Opaque
+----
+
+.. Create the BareMetalHost object:
++
+*Note:* The `infraenvs.agent-install.openshift.io` label is used to specify which `InfraEnv` is used to start the BareMetalHost. The `bmac.agent-install.openshift.io/hostname` label is used to manually set a hostname.
++
+If you want to manually specify the installation disk, you can use the `rootDeviceHints` in the BareMetalHost specification. If `rootDeviceHints` are not provided, the agent picks the installation disk that better suits the installation requirements. For more information about `rootDeviceHints`, see the _rootDeviceHints_ section of the BareMetalHost documentation.
+//lahinson -June 2023 - We need to replace this heredoc syntax with a substep to create the yaml file with the BareMetalHost object info and another substep to enter the command to create the object.
```
swopebe: this comment seems to be off for format

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,188 @@
...
+    bmac.agent-install.openshift.io/hostname: ${WORKER_NAME}
+spec:
+  automatedCleaningMode: disabled
+  bmc:
+    disableCertificateVerification: True
+    address: ${REDFISH}
+    credentialsName: ${WORKER_NAME}-bmc-secret
+  bootMACAddress: ${BOOT_MAC_ADDRESS}
+  online: true
+----
```
swopebe, line 142: as does this code

clusters/hosted_control_planes/hosted_bare_metal_dns.adoc
```diff
@@ -0,0 +1,17 @@
...
+See the following example DNS configuration:
+----
+api.example.krnl.es.    IN A 192.168.122.20
+api.example.krnl.es.    IN A 192.168.122.21
+api.example.krnl.es.    IN A 192.168.122.22
+api-int.example.krnl.es.    IN A 192.168.122.20
+api-int.example.krnl.es.    IN A 192.168.122.21
+api-int.example.krnl.es.    IN A 192.168.122.22
+`*`.apps.example.krnl.es. IN A 192.168.122.23
+----
```
swopebe, line 18: it's likely these pesky astericks

clusters/hosted_control_planes/hosted_bare_metal_infraenv.adoc
```diff
@@ -0,0 +1,36 @@
+[#hosted-control-planes-create-infraenv]
+= Creating an InfraEnv resource for hosted control planes on bare metal (Technology Preview)
+
+An `InfraEnv` is a environment where hosts that are starting the live ISO can join as agents. In this case, the agents are created in the same namespace as your hosted control plane.
+
+. Create an `InfraEnv` resource:
+//lahinson - June 2023 - To make this information easier for users to parse, we need to replace the heredoc syntax with instructions to (1) create the yaml file to define the InfraEnv resource and (2) enter the command to create the InfraEnv resource. Please check my revision for accuracy.
+
```
swopebe, line 7: comment seems visible, code blocks seem to not be grayed out further down.

clusters/hosted_control_planes/hosted_bare_metal_infraenv.adoc
```diff
@@ -0,0 +1,36 @@
...
++
+----
+oc apply -f infraenv-config.yaml
+----
+
+. To fetch the URL to download the live ISO that allows virtual machines or bare metal machines to join as agents, enter the following command:
++
+----
+oc -n ${HOSTED_CONTROL_PLANE_NAMESPACE} get InfraEnv ${HOSTED_CLUSTER_NAME} -ojsonpath="{.status.isoDownloadURL}"
+----
```
swopebe, line 38: here, too

swopebe (changes requested):

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,206 @@
...
+oc patch provisioning provisioning-configuration --type merge -p '{"spec":{"watchAllNamespaces": true }}'
+----
+
++
+The `metal3` pod is restarted in the `openshift-machine-api` namespace.
+
+. Enter the following command, which continuously checks the status of the pod and returns the status when it is ready:
+
++
+----
```
swopebe, line 70: until?

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
```diff
@@ -0,0 +1,220 @@
...
+NAME                                   CLUSTER   APPROVED   ROLE          STAGE
+86f7ac75-4fc4-4b36-8130-40fa12602218             true       auto-assign
+e57a637f-745b-496e-971d-1abbf03341ba             true       auto-assign
+----
+
+[#metal3-adding-agents]
+== Adding agents with Metal#
```
swopebe: 3

swopebe (changes requested):

clusters/hosted_control_planes/configure_hosted_bm.adoc
lahinson: I don't think it will hurt to add that for clarity. It's not obvious to me whether the user knows what the local cluster is at this point.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
lahinson: I agree that it's unusual to see a command start with `until`. I checked the source, though, and it starts with `until`, so I think it's supposed to be there.

clusters/hosted_control_planes/configure_hosted_bm.adoc
lahinson: I added a note to explain the difference between management cluster and managed cluster. I'll be sure to do the same in the intro of any other files where we refer to "management cluster". I don't think there are many in RHACM.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
swopebe: ok, I will let dev and you address it and approve knowing confirmation will come before merge.

swopebe (approved):
Everything looks great. I know this wasn't easy to implement in the portal, thank you so much for your high standards on the doc.

clusters/hosted_control_planes/hosted_bare_metal_agents.adoc
lahinson: Great. I'm just awaiting confirmation from the devs.

lahinson: @swopebe You were right to question that command! The devs just confirmed that it's wrong and sent me an updated version. I just pushed the update.

PR #5213, opened 2023-06-21
https://issues.redhat.com/browse/ACM-5792 Policy automation known issue

dockerymick:
https://issues.redhat.com/browse/ACM-5792

### DIFF
--- a/release_notes/known_issues_governance.adoc
+++ b/release_notes/known_issues_governance.adoc
@@ -177,4 +177,20 @@ When there are templating errors, such as incorrect syntax in a configuration po
 == Duplicate Ansible jobs are created for policy automations  
 //2.7+2.8:ACM-5644
 
-If you have a `PolicyAutomation` that is set to _Run once_ mode and disabled an extra Ansible job is created. 
+If you have a `PolicyAutomation` that is set to _Run once_ mode and disabled an extra Ansible job is created. You can delete the extra Ansible job. Complete the following steps:
+
+. Run the following command to view the Ansible job list:
++
+[source,bash]
+----
+oc get ansiblejob -n {namespace}
+----
+
+. Delete the duplicate Ansible job by using the following command:
++
+[source,bash]
+----
+oc delete ansiblejob {ansiblejob name} -n {namespace}
+----
+
+The duplicate Ansible job is deleted. 
--- a/release_notes/known_issues_governance.adoc
+++ b/release_notes/known_issues_governance.adoc
@@ -192,5 +192,3 @@ oc get ansiblejob -n {namespace}
 ----
 oc delete ansiblejob {ansiblejob name} -n {namespace}
 ----
-
-The duplicate Ansible job is deleted. 
--- a/release_notes/known_issues_governance.adoc
+++ b/release_notes/known_issues_governance.adoc
@@ -177,7 +177,7 @@ When there are templating errors, such as incorrect syntax in a configuration po
 == Duplicate Ansible jobs are created for policy automations  
 //2.7+2.8:ACM-5644
 
-If you have a `PolicyAutomation` that is set to _Run once_ mode and disabled an extra Ansible job is created. You can delete the extra Ansible job. Complete the following steps:
+If you have a `PolicyAutomation` that is set to _Run once_ mode and disabled, an extra Ansible job is created. You can delete the extra Ansible job. Complete the following steps:
 
 . Run the following command to view the Ansible job list:
 +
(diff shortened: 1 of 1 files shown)


release_notes/known_issues_governance.adoc
```diff
@@ -172,3 +172,9 @@ oc delete lease governance-policy-addon-controller-lock -n <namespace>
 
 When there are templating errors, such as incorrect syntax in a configuration policy, objects are deleted. Recreate your deleted object with the correct syntax. 
 
+
+[#duplicate-ansible-jobs]
+== Duplicate Ansible jobs are created for policy automations  
+//2.7+2.8:ACM-5644
+
+If you have a `PolicyAutomation` that is set to _Run once_ mode and disabled an extra Ansible job is created.
```
cadawson: I wasn't sure if `disabled` is a setting, but looked at the issue. Suggestion: If you have a `PolicyAutomation` that is set to _Run once_ mode and the `disabled` option is selected, an extra Ansible job is created. 

What is the result of this? Does the second Ansible job run? Do you have to remove it? Maybe it doesn't matter, and it is just for information.

cadawson (changes requested):
Suggestion for clarification and possibly a next step (if necessary). Thanks!

release_notes/known_issues_governance.adoc
dockerymick: @cadawson great question! the disabled parameter is only within the policy, there isn't a option from the console. The user would have the `disabled` parameter set to `true` to disable the policy 

@yiraeChristineKim does the user need to remove the extra Ansible job? are they able to remove the extra

yiraeChristineKim: you don't have to remove it because it creates the same 2 ansible jobs. However "Run once" mode means it is supposed to make 1 ansibleJob. From a user's perspective, it makes the user confused.

dockerymick: Thanks for clarifying that it is not required to remove the Ansible job. If the user wants to remove it, how can they do that?

yiraeChristineKim: In addition Ansible plays “should” be idempotent, in other words, it is run to achieve a particular state. Let me double check with that

yiraeChristineKim: Surely the user is able to delete using CLI, in console, I tried but I don't have credential so I cannot create this. Sorry about this. I will ping Derek, let you know after I get an answer

yiraeChristineKim: In the console, the user cannot delete

dockerymick: Thanks! Can you share the command that the user should run to delete the extra Ansible job?

yiraeChristineKim: sure!  To get the ansiblejob list `kubectl(or oc) get ansiblejob -n {namespace}` -> 
`kubectl(or oc) delete ansiblejob {ansiblejob name} -n {namespace}`

release_notes/known_issues_governance.adoc
```diff
@@ -172,3 +172,25 @@ oc delete lease governance-policy-addon-controller-lock -n <namespace>
 
 When there are templating errors, such as incorrect syntax in a configuration policy, objects are deleted. Recreate your deleted object with the correct syntax. 
 
+
+[#duplicate-ansible-jobs]
+== Duplicate Ansible jobs are created for policy automations  
+//2.7+2.8:ACM-5644
+
+If you have a `PolicyAutomation` that is set to _Run once_ mode and disabled an extra Ansible job is created. You can delete the extra Ansible job. Complete the following steps:
```
yiraeChristineKim: Do we need a comma between "disabled" and "an"?

yiraeChristineKim (approved):

release_notes/known_issues_governance.adoc
```diff
@@ -172,3 +172,25 @@ oc delete lease governance-policy-addon-controller-lock -n <namespace>
...
+----
+
+. Delete the duplicate Ansible job by using the following command:
++
+[source,bash]
+----
+oc delete ansiblejob {ansiblejob name} -n {namespace}
+----
+
+The duplicate Ansible job is deleted.
```
swopebe: the last line I would avoid. Remember we discussed as a team to only put validation if there is a way for the user to validate.

_Run the following command to see the job is removed from the list._

Without a verification step, we cannot predict that the user may not continue to run into problems for some other reason. So I would consider just removing the line 196.

release_notes/known_issues_governance.adoc
dockerymick: Updated

swopebe (approved):
see comment for consideration

release_notes/known_issues_governance.adoc
dockerymick: Great point Brandi! thanks for the reminder

PR #5215, opened 2023-06-21
Swope sizes ha

swopebe:


install/cluster_size.adoc
```diff
@@ -131,18 +133,21 @@ A {product-title} hub cluster can be installed and supported on {ocp-short} bare
...
 | 3
 | 289 GB
-| 110 GB
+| 64 GB - 110 GB
 | 90 
 | 44 
-|===
-
-*Note:* The scale results reflect Red Hat Performance and Scale lab results of 4 January 2023.
-
```
swopebe, line 148: not sure what writer added this but it's not a good idea to add this because we have to go back and update dates. It's best to operate on releases, like we do for the rest of the doc. We can say results are based on a certain release?

swopebe: The simplest thing is to not put it in the doc, but simpler than dates is the version.

install/cluster_size.adoc
```diff
@@ -131,18 +133,21 @@ A {product-title} hub cluster can be installed and supported on {ocp-short} bare
 [#single-node]
 === Creating and managing single node {ocp-short} clusters
 
+View https://access.redhat.com/documentation/en-us/openshift_container_platform/4.13/html/installing/installing-on-a-single-node[Installing on a single node] to learn about the requirements. Since each cluster is unique, the following guidelines provide only sample deployment requirements that are classified by size and purpose.
```
swopebe, line 136: @bweidenb

install/cluster_size.adoc
swopebe: Slack approval: we should be fine them, thanks

PR #5218, opened 2023-06-22
https://issues.redhat.com/browse/ACM-5544 Ascii path fix test 1

oafischer:
Fixing the build issues we had with images by using 2 main files. This is only for DR4Hub for now. If it looks good, let's merge and test it in the build, preview and potentially portal before switching over the rest of the docs.


dockerymick (approved):
Thanks for making this update! I ran the link check tool and the links look good. Approving

PR #5252, opened 2023-06-29
https://issues.redhat.com/browse/ACM-6119

faaden123:
Will probably need to replace note with yaml callout -- as well reaching out to see if the note is user-friendly.

### DIFF
--- a/networking/submariner/install_subctl.adoc
+++ b/networking/submariner/install_subctl.adoc
@@ -42,5 +42,6 @@ After adding the utility to your path, view the following table for a brief desc
 | link:https://submariner.io/operations/deployment/subctl/#gather[`gather`] | Collects information from the clusters to help troubleshoot a Submariner deployment.
 | link:https://submariner.io/operations/deployment/subctl/#version[`version`] | Displays the version details of the `subctl` binary tool.
 |===
+*Note* The commands listed above can only support downstream subctl. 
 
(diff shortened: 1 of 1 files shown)


networking/submariner/install_subctl.adoc
```diff
@@ -42,5 +42,6 @@ After adding the utility to your path, view the following table for a brief desc
 | link:https://submariner.io/operations/deployment/subctl/#gather[`gather`] | Collects information from the clusters to help troubleshoot a Submariner deployment.
 | link:https://submariner.io/operations/deployment/subctl/#version[`version`] | Displays the version details of the `subctl` binary tool.
 |===
+*Note*: The commands listed above can only support `subctl`.
```
maayanf24:
```suggestion
*Note*: Only the commands listed above are support in the shipped `subctl`. 
```
I'm not sure if we should use here the word `downstream` to make clean that it is not the same as the upstream `subctl`.

maayanf24: Replace L45 with L47.

faaden123: Ah- yes discussed with Brandi and noted that downstream is not necessary/ user-friendly.

dfarrell07:
```suggestion
*Note*: Red Hat's build of `subctl` only includes commands relevant to ACM deployments.
```

I think something like this might be more clear/accurate.

faaden123:
This isssue applies to 2.9. Will be closing this issue as it does not reflect the changes to the 2.9 stage

PR #5266, opened 2023-07-05
https://issues.redhat.com/browse/ACM-6234--version changes

swopebe:


install/uninstall.adoc
```diff
@@ -125,13 +125,13 @@ Replace `<namespace>` in the script with the name of the namespace where {produc
 *Note:* If you plan to reinstall the same {product-title-short} version, you can skip the next steps in this procedure and reinstall the custom resource. Proceed for a complete operator uninstall.
 
 +
-. Enter the following commands to delete the {product-title-short} `ClusterServiceVersion` and `Subscription` in the namespace it is installed in:
+. Enter the following commands to delete the {product-title-short} `ClusterServiceVersion` and `Subscription` in the namespace where it is installed. Replace the `2.x.0` value with the current major or minor release:
```
swopebe, line 129: Just not great to end with preposition; seems left undone and can always create a better sentence.

install/uninstall.adoc
```diff
@@ -125,13 +125,13 @@ Replace `<namespace>` in the script with the name of the namespace where {produc
...
 ----
 ❯ oc get csv
 NAME                                 DISPLAY                                      VERSION   REPLACES   PHASE
-advanced-cluster-management.v2.4.0   Advanced Cluster Management for Kubernetes   2.4.0                Succeeded
+advanced-cluster-management.v2.x.0   Advanced Cluster Management for Kubernetes   2.x.0                Succeeded
 
-❯ oc delete clusterserviceversion advanced-cluster-management.v2.4.0
+❯ oc delete clusterserviceversion advanced-cluster-management.v2.x.0
```
swopebe, line 136: What are these in here for `>` (I didn't write this topic, wondering if it's not needed as we don't do this in other code blocks, we make the command and output cleaner)

install/uninstall.adoc
```diff
@@ -97,6 +97,7 @@ oc get mch -o yaml
 . Remove any potential remaining artifacts by running the clean-up script. Run this clean-up script if you plan to reinstall with an older version of {product-title-short} on the same cluster. 
 
 .. Install the Helm CLI binary version 3.2.0, or later, by following the instructions at link:https://helm.sh/docs/intro/install/[Installing Helm].
+//outdated much???
```
swopebe, line 100: @JakobGray what is this now?

install/uninstall.adoc
```diff
@@ -158,7 +159,7 @@ When you use the {ocp} console to uninstall, you remove the operator. Complete t
 
 . Run the clean-up script according to the procedure in xref:../install/uninstall.adoc#removing-a-multiclusterhub-instance-by-using-commands[Removing a MultiClusterHub instance by using commands].
 +
-*Tip:* If you plan to reinstall the same {product-title-short} version, you can skip the rest of the steps in this procedure and reinstall the custom resource.
+*Note:* If you plan to reinstall the same {product-title-short} version, you can skip the rest of the steps in this procedure and reinstall the custom resource.
```
swopebe, line 163: we agreed to abandon "tip" as only one writer was using it and we had too many options. Can also be "important" here?

install/uninstall.adoc
JakobGray: No need to update the Helm binary version. The uninstall list as a whole is worth revisiting because most of the resources are well out of the release cycle.

install/uninstall.adoc
dockerymick: I think it is used to resemble the terminal. I believe it can be removed, but will let development confirm

dockerymick (approved):
Responded to one of your comments. Great questions. Your changes look good to me!

JakobGray (approved):

PR #5271, opened 2023-07-06
Update special_considerations.md simplify asterisks

swopebe:
With internal doc/process being shared across the team, we want to ensure that we don't document what writers should just know (like what is code and what is not) because most things are in style manuals or taught in collegiate programs (like grammar). We all have a chance to update this topic based on where people are running into issues, but in the simplest way possible. I didn't understand a few lines here, though I knew what we were trying to say. 

Asterisks are either part of the code if not bulleted lists or bold tags, and should always be treated as such (I feel this is just standard), but I added a recent example that wasn't really clear before how to fix lists, and how to avoid formatting issues in DNS examples.


swopebe:
Recent example: https://github.com/stolostron/rhacm-docs/pull/5258/files

cadawson (approved):
Thanks for adding the example.

PR #5274, opened 2023-07-07
https://issues.redhat.com/browse/ACM-4940 Errata update

oafischer:
This one was missed by the release manager so I offered a one time exception to add it late. For the future I have let the issue reporter know to contact the release manager directly if they notice a missing errata label on an engineering issue.


cadawson (approved):
Looks good. Thanks!

PR #5280, opened 2023-07-12
https://issues.redhat.com/browse/ACM-6287 PVC Support for observability

dockerymick:
https://issues.redhat.com/browse/ACM-6287

### DIFF
--- a/observability/observe_environments.adoc
+++ b/observability/observe_environments.adoc
@@ -287,7 +287,7 @@ Observability components require 2701mCPU and 11972Mi memory to install the obse
 [#persistent-stores-observability]
 == Persistent stores used in the observability service
 
-*Important:* Do not use the local storage operator or a storage class that uses local volumes for persistent storage. You can lose data if the pod relaunched on a different node after a restart. When this happens, the pod can no longer access the local storage on the node. Be sure that you can access the persistent volumes of the `receive` and `rules` pods to avoid data loss. When a pod is restarted, Kubernetes automatically attaches previously created volumes to the new pods.
+*Important:* Do not use the local storage operator or a storage class that uses local volumes for persistent storage. You can lose data if the pod relaunched on a different node after a restart. When this happens, the pod can no longer access the local storage on the node. Be sure that you can access the persistent volumes of the `receive` and `rules` pods to avoid data loss.
 
 When you install {product-title-short} the following persistent volumes (PV) must be created so that Persistent Volume Claims (PVC) can attach to it automatically. As a reminder, you must define a storage class in the `MultiClusterObservability` custom resource when there is no default storage class specified or you want to use a non-default storage class to host the PVs. It is recommended to use Block Storage, similar to what Prometheus uses. Also each replica of `alertmanager`, `thanos-compactor`, `thanos-ruler`, `thanos-receive-default` and `thanos-store-shard` must have its own PV. View the following table:
 
--- a/observability/observe_environments.adoc
+++ b/observability/observe_environments.adoc
@@ -290,7 +290,7 @@ Observability components require 2701mCPU and 11972Mi memory to install the obse
 *Important:* Do not use the local storage operator for persistent storage. You can lose data if the pod relaunched on a different node after a restart. When this happens, the pod can no longer access the local storage on the node. Be sure that you can access the persistent volumes of the receive and rules pods to avoid data loss.
 //should we point the user to backup and restore topic? 
 
-When you install {product-title-short} the following persistent volumes (PV) must be created so that Persistent Volume Claims (PVC) can attach to it automatically. As a reminder, you must define a storage class in the `MultiClusterObservability` custom resource when there is no default storage class specified or you want to use a non-default storage class to host the PVs. It is recommended to use Block Storage, similare to what Prometheus uses. Also each replica of `alertmanager`, `thanos-compactor`, `thanos-ruler`, `thanos-receive-default` and `thanos-store-shard` must have its own PV. View the following table:
+When you install {product-title-short} the following persistent volumes (PV) must be created so that Persistent Volume Claims (PVC) can attach to it automatically. As a reminder, you must define a storage class in the `MultiClusterObservability` custom resource when there is no default storage class specified or you want to use a non-default storage class to host the PVs. It is recommended to use Block Storage, similar to what Prometheus uses. Also each replica of `alertmanager`, `thanos-compactor`, `thanos-ruler`, `thanos-receive-default` and `thanos-store-shard` must have its own PV. View the following table:
 
 .Table list of persistent volumes
 |===
--- a/observability/observe_environments.adoc
+++ b/observability/observe_environments.adoc
@@ -287,7 +287,7 @@ Observability components require 2701mCPU and 11972Mi memory to install the obse
 [#persistent-stores-observability]
 == Persistent stores used in the observability service
 
-*Important:* Do not use the local storage operator for persistent storage. You can lose data if the pod relaunched on a different node after a restart. When this happens, the pod can no longer access the local storage on the node. Be sure that you can access the persistent volumes of the receive and rules pods to avoid data loss.
+*Important:* Do not use the local storage operator for persistent storage. You can lose data if the pod relaunched on a different node after a restart. When this happens, the pod can no longer access the local storage on the node. Be sure that you can access the persistent volumes of the `receive` and `rules` pods to avoid data loss.
 //should we point the user to backup and restore topic? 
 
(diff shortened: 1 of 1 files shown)


observability/observe_environments.adoc
```diff
@@ -287,6 +287,9 @@ Observability components require 2701mCPU and 11972Mi memory to install the obse
 [#persistent-stores-observability]
 == Persistent stores used in the observability service
 
+**Important:** Avoid using the local storage operator for persistence. this can lead to data loss situations. The local storage on a node is not accessible by a pod if it is ever relaunched on a different node on a restart. The persistent volumes must be recoverable, especially for receive pods and rules pods if you want to avoid data loss.
```
dockerymick:
```suggestion
**Important:** Avoid using the local storage operator for persistence. This can lead to data loss. The local storage on a node is not accessible by a pod if it is ever relaunched on a different node during restart. The persistent volumes for the `receive` and `rules` pods must be recoverable to avoid data loss.
```

observability/observe_environments.adoc
```diff
@@ -287,6 +287,9 @@ Observability components require 2701mCPU and 11972Mi memory to install the obse
 [#persistent-stores-observability]
 == Persistent stores used in the observability service
 
+**Important:** Avoid using the local storage operator for persistence. This can lead to data loss. The local storage on a node is not accessible by a pod if it is ever relaunched on a different node during restart. The persistent volumes for the `receive` and `rules` pods must be recoverable to avoid data loss.
```
dockerymick: should we point the user to backup and restore topic? Would it be accurate to replace "recoverable" with "restored"? It seems like we are saying, "Be sure that you can access the persistent volumes of the `receive` and `rules` pods to avoid data loss."

How does the user ensure that it's accessible? Do they have to add it to a specific namespace? 

The persistent volumes for the `receive` and `rules` pods must be restored to avoid data loss.

The persistent volumes for the `receive` and `rules` pods must be accessible to avoid data loss.

dockerymick: I'm in favor of this suggestion

```suggestion
**Important:** Do not use the local storage operator for persistent storage. You can lose data if the pod relaunched on a different node after a restart. When this happens, the pod can no longer access the local storage on the node. Be sure that you can access the persistent volumes of the receive and rules pods to avoid data loss. 
```

cadawson: This looks good. A couple of thoughts and suggestions: 

Is this a significant enough issue to say: "Do not use the local storage operator for persistent storage."

Alternate approach: Important: Do not use the local storage operator for persistent storage. You can lose data if the pod is relaunched on a different node after a restart. When this happens, the pod can no longer access the local storage on the node. Be sure that you can access the persistent volumes of the receive and rules pods to avoid data loss.

Or a little softer: Important: You can lose date if you use the local storage operator for persistent storage. If the pod is relaunched on a different node after a restart, the local storage on a node is no longer accessible by the pod. Be sure that you can access the persistent volumes of the receive and rules pods to avoid data loss.

How are the receive and rules pods related to this?

cadawson (approved):
A couple of thoughts, but looks good. Thanks!

observability/observe_environments.adoc
dockerymick: @cadawson thanks for your review and suggestions. Great questions! I really like your Alternate approach so I updated my suggestion. It seems like we should specify the `receiver` and `rules` pods. From my perspective, it seems like the pods we should specify are `thanos-rule` and `thanos-receive-default`

@berenss are those the pods you were referring to?

observability/observe_environments.adoc
```diff
@@ -287,6 +287,9 @@ Observability components require 2701mCPU and 11972Mi memory to install the obse
 [#persistent-stores-observability]
 == Persistent stores used in the observability service
 
+*Important:* Avoid using the local storage operator for persistence. This can lead to data loss. The local storage on a node is not accessible by a pod if it is ever relaunched on a different node during restart. The persistent volumes for the `receive` and `rules` pods must be recoverable to avoid data loss.
```
dockerymick:
```suggestion
*Important:* Do not use the local storage operator for persistent storage. You can lose data if the pod relaunched on a different node after a restart. When this happens, the pod can no longer access the local storage on the node. Be sure that you can access the persistent volumes of the receive and rules pods to avoid data loss.
```

dockerymick: How are the receive and rules pods related to this? ([From Chris](https://github.com/stolostron/rhacm-docs/pull/5280#discussion_r1262917272))

@cadawson thanks for your review and suggestions. Great questions! I really like your Alternate approach so I updated my suggestion. It seems like we should specify the receiver and rules pods. From my perspective, it seems like the pods we should specify are thanos-rule and thanos-receive-default

@berenss are those the pods you were referring to?

observability/observe_environments.adoc
```diff
@@ -287,6 +287,9 @@ Observability components require 2701mCPU and 11972Mi memory to install the obse
...
+//should we point the user to backup and restore topic? 
+
 When you install {product-title-short} the following persistent volumes (PV) must be created so that Persistent Volume Claims (PVC) can attach to it automatically. As a reminder, you must define a storage class in the `MultiClusterObservability` custom resource when there is no default storage class specified or you want to use a non-default storage class to host the PVs. It is recommended to use Block Storage, similare to what Prometheus uses. Also each replica of `alertmanager`, `thanos-compactor`, `thanos-ruler`, `thanos-receive-default` and `thanos-store-shard` must have its own PV. View the following table:
```
berenss: spelling on the word "similar"

observability/observe_environments.adoc
berenss: Mickela I think we've crossed paths on 2 issues. I noticed this PR has FIPS in the title and gets linked to ACM-6285, but the actual PR content deals with Obs PVs and should be for ACM-6287

berenss (changes requested):

observability/observe_environments.adoc
dockerymick: big time lmao, thanks Scott. Let me update the pr

observability/observe_environments.adoc
```diff
@@ -287,7 +287,10 @@ Observability components require 2701mCPU and 11972Mi memory to install the obse
 [#persistent-stores-observability]
 == Persistent stores used in the observability service
 
-When you install {product-title-short} the following persistent volumes (PV) must be created so that Persistent Volume Claims (PVC) can attach to it automatically. As a reminder, you must define a storage class in the `MultiClusterObservability` custom resource when there is no default storage class specified or you want to use a non-default storage class to host the PVs. It is recommended to use Block Storage, similare to what Prometheus uses. Also each replica of `alertmanager`, `thanos-compactor`, `thanos-ruler`, `thanos-receive-default` and `thanos-store-shard` must have its own PV. View the following table:
+*Important:* Do not use the local storage operator for persistent storage. You can lose data if the pod relaunched on a different node after a restart. When this happens, the pod can no longer access the local storage on the node. Be sure that you can access the persistent volumes of the receive and rules pods to avoid data loss.
```
dockerymick:
```suggestion
*Important:* Do not use the local storage operator for persistent storage. You can lose data if the pod relaunched on a different node after a restart. When this happens, the pod can no longer access the local storage on the node. Be sure that you can access the persistent volumes of the `receive` and `rules` pods to avoid data loss.
```

dockerymick:
@subbarao-meduri when you have an opportunity, will you review the PR?

observability/observe_environments.adoc
subbarao-meduri: @dockerymick 
1. This is not related to backup and restore.  
2. A more accurate way of saying this would be `avoid using local volumes or a storage class that uses local volumes.` 

See [Persistent storage using local volumes](https://docs.openshift.com/container-platform/4.13/storage/persistent_storage/persistent_storage_local/persistent-storage-local.html) for full explanation. 

** not for customer **
`receive` and `rules` pods are stateful sets, and if the pods die, and are restored on a different node, they must be attached to the same storage volumes as before, or else data loss occurs. The problem with local storage volumes is that they cannot be accessed across nodes, and they only exist on the nodes they are created on.

The restoration process itself is automatic in Kubernetes. When a pod is restarted, Kubernetes automatically attaches previously created volume to the new pods (no user action is needed). It cannot do this if the persistent volumes are local volumes. 

Hope this helps.

subbarao-meduri: @dockerymick I like this wording as well. 

Yes, there are 5 stateful sets that use local volumes. But `thnaos-rule` and `thanos-receive-default` are critical, and if they do not have access to the same volume after a pod restart, data loss could occur.

subbarao-meduri (commented):
/lgtm

dockerymick:
Received necessary approvals. Merging this PR

PR #5287, opened 2023-07-17
https://issues.redhat.com/browse/ACM-6279 Hive limitation

oafischer:
Applies to 2.7 and 2.6 as well from what I understand.

Not sure if Cluster is the best place for this, open to suggestions if there is a better place in the docs to add this limitation.

### DIFF
--- a/clusters/release_notes/known_issues.adoc
+++ b/clusters/release_notes/known_issues.adoc
@@ -424,3 +424,9 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
 //2.7:ACM-2664
 
 When you use the console to add a new node pool to an existing hosted cluster, the same version of {ocp-short} might appear more than once in the list of options. You can select any instance in the list for the version that you want. 
+
+[#custom-ingress-domain-limitation]
+=== Custom _ingress_ domain is not applied correctly
+//2.8:ACM-6279
+
+You can specify a custom `ingress` domain by using the `ClusterDeployment` resource while installing a managed cluster, but the change is only applied after the installation by using `SyncSet`. As a result, the `spec` field in the `clusterdeployment.yaml` file uses the the custom `ingress` domain you specified, but the `status` still uses the default domain.
--- a/clusters/release_notes/known_issues.adoc
+++ b/clusters/release_notes/known_issues.adoc
@@ -426,7 +426,7 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
 When you use the console to add a new node pool to an existing hosted cluster, the same version of {ocp-short} might appear more than once in the list of options. You can select any instance in the list for the version that you want. 
 
 [#custom-ingress-domain-limitation]
-=== Custom _ingress_ domain is not applied correctly
+=== Custom ingress domain is not applied correctly
 //2.8:ACM-6279
 
-You can specify a custom `ingress` domain by using the `ClusterDeployment` resource while installing a managed cluster, but the change is only applied after the installation by using `SyncSet`. As a result, the `spec` field in the `clusterdeployment.yaml` file uses the custom `ingress` domain you specified, but the `status` still uses the default domain.
+You can specify a custom ingress domain by using the `ClusterDeployment` resource while installing a managed cluster, but the change is only applied after the installation by using `SyncSet`. As a result, the `spec` field in the `clusterdeployment.yaml` file uses the custom ingress domain you specified, but the `status` still uses the default domain.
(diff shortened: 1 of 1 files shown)


oafischer:
Approval: https://redhat-internal.slack.com/archives/CE3ETN3J8/p1689604189437999?thread_ts=1688562352.347059&cid=CE3ETN3J8

clusters/release_notes/known_issues.adoc
```diff
@@ -424,3 +424,9 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
 //2.7:ACM-2664
 
 When you use the console to add a new node pool to an existing hosted cluster, the same version of {ocp-short} might appear more than once in the list of options. You can select any instance in the list for the version that you want. 
+
+[#custom-ingress-domain-limitation]
+=== Custom ingress domain is not applied correctly
+//2.8:ACM-6279
+
+You can specify a custom ingress domain by using the `ClusterDeployment` resource while installing a managed cluster, but the change is only applied after the installation by using `SyncSet`. As a result, the `spec` field in the `clusterdeployment.yaml` file uses the custom ingress domain you specified, but the `status` still uses the default domain.
```
2uasimojo: - Does this need to say _where_ in the ClusterDeployment custom ingress domains need to be specified? Or refer to some other part of the documentation that does so?
- What does it mean to say "the `spec` field ... uses the custom ingress domain" or "the `status` still uses the default domain"? Do we mean the value *displayed* in those fields, or are we implying some kind of use by hive itself?

clusters/release_notes/known_issues.adoc
```diff
@@ -424,3 +424,9 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
 //2.7:ACM-2664
 
 When you use the console to add a new node pool to an existing hosted cluster, the same version of {ocp-short} might appear more than once in the list of options. You can select any instance in the list for the version that you want. 
+
+[#custom-ingress-domain-limitation]
+=== Custom ingress domain is not applied correctly
```
2uasimojo, line 429: Is this being considered a bug? What would "correctly" look like?

clusters/release_notes/known_issues.adoc
```diff
@@ -424,3 +424,9 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
...
+//2.8:ACM-6279
```
2uasimojo, line 430: If it matters, this code has been in place for a long time and thus would affect earlier versions of ACM. Probably all versions currently in support. Let me know if you need to nail that down more precisely.

clusters/release_notes/known_issues.adoc
oafischer: Updated the wording based on your feedback, thanks. For the first part of the question, this is just a known issues and limitations section so we would only describe the actual issue/limitation.

clusters/release_notes/known_issues.adoc
oafischer: It would be a known issue/limitation. I assume the correct behavior would be the custom domain appearing in the `spec` field of the yaml file? Would be good if a developer could confirm that.

2uasimojo: Assuming the default URL in `status` still works (which we're still trying to [determine](https://redhat-internal.slack.com/archives/CE3ETN3J8/p1689691616614389?thread_ts=1688562352.347059&cid=CE3ETN3J8)) I would say "not applied correctly" is a bit strong. Perhaps "Custom ingress domain not reflected in status" would be more appropriate?

oafischer: "Custom ingress domain does not appear in status" would work (avoid passive voice as per IBM style), or alternatively "is partially applied" but I think the former is good. I'll wait and follow the developments on the details of whether it works or not, thanks for bringing that up.

cdoan1 (approved):
LGTM

clusters/release_notes/known_issues.adoc
```diff
@@ -424,3 +424,9 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
...
+
+You can specify a custom ingress domain by using the `ClusterDeployment` resource while installing a managed cluster, but the change is only applied after the installation by using `SyncSet`. As a result, the `spec` field in the `clusterdeployment.yaml` file displays the custom ingress domain you specified, but the `status` still displays the default domain.
```
dockerymick:
```suggestion
You can specify a custom ingress domain by using the `ClusterDeployment` resource while installing a managed cluster, but the change is only applied after the installation by using `SyncSet` resource. As a result, the `spec` field in the `clusterdeployment.yaml` file displays the custom ingress domain you specified, but the `status` still displays the default domain.
```

dockerymick (approved):
left a suggestion, otherwise this looks good!

PR #5297, opened 2023-07-20
https://issues.redhat.com/browse/ACM-6285 Updating FIPS for 2.8

dockerymick:
https://issues.redhat.com/browse/ACM-6285

### DIFF
--- a/release_notes/fips_readiness.adoc
+++ b/release_notes/fips_readiness.adoc
@@ -3,6 +3,9 @@
 
 FIPS readiness has been completed for {product-title}. {product-title-short} uses the same tools to make sure cryptography calls are passed to the Red Hat Enterprise Linux (RHEL) certified cryptographic modules that are used by {ocp}. 
 
+You must deploy your managed clusters from a FIPS-ready hub cluster. You must set the `fips:` parameter to `true` in your {ocp-short} managed cluster. 
+//is the created part important? 
+
 [#fips-limitations]
 == Limitations 
 
@@ -17,3 +20,5 @@ Read the following limitations with {product-title-short} and FIPS.
 ----
 FIPS with information text: Use the Federal Information Processing Standards (FIPS) modules provided with Red Hat Enterprise Linux CoreOS instead of the default Kubernetes cryptography suite file before you deploy the new managed cluster.
 ----
+
+
--- a/release_notes/fips_readiness.adoc
+++ b/release_notes/fips_readiness.adoc
@@ -3,7 +3,7 @@
 
 FIPS readiness has been completed for {product-title}. {product-title-short} uses the same tools to make sure cryptography calls are passed to the Red Hat Enterprise Linux (RHEL) certified cryptographic modules that are used by {ocp}. 
 
-If you plan to manage clusters with FIPS enabled, you must have a FIPS-ready hub cluster because cryptography created on the hub cluster is stored on the managed cluster. Set the following parameter in your {ocp-short} managed cluster, `fips:true`.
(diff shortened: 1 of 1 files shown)


release_notes/fips_readiness.adoc
```diff
@@ -3,6 +3,9 @@
 
 FIPS readiness has been completed for {product-title}. {product-title-short} uses the same tools to make sure cryptography calls are passed to the Red Hat Enterprise Linux (RHEL) certified cryptographic modules that are used by {ocp}. 
 
+You must deploy your managed clusters from a FIPS-ready hub cluster. You must set the `fips:` parameter to `true` in your {ocp-short} managed cluster.
```
berenss: You must deploy your managed clusters from a FIPS-ready hub cluster. You must set the `fips:` parameter to `true` in your {ocp-short} managed cluster with all necessary FIPS configuration and remediations performed.

gparvin: I am restating this some to make the scenario more clear to me.
```suggestion
If you will manage clusters with FIPS enabled, you must have a FIPS-ready hub cluster because cryptography created on the hub is stored on managed cluster. 
```

dockerymick: Thank you Gus for your suggestion! I like that you added the reason for FIPS enablement. Based on your suggestion and the style guide (removed the word "will" to maintain active tone), I suggest the following statement:

 "If you plan to manage clusters with FIPS enabled, you must have a FIPS-ready hub cluster because cryptography created on the hub cluster is stored on the managed cluster. Set the following parameter in your {ocp-short} managed cluster, `fips:true`."

I added the instruction on how to enable FIPS on the managed cluster, should we still remove that part? Is there a resource that we want to point them too or can we just let the user know which resource they need to update in OCP?

release_notes/fips_readiness.adoc
```diff
@@ -3,6 +3,8 @@
 
 FIPS readiness has been completed for {product-title}. {product-title-short} uses the same tools to make sure cryptography calls are passed to the Red Hat Enterprise Linux (RHEL) certified cryptographic modules that are used by {ocp}. 
 
+If you plan to manage clusters with FIPS enabled, you must have a FIPS-ready hub cluster because cryptography created on the hub cluster is stored on the managed cluster. Set the following parameter in your {ocp-short} managed cluster, `fips:true`.
+
```
dockerymick, line 7: Adding the link to Gus' suggestion: https://github.com/stolostron/rhacm-docs/pull/5297#discussion_r1269931550

Since I made updates, his suggestion is not displayed from the _Files changed_ tab

swopebe (approved):

release_notes/fips_readiness.adoc
gparvin: Thanks, I don't think we want to have the `fips: true` sentence but I don't see another location where this is documented right now.  If you want to keep it the wording needs to be changed.  The `fips: true` is not a parameter you can set at any time you like -- it must be set when you provision the managed cluster.  That's the kind of statements I think we would need here.  So something like:
When you require FIPS, be sure the clusters in your fleet are created with the `fips: true` setting.  FIPS cannot be enabled after a cluster has been provisioned.

dockerymick:
@swopebe thanks for your review, update, and approval. I made more updates based on feedback from @gparvin . Gus, will you re-review the PR please?

berenss (approved):
fantastic! lgtm

gparvin (approved):

dockerymick:
Thanks everyone for the reviews and feedback @swopebe @berenss @gparvin merging this PR

PR #5310, opened 2023-07-25
https://issues.redhat.com/browse/ACM-6260 Mirroring known issue

oafischer:
For 2.8 only


clusters/release_notes/known_issues.adoc
```diff
@@ -424,3 +424,12 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
 //2.7:ACM-2664
 
 When you use the console to add a new node pool to an existing hosted cluster, the same version of {ocp-short} might appear more than once in the list of options. You can select any instance in the list for the version that you want. 
+
+[#mirror-ironic-agent-image]
+=== Hub cluster {ocp} image is not mirrored automatically
+
+When you deploy a hub cluster on bare metal by using the Assisted Service, the hub cluster requires an ironic agent image. The hub cluster {ocp-short} version determines which ironic agent image is used.
```
carbonin:
```suggestion
When you deploy a hub cluster on bare metal by using the Assisted Service, spoke cluster installation requires an ironic agent image. The hub cluster {ocp-short} version determines which ironic agent image is used.
```

oafischer: Thanks for the suggestion, I'll add installation but we cannot use "spoke cluster" in the docs, so I'll leave that out.

carbonin: Sure, whatever term you need is fine. I just thought using `hub` made it sound like this was a requirement for installing the Assisted Service, when it's not. It's only required when you go to install an additional cluster.

oafischer: That's good to know, I'll make sure to clarify that, thanks!

clusters/release_notes/known_issues.adoc
```diff
@@ -424,3 +424,15 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
 //2.7:ACM-2664
 
 When you use the console to add a new node pool to an existing hosted cluster, the same version of {ocp-short} might appear more than once in the list of options. You can select any instance in the list for the version that you want. 
+
+[#mirror-ironic-agent-image]
+=== Hub cluster {ocp} image is not mirrored automatically
+//2.8:ACM-6260
+
+When you deploy a hub cluster on bare metal by using the Assisted Service, the hub cluster installation requires an ironic agent image. The hub cluster {ocp-short} version determines which ironic agent image is used.
```
carbonin: This is still very confusing to me. Maybe something like this?

```suggestion
When you deploy a hub cluster on bare metal by using the Assisted Service, cluster installation requires an ironic agent image. The hub cluster {ocp-short} version determines which ironic agent image is used.
```

oafischer: Are we talking about managed clusters? I assumed it was hub cluster, but sounds like it should be managed cluster instead?

carbonin: Yes, when installing a managed cluster using assisted we require this image.
It needs to be able to be pulled to the host that will become a node in a managed cluster.

I think spoke == managed

oafischer: That makes sense, I'll switch it to managed cluster! Thanks for clearing that up.

carbonin (approved):

clusters/release_notes/known_issues.adoc
```diff
@@ -425,6 +425,18 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
 
 When you use the console to add a new node pool to an existing hosted cluster, the same version of {ocp-short} might appear more than once in the list of options. You can select any instance in the list for the version that you want. 
 
+[#mirror-ironic-agent-image]
+=== Hub cluster {ocp} image is not mirrored automatically
+//2.8:ACM-6260
+
+If you deploy a hub cluster on bare metal by using the Assisted Service, installing managed clusters requires an ironic agent image. The hub cluster {ocp-short} version determines which ironic agent image is used.
```
dockerymick: I recommend restating the issue after you explain the cause/case. In this way you are letting the user know that this only effects users in disconnected environments earlier. I was a little confused reading, "The hub cluster {ocp-short} version determines which ironic agent image is used."

```suggestion
If you deploy a hub cluster on bare metal by using the Assisted Service, the {ocp} image is not mirrored automatically in disconnected environments. A matching ironic agent image of {ocp-short} must be used on your hub cluster to install managed clusters. The hub cluster {ocp-short} version detects which ironic agent image is being used.
```

After rereading your changes for a few times, it seems like the user does not have to create the image but they have to check if they have the correct mirrored image. What should the user do after they run the command and find that the agent image version doesn't match OCP ? I found this information in the credential_on_prem file:

"* For a disconnected environment, you must have a configured mirror registry where you can copy the release images for your cluster creation. See link:https://docs.openshift.com/container-platform/4.11/installing/disconnected_install/installing-mirroring-disconnected.html[Mirroring images for a disconnected installation] in the {ocp-short} documentation for more information."

If the user verifies that the images do not match, I would point them to a topic that helps that obtain a mirrored image. Heads up the OCP link is for 4.11, we would also need to update the reference link

oafischer: Thanks for the suggestions! Updated the PR, let me know if it looks good now.

clusters/release_notes/known_issues.adoc
```diff
@@ -425,6 +425,20 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
...
+
+If you deploy a hub cluster on bare metal by using the Assisted Service, the {ocp} image is not mirrored automatically in disconnected environments. A matching ironic agent image of {ocp-short} must be used on your hub cluster to install managed clusters. The hub cluster {ocp-short} version detects which ironic agent image is being used.
+
+In a disconnected environment, you must mirror the {ocp-short} image manually. Make sure that the ironic agent image version matches the hub cluster {ocp-short} version by running the following command:
+
+----
+oc adm release info --image-for=ironic-agent <hub-release-image>
+----
+
+If the image versions do not match, see link:https://docs.openshift.com/container-platform/4.12/installing/disconnected_install/installing-mirroring-disconnected.html[Mirroring images for a disconnected installation] in the {ocp-short} documentation to learn how to obtain a mirrored image.
```
dockerymick: What are your thoughts about us adding a link to [Maintaining a custom list of release images while disconnected](https://github.com/stolostron/rhacm-docs/blob/2.8_stage/clusters/cluster_lifecycle/release_image_disconn.adoc). This way we promote our doc haha 

this is the link from the customer portal: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html-single/clusters/index#release-images-disconnected

oafischer: I'm not familiar enough with this topic to say for sure. From the issue discussion it sounds like oc-mirror or oc adm is required but maybe someone from installer can chime in?

Update: I asked in the issue for confirmation.

dockerymick: @oafischer Yea, either install or cluster team

dockerymick (approved):
The last thing is to confirm which reference link to use by asking dev team, otherwise this looks good! Thanks for the updates and for receiving my suggestions!

clusters/release_notes/known_issues.adoc
carbonin: If we're trying to instruct someone to mirror just the single image I think something more simple like this would be the best option:

```
skopeo copy <image-from-oc-adm-output> <customer-mirror>
```

carbonin: Alternatively if we'd rather them mirror the entire hub release image then we can probably link to some existing doc.

oafischer: Simple is better, thanks for the command. I'll use that instead of linking to the OCP doc.

clusters/release_notes/known_issues.adoc
```diff
@@ -425,6 +425,28 @@ If the annotation and `ManagedCluster` name do not match, the console displays t
...
+
+If the image versions do not match, run the following command to mirror the {ocp-short} image manually:
+
+----
+skopeo copy <image-from-oc-adm-output> <mirror>
+----
+
+Replace `<image-from-oc-adm-output>` with the image from the output of `oc adm`.
+
+Replace `<mirror>` with your mirror.
```
dockerymick:
```suggestion
Replace `<mirror>` with your mirrored image.
```

clusters/install_upgrade/install_disconnected.adoc
```diff
@@ -137,8 +137,41 @@ oc apply -f mce-policy.yaml
...
+
+- x86_64: `quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:d3f1d4d3cd5fbcf1b9249dd71d01be4b901d337fdc5f8f66569eb71df4d9d446`
+- arm64: `quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:cb0edf19fffc17f542a7efae76939b1e9757dc75782d4727fb0aa77ed5809b43`
+
+You can set a different default ironic agent image by overriding the following environment variables:
+
+- x86_64: `IRONIC_AGENT_IMAGE`
+- arm64: `IRONIC_AGENT_IMAGE_ARM`
+
+Alternatively, you can also use the `infraenv.agent-install.openshift.io/ironic-agent-image-override` InfraEnv annotation to override the ironic agent image.
```
dockerymick: Is InfraEnv a parameter? If yes, I recommend that you monospace
```suggestion
Alternatively, you can also use the `infraenv.agent-install.openshift.io/ironic-agent-image-override` `InfraEnv` annotation to override the ironic agent image.
```

dockerymick (approved):
Thanks for making the updates Oliver! I checked to see if we monospaced the architecture names. Based on the [Importing a managed cluster by using the CLI](https://github.com/stolostron/rhacm-docs/blob/2.9_stage/clusters/cluster_lifecycle/import_cli.adoc) topic, it seems like we don't. I will add a commit to this PR to update the FIPS topic. Left a comment btw

clusters/install_upgrade/install_disconnected.adoc
```diff
@@ -137,8 +137,41 @@ oc apply -f mce-policy.yaml
 +
 the {mce-short} is included in the Operator Lifecycle Manager Red Hat Operator catalog.
 
-. Configure the disconnected Operator Lifecycle Manager for the Red Hat Operator catalog. Follow the steps in the link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.11/html/operators/administrator-tasks#olm-restricted-networks[Using Operator Lifecycle Manager on restricted networks] topic of the{ocp} documentation.
+. Configure the disconnected Operator Lifecycle Manager for the Red Hat Operator catalog. Follow the steps in the link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.11/html/operators/administrator-tasks#olm-restricted-networks[Using Operator Lifecycle Manager on restricted networks] topic of the {ocp} documentation.
```
carbonin: Did you also want to change this to a 4.13 link?

clusters/install_upgrade/install_disconnected.adoc
```diff
@@ -137,8 +137,41 @@ oc apply -f mce-policy.yaml
...
 
 . Now that you have the image in the disconnected Operator Lifecycle Manager, continue to install the {mce-short} for Kubernetes from the  Operator Lifecycle Manager catalog.
 
 See xref:./install_connected.adoc#installing-while-connected-online-mce[Installing while connected online] for the required steps.
+
+[#mirror-ironic-agent-image-mce]
+== Mirroring images when using Assisted Installer
+
+If you deploy a hub cluster on bare metal by using `assisted-service`, the {ocp} image is not mirrored automatically in disconnected environments. A matching ironic agent image of {ocp-short} must be used on your hub cluster to install managed clusters. The hub cluster {ocp-short} version detects which ironic agent image is being used.
```
carbonin: >must be used on your hub cluster to install managed clusters

It's not used on the hub, it's pulled by the hosts that are being installed during discovery.

carbonin: >The hub cluster {ocp-short} version detects which ironic agent image is being used.

More like "The hub cluster version _determines_ which ironic agent image _will_ be used"

clusters/install_upgrade/install_disconnected.adoc
```diff
@@ -137,8 +137,41 @@ oc apply -f mce-policy.yaml
...
+
+In a disconnected environment, you must mirror the {ocp-short} image manually. The ironic agent image used is based on the hub cluster release version set in the `ClusterVersion` resource. If the image version set in the resource is not available, one of the following default images is used, depending on your CPU architecture:
+
+- x86_64: `quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:d3f1d4d3cd5fbcf1b9249dd71d01be4b901d337fdc5f8f66569eb71df4d9d446`
+- arm64: `quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:cb0edf19fffc17f542a7efae76939b1e9757dc75782d4727fb0aa77ed5809b43`
```
carbonin: I'd like to update these, but I suppose we will need to create a docs issue each time we do, right?

clusters/install_upgrade/install_disconnected.adoc
```diff
@@ -137,8 +137,41 @@ oc apply -f mce-policy.yaml
...
+
+You can set a different default ironic agent image by overriding the following environment variables:
```
carbonin: Do we have any doc explaining how or where to do this?

Upstream we have https://github.com/openshift/assisted-service/blob/ceb709686799699f088892fa72994a5a60aa2c02/docs/operator.md#specifying-environmental-variables-via-configmap

clusters/install_upgrade/install_disconnected.adoc
```diff
@@ -137,8 +137,41 @@ oc apply -f mce-policy.yaml
...
+
+- x86_64: `IRONIC_AGENT_IMAGE`
+- arm64: `IRONIC_AGENT_IMAGE_ARM`
+
+Alternatively, you can also use the `infraenv.agent-install.openshift.io/ironic-agent-image-override` `InfraEnv` annotation to override the ironic agent image.
```
carbonin: Using "alternatively" and "also" together is confusing.

A user can definitely set the default image env vars and also use the infraenv annotation.

install/install_disconnected.adoc
```diff
@@ -225,3 +225,8 @@ spec: {}
 If you are creating the resource with a CLI, include the `mce-subscription-spec` annotation in the YAML that you apply with the `oc apply` command to create the `MulticlusterHub` resource.
 
 If you create the resource by using the OperatorHub console, switch to the _YAML view_ and insert the annotation as previously displayed. *Important:* There is no field in the OperatorHub console for the annotation in the the _Field view_ panel to create the `MulticlusterHub`.
+
+[#mirror-ironic-agent-image]
+== Mirror ironic agent image manually
+
+If you deploy a hub cluster on bare metal by using `assisted-service`, the {ocp} image is not mirrored automatically in disconnected environments. See link:../clusters/install_upgrade/install_disconnected.adoc#mirror-ironic-agent-image-mce[Mirroring images when using Assisted Installer] to learn how to mirror a matching ironic agent image manually.
```
carbonin: This has nothing to do with how the hub cluster is installed.

I'd say something more like "If you are deploying managed clusters using the infrastructure-operator, an image required for installation will not be mirrored automatically to the managed cluster disconnected environment"

clusters/install_upgrade/install_disconnected.adoc
```diff
@@ -137,8 +137,41 @@ oc apply -f mce-policy.yaml
...
+
+Make sure that the ironic agent image version matches the hub cluster {ocp-short} version by running the following command. Replace `<hub-release-image>` with the image for the hub cluster release:
+
+----
+oc adm release info --image-for=ironic-agent <hub-release-image>
+----
+
+If the image versions do not match, run the following command to mirror the {ocp-short} image manually:
```
carbonin: What image versions? If we detected the image automatically then they will match and also the user won't know. If they are here they likely have run into an issue and need to figure out how to find the image that should be mirrored so this all feels worded a bit strangely.

I think this would be better framed as first, describe what is done out of the box, then if that's not working, describe how the issue can be solved. We seem to be mixing the two concepts together by saying "check something, if it doesn't match do this other thing".

clusters/install_upgrade/install_disconnected.adoc
oafischer: I'll reword it but we cannot use future tense as per IBM Style, so I'll go with "to use" if that's OK.

clusters/install_upgrade/install_disconnected.adoc
oafischer: Do these default images change with every OCP release?

clusters/install_upgrade/install_disconnected.adoc
oafischer: I didn't see anything similar in our docs. If the user would typically not already know this I wouldn't mind linking to upstream in this case.

clusters/install_upgrade/install_disconnected.adoc
oafischer: I'm a bit confused about this. I thought the original issue was that the ironic agent image is not mirrored automatically, so don't we need steps to show how to do that manually? I'm not really familiar with the install component as usually a different writer covers this. If all we need to do is stop after the first oc adm command I'm fine with that.

I agree that we need to describe how the issue can be solved. I thought the manual mirroring was exactly that. If not, how to solve the issue?

install/install_disconnected.adoc
oafischer: I'm OK with that description, it's just missing the action item at the end. I'm still confused based on your other comment but I assume the goal/action is to mirror the ironic agent image manually since we just explained that it is not mirrored automatically?

clusters/install_upgrade/install_disconnected.adoc
carbonin: They will only differ based on MCE version as they exist in the infrastructure-operator code base.

oafischer: I see, this doc topic is for ACM 2.8 (MCE 2.3) and it sounds like we are trying to fix this whole thing by 2.9, so as long as we use the correct MCE 2.3 images we should be fine for now. Are the ones we have in the PR the correct ones?

clusters/install_upgrade/install_disconnected.adoc
carbonin: Maybe then we should get all the bits about the environment variables and default images out of this doc and just say something like:
1. The image isn't mirrored into the managed cluster environment automatically
2. If the managed cluster matches the hub cluster release architecture the image can be found by `oc adm release info --image-for=ironic-agent <hub-release-image>`
3. If the managed cluster is a different cpu architecture than the hub, use the same command but with the managed cluster arch release image and set the infraenv annotation to use this image.
4. Mirror the image to the managed cluster environment.

Does that make sense?

clusters/install_upgrade/install_disconnected.adoc
carbonin: Yup, that's what we have currently in our 2.8 branch.

install/install_disconnected.adoc
carbonin: Yup, the rest of the original text is fine I think. "See link:../clusters/install_upgrade/install_disconnected.adoc#mirror-ironic-agent-image-mce[Mirroring images when using Assisted Installer] to learn how to mirror a matching ironic agent image manually."

clusters/install_upgrade/install_disconnected.adoc
oafischer: That makes sense, but for 4, we still need to explain how to mirror the image to the managed cluster environment. That's what I thought we were explaining with the steps we have now, but it sounds like we are not? If so, would appreciate if you could provide the steps for mirroring the image to the managed cluster environment.

carbonin: >That makes sense, but for 4, we still need to explain how to mirror the image to the managed cluster environment. That's what I thought we were explaining with the steps we have now

Yup, that's what this currently is, the hard part is explaining what image to mirror.

oafischer: If we need to add more detailed descriptions that's fine, but we do need to explain how to do 4 if we bring it up. The alternative would be to link to OCP docs, if you think the information there is sufficient?

https://docs.openshift.com/container-platform/4.13/installing/disconnected_install/installing-mirroring-disconnected.html

oafischer:
We do need to wrap this up soon, for now I'll link out to OCP docs.

clusters/install_upgrade/install_disconnected.adoc
carbonin: No, I think the line we have with `skopeo` should be enough

carbonin: My only issue here is really "If the image versions do not match".
There's no matching, it's just mirror the hub ironic agent image.

If the hub is a different CPU architecture than the spoke mirror the default for your spoke or use the annotation to override the image to use the correct agent image (from the release image matching the hub version and spoke CPU architecture)

oafischer: Thanks for the reply, I think I finally get it... We kept talking about image versions, but it sounds like what you meant was image architecture? Those are 2 different things.

The whole topic makes much more sense when I replace "version" with "architecture" or "architecture version". Can you confirm if that is what's happening here? Or am I misinterpreting it again?

oafischer:
@carbonin Made some updates based on our meeting, let me know how it looks now!

clusters/install_upgrade/install_disconnected.adoc
```diff
@@ -137,8 +137,53 @@ oc apply -f mce-policy.yaml
...
+. Configure the disconnected Operator Lifecycle Manager for the Red Hat Operator catalog. Follow the steps in the link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.13/html/operators/administrator-tasks#olm-restricted-networks[Using Operator Lifecycle Manager on restricted networks] topic of the {ocp} documentation.
 
 . Now that you have the image in the disconnected Operator Lifecycle Manager, continue to install the {mce-short} for Kubernetes from the  Operator Lifecycle Manager catalog.
 
 See xref:./install_connected.adoc#installing-while-connected-online-mce[Installing while connected online] for the required steps.
+
+[#mirror-ironic-agent-image-mce]
+== Mirroring images when using Assisted Installer
+
+If you deploy a hub cluster on bare metal by using `assisted-service`, the {ocp} image is not mirrored automatically in disconnected environments. You must use a matching {ocp-short} ironic agent image to install managed clusters. The hub cluster {ocp-short} version determines which ironic agent image to use. In a disconnected environment, you must mirror the {ocp-short} image manually.
```
carbonin:
```suggestion
If you deploy managed clusters by using `assisted-service`, the {ocp} image is not mirrored automatically in disconnected environments. You must use a matching {ocp-short} ironic agent image to install managed clusters. The hub cluster {ocp-short} version determines which ironic agent image to use. In a disconnected environment, you must mirror the {ocp-short} image manually.
```

I think we made the same change earlier in the other file [here](https://github.com/stolostron/rhacm-docs/pull/5310/files#diff-92a329ec2f33941152591f38c1c6f38beb83c400acae7e3e44f18032f3494a8fR232)

carbonin (commented):
Looks good, just one minor comment

carbonin (approved):

cadawson (approved):
Looks good. Thanks!

PR #5322, opened 2023-07-27
https://issues.redhat.com/browse/ACM-6340 Adding object template example

dockerymick:
https://issues.redhat.com/browse/ACM-6340

### DIFF
--- a/governance/adv_template_process.adoc
+++ b/governance/adv_template_process.adoc
@@ -53,7 +53,7 @@ kind: ConfigurationPolicy
 metadata:
   name: create-infra-machineset
 spec:
-  remediationAction: inform
+  remediationAction: enforce
   severity: low
   object-templates-raw: |
     {{- /* Specify the parameters needed to create the MachineSet  */ -}}
--- a/governance/adv_template_process.adoc
+++ b/governance/adv_template_process.adoc
@@ -43,15 +43,19 @@ object-templates-raw: |
 ----
 
 *Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, one of the two parameter fields must be set.
-
-//Multiline is not clear to me. The YAML is using multiple lines in general. Or do we mean multiple function use is supported for object template processing . Object templates support processing multiple functions
-//Are we able to shorten this example?
-//Why might the user want to use multiple functions or the multiline templating in the object-templates-raw section? 
-Multiline templating is supported. View the following partial policy example that is configured to create a `MachineSet` resource:
+ 
+View the following policy example that is a configuration template to create a `MachineSet` resource:
 
 [source,yaml]
 ----
-object-templates-raw: |
+apiVersion: policy.open-cluster-management.io/v1
+kind: ConfigurationPolicy
+metadata:
+  name: create-infra-machineset
+spec:
+  remediationAction: inform
+  severity: low
+  object-templates-raw: |
     {{- /* Specify the parameters needed to create the MachineSet  */ -}}
     {{- $machineset_role := "infra" }}
     {{- $region := "ap-southeast-1" }}
--- a/governance/adv_template_process.adoc
+++ b/governance/adv_template_process.adoc
@@ -44,7 +44,7 @@ object-templates-raw: |
 
 *Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, one of the two parameter fields must be set.
 
-//Multiline is not clear to me. The YAML is using multiple lines in general. Or do we mean multiple functions are supported for object template processing . Object templates support processing multiple functions
+//Multiline is not clear to me. The YAML is using multiple lines in general. Or do we mean multiple function use is supported for object template processing . Object templates support processing multiple functions
 //Are we able to shorten this example?
 Multiline templating is supported. View the following partial policy example that is configured to create a `MachineSet` resource:
 
--- a/governance/adv_template_process.adoc
+++ b/governance/adv_template_process.adoc
@@ -44,7 +44,7 @@ object-templates-raw: |
 
 *Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, one of the two parameter fields must be set.
  
-View the following policy example that uses advanced templates to create a `MachineSet` objects:
+View the following policy example that uses advanced templates to create `MachineSet` objects:
 
 [source,yaml]
 ----
--- a/governance/adv_template_process.adoc
+++ b/governance/adv_template_process.adoc
@@ -43,7 +43,92 @@ object-templates-raw: |
 ----
 
 *Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, one of the two parameter fields must be set.
- 
+
+//Multiline is not clear to me. The YAML is using multiple lines in general. Or do we mean multiple functions are supported for object template processing . Object templates support processing multiple functions
+//Are we able to shorten this example?
+Multiline templating is supported. View the following partial policy example that is configured to create a `MachineSet` resource:
+
+[source,yaml]
+----
+object-templates-raw: |
+    {{- /* Specify the parameters needed to create the MachineSet  */ -}}
+    {{- $machineset_role := "infra" }}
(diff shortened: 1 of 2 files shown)


governance/adv_template_process.adoc
```diff
@@ -43,7 +43,92 @@ object-templates-raw: |
 ----
 
 *Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, one of the two parameter fields must be set.
- 
+
+//Multiline is not clear to me. The YAML is using multiple lines in general. Or do we mean multiple functions are supported for object template processing . Object templates support processing multiple functions
```
dockerymick:
```suggestion
//Multiline is not clear to me. The YAML is using multiple lines in general. Or do we mean multiple function use is supported for object template processing . Object templates support processing multiple functions
```

governance/adv_template_process.adoc
```diff
@@ -43,7 +43,92 @@ object-templates-raw: |
 ----
 
 *Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, one of the two parameter fields must be set.
- 
+
+//Multiline is not clear to me. The YAML is using multiple lines in general. Or do we mean multiple function use is supported for object template processing . Object templates support processing multiple functions
+//Are we able to shorten this example?
+Multiline templating is supported. View the following partial policy example that is configured to create a `MachineSet` resource:
```
dockerymick:
```suggestion
//Why might the user want to use multiple functions or the multiline templating in the object-templates-raw section? 
Multiline templating is supported. View the following partial policy example that is configured to create a `MachineSet` resource:
```

governance/adv_template_process.adoc
```diff
@@ -43,7 +43,93 @@ object-templates-raw: |
 ----
 
 *Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, one of the two parameter fields must be set.
- 
+
+//Multiline is not clear to me. The YAML is using multiple lines in general. Or do we mean multiple function use is supported for object template processing . Object templates support processing multiple functions
```
mprahl: Before the `object-templates-raw` feature, the policy, including templates, had to be valid YAML at submission time. This meant that the example below wouldn't work because `{{- $machineset_role := "infra" }}` is not valid YAML. Something like `data1: '{{ <my template here> }}'` worked because it is valid YAML.

`object-templates-raw` allows a YAML string which can include any kind of templating, including loops over large chunks of YAML. Once the templates are executed, however, the format of that YAML should match the schema/structure of `object-templates`.

governance/adv_template_process.adoc
```diff
@@ -43,7 +43,93 @@ object-templates-raw: |
...
+//Are we able to shorten this example?
```
mprahl: Not sure if we can but it's a good example for a common OpenShift configuration.

governance/adv_template_process.adoc
```diff
@@ -43,7 +43,93 @@ object-templates-raw: |
...
+//Why might the user want to use multiple functions or the multiline templating in the object-templates-raw section?
```
mprahl: I covered this in an earlier comment, but the main reason is it allows you to have dynamic YAML definitions beyond primitive types such as strings, booleans, and integers.  This is often useful when you need a loop or conditional logic for non-primitive YAML types.

dockerymick:
@mprahl thanks for your feedback. I will address them ASAP

dockerymick:
@mprahl thanks for your patience. When you have an opportunity, will you re-review the PR?

governance/adv_template_process.adoc
```diff
@@ -44,6 +44,96 @@ object-templates-raw: |
 
 *Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, one of the two parameter fields must be set.
  
+View the following policy example that is a configuration template to create a `MachineSet` resource:
```
mprahl:
```suggestion
View the following policy example with advanced templating that creates `MachineSet` objects:
```

dockerymick: @mprahl thanks for your review and suggestion. How about this? 
```suggestion
View the following policy example that uses advanced templates to create a `MachineSet` objects:
```

dockerymick:
@mprahl When you have a chance, will you re-review the PR

dockerymick:
@jc-berger I am having issues with adding you as a reviewer. When you have a chance will you take a look at my PR as well?

governance/adv_template_process.adoc
```diff
@@ -44,6 +44,96 @@ object-templates-raw: |
 
 *Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, one of the two parameter fields must be set.
  
+View the following policy example that uses advanced templates to create a `MachineSet` objects:
```
jc-berger: Singular to plural. I think it should be one of the following:

1) ...example that uses advanced templates to create **a** `MachinSet` **object.** 

or 

2) ...example that uses advanced templates to create `MachinSet` **objects.**

jc-berger: I'm still learning the "extent" of the doc we're wanting to improve. 

Reading the linked issue, I think this PR is great and addresses what it needs to (besides that one grammatical nitpick I have above) :) 

When we edit docs, how much do we "add" to it? 

For example, here we could state the user benefit to a MachinSet object, something like:

View the following policy example that uses advanced templates to create a `MachineSet` object so that you can < name user benefit >:

dockerymick: Great point! Going with option 2

dockerymick:
```suggestion
View the following policy example that uses advanced templates to create `MachineSet` objects:
```

dockerymick: I agree with you! And to answer your question, it depends on how much work would need to be completed based on a specific ask. For example, when I notice that there is a broken link or something can be clarified outside of what is requested on the issue, I make the change. If I notice something that needs to be addressed, but is out of scope I add a hidden comment 

We should definitely be sure to document the user benefit. In this case, it seems like we wanted to add the example to simply show the user a "real-world" case. I think we can get more specific with this real-world case. 

As Jake mentioned through Slack, his suggestion is not a blocker to merge but absolutely something to consider. 

@mprahl @jc-berger thoughts on the following statement:

View the following policy example that uses advanced templates to create and configure infrastructure `MachineSet` objects for your managed clusters.

jc-berger: @dockerymick I love your rewrite! User benefit is clear, but it's still concise and minimal and action oriented. I say make the change :)

governance/adv_template_process.adoc
```diff
@@ -44,6 +44,96 @@ object-templates-raw: |
...
+View the following policy example that uses advanced templates to create `MachineSet` objects:
+
+[source,yaml]
+----
+apiVersion: policy.open-cluster-management.io/v1
+kind: ConfigurationPolicy
+metadata:
+  name: create-infra-machineset
+spec:
+  remediationAction: inform
```
mprahl: Based on the language above, it implies enforce. We can leave it as inform if you add additional text pointing this out.

```suggestion
  remediationAction: enforce
```

governance/adv_template_process.adoc
dockerymick: Thanks! Making that update now

governance/adv_template_process.adoc
```diff
@@ -44,6 +44,96 @@ object-templates-raw: |
 
 *Note:* While `spec.object-templates` and `spec.object-templates-raw` are optional, one of the two parameter fields must be set.
  
+View the following policy example that uses advanced templates to create `MachineSet` objects:
```
dockerymick:
```suggestion
View the following policy example that uses advanced templates to create and configure infrastructure `MachineSet` objects for your managed clusters.
```

dockerymick:
@mprahl @jc-berger thanks to you both again for your feedback and confirmation. I have accepted Matt's suggestion to change the remediation action and updated the statement based on Jake's feedback.

mprahl (approved):

PR #5326, opened 2023-07-28
https://issues.redhat.com/browse/ACM-6366 Policy KI 2.8

oafischer:
2.8 only.

### DIFF
--- a/release_notes/known_issues_governance.adoc
+++ b/release_notes/known_issues_governance.adoc
@@ -212,4 +212,4 @@ oc delete ansiblejob {ansiblejob name} -n {namespace}
 == Compliance state of policies does not appear in _PolicyReport_
 //2.8:ACM-6366
 
-The compliance state of policies that use the Gatekeeper integration does not appear in the `PolicyReport`. As a result, you do not receive alerts for these policies.
+The compliance state of policies that use the Gatekeeper integration does not appear in the `PolicyReport` resource. As a result, you do not receive alerts for these policies.
--- a/release_notes/known_issues_governance.adoc
+++ b/release_notes/known_issues_governance.adoc
@@ -207,3 +207,9 @@ oc get ansiblejob -n {namespace}
 ----
 oc delete ansiblejob {ansiblejob name} -n {namespace}
 ----
+
+[#duplicate-ansible-jobs]
+== Coompliance state of policies does not appear in _PolicyReport_
+//2.8:ACM-6366
+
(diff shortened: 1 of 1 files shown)


release_notes/known_issues_governance.adoc
```diff
@@ -207,3 +207,9 @@ oc get ansiblejob -n {namespace}
 ----
 oc delete ansiblejob {ansiblejob name} -n {namespace}
 ----
+
+[#compliance-state-gatekeeper]
+== Compliance state of policies does not appear in _PolicyReport_
+//2.8:ACM-6366
+
+The compliance state of policies that use the Gatekeeper integration does not appear in the `PolicyReport`. As a result, you do not receive alerts for these policies.
```
dockerymick:
```suggestion
The compliance state of policies that use the Gatekeeper integration does not appear in the `PolicyReport` resource. As a result, you do not receive alerts for these policies.
```

dockerymick (approved):

release_notes/known_issues_governance.adoc
```diff
@@ -207,3 +207,9 @@ oc get ansiblejob -n {namespace}
 ----
 oc delete ansiblejob {ansiblejob name} -n {namespace}
 ----
+
+[#compliance-state-gatekeeper]
+== Compliance state of policies does not appear in _PolicyReport_
```
JustinKuli: Maybe "gatekeeper resources" instead of "policies"? Just that this heading makes it sound worse than it is. Granted, the actual section makes it clear...

release_notes/known_issues_governance.adoc
```diff
@@ -207,3 +207,9 @@ oc get ansiblejob -n {namespace}
...
+//2.8:ACM-6366
+
+The compliance state of policies that use the Gatekeeper integration does not appear in the `PolicyReport` resource. As a result, you do not receive alerts for these policies.
```
JustinKuli: I wonder if we should link to the integration's docs: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/governance/governance#gatekeeper-policy

release_notes/known_issues_governance.adoc
oafischer: Sounds good to me, I'll update the heading

release_notes/known_issues_governance.adoc
oafischer: Not sure, looks like mostly background info. If the user would really benefit from it I can add it, otherwise I'd be inclined to leave it as is.

PR #5367, opened 2023-08-09
Not worked:  (2.9 work to remain, see issue: https://issues.redhat.com/browse/ACM-1869: Chris)

swopebe:
Recognizing cluster lifecycle may need to actually go into MCE, but let's just keep together until we know what we are going to do.


cadawson:
This is not happening in 2.9. PM, development agreed. Closing this for now.

PR #5397, opened 2023-08-14
[ACM-6433]: Removing references to enabling the hosted control planes feature

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-6433


lahinson:
@rokej This PR removes any references to the "Enabling the hosted control planes feature" section of the ACM docs. PTAL when you can. 

I have a separate Jira in progress to remove the Technology Preview wording where appropriate: https://issues.redhat.com/browse/ACM-6859

Finally, I have one question: If hosted control planes on AWS will remain in Technology Preview for OCP 4.14/ACM 2.9, will HyperShift be enabled by default on that platform as well?

rokej (commented):
@lahinson The change looks good to me. 

One suggestion is that under https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/clusters/cluster_mce_overview#hosted-control-planes-intro, we should document that the feature is enabled by default. 

And we should have a separate section called (disabling the hosted control planes feature). This section should be similar to the removed "hosted-enable-feature-aws" section but setting those components to "false" instead. The component name "hypershift-preview" is now changed to "hypershift".

lahinson:
> @lahinson The change looks good to me.
> 
> One suggestion is that under https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/clusters/cluster_mce_overview#hosted-control-planes-intro, we should document that the feature is enabled by default.
> 
> And we should have a separate section called (disabling the hosted control planes feature). This section should be similar to the removed "hosted-enable-feature-aws" section but setting those components to "false" instead. The component name "hypershift-preview" is now changed to "hypershift".

@rokej Good suggestions! I'll let you know when those updates are ready for review. Thank you!

lahinson:
@rokej I made the updates that you suggested. We already had a topic about how to disable the feature, so I updated it to refer to the `hypershift` feature instead of `hypershift-preview`, and I made sure that we linked to that doc from the `configure_hosted_aws.adoc` file.

clusters/hosted_control_planes/configure_hosted_aws.adoc
```diff
@@ -7,8 +7,6 @@ To configure hosted control planes, you need a hosting cluster and a hosted clus
```
lahinson: @rokej I just realized that the *Important* note on line 6 might cause concern once the feature is automatically enabled. Should we update that note to instruct users to disable the feature if the cluster is configured for FIPS mode or is in a disconnected environment?

rokej: @lahinson Good catch. We should remove the Important note. Hypershift should be supported in both environments.

rokej (approved):

clusters/hosted_control_planes/enable_or_disable_hosted.adoc
```diff
@@ -0,0 +1,151 @@
...
+
+. Disable the `hypershift-addon` by running the following command:
++
+----
+oc patch mce multiclusterengine --type=merge -p '{"spec":{"overrides":{"components":[{"name":"hypershift-local-hosting","enabled": false}]}}}' <1>
+----
++
+<1> The default `MultiClusterEngine` resource instance name is `multiclusterengine`, but you can get the `MultiClusterEngine` name from your cluster by running the following command: `$ oc get mce`.
++
+*Tip:* You can also disable the `hypershift-addon` for the `local-cluster` from the {mce-short} console after disabling the `hypershift-addon`.
```
cadawson:
```suggestion
*Note:* You can also disable the `hypershift-addon` for the `local-cluster` from the {mce-short} console after disabling the `hypershift-addon`.
```

cadawson: We are moving away from using a Tip.

clusters/hosted_control_planes/managing_hosted_kubevirt.adoc
```diff
@@ -29,8 +29,7 @@ oc patch ingresscontroller -n openshift-ingress-operator default --type=json -p
 oc patch storageclass ocs-storagecluster-ceph-rbd -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
 ----
 - You need a valid pull secret file for the `quay.io/openshift-release-dev` repository. For more information, see link:https://console.redhat.com/openshift/install/platform-agnostic/user-provisioned[Install OpenShift on any x86_64 platform with user-provisioned infrastructure].
-- You need to enable the hosted control planes feature. For information, see xref:../hosted_control_planes/configure_hosted_aws.adoc#hosted-enable-feature-aws[Enabling the hosted control planes feature].
-- After you enable the hosted control planes feature, you need to xref:../hosted_control_planes/configure_hosted_aws.adoc#hosted-install-cli[install the hcp command line interface binary].
+- You need to xref:../hosted_control_planes/configure_hosted_aws.adoc#hosted-install-cli[install the hcp command line interface binary].
```
cadawson: Capitalize hcp?

clusters/hosted_control_planes/configure_hosted_aws.adoc
```diff
@@ -496,6 +413,6 @@ See link:https://access.redhat.com/documentation/en-us/openshift_container_platf
 [#additional-resources-configure-hosted-cluster-aws]
 == Additional resources
 
-For more information about hosted control planes on AWS, see the following resource:
+* To disable the hosted control planes feature, see xref:../hosted_control_planes/disable_hosted.adoc#hosted-disable-feature[Disabling the hosted control planes feature].
```
cadawson: I don't think the disable_hosted.adoc file is listed in the main file. I think this link will break. Maybe the following: 

```suggestion
* To disable the hosted control planes feature, see xref:../hosted_control_planes/enable_or_disable_hosted.adoc#disable-hosted-control-planes[Disabling the hosted control planes feature].
```

cadawson (changes requested):
It looks like there might be a broken link here.I've suggested what I think is the correct link. If the link is correct as it is, that file needs to be added to the main file. 

A couple of other minor things. 

Looks good! Thanks!

clusters/hosted_control_planes/enable_or_disable_hosted.adoc
lahinson: Good catch!

clusters/hosted_control_planes/configure_hosted_aws.adoc
lahinson: I'll fix that. Thanks!

clusters/hosted_control_planes/managing_hosted_kubevirt.adoc
lahinson: I believe it stays lowercase. I can confirm with Roke.

lahinson: I got confirmation that it's "hcp" -- lowercase.

lahinson:
Thanks so much for your review, @cadawson! I fixed the link and updated "Tip" to "Note". I also confirmed that "hcp" should be lowercase.

PR #5414, opened 2023-08-17
Fix GitOps book

oafischer:


cadawson (approved):
Not ideal, but hopefully will work temporarily. Thanks for looking into this.

PR #5425, opened 2023-08-24
https://issues.redhat.com/browse/ACM-6423 Add API

oafischer:
Also for 2.8 and 2.7.

Please review as I'm not familiar with the APIs, but from what I could tell there are no required spec fields in this case. If there are, let me know which ones we would like to add and what they do.


apis/managedclusterset.json.adoc
```diff
@@ -0,0 +1,244 @@
...
+|===
+|Name|Description|Schema
+|*apiVersion* +
+__required__|Versioned schema of the `ManagedClusterSet`.|string
+|*kind* +
+__required__|String value that represents the REST resource.|string
+|*metadata* +
+__required__|Metadata of the `ManagedClusterSet`.|object
+|*spec* +
+__required__|Specification of the `ManagedClusterSet`.|<<_rhacm-docs_apis_clustersetbinding_jsonclustersetbinding_spec,spec>>
```
brian-jarvis: This creates a link to the spec, but I do not see the spec in the file.

oafischer: Fixed, thanks for catching that!

brian-jarvis: @oafischer  the fix was to remove the link?  Why aren't we documenting the spec?

oafischer: Sorry, I'm probably missing something and if so let me know, but I don't see anything that needs to be documented, based on how we treat other spec fields in our existing documentation.

brian-jarvis (approved):

dockerymick (approved):
I noticed that ManagedClusterSet is written a couple of ways. I appreciate that the `kind` name matches when there is a hyperlink. We can improve the name consistency as tech debt for 2.9. Let me know your thoughts.

oafischer:
Looks like a copy/paste error, I'll fix it!

oafischer:
@dockerymick I just checked, all of our API docs have the same inconsistencies. I think it's to avoid lower case in the title and main paragraph. My suggestion is to keep the version I had at the time of review and eventually fix all the API docs to make them consistent.

PR #5426, opened 2023-08-24
[ACM-6858]: Add docs to disable auto import of hosted clusters

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-6858


clusters/hosted_control_planes/configure_hosted_aws.adoc
```diff
@@ -371,6 +371,51 @@ oc create secret generic hypershift-operator-private-link-credentials --from-lit
 oc label secret hypershift-operator-private-link-credentials -n local-cluster cluster.open-cluster-management.io/backup=""
 ----
 
+[#hosted-disable-auto-import]
+== Disabling the automatic import of hosted clusters into {mce-short}
+
+Hosted clusters are automatically imported into {mce-short} after the control plane becomes available. All {product-title-short} add-ons are also enabled if {product-title-short} is installed.
+
+However, if needed, you can disable the automatic import of hosted clusters. When automatic import is disabled, only newly created hosted clusters are not automatically imported. Hosted clusters that were already imported are not affected. You can still manually import clusters by using the UI or the `hcp` command line interface.
```
dtthuynh: Should we mention clusters that were in the process of being created already or is it implied?
`Hosted clusters already being created and hosted clusters previously imported are not affected`
Actually, I'm not sure if this is true for clusters creating when this switch happens, have to check this edge case maybe.

When we mention `You can still manually import clusters by using the UI or the `hcp` command line interface.`, is it worth linking to that section?

clusters/hosted_control_planes/configure_hosted_aws.adoc
```diff
@@ -371,6 +371,51 @@ oc create secret generic hypershift-operator-private-link-credentials --from-lit
...
+
+To disable the automatic import of hosted clusters, complete the following steps:
+
+. On the hub cluster, open the `hypershift-addon-deploy-config` specification that is in the `AddonDeploymentConfig` resource in the `hypershift` namespace by entering the following command:
```
dtthuynh: It's not in the hypershift namespace but the namespace where MCE is installed in, in the example below its `multicluster-engine`

clusters/hosted_control_planes/configure_hosted_aws.adoc
```diff
@@ -371,6 +371,51 @@ oc create secret generic hypershift-operator-private-link-credentials --from-lit
...
+  customizedVariables:
+   - name: hcMaxNumber
+     value: "80"
+  - name: hcThresholdNumber
+    value: "60"
+  - name: autoImportDisabled
+    value: "true"
+----
+  
+. If the `autoImportDisabled` variable does not exist yet in the `AddonDeploymentConfig` resource, you can patch it by entering the following command:
```
dtthuynh: I added this command below as a shortcut to not needing to open the config to edit, maybe we can list this first and if the `autoImportDisabled` variable already exists (the command would fail in that case), we can ask them to manually update the config? @rokej what do you think?

clusters/hosted_control_planes/configure_hosted_aws.adoc
```diff
@@ -371,6 +371,51 @@ oc create secret generic hypershift-operator-private-link-credentials --from-lit
...
+
++
+----
+oc patch addondeploymentconfig hypershift-addon-deploy-config -n multicluster-engine --type=json -p='[{"op": "add", "path": "/spec/customizedVariables/-","value":{"name":"autoImportDisabled","value":"true"}}]'
+----
+
+. To re-enable automatic import, remove the `autoImportDisabled` variable from the `AddonDeploymentConfig` resource.
```
dtthuynh: I think we can also just set it to a falsy value, but not sure if worth noting.

lahinson: I updated the text to indicate that you can either remove the variable or change the value to false.

clusters/hosted_control_planes/configure_hosted_aws.adoc
lahinson: I'll wait for Roke to weigh in before I make any changes here.

clusters/hosted_control_planes/configure_hosted_aws.adoc
lahinson: Thanks for the clarification! I updated the text.

clusters/hosted_control_planes/configure_hosted_aws.adoc
lahinson: I do think we need more clarity here.  I'm assuming that starting in MCE 2.4/ACM 2.9, hosted clusters are automatically imported, and any hosted clusters that were created in previous MCE versions are not affected. Should we state that explicitly? For example:

>Hosted clusters that were imported in versions of MCE 2.3 or earlier are not affected.

Good call on adding a link to the manual import docs. I can add that.

rokej: I think it is a good idea to clarify these points.

- starting in MCE 2.4/ACM 2.9, hosted clusters are automatically imported
- Hosted clusters previously imported are not affected whether the auto-import is enabled or disabled.
- Hosted clusters being imported will eventually be imported even when the auto-import is disabled.
- when you upgrade to MCE 2.4/ACM 2.9 and the auto-import is enabled, all un-imported hosted clusters will be auto-imported if their control planes are available.

You can still manually import clusters by using the UI or the `hcp` command line interface.   ---> This should be

"You can still manually import clusters by using the UI or by creating the ManagedCluster and KlusterletAddonConfig custom resources which can be a link to https://github.com/stolostron/rhacm-docs/blob/2.9_stage/clusters/hosted_control_planes/importing_hosted_cluster_aws.adoc. By the way, this section is common for all platforms (AWS, KubeVirt and BareMetal agent).

clusters/hosted_control_planes/configure_hosted_aws.adoc
rokej: Although it is a convenient command to quickly add the configuration to the `AddonDeploymentConfig` initially, in order to avoid any confusion, I think we can simply remove the example `oc patch` command and let user decide how to update the `AddonDeploymentConfig`.

rokej (commented):
Added replies to the comments.

clusters/hosted_control_planes/configure_hosted_aws.adoc
lahinson: @rokej Thanks for your response. It sounds like I should remove lines 410 - 415. Is that right?

rokej: @lahinson I was suggesting that we remove lines 402 ~ 407.

lahinson: Done. Thank you, @rokej!

rokej (approved):
Thank you

clusters/hosted_control_planes/configure_hosted_aws.adoc
```diff
@@ -363,6 +363,49 @@ oc create secret generic hypershift-operator-private-link-credentials --from-lit
...
 ----
 
+[#hosted-disable-auto-import]
+== Disabling the automatic import of hosted clusters into {mce-short}
+
+Starting with {mce-short} 2.4, hosted clusters are automatically imported into {mce-short} after the control plane becomes available. All {product-title-short} add-ons are also enabled if {product-title-short} is installed. Any hosted clusters that were previously imported are not affected, even if you disable automatic import. When you upgrade to {mce-2.4} and automatic import is enabled, all hosted clusters that are not imported are automatically imported if their control planes are available.
+
+However, if needed, you can disable the automatic import of hosted clusters. When automatic import is disabled, only newly created hosted clusters are not automatically imported. Hosted clusters that were already imported are not affected. You can still manually import clusters by using the UI or by creating the `ManagedCluster` and `KlusterletAddonConfig` custom resources. For more information about manual import, see _Importing a hosted control plane cluster_.
```
cadawson:
```suggestion
If needed, you can disable the automatic import of hosted clusters. When automatic import is disabled, only newly created hosted clusters are not automatically imported. Hosted clusters that were already imported are not affected. You can still manually import clusters by using the console or by creating the `ManagedCluster` and `KlusterletAddonConfig` custom resources. For more information about manual import, see _Importing a hosted control plane cluster_.
```

clusters/hosted_control_planes/configure_hosted_aws.adoc
```diff
@@ -363,6 +363,49 @@ oc create secret generic hypershift-operator-private-link-credentials --from-lit
 oc label secret hypershift-operator-private-link-credentials -n local-cluster cluster.open-cluster-management.io/backup=""
 ----
 
+[#hosted-disable-auto-import]
+== Disabling the automatic import of hosted clusters into {mce-short}
+
+Starting with {mce-short} 2.4, hosted clusters are automatically imported into {mce-short} after the control plane becomes available. All {product-title-short} add-ons are also enabled if {product-title-short} is installed. Any hosted clusters that were previously imported are not affected, even if you disable automatic import. When you upgrade to {mce-2.4} and automatic import is enabled, all hosted clusters that are not imported are automatically imported if their control planes are available.
```
cadawson:
```suggestion
Starting with {mce-short} 2.4, hosted clusters are automatically imported into {mce-short} after the control plane becomes available. All {product-title-short} add-ons are also enabled if {product-title-short} is installed. Any hosted clusters that were previously imported are not affected, even if you disable automatic import. When you upgrade to {mce-short} {mce-version} and automatic import is enabled, all hosted clusters that are not imported are automatically imported if their control planes are available.
```

cadawson (approved):
A couple of suggestions, but looks good. I don't think the conref will work as it is. Thanks!

clusters/hosted_control_planes/configure_hosted_aws.adoc
lahinson: Good catch! I must have been thinking "MCE 2.4" when I was typing {mce-short}, and it came out as {mce-2.4}. 🤪

clusters/hosted_control_planes/configure_hosted_aws.adoc
lahinson: Excellent suggestions. Fixing now.

PR #5429, opened 2023-08-24
https://issues.redhat.com/browse/ACM-7007 Removal (ingress)

oafischer:
Is this enough? We still have some topics like https://github.com/stolostron/rhacm-docs/blob/2.9_stage/install/adv_config_install.adoc#sslciphers-deprecated floating around. As well as a FIPS notice about sslCiphers and various mentions here and there in the doc.

### DIFF
--- a/release_notes/deprecate_remove.adoc
+++ b/release_notes/deprecate_remove.adoc
@@ -175,12 +175,6 @@ A _deprecated_ component, feature, or service is supported, but no longer recomm
 | While `PlacementRule` is still available, it is not supported and the console displays `Placement` by default.
 
 | Installer
-| `ingress.sslCiphers` field in `operator.open-cluster-management.io_multiclusterhubs_crd.yaml`
-| 2.7
-| None
-| See link:../install/adv_config_install.adoc[Advanced Configuration] for configuring install.
-
-| Installer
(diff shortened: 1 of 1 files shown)


dislbenn (approved):
The changes look good to me. The goal of the initial task was to stop auto-populating the `spec.ingress.sslCiphers` field in the MCH CR. If a customer or user upgrade their ACM version and originally had an MCH CR with the `spec.ingress.sslCiphers` field defined, we will still recognize the field; however, it is `deprecated` and serves no purpose.

dockerymick:
Thanks for making these changes @oafischer ! based on @dislbenn [comment](https://github.com/stolostron/rhacm-docs/pull/5429#pullrequestreview-1594116431), it seems like the entry should remain in the deprecations table. 

`ingress.sslCiphers` field in `operator.open-cluster-management.io_multiclusterhubs_crd.yaml` is deprecated only. Will the field be displayed for the user still or is the field completely removed? It would be confusing for the user to read that the field is removed, while also seeing the field in their environment (if they do)
@dislbenn will you confirm?  

From my perspective, I would also add Disaiah's explanation to the _More details_ column: "If a customer or user upgrade their ACM version and originally had an MCH CR with the spec.ingress.sslCiphers field defined, we will still recognize the field; however, it is deprecated and serves no purpose."

oafischer:
Updated, thanks for checking in @dockerymick!

PR #5430, opened 2023-08-24
[ACM-6974]: Adding supported OCP versions for hosted control planes

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-6974


clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -5,6 +5,15 @@ With {mce-short} cluster management, you can deploy {ocp-short} clusters by usin
 
 Hosted control planes for {ocp-short} are available on bare metal and {ocp-virt}, and as a Technology Preview feature on Amazon Web Services (AWS). You can host the control planes on {ocp-short} version 4.14. The hosted control planes feature is enabled by default.
 
+The following table indicates which {ocp-short} versions are supported for each platform:
+
+|===
+| Platform | HyperShift Operator versions | Minimum hosted {ocp-short} version
```
rokej: Can you please rename `HyperShift Operator versions` to something like `Hosting OCP version`. `Hosting OCP` is the OCP where MCE or ACM is installed.

lahinson: Will do.

lahinson: @rokej I updated `HyperShift Operator version` to `Hosting OpenShift Container Platform version`.

rokej: @lahinson Thank you. Should we add a little note about what `Hosting OpenShift Container Platform` means? Like... this is the OCP where the MCE hypershift feature is enabled?

lahinson: @rokej Good point. I added a line just above the table to clarify what Hosting OCP version means.

rokej (approved):
Thank you

clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -5,6 +5,15 @@ With {mce-short} cluster management, you can deploy {ocp-short} clusters by usin
 
 Hosted control planes for {ocp-short} are available on bare metal and {ocp-virt}, and as a Technology Preview feature on Amazon Web Services (AWS). You can host the control planes on {ocp-short} version 4.14. The hosted control planes feature is enabled by default.
 
+The following table indicates which {ocp-short} versions are supported for each platform. In the table, _Hosting {ocp-short} version_ refers to the {ocp-short} version where {mce-short} is enabled:
+
+|===
+| Platform | Hosting {ocp-short} version | Minimum hosted {ocp-short} version
+| Bare metal | 4.11, 4.12, 4.13, 4.14 | 4.14
```
dockerymick: Small suggestion, not required. Would it make sense to write a range instead of listing all of the versions ? For example, 4.11 - 4.14

dockerymick (approved):
Quick question, but this looks good!

clusters/hosted_control_planes/hosted_intro.adoc
lahinson: Valid suggestion - will apply 🚀

PR #5431, opened 2023-08-24
fix build

cadawson:


cadawson:
Add gitops_policy_operator 
and 
gitops_manage_policy_def

PR #5445, opened 2023-08-28
[ACM-7126]: Node labeling guidelines

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-7126


lahinson:
@zanetworker When you can, would you comment on what we need to add for node-labeling guidelines for OCP 4.14/MCE 2.4? The current ACM docs contain https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/clusters/cluster_mce_overview#hosted-cluster-workload-distributing, but I assume we need to add more details. 

I'll also plan to link to these docs from the official OCP docs, most likely on this page: https://docs.openshift.com/container-platform/4.13/hosted_control_planes/hcp-configuring.html

zanetworker:
@lahinson So distributing is one use-case, runnong on infra is a special piece of it that serves a different purpose. On its own its an important user-task that should be mentioned early in our docs.

This part: 

`To enable a hosted cluster to require its pods to be scheduled into infra nodes, set HostedCluster.spec.nodeSelector, as shown in the following example:

  spec:
    nodeSelector:
      role.kubernetes.io/infra: ""`
      
 Describes a small part of the How but not the full How (we still need to tell users to label all the management cluster's nodes as infra) or the why (the part about the subscription). More details here: https://issues.redhat.com/browse/HOSTEDCP-1053 
 
 Happy to chat about it too.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,37 @@
...
+
+Before you get started with hosted control planes for {ocp-short}, you must properly label nodes so that the pods of hosted clusters can be scheduled into infrastructure nodes. Node labeling is also important for the following reasons:
+
+* To ensure high availability and proper workload deployment
+* To ensure that control planes are well contained
+
+[#hosted-cluster-labels-taints-overview]
+== Labels and taints for management cluster nodes
+
+//lahinson - sept. 2023 - do we need to add a note to prevent customers from using the management cluster for their workload?
```
zanetworker: Yes please :)

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,37 @@
...
 
 As a management cluster administrator, you can use the following labels and taints in the management cluster nodes to schedule the control plane workload:
 
-* `hypershift.openshift.io/control-plane: true`
+//lahinson - sept. 2023 - If we want to explain the implications of each label and taint, I'll need that info.
+
+* `hypershift.openshift.io/hypershift.openshift.io/control-plane: true`
+* `node-role.kubernetes.io/infra`
+* `topology.kubernetes.io/zone`
 * `hypershift.openshift.io/cluster: ${HostedControlPlane Namespace}`
```
zanetworker: @enxebre can you help @lahinson here explain the implications of each label and taint?

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,37 @@
...
 
 Pods for a hosted cluster have tolerations, and the scheduler uses affinity rules to schedule them. Pods tolerate taints for `control-plane` and the `cluster` for the pods. The scheduler prioritizes the scheduling of pods into nodes that are labeled with `hypershift.openshift.io/control-plane` and `hypershift.openshift.io/cluster: ${HostedControlPlane Namespace}`.
 
 For the `ControllerAvailabilityPolicy` option, use `HighlyAvailable`. When you use that option, you can schedule pods for each deployment within a hosted cluster across different failure domains by setting `topology.kubernetes.io/zone` as the topology key.
 
-To enable a hosted cluster to require its pods to be scheduled into infra nodes, set `HostedCluster.spec.nodeSelector`, as shown in the following example:
+//lahinson - sept. 2023 - If we want to explain zone versus hostname, I'll need info for that.
```
zanetworker: @lahinson labeling by hostname is useful in on-prem deployments (e.g., agent/kubevirt) as this deployments has not concept of zones, unlike AWS for example. 

@csrwng @enxebre can you help clarify this as well?

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -21,6 +42,8 @@ To enable a hosted cluster to require its pods to be scheduled into infra nodes,
 
 This way, hosted control planes for each tenant are eligible infrastructure node workloads, and you do not need to entitle the underlying {ocp-short} nodes.
 
+//lahinson - sept. 2023 - in the line above, should we replace "tenant" with something else? Maybe "hosted cluster"?
```
zanetworker: Makes sense :)

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,37 @@
...
+== Labels and taints for management cluster nodes
+
+//lahinson - sept. 2023 - do we need to add a note to prevent customers from using the management cluster for their workload?
 
 As a management cluster administrator, you can use the following labels and taints in the management cluster nodes to schedule the control plane workload:
 
-* `hypershift.openshift.io/control-plane: true`
+//lahinson - sept. 2023 - If we want to explain the implications of each label and taint, I'll need that info.
+
+* `hypershift.openshift.io/hypershift.openshift.io/control-plane: true`
```
csrwng: s / hypershift.openshift.io/hypershift.openshift.io/control-plane: true / hypershift.openshift.io/control-plane: true /

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,37 @@
 [#hosted-cluster-workload-distributing]
-= Distributing hosted cluster workloads (Technology Preview)
+= Distributing hosted cluster workloads
+
+Before you get started with hosted control planes for {ocp-short}, you must properly label nodes so that the pods of hosted clusters can be scheduled into infrastructure nodes. Node labeling is also important for the following reasons:
```
csrwng, line 4: @zanetworker why do we want pods of hosted clusters to be scheduled into infrastructure nodes? This seems confusing to me. They're not infrastructure nodes from the management cluster's perspective for sure.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,37 @@
...
+
+* To ensure high availability and proper workload deployment
+* To ensure that control planes are well contained
```
csrwng: I would change the above sentence to something like:
* To ensure that control plane workloads have the proper isolation level

This requires some explanation as well. By isolation level I mean one of:
1. Everything shared (control planes for hosted clusters can run on any node designated for control planes.
2. Request serving isolation (request serving pods in their own dedicated nodes)
3. Nothing shared (every control plane in its own dedicated nodes)

I would also add:
* To ensure that control plane workloads are isolated from other workloads in the management cluster.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
csrwng: I would temper that by saying that they could run other workloads as long as those workloads do not run on nodes where control planes run (which is why it is important to taint nodes for control planes)

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
csrwng: For this one, it would be good to explain that you need *both* the label and the taint. The effect of this one is that it dedicates a node to running hosted control plane workloads only. This is important if you don't want other workloads in the management cluster (including MCE/ACM, monitoring, etc) to run in your control plane nodes. I would almost say it's required.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,37 @@
...
+
+//lahinson - sept. 2023 - do we need to add a note to prevent customers from using the management cluster for their workload?
 
 As a management cluster administrator, you can use the following labels and taints in the management cluster nodes to schedule the control plane workload:
 
-* `hypershift.openshift.io/control-plane: true`
+//lahinson - sept. 2023 - If we want to explain the implications of each label and taint, I'll need that info.
+
+* `hypershift.openshift.io/hypershift.openshift.io/control-plane: true`
+* `node-role.kubernetes.io/infra`
```
csrwng: Not sure why this one is here. @zanetworker can you remind me why we wanted control plane workloads in infra nodes?

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,37 @@
...
+* `topology.kubernetes.io/zone`
```
csrwng: This one is only the label that is needed. It is a required label if you are running control planes with a controllerAvailabilityPolicy of HighlyAvailable. For environments where zones don't make sense, you can set this label to the host name of the node. If this label is not present, HighlyAvailable control planes will not deploy successfully.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
csrwng: This one is both a label and a taint. It should only be used when you want to dedicate a node to a single HostedCluster.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
csrwng: The zone label is required for us to successfully deploy HighlyAvailable clusters. It is up to the customer to decide what zone means. It could be a location or rack name, or at the very least, the hostname of the node where it is set. Without the zone label we cannot successfully deploy HighlyAvailable control planes.

csrwng (commented):
Some comments

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
zanetworker: I think "isolation" might be mis-interpreted here. OpenShift already provides necessary means for isolation via SELinux, Network Policies,... which the control planes of hosted clusters are consumers of being a workload. To avoid confusion, I would opt-out of mentioning "contained" or "isolation" except for special cases where it makes sense for the Hosted cluster tenants.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
zanetworker: @csrwng what about the impact on the API-server, etcd of the management cluster?

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
zanetworker: Not to count towards the subscription (additional cores).

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
csrwng: There is definitely an impact, but that is true whether you only run control plane workloads or other workloads as well.. it's more of what it is that you're running. You need to properly size your MC's control plane for whatever it is that you want to run.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
csrwng: For this one we only need the label

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
enxebre: this all makes sense. I'd have clear separated points one for
"To ensure that control plane workloads are isolated from other workloads in the management cluster." via `hypershift.openshift.io/hypershift.openshift.io/control-plane: true`
an another one for the different multi-tenancy workload distribution modes Cesar mentions above.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
zanetworker: @csrwng except that the control-planes are relatively vetted (and perf/scaled) workloads that are more predictible than wild-card workloads a customer might decide to run on the cluster.

zanetworker: Knowing how the CP behaves understress, configuring it with the right reqs/limits (which we do), providing the right quotas,... is something we can do for workloads we are familiar with and have properly tested, unlike just any other workloads which might not be properly configured and yields more surprises.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
zanetworker: To avoid charing them towards cores of a subscription.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,37 @@
...
-= Distributing hosted cluster workloads (Technology Preview)
+= Distributing hosted cluster workloads
+
+Before you get started with hosted control planes for {ocp-short}, you must properly label nodes so that the pods of hosted clusters can be scheduled into infrastructure nodes. Node labeling is also important for the following reasons:
+
+* To ensure high availability and proper workload deployment
+* To ensure that control plane workloads are configured at one of the following multi-tenancy distribution levels:
+** Everything shared: Control planes for hosted clusters can run on any node that is designated for control planes.
+** Request serving isolation: Serving pods are requested in their own dedicated nodes.
+** Nothing shared: Every control plane has its own dedicated nodes.
```
zanetworker: Request serving/nothing is shared can be confusing for self-managed. I would not add them before we have made sure we have that as a supported configuration in self-managed.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,37 @@
...
+
+[#hosted-cluster-labels-taints-overview]
+== Labels and taints for management cluster nodes
 
 As a management cluster administrator, you can use the following labels and taints in the management cluster nodes to schedule the control plane workload:
 
-* `hypershift.openshift.io/control-plane: true`
-* `hypershift.openshift.io/cluster: ${HostedControlPlane Namespace}`
+* `hypershift.openshift.io/control-plane: true`: Use this label and taint to dedicate a node to running hosted control plane workloads. Setting this label and taint is required if you do not want other workloads on the management cluster to run in your control plane nodes.
+* `node-role.kubernetes.io/infra`: Use this label to avoid having a node count toward your subscription.
```
zanetworker: I think we can try to call this out earlier.
```suggestion
* `node-role.kubernetes.io/infra`: Use this label to avoid having the control-plane workload count toward your subscription.
```

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,37 @@
...
+* `topology.kubernetes.io/zone`: Use this label to deploy highly available clusters. The zone might be a location, rack name, or the hostname of the node where the zone is set.
```
zanetworker: Makes sense. 

It would be great if we can summon an example and add it here. @enxebre @csrwng can you help with that?

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,37 @@
...
+*Important:* Do not use the management cluster for your workload. Workloads must not run on nodes where control planes run.
+
+[#hosted-cluster-labels-taints-overview]
+== Labels and taints for management cluster nodes
 
 As a management cluster administrator, you can use the following labels and taints in the management cluster nodes to schedule the control plane workload:
 
-* `hypershift.openshift.io/control-plane: true`
-* `hypershift.openshift.io/cluster: ${HostedControlPlane Namespace}`
+* `hypershift.openshift.io/control-plane: true`: Use this label and taint to dedicate a node to running hosted control plane workloads. Setting this label and taint is required if you do not want other workloads on the management cluster to run in your control plane nodes.
```
zanetworker: We should then try to articulate that we need to label the nodes that will host the control-plane pods by two things: 

- `hypershift.openshift.io/control-plane`: true to avoid sharing those nodes with others components (e.g., the management cluster's infra components OR any other mistakenly deployed workload).
- `node-role.kubernetes.io/infra`

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,37 @@
...
+* `node-role.kubernetes.io/infra`: Use this label to avoid having a node count toward your subscription.
+* `topology.kubernetes.io/zone`: Use this label to deploy highly available clusters. The zone might be a location, rack name, or the hostname of the node where the zone is set.
+* `hypershift.openshift.io/cluster: ${HostedControlPlane Namespace}`: Use this label and taint when you want to dedicate a node to a single hosted cluster.
 
 Pods for a hosted cluster have tolerations, and the scheduler uses affinity rules to schedule them. Pods tolerate taints for `control-plane` and the `cluster` for the pods. The scheduler prioritizes the scheduling of pods into nodes that are labeled with `hypershift.openshift.io/control-plane` and `hypershift.openshift.io/cluster: ${HostedControlPlane Namespace}`.
 
 For the `ControllerAvailabilityPolicy` option, use `HighlyAvailable`. When you use that option, you can schedule pods for each deployment within a hosted cluster across different failure domains by setting `topology.kubernetes.io/zone` as the topology key.
```
zanetworker: We need to call out that `HighleyAavilable` is the default and what is the `HCP` CLI deploys. Non Highley Available control-planes are not supported beyond testing.

clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -35,11 +35,19 @@ See the following highly available hosted control plane requirements, which were
...
+== Prerequisites
+
+To use hosted control planes for {ocp-short}, you must ensure that your nodes are properly labeled. For more information, see xref:../hosted_control_planes/hosted-cluster-workload-distributing.adoc#hosted-cluster-workload-distributing[Distributing hosted cluster workloads].
+
+[#hosted-control-planes-get-started]
+== Getting started with hosted control planes
+
+To get started with hosted control planes, read the topics related to the platform that you plan to use:
 
 * xref:../hosted_control_planes/configure_hosted_aws.adoc#hosting-service-cluster-configure-aws[Configuring the hosting cluster on AWS (Technology Preview)]
```
zanetworker, line 48: Should we start to use `management cluster` as the the term to describe that hosted cluster in all occurances?

zanetworker (commented):
Thank you @lahinson, added some comments.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
lahinson: Commented out for now.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
lahinson: I added a mention of this to the first bullet point in the list that starts on line 6.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
lahinson: Agreed!

clusters/hosted_control_planes/hosted_intro.adoc
lahinson: @zanetworker I would really love to settle on one, consistently used term for _management cluster_/_hosting cluster._ The ACM team uses "hosting cluster" consistently, and the OCP team uses "management cluster" consistently, and frankly, it's been confusing for me, so I'm sure it will be for customers, too. I don't have a strong preference for which term to use. Perhaps this is a good topic to bring up in the HyperShift office hours?

zanetworker: We use management cluster in our upstream so I would opt for `management` to change in one place instead of two :)

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
lahinson: I updated the text as follows:
>For the `ControllerAvailabilityPolicy` option, use `HighlyAvailable`, which is the default value that the hosted control planes command line interface (`hcp`) deploys. When you use that option, you can schedule pods for each deployment within a hosted cluster across different failure domains by setting `topology.kubernetes.io/zone` as the topology key. Control planes that are not highly available are not supported.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
lahinson: I made some updates to the text on lines 18 - 26. PTAL.

clusters/hosted_control_planes/hosted_intro.adoc
lahinson: IIRC, I believe that ACM prefers the use of "hosting cluster" because "management cluster" sounds too similar to "managed cluster", which is the cluster that the hub manages in ACM.

@rokej When you have a moment, I'd appreciate your feedback. We'd like to settle on either "management cluster" or "hosting cluster" and use one term consistently. Currently, the ACM docs use "hosting cluster" more often, but the OCP docs use "management cluster". As Adel mentioned above, the HyperShift OCP team prefers "management cluster" because that's the term that is used upstream. Do you think it would be problematic to use "management cluster" consistently instead of "hosting cluster"?

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
csrwng: Would this be sufficient?

Given a management cluster with nodes: `worker-1a`, `worker-1b`, `worker-2a`, `worker-2b` where `worker1-a` and `worker-1b` live in `rack1`; and `worker-2a` and `worker-2b` live in `rack2`, use each rack as an availability zone:

```
oc label node/worker-1a node/worker-1b topology.kubernetes.io/zone=rack1
oc label node/worker-2a node/worker-2b topology.kubernetes.io/zone=rack2
```

lahinson: Works for me. Thank you!

lahinson: I added Cesar's example.

clusters/hosted_control_planes/hosted_intro.adoc
rokej: I agree with @lahinson that `management cluster` sounds similar to MCE's `managed cluster`. I prefer to use the term `hosting cluster` because I am sure MCE/ACM users will get confused between `managed cluster` and `management cluster`. In MCE context, a `managed cluster` becomes a `hosting cluster` when the hypershift addon is enabled to install the hypershift operator in that cluster. I think it will be confusing to say a `managed cluster` becomes a `management cluster`.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,46 @@
...
-= Distributing hosted cluster workloads (Technology Preview)
+= Distributing hosted cluster workloads
 
-As a management cluster administrator, you can use the following labels and taints in the management cluster nodes to schedule the control plane workload:
+Before you get started with hosted control planes for {ocp-short}, you must properly label nodes so that the pods of hosted clusters can be scheduled into infrastructure nodes. Node labeling is also important for the following reasons:
 
-* `hypershift.openshift.io/control-plane: true`
-* `hypershift.openshift.io/cluster: ${HostedControlPlane Namespace}`
+* To ensure high availability and proper workload deployment. For example, you can set the `node-role.kubernetes.io/infra` label to avoid having the control-plane workload count toward your {ocp-short} subscription.
+* To ensure that control plane workloads are configured at the "Everything shared" multi-tenancy distribution level, where control planes for hosted clusters can run on any node that is designated for control planes
```
zanetworker: I would remove this completely. We haven't defined "Everything Shared" anywhere, and we are not explaining it enough (makes sense not to because its only relevant for request serving components in a different RH context).

zanetworker (commented):
Tiny nit, otherwise LGTM!

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
lahinson: @zanetworker Removed ✅

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,47 @@
...
 
-* `hypershift.openshift.io/control-plane: true`
-* `hypershift.openshift.io/cluster: ${HostedControlPlane Namespace}`
+* To ensure high availability and proper workload deployment. For example, you can set the `node-role.kubernetes.io/infra` label to avoid having the control-plane workload count toward your {ocp-short} subscription.
+//lahinson - sept. 2023 - commenting out the following lines until those levels are supported for self-managed hypershift
+//* To ensure that control plane workloads are configured at one of the following multi-tenancy distribution levels:
+//** Everything shared: Control planes for hosted clusters can run on any node that is designated for control planes.
+//** Request serving isolation: Serving pods are requested in their own dedicated nodes.
+//** Nothing shared: Every control plane has its own dedicated nodes. 
+* To ensure that control plane workloads are separate from other workloads in the management cluster
```
jc-berger: missing period at the end. I believe this bullet point on line 12 is _not_ commented out, therefore, please include a period at the end.

clusters/hosted_control_planes/hosted-cluster-workload-distributing.adoc
```diff
@@ -1,16 +1,47 @@
...
+
+*Important:* Do not use the management cluster for your workload. Workloads must not run on nodes where control planes run.
+
+[#hosted-cluster-labels-taints-overview]
+== Labels and taints for management cluster nodes
+
+As a management cluster administrator, you can use the following labels and taints in management cluster nodes to schedule the control plane workload:
```
jc-berger: Optional:

For minimalism purposes, you can remove `you can` especially since you already specified who the user is. Also, can we say, `your control plan workload` or does the user not own this workload? So it'd be: 

As a management cluster administrator, use the following labels and taints in management cluster nodes to schedule your control plane workload:

lahinson: I like your revision! I'm not sure if it's entirely correct to refer to the control plane workload as "your control plane workload", although I like your intention of being more engaging with the reader. I think changing it to "a control plane workload" might work.

PR #5446, opened 2023-08-28
create-via annotation should be set to hypershift for hosted cluster …

rokej:
This is for story https://issues.redhat.com/browse/ACM-6547 about setting the `open-cluster-management/created-via: hypershift` annotation in the ManagedCluster CR to import a hosted cluster.


lahinson (approved):
lgtm

PR #5448, opened 2023-08-28
[ACM-6751] Updated uninstall topic for MCH

dislbenn:
Removing deprecated resources from uninstall topic.

### DIFF
--- a/install/uninstall.adoc
+++ b/install/uninstall.adoc
@@ -105,14 +105,12 @@ oc get mch -o yaml
 ACM_NAMESPACE=<namespace>
 oc delete mch --all -n $ACM_NAMESPACE
 helm ls --namespace $ACM_NAMESPACE | cut -f 1 | tail -n +2 | xargs -n 1 helm delete --namespace $ACM_NAMESPACE
-oc delete apiservice v1beta2.webhook.certmanager.k8s.io v1.admission.cluster.open-cluster-management.io v1.admission.work.open-cluster-management.io
+oc delete apiservice v1.admission.cluster.open-cluster-management.io v1.admission.work.open-cluster-management.io
 oc delete clusterimageset --all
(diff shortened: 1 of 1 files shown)


jc-berger (approved):
/lgtm

jc-berger:
For Jira Issue 6751: https://issues.redhat.com/browse/ACM-6751 

Specifically, modifying and updating script in step 4 of the document. See the current version here: https://github.com/stolostron/rhacm-docs/blob/2.9_stage/install/uninstall.adoc#removing-resources-by-using-commands

install/uninstall.adoc
```diff
@@ -105,17 +105,12 @@ oc get mch -o yaml
...
+oc delete crd klusterletaddonconfigs.agent.open-cluster-management.io placementbindings.policy.open-cluster-management.io policies.policy.open-cluster-management.io userpreferences.console.open-cluster-management.io discoveredclusters.discovery.open-cluster-management.io discoveryconfigs.discovery.open-cluster-management.io
+oc delete mutatingwebhookconfiguration ocm-mutating-webhook managedclustermutators.admission.cluster.open-cluster-management.io multicluster-observability-operator
+oc delete validatingwebhookconfiguration channels.apps.open.cluster.management.webhook.validator application-webhook-validator multiclusterhub-operator-validating-webhook ocm-validating-webhook multicluster-observability-operator multiclusterengines.multicluster.openshift.io
 ----
 +
 Replace `<namespace>` in the script with the name of the namespace where {product-title-short} was installed. Ensure that you specify the correct namespace, as the namespace is cleaned out and deleted.
```
cadawson: Something not related to your change here: Can we make this an *Important:* note about deleting the namespace? The concern is that if they delete the wrong namespace, they might lose data. I ran it by Mikela, and she agreed that moving it to the following line and adding the bold important callout would be beneficial. Thanks!

cadawson:
```suggestion
Replace `<namespace>` in the script with the name of the namespace where {product-title-short} was installed. 

*Important:* Ensure that you specify the correct namespace, as the namespace is cleaned out and deleted. 
```

cadawson (approved):
Your updates look good. Added a note and an update to some other content. I don't need to review again. Thanks!

install/uninstall.adoc
dislbenn: @cadawson, yes I agree with that suggestion. I will make those changes.

dislbenn: @cadawson @jc-berger changes has been made.

jc-berger (approved):
/lgtm

PR #5466, opened 2023-09-05
No Jira: Restructuring bare metal docs for HyperShift

lahinson:
This PR breaks the managing_hosted_bm.adoc file, which was getting very long, into multiple files. Very little content was changed, so a thorough peer review for style, terminology, etc. is not needed.

To make the review easier, no content was changed in the following files -- the content was simply moved from sections within one file to separate files:
- destroying_hosted_bare_metal.adoc (Destroying a hosted cluster on bare metal)
- handling_ingress_hosted.adoc (Handling Ingress in a hosted cluster on bare metal)
- scaling_nodepool_hosted.adoc (Scaling the NodePool object for a hosted cluster)


clusters/hosted_control_planes/handling_ingress_hosted.adoc
```diff
@@ -0,0 +1,213 @@
...
+
++
+.. Create the resources by entering the following command:
+
++
+----
+oc apply -f ipaddresspool-bgpadvertisement-config.yaml
+----
+
+. Expose the {ocp-short} Router via MetalLB by following these steps:
```
cadawson:
```suggestion
. Expose the {ocp-short} Router through MetalLB by following these steps:
```

cadawson (approved):
One suggestion, as I don't think we use "via". Let me know if that is not correct. Otherwise, looks good. Thanks!

clusters/hosted_control_planes/handling_ingress_hosted.adoc
lahinson: Good catch. I can make that update. Thanks!

PR #5489, opened 2023-09-08
Fixing links and reorganizing manual import topics for hosted control planes

lahinson:
This PR reverts a link back to an xref and moves the "Manually importing a hosted control plane cluster" out of the AWS book because that procedure is common across platforms. I also moved the file lower in the ToC structure because it shouldn't be a task that users typically need to do, as hosted clusters are auto-imported as of MCE 2.4.


dockerymick (approved):
All links work in this PR, thanks for the updates!

PR #5532, opened 2023-09-20
added more content for argo cd push model

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-7009


dockerymick (commented):
@jc-berger it seems like there will be a few new files added for this PR. I recommend that you create a new file name cluster_permissions_intro.adoc. Add the background info there and then direct users to other new files cluster_permission_policy.adoc

The end of the Google doc says, "We are now ready to deploy apps to the managed clusters using our new managed-sa-sample service account and the permissions that we granted it." 

Which tasks are a part of the objective to Deploy applications by using the ManagedServiceAccount add-on?

gitops/gitops_push_pull.adoc
```diff
@@ -81,6 +81,10 @@ For both Push and Pull model, the _Argo CD ApplicationSet controller_ on the hub
 
 - By default, the Push model is used to deploy the application unless you add the `apps.open-cluster-management.io/ocm-managed-cluster` and `apps.open-cluster-management.io/pull-to-ocm-managed-cluster` annotations to the template section of the `ApplicationSet`.
 
+- The Argo CD controller pushes resources from the centralized hub cluster to the managed cluster. For every managed cluster, there is an application manager add-on that runs on the managed cluster. By default, the Argo CD controller uses the service account application manager to push these resources to the managed cluster. 
+
+- Now, you can specify the different service account you want the Argo CD push model to use, based on your individual developer needs. Therefore, when the Argo CD  controller pushes resources from the hub to the managed cluster, you can use a different service account than the default application manager.
```
philipwu08: Instead of "Now", should be specify the actual release number, like ACM 2.9. In ACM 2.10, "now" gives the impression the feature is new.

philipwu08: "you can specify the different service account" -> "you can specify a different service account"

philipwu08: ", based on your individual developer needs" 

Typically, I can it's based on the company's needs but not the individual developer. Might be better to just take this part out though.

philipwu08: "Therefore, when the Argo CD  controller pushes resources from the hub to the managed cluster, you can use a different service account than the default application manager."

Maybe too push duplication here? Maybe a point on why they would want to use a different service account? Like: By using a different service account, you have full control over the permissions that is granted to this service account.

That is a good entry to talk about cluster permissions. Maybe add a sentence to introduce cluster permissions and how it could be used to customize the permissions for the service account, like:

The service account must exist on the managed cluster. You may want to use the managedserviceaccount resource, and the new clusterpermission resource, on the hub to facilitate with the creation of the service account with the associated permissions.

gitops/gitops_push_pull.adoc
```diff
@@ -81,6 +81,10 @@ For both Push and Pull model, the _Argo CD ApplicationSet controller_ on the hub
 
 - By default, the Push model is used to deploy the application unless you add the `apps.open-cluster-management.io/ocm-managed-cluster` and `apps.open-cluster-management.io/pull-to-ocm-managed-cluster` annotations to the template section of the `ApplicationSet`.
 
+- The Argo CD controller pushes resources from the centralized hub cluster to the managed cluster. For every managed cluster, there is an application manager add-on that runs on the managed cluster. By default, the Argo CD controller uses the service account application manager to push these resources to the managed cluster.
```
philipwu08: "The Argo CD controller pushes resources from the centralized hub cluster to the managed cluster"

Maybe this this out? It's seem to be a duplicate of the first bullet.

philipwu08: We may want to add a note on why the customer would want to use a different service account, like:

The application manager service account has a broad set of permissions since it’s used by the application subscription addon to deploy applications on the managed cluster. For some users, this may not be desirable, and prefer to use a service account with a limited set of permissions.

xiangjingli: `The Argo CD controller` should be `The ArgoCD application controller` specifically.

dockerymick: Based on the Argo CD doc, https://argo-cd.readthedocs.io/en/stable/, please keep my space between Argo and CD `Argo CD application controller`

gitops/gitops_push_pull.adoc
```diff
@@ -81,6 +81,14 @@ For both Push and Pull model, the _Argo CD ApplicationSet controller_ on the hub
 
 - By default, the Push model is used to deploy the application unless you add the `apps.open-cluster-management.io/ocm-managed-cluster` and `apps.open-cluster-management.io/pull-to-ocm-managed-cluster` annotations to the template section of the `ApplicationSet`.
 
+- For every managed cluster, there is an application manager add-on that runs on the managed cluster. By default, the Argo CD controller uses the service account application manager to push these resources to the managed cluster.
```
swopebe: For every managed cluster, there is an application manager add-on that runs on the managed cluster.

Recommending concise:

An application manager add-on runs on each managed cluster. (No need to say "there is an" and I don't think we need to say managed cluster twice)

gitops/gitops_push_pull.adoc
```diff
@@ -81,6 +81,14 @@ For both Push and Pull model, the _Argo CD ApplicationSet controller_ on the hub
 
 - By default, the Push model is used to deploy the application unless you add the `apps.open-cluster-management.io/ocm-managed-cluster` and `apps.open-cluster-management.io/pull-to-ocm-managed-cluster` annotations to the template section of the `ApplicationSet`.
 
+- For every managed cluster, there is an application manager add-on that runs on the managed cluster. By default, the Argo CD controller uses the service account application manager to push these resources to the managed cluster. 
+
+- The application manager service account has a broad set of permissions because the application subscription add-on uses the application manager service to deploy applications on the managed cluster. If you prefer to use a service account with a limited set of permissions, do not use the application manager service account.
```
swopebe: The application manager service account uses a large set of permissions because the application subscription add-on uses the application manager service to deploy applications on the managed cluster. (concise from: has a broad set of, changes word usage and tone)

Do not use the application manager service account if you want a limited set of permissions. (imperative mood is a verb form used to make a demand, issue a warning, or give advice or instructions, better than "preference" and almost always the best choice for tech doc.)

gitops/gitops_push_pull.adoc
```diff
@@ -81,6 +81,14 @@ For both Push and Pull model, the _Argo CD ApplicationSet controller_ on the hub
...
+
+- You can specify a different service account you want the Argo CD push model to use. Therefore, when the Argo CD controller pushes resources from the centralized hub cluster to the managed cluster, you can use a different service account than the default application manager. By using a different service account, you have full control over the permissions that are granted to this service account.
```
swopebe: that you want the Argo CD...

Please take out Therefore, not needed.

please change: have full control over to a simple "can"

install/adv_config_install.adoc
```diff
@@ -259,7 +259,7 @@ oc patch MultiClusterHub multiclusterhub -n <namespace> --type=json -p='[{"op":
 
 [#serviceaccount-addon-intro]
 == ManagedServiceAccount add-on (Technology Preview)
-
+//should this be Technology Preview? This is inconsistent with what is doc'ed in install_upgrade/adv_config_install.adoc. I think it isn't TP for 2.9, but we need to confirm
```
swopebe, line 262: Good catch, please check with Chris who I think wrote this.

swopebe: I see some inconsistency in naming (capital letters vs not). I wrote this to start and I think when it is not clear, the general rule is to do the same everywhere: 

Push model/push model
Pull model/pull model

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
+[#gitops-service-account-argo-cd]
+= Creating a customized service account with Argo CD Push model 
+
+When you use the default Argo CD cluster secret, you might encounter limitation like the following: 
+
+* The application subscription addon, or application manager, must be installed on your managed cluster because it creates a secret in the managed cluster namespace. This secret contains the access token for the application manager service account.
```
swopebe: we need to make sure we are writing add-on the same throughout the doc:

Preferred add-on noun
Write as shown: hyphenated.

swopebe: I also think we were writing Application manager (not sure, let's look into that)

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
+[#gitops-service-account-argo-cd]
+= Creating a customized service account with Argo CD Push model 
+
+When you use the default Argo CD cluster secret, you might encounter limitation like the following: 
+
+* The application subscription addon, or application manager, must be installed on your managed cluster because it creates a secret in the managed cluster namespace. This secret contains the access token for the application manager service account. 
+* You need a cluster admin role service account for your application manager service account. The application manager service account has a broad set of permissions since the account is used by the application subscription addon to deploy applications on the managed cluster.
```
swopebe: Same here, please rewrite for conciseness, but why do we say the same thing twice? (line 7)

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+When you use the default Argo CD cluster secret, you might encounter limitation like the following: 
+
+* The application subscription addon, or application manager, must be installed on your managed cluster because it creates a secret in the managed cluster namespace. This secret contains the access token for the application manager service account. 
+* You need a cluster admin role service account for your application manager service account. The application manager service account has a broad set of permissions since the account is used by the application subscription addon to deploy applications on the managed cluster. 
+
+To work around these limitations, use a customized service account on the managed clusters. Create a service account on a managed cluster by creating a `managedserviceaccount` resource on the hub. Then, use the `clusterpermission` resource to grant specific permissions to the service account.
+
+By granting cluster permissions to your managed service accounts ,you can deploy applications across managed clusters. 
+
+Complete the following procedures in the following order:
```
swopebe: we don't ever say "in order" as that is a given with the task.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
+[#gitops-service-account-argo-cd]
+= Creating a customized service account with Argo CD Push model 
+
+When you use the default Argo CD cluster secret, you might encounter limitation like the following: 
+
+* The application subscription addon, or application manager, must be installed on your managed cluster because it creates a secret in the managed cluster namespace. This secret contains the access token for the application manager service account. 
+* You need a cluster admin role service account for your application manager service account. The application manager service account has a broad set of permissions since the account is used by the application subscription addon to deploy applications on the managed cluster. 
+
+To work around these limitations, use a customized service account on the managed clusters. Create a service account on a managed cluster by creating a `managedserviceaccount` resource on the hub. Then, use the `clusterpermission` resource to grant specific permissions to the service account.
```
swopebe: what limitations? We don't usually discuss limitations in true procedures...we put those in the release notes. It seems this is a requirement? Simply: (please write out "hub cluster")

Create a service account on a managed cluster by creating a `managedserviceaccount` resource on the hub cluster. Then, use the `clusterpermission` resource to grant specific permissions to the service account.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+
+By granting cluster permissions to your managed service accounts ,you can deploy applications across managed clusters.
```
swopebe: space

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+* The application subscription addon, or application manager, must be installed on your managed cluster because it creates a secret in the managed cluster namespace. This secret contains the access token for the application manager service account. 
+* You need a cluster admin role service account for your application manager service account. The application manager service account has a broad set of permissions since the account is used by the application subscription addon to deploy applications on the managed cluster. 
+
+To work around these limitations, use a customized service account on the managed clusters. Create a service account on a managed cluster by creating a `managedserviceaccount` resource on the hub. Then, use the `clusterpermission` resource to grant specific permissions to the service account.
+
+By granting cluster permissions to your managed service accounts ,you can deploy applications across managed clusters. 
+
+Complete the following procedures in the following order:
+
+. <<Creating a managed service account>>
```
swopebe, line 15: <<Creating a managed service account, Creating a managed service account >> is how you build an anchor and our rule is if we have three or more sections, so this is great.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+    - enabled: true     =====> set enabled=true
+      name: managedserviceaccount-preview
+    ...
+----
+
+.Procedure
+
+. Go to the hub cluster > `managed cluster` > `cluster1`. 
+. To install the `managedserviceaccount` addon on `cluster1`, apply the following YAML:
+
```
swopebe: There is usually a command to apply, as we would not just assume they have it?

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+apiVersion: authentication.open-cluster-management.io/v1alpha1
+kind: ManagedServiceAccount
+metadata:
+  name: managed-sa-sample
+  namespace: cluster1
+spec:
+  rotation: {}
+----
+
+.Verification
```
swopebe: we don't usually use these ^^^ What is the benefit of this?

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+% oc get secrets -n cluster1 managed-sa-sample
+----
+
+== Creating a cluster permission 
+
+To grant permissions to the new service account, use the `clusterpermission` resource. The `clusterpermission` resource helps you deploy your applications. 
+
+.Prerequisites 
+
+* You have completed <<Creating a managed service account>>
```
swopebe: We don't need this, as it is understood in the procedure order.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+      apiGroup: authentication.open-cluster-management.io
+      kind: ManagedServiceAccount
+      name: managed-sa-sample
+  clusterRoleBinding:
+    subject:
+      apiGroup: authentication.open-cluster-management.io
+      kind: ManagedServiceAccount
+      name: managed-sa-sample
+----
+
```
swopebe, line 88: Usually there is a command in these examples with two steps. 

Create the YAML
Apply the file

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+. Create a NC  `mortage` if one does not already exist.
```
swopebe: NC?

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+    argoNamespace: openshift-gitops
+    cluster: local-cluster
+  managedServiceAccountRef: managed-sa-sample
+  placementRef:
+    apiVersion: cluster.open-cluster-management.io/v1beta1
+    kind: Placement
+    name: aws-app-placement
+----
+
+. To create a GitOpsCluster CR, run the following full YAML:
```
swopebe: We spell out CR usually.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+== Creating an Argo CD application 
+
+Use your cluster secret to deliver an Argo CD application from the Argo CD UI. The Argo CD application is delivered by the managed service account, `managed-sa-sample`. 
+
+.Prerequisites 
+
+. You have completed <<Creating a managed service account>> 
+. You have completed <<Creating a cluster permission>> 
+. You have completed <<Creating a managed service account in GitOpsCluster>>
+
```
swopebe, line 152: we don't need this

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+.Procedure 
+
+. Log into the Argo CD UI.
```
swopebe: Console, instead of UI

swopebe (changes requested):

swopebe:
Some changes needed and some recommended, as well as link to voice and tone guidance. I think this needs to be structured into multiple files within the GitOps guide, not just for easier digest, but easier organization.

gitops/gitops_service_account_argo_cd.adoc
jc-berger: Yes, I've made it `add-on` 

For now, I'm keeping it `application manager` because it's lowercased in our `gitops_push-pull.adoc` so I think lowercase in this doc would at least be consistent with other docs within the gitops folder. 

What do you think?

gitops/gitops_service_account_argo_cd.adoc
jc-berger: Philip, I didn't see a command in the Gdoc template, do you know if we have one?

gitops/gitops_service_account_argo_cd.adoc
jc-berger: I've removed them :)

gitops/gitops_service_account_argo_cd.adoc
jc-berger: Philip, again, I didn't see a command in the Gdoc template. Am I overlooking anything? Do you know if we have one to use here like Brandi suggests?

gitops/gitops_service_account_argo_cd.adoc
jc-berger: Philip, can `NC` be removed? Or does it stand for something and determines what kind of mortage the user will create?

gitops/gitops_service_account_argo_cd.adoc
jc-berger: I've removed these bullets all together. 

What do you think? 

Philip and/or SMEs, is there important meaning loss by removing the content that I did?

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
+
+.  To verify the `managed-serviceaccount` add-on is running on the managed cluster, `cluster1`, run the following command: 
+
++
+[source,yaml]
+----
+% oc get pods -n open-cluster-management-agent-addon
+managed-serviceaccount-addon-agent-bb9799fbc-dwxzp   1/1     Running   0             47s
+----
+
```
xiangjingli: @philipwu08  @fmt47  I remember foundation has changed the way for deploying managed service account component.  Could you double check it and provide the new way if this is not working any more

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
+    subject:
+      apiGroup: authentication.open-cluster-management.io
+      kind: ManagedServiceAccount
+      name: managed-sa-sample
+----
+
+. Create a NC  `mortage` if one does not already exist. 
+. Review the resources that are created on the `cluster1`. 
+
+== Creating a managed service account in GitOpsCluster
```
xiangjingli: The title may be a bit confusing. The below description is accurate
```
Use the GitOpsCluster to create the Argo CD cluster secret using the `managed-sa-sample` service account.
```

gitops/gitops_push_pull.adoc
```diff
@@ -81,6 +81,14 @@ For both Push and Pull model, the _Argo CD ApplicationSet controller_ on the hub
 
 - By default, the Push model is used to deploy the application unless you add the `apps.open-cluster-management.io/ocm-managed-cluster` and `apps.open-cluster-management.io/pull-to-ocm-managed-cluster` annotations to the template section of the `ApplicationSet`.
 
+- An application manager add-on runs on each managed cluster. By default, the Argo CD controller uses the service account application manager to push these resources to the managed cluster. 
+
+- The application manager service account uses a large set of permissions because the application subscription add-on uses the application manager service to deploy applications on the managed cluster. Do not use the application manager service account if you want a limited set of permissions.
```
philipwu08: "The application manager service account uses a large set"

Instead of "uses", maybe use "has" or "is granted"

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
+[#gitops-service-account-argo-cd]
+= Creating a customized service account with Argo CD
```
philipwu08: "Creating a customized service account with Argo CD"

"With" seems to give the impression that ArgoCD is used to create the service account.

I think it's better either:
"Using a customized service account with the Argo CD push model" or "Creating a customized service account for use with the Argo CD push model"

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
+[#gitops-service-account-argo-cd]
+= Creating a customized service account with Argo CD 
+
+Create a service account on a managed cluster by creating a `managedserviceaccount` resource on the hub cluster. Use the `clusterpermission` resource to grant specific permissions to the service account.
+
+By granting cluster permissions to your managed service accounts, you can deploy applications across managed clusters.
```
philipwu08: "you can deploy applications across managed clusters.", 

maybe more specific? ==>

"it will have the necessary permission to deploy the application resources on the managed clusters".

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
+
+Create a service account on a managed cluster by creating a `managedserviceaccount` resource on the hub cluster. Use the `clusterpermission` resource to grant specific permissions to the service account.
+
+By granting cluster permissions to your managed service accounts, you can deploy applications across managed clusters. 
+
+Complete the following procedures:
+
+. <<Creating a managed service account>> 
+. <<Creating a cluster permission>> 
+. <<Creating a managed service account in GitOpsCluster>>
```
philipwu08: <<Creating a managed service account in GitOpsCluster>> 

We're not creating the managed service account in the gitopscluster resource. We're "using" or "referencing" the service account in the gitopscluster resource.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
+Complete the following procedures:
+
+. <<Creating a managed service account>> 
+. <<Creating a cluster permission>> 
+. <<Creating a managed service account in GitOpsCluster>> 
+. <<Creating an Argo CD application>> 
+
+== Creating a managed service account 
+
+Create a managed service account so you can have cluster permissions for your other service accounts. When you create a managed service account, you simultaneously create a secret with the same name as the managed service account. This secret is in the cluster namespace for the hub cluster and contains the access token for the `managed-sa-sample` service account that is used in the Argo CD cluster secret.
```
philipwu08: "Create a managed service account so you can have cluster permissions for your other service accounts."

Not sure if that's correct.  There must be existing doc in ACM that talks about managed service accounts already that we could reference: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/multicluster_engine/managed-serviceaccount-addon#doc-wrapper

Basically, what we want to say here is something like:
The ManagedServiceAccount custom resource on the hub provides a convenient way to create ServiceAccounts on the managed clusters; When a ManagedServiceAccount custom resource is created in the <managed_cluster> namespace on the hub cluster, a ServiceAccount is created on the managed cluster.

philipwu08: "When you create a managed service account, you simultaneously create a secret with the same name as the managed service account. This secret is in the cluster namespace for the hub cluster and contains the access token for the `managed-sa-sample` service account that is used in the Argo CD cluster secret."

This is implementation detail. I think we could omit this. Check with the team though.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
+
+. Ensure the `managedserviceaccount-preview` component is enabled by running, `oc edit mce multiclusterengine`.
```
philipwu08: Is it better to reference the managed service account doc here:
https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/multicluster_engine/managed-serviceaccount-addon#doc-wrapper

The doc above talks about enabling the ManagedClusterAddon for the managed-serviceaccount. It does not talk about enabling the managedserviceaccount-preview feature in MCE. Either case, I think we should reference the doc above on how to enable managed service account and create a managed service account.

We should ask the foundation team if the above doc is correct - is it missing the section to enable the managedserviceaccount-preview feature in MCE.

gitops/gitops_service_account_argo_cd.adoc
philipwu08: We could provide an `oc patch` command to make it simplier, but might be better to take this out and reference the existing managed service account doc.

gitops/gitops_service_account_argo_cd.adoc
philipwu08: @xiangjingli I could verify but I think this is another reason to reference the managed service account doc: https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/multicluster_engine/managed-serviceaccount-addon#doc-wrapper

This way we're not out-of-sync if there are changes in the future.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
+. To verify that the `managed-sa-sample` service account is created on `cluster1`, run the following command:  
+
++
+[source,yaml]
+----
+// go to cluster1
+% oc get sa -n open-cluster-management-agent-addon managed-sa-sample
+----
+
+. To access the secret for your managed service account, run the following command:
```
philipwu08: "To access the secret for your managed service account, run the following command: "

I think we can skip this. This is detail I don't think the user needs know about.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
++
+[source,yaml]
+----
+// go to the hub cluster
+% oc get secrets -n cluster1 managed-sa-sample
+----
+
+== Creating a cluster permission 
+
+To grant permissions to the new service account, use the `clusterpermission` resource. The `clusterpermission` resource helps you deploy your applications.
```
philipwu08: Maybe add in the beginning:

When the service account is created, is does not have any permission associated to it. To grant permissions ...

philipwu08: "The `clusterpermission` resource helps you deploy your applications. "

clusterpermission does not help you deploy your application, it's basically a way to grant permission to your service account, so maybe something like this?

The clusterpermission resource is created in the managed cluster namespace on the hub. It provides a convenient way to create role, clusterrole resources on the managed clusters, and bind them to a service account through a
rolebinding or clusterrolebinding resource.

gitops/gitops_service_account_argo_cd.adoc
philipwu08: You can wrap any yaml with the oc command, like this:
oc apply -f - <<EOF

yaml content in the middle

EOF

For example:
oc apply -f - <<EOF
apiVersion: rbac.open-cluster-management.io/v1alpha1
kind: ClusterPermission
metadata:
  name: clusterpermission-msa-subject-sample
  namespace: cluster1
spec:
  roles:
  - namespace: default
    rules:
    - apiGroups: ["apps"]
      resources: ["deployments"]
      verbs: ["get", "list", "create", "update", "delete", "patch"]
    - apiGroups: [""]
      resources: ["configmaps", "secrets", "pods", "podtemplates", "persistentvolumeclaims", "persistentvolumes"]
      verbs: ["get", "update", "list", "create", "delete", "patch"]
    - apiGroups: ["storage.k8s.io"]
      resources: ["*"]
      verbs: ["list"]
  - namespace: mortgage
    rules:
    - apiGroups: ["apps"]
      resources: ["deployments"]
      verbs: ["get", "list", "create", "update", "delete", "patch"]
    - apiGroups: [""]
      resources: ["configmaps", "secrets", "pods", "services", "namespace"]
      verbs: ["get", "update", "list", "create", "delete", "patch"]
  clusterRole:
    rules:
    - apiGroups: ["*"]
      resources: ["*"]
      verbs: ["get", "list"]
  roleBindings:
  - namespace: default
    roleRef:
      kind: Role
    subject:
      apiGroup: authentication.open-cluster-management.io
      kind: ManagedServiceAccount
      name: managed-sa-sample
  - namespace: mortgage
    roleRef:
      kind: Role
    subject:
      apiGroup: authentication.open-cluster-management.io
      kind: ManagedServiceAccount
      name: managed-sa-sample
  clusterRoleBinding:
    subject:
      apiGroup: authentication.open-cluster-management.io
      kind: ManagedServiceAccount
      name: managed-sa-sample
EOF

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
+----
+// go to the hub cluster
+% oc get secrets -n cluster1 managed-sa-sample
+----
+
+== Creating a cluster permission 
+
+To grant permissions to the new service account, use the `clusterpermission` resource. The `clusterpermission` resource helps you deploy your applications. 
+
+. To grant the `managed-sa-sample` service account permission to deploy your sample application, apply the following YAML:
```
philipwu08: To grant the `managed-sa-sample` service account permission to deploy your sample application

Instead of "your sample application", I think we could be more specific here and say "the sample mortgage application to the mortgage namespace on cluster1", since the permission here is specific to that application

gitops/gitops_service_account_argo_cd.adoc
philipwu08: NC stands for namespace. So this needs a bit more explanation.

The sample clusterpermission above creates a role called clusterpermission-msa-subject-sample in the mortgage namespace.

gitops/gitops_service_account_argo_cd.adoc
philipwu08: Agree, see comment in line 12.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
+      apiGroup: authentication.open-cluster-management.io
+      kind: ManagedServiceAccount
+      name: managed-sa-sample
+----
+
+. Create a NC  `mortage` if one does not already exist. 
+. Review the resources that are created on the `cluster1`. 
+
+== Creating a managed service account in GitOpsCluster 
+
```
philipwu08, line 95: Maybe a little intro here about the gitopscluster resource first?

The gitopscluster resource uses placement to import selected managed clusters into Argo CD. The includes the creation of the Argo CD cluster secrets, which contains info used to access the clusters. By default, the Argo CD cluster secret uses the application manager service account to access the managed clusters.

You could also reference the doc here: https://github.com/stolostron/rhacm-docs/blob/2.9_stage/gitops/gitops_registering.adoc

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
+Use the GitOpsCluster to create the Argo CD cluster secret using the `managed-sa-sample` service account.
```
philipwu08: With the intro, we could remove this since it's not exactly correct.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
+
+. To update the GitOpsCluster resource to add the `managedServiceAccountRef`, run the following YAML:
```
philipwu08: To update the GitOpsCluster resource to use the managed service account , add the `managedServiceAccountRef`  property with the name of the managed service account, and run the following YAML.

The yaml could be wraped in the "oc apply -f - << EOF" to get the command

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
+
+== Creating a managed service account in GitOpsCluster 
+
+Use the GitOpsCluster to create the Argo CD cluster secret using the `managed-sa-sample` service account.
+
+. To update the GitOpsCluster resource to add the `managedServiceAccountRef`, run the following YAML: 
+
++
+[source,yaml]
+----
```
philipwu08: This gitopscluster resource yaml is not complete.

See doc: https://github.com/stolostron/rhacm-docs/blob/2.9_stage/gitops/gitops_registering.adoc

Actually, looks like we could delete this partial yaml completely. The YAML below contains the full YAML for the gitopscluster resource.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
+      labelSelector:
+        matchExpressions:
+        - key: name
+          operator: "In"
+          values:
+          - "cluster1"
+---
+
+
+apiVersion: apps.open-cluster-management.io/v1beta1
```
philipwu08, line 114: FYI, gitopslcuster resource is here already. we can take it out from line 170

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,247 @@
...
+[source,yaml]
+----
+% oc get secrets -n openshift-gitops cluster1-managed-sa-sample-cluster-secret    
+NAME                                        TYPE     DATA   AGE
+cluster1-managed-sa-sample-cluster-secret   Opaque   3      4m2s
+----
+
+== Creating an Argo CD application 
+
+Use your cluster secret to deliver an Argo CD application from the Argo CD UI. The Argo CD application is delivered by the managed service account, `managed-sa-sample`.
```
philipwu08: Is "deploy" better than "deliver"?

gitops/gitops_service_account_argo_cd.adoc
jc-berger: Please let me know if I applied the feedback correctly by adding the oc command to the yaml.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+. <<Creating a cluster permission>> 
+. <<Using a managed service account in the GitOpsCluster resource>> 
+. <<Creating an Argo CD application>> 
+. <<Using policy to create managed service accounts and cluster permissions>>
+
+== Creating a managed service account 
+
+The `managedserviceaccount` custom resource on the hub provides a convenient way to create `serviceaccounts` on the managed clusters. When a `managedserviceccount` custom resource is created in the `<managed_cluster>` namespace on the hub cluster, a `serviceccount` is created on the managed cluster.
+
+To create a manages service account, see https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/multicluster_engine/managed-serviceaccount-addon#doc-wrapper[Enabling manageserviceaccoutn add-ons]
```
jc-berger: Docs, please help me locate the proper file so I can link to it as an xref, not an external link.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+      apiGroup: authentication.open-cluster-management.io
+      kind: ManagedServiceAccount
+      name: managed-sa-sample
+  clusterRoleBinding:
+    subject:
+      apiGroup: authentication.open-cluster-management.io
+      kind: ManagedServiceAccount
+      name: managed-sa-sample
+----
+
```
philipwu08, line 89: Need an EOF here (at the end)

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+    subject:
+      apiGroup: authentication.open-cluster-management.io
+      kind: ManagedServiceAccount
+      name: managed-sa-sample
+  clusterRoleBinding:
+    subject:
+      apiGroup: authentication.open-cluster-management.io
+      kind: ManagedServiceAccount
+      name: managed-sa-sample
+----
```
philipwu08, line 87: Need an EOF here (at the end) 

Replace ----- with EOF

gitops/gitops_service_account_argo_cd.adoc
philipwu08: Missing EOF at the end, see comment above for location

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+
+. The sample `clusterpermission` creates a role called `clusterpermission-msa-subject-sample` in the mortgage namespace. If one does not already exist, create a namespace `mortage`.
```
philipwu08: namespace `mortage` => namespace called `mortgage`

Note the missing `g` in mortgage.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+      kind: ManagedServiceAccount
+      name: managed-sa-sample
+----
+
+. The sample `clusterpermission` creates a role called `clusterpermission-msa-subject-sample` in the mortgage namespace. If one does not already exist, create a namespace `mortage`. 
+. Review the resources that are created on the `cluster1`. 
+
+== Using a managed service account in the GitOpsCluster resource
+
+The GitOpsCluster resource uses placement to import selected managed clusters into the Argo CD push model, including the the creation of the Argo CD cluster secrets which contains information used to access the clusters. By default, the Argo CD cluster secret uses the application manager service account to access the managed clusters.
```
philipwu08: The GitOpsCluster resource uses placement to import selected managed clusters into the Argo CD push model

Remove "push model" at the end. We're importing the clusters into Argo CD only.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+
+. The sample `clusterpermission` creates a role called `clusterpermission-msa-subject-sample` in the mortgage namespace. If one does not already exist, create a namespace `mortage`. 
+. Review the resources that are created on the `cluster1`. 
+
+== Using a managed service account in the GitOpsCluster resource
+
+The GitOpsCluster resource uses placement to import selected managed clusters into the Argo CD push model, including the the creation of the Argo CD cluster secrets which contains information used to access the clusters. By default, the Argo CD cluster secret uses the application manager service account to access the managed clusters. 
+
+. To update the GitOpsCluster resource to use the managed service account , add the `managedServiceAccountRef` property with the name of the managed service account. 
+. To create a GitOpsCluster custom resource, run the following full YAML:
```
philipwu08: Insert empty line?

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+== Using a managed service account in the GitOpsCluster resource
+
+The GitOpsCluster resource uses placement to import selected managed clusters into the Argo CD push model, including the the creation of the Argo CD cluster secrets which contains information used to access the clusters. By default, the Argo CD cluster secret uses the application manager service account to access the managed clusters. 
+
+. To update the GitOpsCluster resource to use the managed service account , add the `managedServiceAccountRef` property with the name of the managed service account. 
+. To create a GitOpsCluster custom resource, run the following full YAML: 
+
++
+[source,yaml]
+----
```
philipwu08, line 112: Duplicate, we can remove this

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+---
```
philipwu08, line 113: Do we want to be consistent and wrap this YAML is "oc apply -f - <<EOF

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+    name: all-openshift-clusters
+    namespace: openshift-gitops
+----
+
+. Go to the `openshift-gitops` namespace and verify that there is a new Argo CD cluster secret with the name `cluster1-managed-sa-sample-cluster-secret`:
+
++
+[source,yaml]
+----
+% oc get secrets -n openshift-gitops cluster1-managed-sa-sample-cluster-secret
```
philipwu08: In the samples we used "cluster1" for the managed cluster. Is this ok or do we use some place holder like {managed cluster}?

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+
+To create a manages service account, see https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/multicluster_engine/managed-serviceaccount-addon#doc-wrapper[Enabling manageserviceaccoutn add-ons] 
+
+== Creating a cluster permission 
+
+When the service account is created, is does not have any permission associated to it. To grant permissions to the new service account, use the `clusterpermission` resource. The `clusterpermission` resource is created in the managed cluster namespace on the hub. It provides a convenient way to create roles, cluster roles resources on the managed clusters, and bind them to a service account through a `rolebinding` or `clusterrolebinding` resource.
+
+. To grant the `managed-sa-sample` service account permission to the sample mortgage application to the mortgage namespace on `cluster1`, apply the following YAML:
```
philipwu08: to the sample mortgage application to the mortgage namespace on `cluster1`

maybe 

to the sample mortgage application that is deployed to the mortgage namespace on `cluster1`

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+[source,yaml]
+----
+% oc get secrets -n openshift-gitops cluster1-managed-sa-sample-cluster-secret    
+NAME                                        TYPE     DATA   AGE
+cluster1-managed-sa-sample-cluster-secret   Opaque   3      4m2s
+----
+
+== Creating an Argo CD application 
+
+Use your cluster secret to deploy an Argo CD application from the Argo CD UI. The Argo CD application is deployed by the managed service account, `managed-sa-sample`.
```
philipwu08: Use your cluster secret to deploy an Argo CD application from the Argo CD UI.

The cluster secret is used automatically, so maybe just say:

Deploy an Argo CD application from the Argo CD UI using the push model.

philipwu08: The Argo CD application is deployed by the managed service account,

Maybe replace "by" with "with" or "using"?

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+Use your cluster secret to deploy an Argo CD application from the Argo CD UI. The Argo CD application is deployed by the managed service account, `managed-sa-sample`. 
+
+. Log into the Argo CD console. 
+. Click *Create a new application*. 
+. Choose the cluster URL. 
+. Go to your Argo CD application and verify that it has the given permissions, like roles and cluster roles, that you propagated to `cluster1`. 
+
+== Using policy to create managed service accounts and cluster permissions 
+
+Use a single policy to create the managed service account and cluster permission for all the managed clusters. When the GitOpsCluster resource is updated with the `managedServiceAccountRef`, each managed cluster in the placement of this GitOpsCluster needs to have the service account. If you have several managed clusters, it becomes tedious for you to create the managed service account and cluster permission for each managed cluster. Therefore, use a single policy.
```
philipwu08: Maybe take this out and add something to the end of the paragraph instead?

"Use a single policy to create the managed service account and cluster permission for all the managed clusters."

philipwu08: "Therefore, use a single policy."

Something like:

This could be simplified using a policy to create the managed service account and cluster permission for all the managed clusters.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+. Log into the Argo CD console. 
+. Click *Create a new application*. 
+. Choose the cluster URL. 
+. Go to your Argo CD application and verify that it has the given permissions, like roles and cluster roles, that you propagated to `cluster1`. 
+
+== Using policy to create managed service accounts and cluster permissions 
+
+Use a single policy to create the managed service account and cluster permission for all the managed clusters. When the GitOpsCluster resource is updated with the `managedServiceAccountRef`, each managed cluster in the placement of this GitOpsCluster needs to have the service account. If you have several managed clusters, it becomes tedious for you to create the managed service account and cluster permission for each managed cluster. Therefore, use a single policy. 
+
+When you apply the `managedServiceAccount` and `clusterPermission` resources to the hub cluster, the placement of this policy is bound to the local cluster. You want to replicate those resources to the managed cluster namespace for all the managed clusters in the placement of the GitOpsCluster resource, making it easier to manage the namespaces and clusters.
```
philipwu08: Make take out the last part 

", making it easier to manage the namespaces and clusters"

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+
+Using a policy to create the `managedServiceAccount` and `clusterPermission` resources include the following benefits:
```
philipwu08: I'm not sure what is the right word here, but not "benfits". 

"behaviours", "attributes", .... basically we're describing how the policy works/behaves.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+
+* Updating the `managedServiceAccount` and `clusterPermission` object templates in the policy results in updates to all the `managedServiceAccount` and `clusterPermission` resources.
```
philipwu08: to all the `managedServiceAccount` and `clusterPermission` resources

maybe more specific:

to the `managedServiceAccount` and `clusterPermission` resources in all the managed clusters.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+* Updating directly to the `managedServiceAccount` and `clusterPermission` resources becomes reverted because it is enforced by the policy.
```
philipwu08: "Updating directly to the `managedServiceAccount` and `clusterPermission` resources becomes reverted"

maybe:

 "Updates to the `managedServiceAccount` and `clusterPermission` resources directly will be reverted back because

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+* If the placement decision for the GitOpsCluster placement changes, the policy manages the creation and deletion of the resources in the managed cluster namespaces.
+
+. To create a policy for a YAML to create a managed service account and cluster permission, run the following YAML: 
+
++
+[source,yaml]
+----
```
philipwu08, line 167: Should we be consistent and wrap it with 

oc apply -f - <<EOF

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,307 @@
...
+spec:
+  numberOfClusters: 1
+  predicates:
+  - requiredClusterSelector:
+      labelSelector:
+        matchLabels:
+          name: local-cluster
+----
+
+. In the object template of the policy, iterate through the placement decision of the GitOpsCluster associated placement and apply the following `managedServiceAccount` and `clusterPermission` templates:
```
philipwu08: "In the object template of the policy, iterate through the placement decision of the GitOpsCluster associated placement and apply the following `managedServiceAccount` and `clusterPermission` templates"

maybe

In the object template of the policy, it iterates through the placement decision of the GitOpsCluster associated placement and applies the following `managedServiceAccount` and `clusterPermission` templates

gitops/gitops_service_account_argo_cd.adoc
cadawson:
```suggestion
To create a managed service account, see link:../clusters/cluster_lifecycle/addon_managed_service.adoc#managed-serviceaccount-addon[Enabling managedserviceaccount add-ons].
```

cadawson, line 25:
```suggestion
To create a managed service account, see https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.5/html/multicluster_engine/managed-serviceaccount-addon#doc-wrapper[Enabling managederviceaccount add-ons] 
```

gitops/gitops_service_account_argo_cd.adoc
cadawson: Hi @philipwu08. You are correct that you can use the "EOF" format, but for consistency, we break it up in our docs. For example: 

. Create the following YAML file named `yaml.yaml`

----
yaml content here
----

. Apply the file by entering the following command:

----
oc apply -f yaml.yaml
----

gitops/gitops_service_account_argo_cd.adoc
cadawson:
```suggestion
. The sample `clusterpermission` creates a role called `clusterpermission-msa-subject-sample` in the mortgage namespace. If one does not already exist, create a namespace `mortgage`. 
```

gitops/gitops_service_account_argo_cd.adoc
cadawson: For consistency, we should probably console here instead of UI, unless there is a compelling reason not to do that.

gitops/gitops_service_account_argo_cd.adoc
cadawson:
```suggestion
When you apply the `managedServiceAccount` and `clusterPermission` resources to the hub cluster, the placement of this policy is bound to the local cluster. Replicate those resources to the managed cluster namespace for all of the managed clusters in the placement of the GitOpsCluster resource, which makes it easier to manage the namespaces and clusters.
```

gitops/gitops_service_account_argo_cd.adoc
swopebe: Sure, but it is probably wrong. :-)

gitops/gitops_service_account_argo_cd.adoc
swopebe: We don't usually write EOF in the docs, @philipwu08 bc we want the YAML as clean as possible, does it break the user here without it?

gitops/gitops_service_account_argo_cd.adoc
swopebe: We need to ensure we are not using shortcuts or "developer speak" and we write the same way across the whole docs, @philipwu08 -- so here would would write out namespace.

gitops/gitops_push_pull.adoc
swopebe: either _contains_ or _has_ on line 86, it is already a bit wordy and hard to understand. We want to make it as concise as possible.

gitops/gitops_service_account_argo_cd.adoc
swopebe: The title needs to match the actual task. Please be careful with "Using" and we can end up calling everything "Using." :-) We also need to keep it concise, if possible.

Since the user is "Creating" in this task, we need to keep that in the title.

_Creating a customized service account for Argo CD push model_

Does this work?

We don't normally write: _for use with_

gitops/gitops_service_account_argo_cd.adoc
swopebe: _it will have the_ 

We should not use "to be" phrases in the doc for many reasons, that is why we take them out per our tech writing standards.

It really needs to be rewritten and straight to the point and probably added to the top of the file as the introduction. We tell the user why they need to do it AFTER we tell them what to do...I think we can flip that:

_Grant cluster permissions to your managed service accounts so that you can deploy applications across managed clusters._

Remember: Active voice almost always, and imperative almost always.

gitops/gitops_service_account_argo_cd.adoc
swopebe: I agree if it is in the doc, always we don't want to rewrite it.

gitops/gitops_service_account_argo_cd.adoc
swopebe: This only works if the sample does not change @philipwu08 -- if you guys change the sample, our doc is out of date.

gitops/gitops_service_account_argo_cd.adoc
swopebe: Again, we don't want to use EOF in official doc

gitops/gitops_service_account_argo_cd.adoc
swopebe: please remove UI, as we need to use the word console

gitops/gitops_service_account_argo_cd.adoc
swopebe: We want to clearly separate YAML samples and not use EOF in official doc

gitops/gitops_service_account_argo_cd.adoc
swopebe: @jc-berger you can see where and how it is done throughout the doc so that we are being consistent, but I think it is fair to add a placeholder. I think we normally do <managed-cluster> but please check and let me know if there is inconsistencies.

gitops/gitops_service_account_argo_cd.adoc
swopebe: console or terminal is how we use it in all of RHACM doc, never UI

gitops/gitops_service_account_argo_cd.adoc
philipwu08: Yes, if we use `oc apply -f - <<EOF` then `EOF` is also needed at the end. Or we could change it to what Chris proposed, which is to just include the YAML content, then provide the command separately.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+The GitOpsCluster resource uses placement to import selected managed clusters into the Argo CD, including the the creation of the Argo CD cluster secrets which contains information used to access the clusters. By default, the Argo CD cluster secret uses the application manager service account to access the managed clusters. 
+
+. To update the GitOpsCluster resource to use the managed service account , add the `managedServiceAccountRef` property with the name of the managed service account. 
+. To create a GitOpsCluster custom resource, run the following full YAML: 
+
++
+[source,yaml]
+----
+---
+oc apply -f yaml.yaml
```
philipwu08: If we go with Chris's suggestion, this command needs to be moved after the yaml.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+* Updating the `managedServiceAccount` and `clusterPermission` object templates in the policy results in updates to all of the `managedServiceAccount` and `clusterPermission` resources in all of the managed clusters.
+* Updating directly to the `managedServiceAccount` and `clusterPermission` resources becomes reverted back to the original state because it is enforced by the policy.
+* If the placement decision for the GitOpsCluster placement changes, the policy manages the creation and deletion of the resources in the managed cluster namespaces.
+
+. To create a policy for a YAML to create a managed service account and cluster permission, run the following YAML: 
+
++
+[source,yaml]
+----
+oc apply -f yaml.yaml
```
philipwu08: Move this command after the yaml for consistency

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+
+When you apply the `managedServiceAccount` and `clusterPermission` resources to the hub cluster, the placement of this policy is bound to the local cluster. Replicate those resources to the managed cluster namespace for all of the managed clusters in the placement of the GitOpsCluster resource.
+
+Using a policy to create the `managedServiceAccount` and `clusterPermission` resources include the following attributes: 
+
+* Updating the `managedServiceAccount` and `clusterPermission` object templates in the policy results in updates to all of the `managedServiceAccount` and `clusterPermission` resources in all of the managed clusters.
+* Updating directly to the `managedServiceAccount` and `clusterPermission` resources becomes reverted back to the original state because it is enforced by the policy.
+* If the placement decision for the GitOpsCluster placement changes, the policy manages the creation and deletion of the resources in the managed cluster namespaces.
+
+. To create a policy for a YAML to create a managed service account and cluster permission, run the following YAML:
```
philipwu08: Since the command comes after the YAML, maybe instead of "run the following YAML", 

==> "create a YAML with the following content"

We save this YAML as policy.yaml?

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+== Creating a cluster permission 
+
+When the service account is created, is does not have any permission associated to it. To grant permissions to the new service account, use the `clusterpermission` resource. The `clusterpermission` resource is created in the managed cluster namespace on the hub. It provides a convenient way to create roles, cluster roles resources on the managed clusters, and bind them to a service account through a `rolebinding` or `clusterrolebinding` resource.
+
+. To grant the `<managed-sa-sample>` service account permission to the sample mortgage application that is deployed to the mortgage namespace on `<managed cluster>`, apply the following YAML: 
+
++
+[source,yaml]
+----
+oc apply -f yaml.yaml
```
philipwu08: Move command after the YAML?

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+
+The `managedserviceaccount` custom resource on the hub provides a convenient way to create `serviceaccounts` on the managed clusters. When a `managedserviceccount` custom resource is created in the `<managed_cluster>` namespace on the hub cluster, a `serviceccount` is created on the managed cluster.
+
+To create a managed service account, see link:../clusters/cluster_lifecycle/addon_managed_service.adoc#managed-serviceaccount-addon[Enabling managedserviceaccount add-ons].
+
+== Creating a cluster permission 
+
+When the service account is created, is does not have any permission associated to it. To grant permissions to the new service account, use the `clusterpermission` resource. The `clusterpermission` resource is created in the managed cluster namespace on the hub. It provides a convenient way to create roles, cluster roles resources on the managed clusters, and bind them to a service account through a `rolebinding` or `clusterrolebinding` resource.
+
+. To grant the `<managed-sa-sample>` service account permission to the sample mortgage application that is deployed to the mortgage namespace on `<managed cluster>`, apply the following YAML:
```
philipwu08: "apply the following YAML" 
 
maybe change to "create a YAML with the following content":

Our command "oc apply -f yaml.yaml", expects the YAML to be saved in a file called yaml.yaml. Maybe ask them to save the YAML to a file called "cluster-permission.yaml". This way the yaml will not conflict with other YAMLs they will be saving.

The command is then:
oc apply -f cluster-permission.yaml

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+
+. The sample `<clusterpermission>` creates a role called `<clusterpermission-msa-subject-sample>` in the mortgage namespace. If one does not already exist, create a namespace `mortgage`. 
+. Review the resources that are created on the `<managed cluster>`. 
+
+== Using a managed service account in the GitOpsCluster resource
+
+The GitOpsCluster resource uses placement to import selected managed clusters into the Argo CD, including the the creation of the Argo CD cluster secrets which contains information used to access the clusters. By default, the Argo CD cluster secret uses the application manager service account to access the managed clusters. 
+
+. To update the GitOpsCluster resource to use the managed service account , add the `managedServiceAccountRef` property with the name of the managed service account. 
+. To create a GitOpsCluster custom resource, run the following full YAML:
```
philipwu08: Save this YAML as gitops.yaml, then the command after the YAML is:

oc apply -f gitops.yaml

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,286 @@
...
+        matchLabels:
+          name: local-cluster
+----
+
+. In the object template of the policy, it iterates through the placement decision of the GitOpsCluster associated placement and applies the following `managedServiceAccount` and `clusterPermission` templates: 
+
++
+[source,yaml]
+----
+oc apply -f yaml.yaml
```
philipwu08: This is not needed

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,279 @@
...
+The `managedserviceaccount` custom resource on the hub provides a convenient way to create `serviceaccounts` on the managed clusters. When a `managedserviceccount` custom resource is created in the `<managed_cluster>` namespace on the hub cluster, a `serviceccount` is created on the managed cluster.
+
+To create a managed service account, see link:../clusters/cluster_lifecycle/addon_managed_service.adoc#managed-serviceaccount-addon[Enabling managedserviceaccount add-ons].
+
+== Creating a cluster permission 
+
+When the service account is created, is does not have any permission associated to it. To grant permissions to the new service account, use the `clusterpermission` resource. The `clusterpermission` resource is created in the managed cluster namespace on the hub. It provides a convenient way to create roles, cluster roles resources on the managed clusters, and bind them to a service account through a `rolebinding` or `clusterrolebinding` resource.
+
+. To grant the `<managed-sa-sample>` service account permission to the sample mortgage application that is deployed to the mortgage namespace on `<managed cluster>`, create a YAML with the following content: 
+
```
swopebe, line 32: We can do this in another round but I agree that we need split the YAML like we do throughout the doc and provide clear steps to apply them. It seems there may be a formatting issue in the editor that may resolve when we do that or may not, we can investigate. Approving to get us closer to reorg steps.

swopebe (approved):
Commented

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,279 @@
...
+      kind: ManagedServiceAccount
+      name: <managed-sa-sample>
+  clusterRoleBinding:
+    subject:
+      apiGroup: authentication.open-cluster-management.io
+      kind: ManagedServiceAccount
+      name: <managed-sa-sample>
+----
+
+. Run `oc apply -f cluster-permission.yaml`.
```
philipwu08, line 90: We should probably ask them to save the YAML in a file called cluster-permission.yaml, since we're referring to a file by this name for the apply command.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,279 @@
...
+spec:
+  numberOfClusters: 1
+  predicates:
+  - requiredClusterSelector:
+      labelSelector:
+        matchLabels:
+          name: local-cluster
+----
+
+. Run `oc apply -f policy.yaml`.
```
philipwu08, line 288: Same comment. We're referring to policy.yaml but we never asked user to save the YAML in a file with this name.

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,279 @@
...
+    cluster: notused
+    argoNamespace: openshift-gitops
+  placementRef:
+    kind: Placement
+    apiVersion: cluster.open-cluster-management.io/v1beta1
+    name: all-openshift-clusters
+    namespace: openshift-gitops
+----
+
+. Run `oc apply -f gitops.yaml`.
```
philipwu08, line 131: Ask user to save YAML to file called gitops.yaml

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,283 @@
...
+  clusterRoleBinding:
+    subject:
+      apiGroup: authentication.open-cluster-management.io
+      kind: ManagedServiceAccount
+      name: <managed-sa-sample>
+----
+
+. Save the YAML file in a file called, `cluster-permission.yaml`.
+. Run `oc apply -f cluster-permission.yaml`. 
+. The sample `<clusterpermission>` creates a role called `<clusterpermission-msa-subject-sample>` in the mortgage namespace. If one does not already exist, create a namespace `mortgage`.
```
xiangjingli, line 91: Once the sample `<clusterpermission>` is created, the following resources are created in the given managed cluster
- one role called `<clusterpermission-msa-subject-sample>` in the default namespace
- one rolebinding called `<clusterpermission-msa-subject-sample>` in the default namespace for binding the role to the managed service account

- one role called `<clusterpermission-msa-subject-sample>` in the mortgage namespace
- one rolebinding called `<clusterpermission-msa-subject-sample>` in the mortgage namespace for binding the role to the managed service account

-  one clusterrole called `<clusterpermission-msa-subject-sample>`
- one clusterrolebinding called `<clusterpermission-msa-subject-sample>` for binding the clusterRole to the managed service account

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,283 @@
...
++
+[source,yaml]
+----
+% oc get secrets -n openshift-gitops <managed cluster-managed-sa-sample-cluster-secret>    
+NAME                                        TYPE     DATA   AGE
+<managed cluster-managed-sa-sample-cluster-secret>   Opaque   3      4m2s
+----
+
+. Save the YAML file in a file called, `cluster-permission.yaml`.
+. Run `oc apply -f cluster-permission.yaml`.
```
xiangjingli: delete the two lines 
```
. Save the YAML file in a file called, `cluster-permission.yaml`.
. Run `oc apply -f cluster-permission.yaml`. 
```

gitops/gitops_service_account_argo_cd.adoc
```diff
@@ -0,0 +1,289 @@
...
+
+== Creating a managed service account 
+
+The `managedserviceaccount` custom resource on the hub provides a convenient way to create `serviceaccounts` on the managed clusters. When a `managedserviceccount` custom resource is created in the `<managed_cluster>` namespace on the hub cluster, a `serviceccount` is created on the managed cluster.
+
+To create a managed service account, see link:../clusters/cluster_lifecycle/addon_managed_service.adoc#managed-serviceaccount-addon[Enabling managedserviceaccount add-ons].
+
+== Creating a cluster permission 
+
+When the service account is created, is does not have any permission associated to it. To grant permissions to the new service account, use the `clusterpermission` resource. The `clusterpermission` resource is created in the managed cluster namespace on the hub. It provides a convenient way to create roles, cluster roles resources on the managed clusters, and bind them to a service account through a `rolebinding` or `clusterrolebinding` resource.
```
xiangjingli: When the service account is created, **is** does not have any permission associated to it.
====>
When the service account is created, **it** does not have any permission associated to it.

xiangjingli:
just one typo left. Once it is fixed, it is good to merge. Great work! @jc-berger

PR #5552, opened 2023-09-28
https://issues.redhat.com/browse/ACM-7523 Deprecate MSA API 2.9

oafischer:


clusters/api/managed_serviceaccount.json.adoc
```diff
@@ -1,5 +1,5 @@
 [#serviceaccount-api]
-= ManagedServiceAccount API
+= ManagedServiceAccount API (v1beta1)
```
swopebe, line 3: If we are actually going to replace, essentially remove the former API, that is not a true deprecation. We should not be removing something from the doc that was not even deprecated. We can however deprecate it and put:

v1alpha is not removed, but is deprecated. Best practice: Use v1beta1 instead.

swopebe (changes requested):

swopebe (approved):
Made one small change.

zhujian7:
LGTM

PR #5512, opened 2023-09-13
https://issues.redhat.com/browse/ACM-6737 add Nutanix

cadawson:
https://issues.redhat.com/browse/ACM-6737

https://issues.redhat.com/browse/MGMT-13374

PR #5153, opened 2023-06-02
https://issues.redhat.com/browse/ACM-5430 deploy cluster cli 2.8

cadawson:
Issue: https://issues.redhat.com/browse/ACM-5430

PR #5562, opened 2023-09-29
https://issues.redhat.com/browse/ACM-5430 and ACM-6737 Add CLI instru…

cadawson:
…ctions for create cluster on premises and Nutanix


cadawson:
This is a front-port from https://github.com/stolostron/rhacm-docs/pull/5153 and https://github.com/stolostron/rhacm-docs/pull/5512. Should be ready to go, but would like final confirmation.

cadawson:
We merged this into 2.8 along with the Nutanix instructions. Lookin to frontport it here into 2.9. I just want to confirm that the procedure is the same. Thanks @filanov @avishayt @chadcrum!

filanov:
between 2.9 and 2.8 it is the same

PR #5575, opened 2023-10-03
[ACM-7382]: Adding instructions to create a hosted cluster from the console

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-7382

This PR adds the following information:
- Instructions to create a hosted cluster on bare metal by using the OpenShift web console
- Instructions to create an InfraEnv resource on bare metal by using the console


lahinson:
Hi @zanetworker and @eranco74 - PTAL at this PR when you have the chance. This is related to a Slack thread from last month about adding documentation about using the console to create hosted clusters: https://redhat-internal.slack.com/archives/C058TF9K37Z/p1694095268422769

clusters/hosted_control_planes/hosted_bare_metal_infraenv.adoc
```diff
@@ -1,5 +1,5 @@
 [#hosted-control-planes-create-infraenv]
-= Creating an InfraEnv resource for hosted control planes on bare metal (Technology Preview)
+= Creating an InfraEnv resource for hosted control planes on bare metal
```
avishayt: I guess this should now just point to the CIM docs on creating a host inventory:
https://github.com/stolostron/rhacm-docs/pull/5605/files

The plan is to have one chapter that discusses the host inventory (creating an infrastructure environment and booting hosts with the Discovery Image).  Then we have separate chapters for installing standalone OCP clusters and hosted clusters (each of those should refer to the host inventory chapter and not duplicate it).

FYI @oafischer

clusters/hosted_control_planes/hosted_create_bare_metal.adoc
```diff
@@ -69,6 +69,33 @@ redhat-marketplace-catalog-7dd77d896-jtxkd       1/1     Running   0          2m
...
+. In the console header, ensure that **All Clusters** is selected.
+
+. Click **Infrastructure > Clusters**.
+
+. Click **Create cluster > Host inventory > Hosted control plane**.
+
++
+The **Create cluster** page is displayed.
+
+. On the **Create cluster** page, follow the prompts to enter details about the cluster, node pools, networking, and automation.
```
avishayt, line 86: Do we want to go into details about each screen and field?

clusters/hosted_control_planes/hosted_bare_metal_infraenv.adoc
zanetworker: Avishay, if the CIM docs will include details on HCP then that would make sense.

clusters/hosted_control_planes/hosted_create_bare_metal.adoc
lahinson: @avishayt Good question. I was hoping that you and @zanetworker might have an opinion on this :) 

I'd assume/hope that the UI is intuitive enough that we don't need to explain each screen and field, but if we know that there are areas where people might struggle, it makes sense to me to document those areas (at least until the UI can be improved). 

Do we have user feedback about any screens or fields that we need to provide more guidance about in the docs?

clusters/hosted_control_planes/hosted_bare_metal_infraenv.adoc
oafischer: I'll make sure not to change the [#create-host-inventory-console] handle if you want to link to that topic.

lahinson: @avishayt Can you confirm whether the new CIM docs will contain all the relevant details for hosted control planes? If so, should I update this page to replace the first half of the content (about creating an InfraEnv resource by using the command line) with a pointer from this page to `cluster_lifecycle/cim_intro.adoc`?

avishayt: We already have docs for creating an InfraEnv using the console:
https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/clusters/cluster_mce_overview#creating-your-infra-env-with-the-console

We also have ACM-7816 to add instructions on creating an InfraEnv using the CLI, and a google doc ready with the content:
https://docs.google.com/document/d/1pOV_rU8sebHHX-axQnrW_z_8Piw-cnR1ucXS9lEEFMo/edit?usp=sharing

clusters/hosted_control_planes/hosted_create_bare_metal.adoc
zanetworker: I understood Avishay's question as a hint to add more details (I could be wrong). I would encourage more details whenever possible, at least for the non-obvious areas. Ideally the UX is self-descriptive, but there are times as a user I would have appreciated the "extra" detail that would save me 5 minutes especially when getting started. 

>Do we have user feedback about any screens or fields that we need to provide more guidance about in the docs?

We do have a couple, will forward to you on slack Laura :)

avishayt: It wasn't a hint - I would have been more direct :)
If it's self-explanatory, then probably nobody will read it, and we can save the resources for writing and maintaining docs for it.  If users might get stuck there, then it's worth investing in it.

clusters/hosted_control_planes/hosted_bare_metal_infraenv.adoc
lahinson: Thanks for the clarification, @avishayt. I updated the file structure so that within the "Hosted control planes" book, the instructions for bare metal will point to https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/clusters/cluster_mce_overview#creating-your-infra-env-with-the-console instead of https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/clusters/cluster_mce_overview#hosted-control-planes-create-infraenv.

clusters/hosted_control_planes/hosted_create_bare_metal.adoc
lahinson: I'm reading through some user feedback that Adel sent me related to creating a hosted cluster with the console. It looks like we'll need to add some details in the docs about the pull secret, the namespace on the Node pools page, and the Networking page. I will likely have questions soon about what exactly we want to say about those screens and fields :)

avishayt: Happy to answer any questions

clusters/hosted_control_planes/hosted_bare_metal_infraenv.adoc
oafischer: I think the link you are looking for is https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.8/html/clusters/cluster_mce_overview#create-host-inventory-console

This one is definitely final and will not be changed anymore since we need to lock in the UI links as well.

lahinson: Thanks @oafischer! I'll point to that page instead.

clusters/hosted_control_planes/hosted_create_bare_metal.adoc
lahinson: Thanks @avishayt! I noted my questions in this Google Doc: https://docs.google.com/document/d/1sNtpyU3YG2ryAISw7QdHAqUEo4m44P97HSuItXMrX2U/edit?usp=sharing, which is based on a document that Adel shared with me that had user feedback about the console. 

cc @zanetworker

lahinson: @avishayt Thanks so much for your feedback in the Google Doc. Based on the information that you provided, I added some tips in the note starting on line 89. PTAL when you can and let me know what you think. cc @zanetworker

clusters/hosted_control_planes/hosted_create_bare_metal.adoc
```diff
@@ -69,9 +69,49 @@ redhat-marketplace-catalog-7dd77d896-jtxkd       1/1     Running   0          2m
...
+
+** If you want to use predefined values to automatically populate fields in the console, you can create a host inventory credential. For more information, see _Creating a credential for an on-premises environment_.
+
+** On the *Cluster details* page, the pull secret is your {ocp-short} pull secret that you use to access {ocp-short} resources. If you selected a host inventory credential, the pull secret is automatically populated.
+
+** On the *Node pools* page, the namespace contains the hosts for the node pool. If you created a host inventory by using the console, the console creates a dedicated namespace.
+
+** On the *Networking* page, you select an API server publishing strategy. The API server for the hosted cluster is exposed as a service of the `NodePort` type. A DNS entry must exist for the `api.${HOSTED_CLUSTER_NAME}.${BASEDOMAIN}` setting that points to the destination where the API server can be reached. This entry can be a record that points to one of the nodes in the management cluster or a record that points to a load balancer that redirects incoming traffic to the Ingress pods.
```
avishayt:
```suggestion
** On the *Networking* page, you select an API server publishing strategy. The API server for the hosted cluster can be exposed either using an existing load balancer or as a service of the `NodePort` type. A DNS entry must exist for the `api.${HOSTED_CLUSTER_NAME}.${BASEDOMAIN}` setting that points to the destination where the API server can be reached. This entry can be a record that points to one of the nodes in the management cluster or a record that points to a load balancer that redirects incoming traffic to the Ingress pods.
```

avishayt (commented):
Small comment, otherwise looks great. Thanks!

clusters/hosted_control_planes/configure_hosted_bm.adoc
```diff
@@ -61,7 +61,7 @@ The Agent platform does not create any infrastructure, but it does have the foll
 After you meet the prerequisites, complete the following steps to configure hosted control planes on bare metal:
 
 . xref:../hosted_control_planes/hosted_bare_metal_dns.adoc#configure-dns-bm[Configure DNS on bare metal]
-. xref:../hosted_control_planes/hosted_bare_metal_infraenv.adoc#hosted-control-planes-create-infraenv[Create an InfraEnv resource for hosted control planes on bare metal]
+. xref:../cluster_lifcycle/cim_create_console.adoc#create-host-inventory-console[Create an InfraEnv resource for hosted control planes on bare metal]
```
oafischer: If that's the correct link the title should be: Creating a host inventory by using the console

clusters/hosted_control_planes/hosted_bare_metal_infraenv.adoc
oafischer:
```suggestion
= Creating an _InfraEnv_ resource for hosted control planes on bare metal
```

clusters/hosted_control_planes/hosted_bare_metal_infraenv.adoc
```diff
@@ -38,4 +38,17 @@ oc apply -f infraenv-config.yaml
...
 ----
 oc -n ${HOSTED_CONTROL_PLANE_NAMESPACE} get InfraEnv ${HOSTED_CLUSTER_NAME} -ojsonpath="{.status.isoDownloadURL}"
-----
\ No newline at end of file
+----
+
+where `{$HOSTED_CONTROL_PLANE_NAMESPACE}` is the namespace for hosted control plane pods.
+
+[#hosted-bare-metal-infraenv-console]
+== Creating an InfraEnv resource by using the console
```
oafischer:
```suggestion
== Creating an _InfraEnv_ resource by using the console
```

oafischer (approved):
Some small comments, otherwise looks great.

Important note that doesn't necessarily require changes: We use host inventory and infrastructure environment interchangeably because both terms appear in the console, but we want to primarily use host inventory because it is the first term you encounter in the console.

I was also told it might be easier to understand for the user, at least initially when they don't know what an infrastructure environment is yet. The way I approach it is: On first use, say `host inventory (infrastructure environment)`, from then on you can use infrastructure environment.

clusters/hosted_control_planes/configure_hosted_bm.adoc
lahinson: Good catch! Will fix.

lahinson:
@oafischer Thanks for the edits and the helpful info about host inventory/infrastructure environment. I'm incorporating your feedback, and I'll keep the terminology guidance in mind. I appreciate it!

PR #5576, opened 2023-10-03
modified notes

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-7807

### DIFF
--- a/business_continuity/backup_restore/backup_install.adoc
+++ b/business_continuity/backup_restore/backup_install.adoc
@@ -7,7 +7,7 @@ The cluster backup and restore operator is not installed automatically. Continue
 
 - Custom resource definitons are cluster-scoped, so you cannot have two different versions of OADP or Velero installed on the same cluster. If you have two different versions, one version runs with the wrong custom resource definitions.
 
-- If you did not enable the cluster backup and restore operator, the `MultiClusterHub` resource installs on the hub for the OADP Operator and Velero Custome Resource Definition (CRD) resources. The `MultiClusterHub` reconciles the OADP and Velero CRD resources with the version that is used by the OADP Operator that was installed when you enabled the cluster backup and restore operator. You cannot install another version of OADP or Velero on the clusterhub, unless you use the same CRD resources that is used by the OADP Operator that was installed when you enabled the backup and restore operator.
+- If you did not enable the cluster backup and restore operator, the `MultiClusterHub` resource installs on the hub cluster for the OADP Operator and Velero Custom Resource Definition (CRD) resources. The `MultiClusterHub` reconciles the OADP and Velero CRD resources with the version that is used by the OADP Operator that was installed when you enabled the cluster backup and restore operator. You cannot install another version of OADP or Velero on the clusterhub, unless you use the same CRD resources that is used by the OADP Operator that was installed when you enabled the backup and restore operator.
 
 - The backup component works with OADP installed in the component namespace. 
 
--- a/business_continuity/backup_restore/backup_install.adoc
+++ b/business_continuity/backup_restore/backup_install.adoc
@@ -7,7 +7,7 @@ The cluster backup and restore operator is not installed automatically. Continue
 
 - Custom resource definitons are cluster-scoped, so you cannot have two different versions of OADP or Velero installed on the same cluster. If you have two different versions, one version runs with the wrong custom resource definitions.
 
-- If you did not enable the cluster backup and restore operator on the `MultiClusterHub` resource, the OADP Operator and Velero Custom Resource Definition (CRD) are still installed on the hub cluster. The `MultiClusterHub` reconciles the OADP and Velero CRD resources with the version that is used by the OADP Operator which is installed when you enabled the cluster backup and restore operator. As a result, you cannot install another version of OADP or Velero on the hub cluster, unless your version uses the same CRD definitions as the OADP Operator installed when you enabled the backup and restore operator.
+- If you did not enable the cluster backup and restore operator on the `MultiClusterHub` resource, the OADP Operator and Velero custom resource definition are still installed on the hub cluster. The `MultiClusterHub` resource reconciles the OADP and Velero custom resource definitions with the version that is used by the OADP Operator, which is installed when you enabled the cluster backup and restore operator. As a result, you cannot install another version of OADP or Velero on the hub cluster, unless your version uses the same custom resource definitions as the OADP Operator that is installed when you enabled the backup and restore operator.
 
 - The backup component works with the OADP Operator that is installed in the component namespace. 
 
--- a/business_continuity/backup_restore/backup_install.adoc
+++ b/business_continuity/backup_restore/backup_install.adoc
@@ -7,9 +7,9 @@ The cluster backup and restore operator is not installed automatically. Continue
 
 - Custom resource definitons are cluster-scoped, so you cannot have two different versions of OADP or Velero installed on the same cluster. If you have two different versions, one version runs with the wrong custom resource definitions.
 
-- If you did not enable the cluster backup and restore operator, the `MultiClusterHub` resource installs on the hub cluster for the OADP Operator and Velero Custom Resource Definition (CRD) resources. The `MultiClusterHub` reconciles the OADP and Velero CRD resources with the version that is used by the OADP Operator that was installed when you enabled the cluster backup and restore operator. You cannot install another version of OADP or Velero on the clusterhub, unless you use the same CRD resources that is used by the OADP Operator that was installed when you enabled the backup and restore operator.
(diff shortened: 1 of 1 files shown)


business_continuity/backup_restore/backup_install.adoc
```diff
@@ -5,9 +5,11 @@ The cluster backup and restore operator is not installed automatically. Continue
...
+- Custom resource definitons are cluster-scoped, so you cannot have two different versions of OADP or Velero installed on the same cluster. If you have two different versions, one version runs with the wrong custom resource definitions.
 
-- The backup component works with OADP installed in the component namespace. Before installing the OADP Operator in the backup component namespace, you might need to manually delete the OADP and Velero custom resource definitions installed by previous versions of the OADP Operator.
+- Even if you have not enabled the cluster backup and restore operator, the MultiClusterHub resource installs on the hub for the OADP Operator and Velero CRD resources. The MultiClusterHub reconciles the OADP and Velero CRD resources with the version used by the OADP Operator that was installed when the cluster backup and restore operator was enabled. For this reason, you cannot install another version of OADP or Velero on the cluster hub, unless you use the same CRD resources used by the OADP Operator that was installed when the backup and restore operator was enabled.
```
jc-berger: Am I correct that in this instance we should use past tense since the note is detailing what has happened? 

Happy to change verb tense if I am incorrect. Thanks!

swopebe: I think you can switch to "If you did not enable" to remove the participle phrase, which is not really needed here and not as concise. (Sometimes participles are needed, often not; the best thing to do is tighten the verb tense without it when and where you can.)

MultiClusterHub as a resource we are writing within code like `MultiClusterHub`, if I am not mistaking.

Not sure we need: For this reason,

hub cluster (together always)

CRD resources? Is that Custom Resource Definition resources?

_that is_ used by the OADP....

swopebe (changes requested):
couple small things.

business_continuity/backup_restore/backup_install.adoc
```diff
@@ -5,9 +5,11 @@ The cluster backup and restore operator is not installed automatically. Continue
...
+
+- The backup component works with OADP installed in the component namespace.
```
swopebe: that is installed?

business_continuity/backup_restore/backup_install.adoc
jc-berger: awesome, thanks, see my changes!

business_continuity/backup_restore/backup_install.adoc
```diff
@@ -5,9 +5,11 @@ The cluster backup and restore operator is not installed automatically. Continue
...
+- Custom resource definitons are cluster-scoped, so you cannot have two different versions of OADP or Velero installed on the same cluster. If you have two different versions, one version runs with the wrong custom resource definitions.
 
-- The backup component works with OADP installed in the component namespace. Before installing the OADP Operator in the backup component namespace, you might need to manually delete the OADP and Velero custom resource definitions installed by previous versions of the OADP Operator.
+- If you did not enable the cluster backup and restore operator, the `MultiClusterHub` resource installs on the hub cluster for the OADP Operator and Velero Custom Resource Definition (CRD) resources. The `MultiClusterHub` reconciles the OADP and Velero CRD resources with the version that is used by the OADP Operator that was installed when you enabled the cluster backup and restore operator. You cannot install another version of OADP or Velero on the clusterhub, unless you use the same CRD resources that is used by the OADP Operator that was installed when you enabled the backup and restore operator.
```
jc-berger: SMEs 

Should it be `on the clusterhub` 

or, `on the hub on the cluster`?

birsanv: Here: >>If you did not enable the cluster backup and restore operator, the `MultiClusterHub` resource installs on the hub cluster for the OADP Operator and Velero Custom Resource Definition (CRD) resources.<< 

 I want to convey that **even if you don't have backup enabled on the MCH resource** , the OADP Operator and Velero Custom Resource Definition (CRD) resources are being installed on the hub cluster.
This is lost with the current wording :) I read that the CRDs are installed if you don't enable the backup

Can we reword it ?

birsanv: For this sentence : >> The `MultiClusterHub` reconciles the OADP and Velero CRD resources with the version that is used by the OADP Operator that was installed when you enabled the cluster backup and restore operator. You cannot install another version of OADP or Velero on the clusterhub, unless you use the same CRD resources that is used by the OADP Operator that was installed when you enabled the backup and restore operator. <<

1.  ( avoid _that_ twice .. )
The `MultiClusterHub` reconciles the OADP and Velero CRD resources with the version that is used by the OADP Operator **which is** installed when you enabled the cluster backup and restore operator. You cannot install another version of OADP or Velero on the clusterhub, unless you use the same CRD resources that is used by the OADP Operator that was installed when you enabled the backup and restore operator.

2. Should be hub cluster, not cluster hub (.. on the clusterhub, unless ..)

3. Change: 
..unless you use the same CRD resources that is used by the OADP Operator that was installed when you enabled the backup and restore operator.
 with : 
 ..unless **your version** uses the same CRD **definitions** as the OADP Operator installed when you enabled the backup and restore operator.
 
4. Is it possible to use the _As a result_  as I put in the jirra ?( I want to convey that because MCH reconciles the CRDs and replaces them with the version installed by the backup operator, as a result .. you can't install another OADP Operator unless it uses the same CRDs as the ones controlled by the MCH ). If it's not following the writing rule is fine but we loose this subtle connection between the MCH CRD and the fact that you can't install any version you like

Something like : The `MultiClusterHub` reconciles the OADP and Velero CRD resources with the version that is used by the OADP Operator **which is** installed when you enabled the cluster backup and restore operator. **As a result**, you cannot install another version of OADP...

business_continuity/backup_restore/backup_install.adoc
birsanv: I would also suggest 

The backup component works with OADP Operator that is installed in the component namespace.

birsanv (changes requested):

business_continuity/backup_restore/backup_install.adoc
```diff
@@ -7,9 +7,9 @@ The cluster backup and restore operator is not installed automatically. Continue
...
+- If you did not enable the cluster backup and restore operator on the MCH resource, the OADP Operator and Velero Custom Resource Definition (CRD) still installs on the hub cluster. The `MultiClusterHub` reconciles the OADP and Velero CRD resources with the version that is used by the OADP Operator which is installed when you enabled the cluster backup and restore operator. As a result, you cannot install another version of OADP or Velero on the hub cluster, unless your version uses the same CRD definitions as the OADP Operator installed when you enabled the backup and restore operator.
```
birsanv: The message is the right one, just a few small comments : 

1. MCH should say MultiClusterHub - I don't think we shortname this in the docs
2. Change : the OADP Operator and Velero Custom Resource Definition (CRD) still installs on the hub cluster.
with : the OADP Operator and Velero Custom Resource Definition (CRD) are still installed on the hub cluster.

birsanv (changes requested):

birsanv (approved):
looks good, thanks!

business_continuity/backup_restore/backup_install.adoc
```diff
@@ -5,9 +5,11 @@ The cluster backup and restore operator is not installed automatically. Continue
...
+- Custom resource definitons are cluster-scoped, so you cannot have two different versions of OADP or Velero installed on the same cluster. If you have two different versions, one version runs with the wrong custom resource definitions.
 
-- The backup component works with OADP installed in the component namespace. Before installing the OADP Operator in the backup component namespace, you might need to manually delete the OADP and Velero custom resource definitions installed by previous versions of the OADP Operator.
+- If you did not enable the cluster backup and restore operator on the `MultiClusterHub` resource, the OADP Operator and Velero Custom Resource Definition (CRD) are still installed on the hub cluster. The `MultiClusterHub` reconciles the OADP and Velero CRD resources with the version that is used by the OADP Operator which is installed when you enabled the cluster backup and restore operator. As a result, you cannot install another version of OADP or Velero on the hub cluster, unless your version uses the same CRD definitions as the OADP Operator installed when you enabled the backup and restore operator.
```
dockerymick:
```suggestion
- If you did not enable the cluster backup and restore operator on the `MultiClusterHub` resource, the OADP Operator and Velero custom resource definition are still installed on the hub cluster. The `MultiClusterHub` resource reconciles the OADP and Velero custom resource definitions with the version that is used by the OADP Operator, which is installed when you enabled the cluster backup and restore operator. As a result, you cannot install another version of OADP or Velero on the hub cluster, unless your version uses the same custom resource definitions as the OADP Operator that is installed when you enabled the backup and restore operator.
```

dockerymick (approved):
Left a suggestion. Otherwise, this looks good

dockerymick (approved):

PR #5607, opened 2023-10-10
Update trouble_network_config_bm.adoc

jgato:
clarifications and context about where/how find the symptoms. working with @swopebe

### DIFF
--- a/clusters/support_troubleshooting/trouble_network_config_bm.adoc
+++ b/clusters/support_troubleshooting/trouble_network_config_bm.adoc
@@ -3,7 +3,7 @@
 
 Adding a node, or scaling out, to your existing cluster that is created by the {mce} with Zero Touch Provisioning or Host inventory create methods fails during installation. The installation process works correctly during the Discovery phase, but fails on the installation phase. 
 
-The configuration of the network is failing. In the console, you see a `Pending` user action. In the description, you can see it failing on the rebooting step.
+The configuration of the network is failing. From the hub cluster in the integrated console, you see a `Pending` user action. In the description, you can see it failing on the rebooting step.
 
 The error message about failing is not very accurate, since the agent that is running in the installing host cannot report information.
  
@@ -12,7 +12,7 @@ The error message about failing is not very accurate, since the agent that is ru
 
 After the Discover phase, the host reboots to continue the installation, but it cannot configure the network. Check for the following symptoms and messages:
 
-* From the managed cluster in the integrated console, check for `Pending` user action on the adding node, with the `Rebooting` indicator:
+* From the hub cluster in the integrated console, check for `Pending` user action on the adding node, with the `Rebooting` indicator:
 +
 ----
 This host is pending user action. Host timed out when pulling ignition. Check the host console... Rebooting
--- a/clusters/support_troubleshooting/trouble_network_config_bm.adoc
+++ b/clusters/support_troubleshooting/trouble_network_config_bm.adoc
@@ -23,7 +23,7 @@ This host is pending user action. Host timed out when pulling ignition. Check th
  ** `/sysroot/etc/NetworkManager/system-connections/` 
  ** `/sysroot/etc/sysconfig/network-scripts/` 
 
-* Installing host: in a terminal, check the failing host for the following messages. You can use `journalctl` to see the log messages:
+* From the terminal of the installing host, check the failing host for the following messages. You can use `journalctl` to see the log messages:
 
 ----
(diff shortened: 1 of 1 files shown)


clusters/support_troubleshooting/trouble_network_config_bm.adoc
```diff
@@ -1,29 +1,29 @@
 [#troubleshooting-network-config-fail]
-= Troubleshooting adding day-two nodes to an existing cluster fails with pending user action
+= Troubleshooting: Adding day-two nodes to an existing cluster fails with pending user action
```
swopebe, line 2: Ok, I see the issue here. I agree we need to use a colon, the problem is all the troubleshooting files are not set up this way, and we keep the same format in the files. (This sets up user expectation and makes it easier for any writer to pick up a JIRA and write one)

I will change the other files to match what you changed back.

jgato: so... should I change it back to not use semicolon, or you will change the other files?

swopebe: just leave it, I will make the other files look like this.

clusters/support_troubleshooting/trouble_network_config_bm.adoc
```diff
@@ -1,29 +1,29 @@
...
 
 Adding a node, or scaling out, to your existing cluster that is created by the {mce} with Zero Touch Provisioning or Host inventory create methods fails during installation. The installation process works correctly during the Discovery phase, but fails on the installation phase. 
 
-The configuration of the network is failing. Using the console, you see a `pending` user action. In the description, you can see it failing on the rebooting step.
+The configuration of the network is failing. In the console, you see a `Pending` user action. In the description, you can see it failing on the rebooting step.
 
-The error message about failing is not very accurate, since the agent cannot report information.
+The error message about failing is not very accurate. Since the agent that is running in the installing host cannot report information.
```
swopebe: These are now not complete sentences. We need the structure to remain as I have written it. Complete sentence: Subject and an action with the phrase.

clusters/support_troubleshooting/trouble_network_config_bm.adoc
```diff
@@ -1,29 +1,29 @@
...
+The error message about failing is not very accurate, since the agent that is running in the installing host cannot report information.
  
 [#symptom-network-config-fail]
 == Symptom: Installation for day two workers fails
 
-At the Discover phase, the host cannot configure the network. Check for the following symptoms and messages:
+After the Discover phase, the host reboots to continue the installation, but it cannot configure the network. Check for the following symptoms and messages:
 
-* In the console, check for `Pending` user action on the adding node, with the `Rebooting` indicator:
+* From the managed cluster in the integrated console, check for `Pending` user action on the adding node, with the `Rebooting` indicator:
```
swopebe: These need to also be complete sentences,, as they already have a colon to introduce the next thing.

clusters/support_troubleshooting/trouble_network_config_bm.adoc
```diff
@@ -1,29 +1,29 @@
...
 ----
 
-* Check all the `MachineConfings` of the existing cluster. Check if any of the `MachineConfigs` create any file on the following directories: 
+* From the {ocp} configuration managed cluster, check the `MachineConfigs` of the existing cluster. Check if any of the `MachineConfigs` create any file on the following directories: 
 
  ** `/sysroot/etc/NetworkManager/system-connections/` 
  ** `/sysroot/etc/sysconfig/network-scripts/` 
 
-In the console, check the failing host for the following messages. You can also use `journalctl` to see the log messages:
+* From the terminal of the installing host, check the failing host for the following messages. You can use `journalctl` to see the log messages:
```
swopebe, line 26: How we normally write this is: Run the following command in the terminal to see the log messages:

swopebe: Here is an example (older version) of how we write procedures in the official Red Hat documentation:

https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.6/html-single/multicluster_engine/index#troubleshooting-klusterlet-with-degraded-conditions-mce

swopebe: We don't always even say terminal because you run the command in the terminal, but I am fine to leave when it helps to clarify. Is it better to say:

In the terminal, run the command.

PR #5639, opened 2023-10-16
https://issues.redhat.com/browse/ACM-5520--2.7

swopebe:
… need more clarity

### DIFF
--- a/troubleshooting/trouble_cluster_offline_avail.adoc
+++ b/troubleshooting/trouble_cluster_offline_avail.adoc
@@ -17,7 +17,7 @@ For example, the cluster is marked offline after five minutes with a `leaseDurat
 
 The five validation attempts is default and cannot be changed, but you can change the `leaseDurationSeconds` interval. 
 
-Determine the amount of time (in minutes) that you want the cluster to be marked as `offline`, then multiply that value by 60 (to convert to seconds). Then divide by the default five number of attemps. The result is your `leaseDurationSeconds` value.
+Determine the amount of time (in minutes) that you want the cluster to be marked as `offline`, then multiply that value by 60 (to convert to seconds). Then divide by the default five number of attempts. The result is your `leaseDurationSeconds` value.
 
 . Edit your `ManagedCluster` specification on the hub cluster by entering the following command, but replace `cluster-name` with the name of your managed cluster:
 +
--- a/troubleshooting/trouble_cluster_offline_avail.adoc
+++ b/troubleshooting/trouble_cluster_offline_avail.adoc
@@ -8,19 +8,31 @@ The status of the managed cluster alternates between `offline` and `available` w
 
 When the network that connects the managed cluster to the hub cluster is unstable, the status of the managed cluster that is reported by the hub cluster cycles between `offline` and `available`. 
 
+The connection between the hub cluster and managed cluster is maintained through a lease that is validated at the `leaseDurationSeconds` interval value. If the lease is not validated within five consecutive attempts of the `leaseDurationSeconds` value, then the cluster is marked offline. 
+
+For example, the cluster is marked offline after five minutes with a `leaseDurationSeconds` interval of `60 seconds`. This configuration can be inadequate for reasons such as connectivity issues or latency, causing instability.
+
 [#resolving-the-problem-cluster-status-changing-from-offline-to-available]
 == Resolving the problem: Cluster status changing from offline to available
 
-To attempt to resolve this issue, complete the following steps:
+The five validation attempts is default and cannot be changed, but you can change the `leaseDurationSeconds` interval. 
 
-. Edit your `ManagedCluster` specification on the hub cluster by entering the following command:
+Determine the amount of time (in minutes) that you want the cluster to be marked as `offline`, then multiply that value by 60 (to convert to seconds). Then divide by the default five number of attemps. The result is your `leaseDurationSeconds` value.
+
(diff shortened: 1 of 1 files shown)


troubleshooting/trouble_cluster_offline_avail.adoc
```diff
@@ -20,4 +20,7 @@ oc edit managedcluster <cluster-name>
 ----
 Replace _cluster-name_ with the name of your managed cluster.
 
-. Increase the value of `leaseDurationSeconds` in your `ManagedCluster` specification. The default value is 5 minutes, but that might not be enough time to maintain the connection with the network issues. Specify a greater amount of time for the lease. For example, you can raise the setting to 20 minutes.
+. Increase the value of `leaseDurationSeconds` in your `ManagedCluster` specification. The default value is 5 minutes, but that might not be enough time to maintain the connection with the network issues. Specify a greater amount of time for the lease. For example, you can raise the setting to 20 minutes. 
+///does this need to change if we are saying what is in the next line?
+
+You can set the lease duration to _____, but if you want the cluster to be detected as offline in less than 30 minutes, you should set it to ____.
```
showeimer: @swopebe kind of agree with you that the two sections are a bit similar.  Do you think it might be a bit more helpful to explain how this piece works, let me know if you think it makes sense or not to provide the additional detail (@qiujian16):

The connection between the hub and managed cluster is maintained through a lease; which is validated at the interval specified in the `leaseDurationSeconds`.  If the lease is not validated within 5 consecutive attempts (five times the `leaseDurationSeconds`), then the cluster will be marked offline.

By default, the cluster will be marked offline after 5 minutes with a `leaseDurationSeconds` interval of 60 seconds. In some network situations, the default configuration may be inadequate for various reasons like connectivity issues or latency.  

If you want to change this setting, first determine the amount of time (in minutes) you would like the cluster to be marked as offline after; then multiply that value by 60 (to convert to seconds) and then divide by 5 (# of attempts). This will result in the necessary value to configure for `leaseDurationSeconds`.

You can change the `leaseDurationSeconds` by configuring the respective ManagedCluster resource for the cluster, see the following sample YAML:

```yaml
apiVersion: cluster.open-cluster-management.io/v1
kind: ManagedCluster
metadata:
  name: <cluster-name>
spec:
  hubAcceptsClient: true
  leaseDurationSeconds: 60
```

---

There was a relevant thread several months ago here:  https://redhat-internal.slack.com/archives/GUWQE4D2N/p1684152296341969

swopebe: this is perfect @showeimer -- I won't use all of it but it give a better symptom.

showeimer (approved):

troubleshooting/trouble_cluster_offline_avail.adoc
qiujian16: @showeimer this is create clarification, thanks!

troubleshooting/trouble_cluster_offline_avail.adoc
```diff
@@ -8,16 +8,33 @@ The status of the managed cluster alternates between `offline` and `available` w
 
 When the network that connects the managed cluster to the hub cluster is unstable, the status of the managed cluster that is reported by the hub cluster cycles between `offline` and `available`. 
 
+The connection between the hub cluster and managed cluster is maintained through a lease that is validated at the `leaseDurationSeconds` interval value. If the lease is not validated within five consecutive attempts of the `leaseDurationSeconds` value, then the cluster is marked offline.
```
jc-berger: Optional. 

It might benefit users to know what "offline" means for their development. 

Like, 

`...then the cluster is marked offline. When you clusters is offline, you cannot deploy your application.`

jc-berger: since you later you use ` ` for offline, `offline` do so here, for consistency. 

...then the cluster is marked `offline`.  

For example, the cluster is marked `offline` after five minutes with a,..

troubleshooting/trouble_cluster_offline_avail.adoc
```diff
@@ -8,16 +8,33 @@ The status of the managed cluster alternates between `offline` and `available` w
...
+
+For example, the cluster is marked offline after five minutes with a `leaseDurationSeconds` interval of `60 seconds`. This configuration can be inadequate for reasons such as connectivity issues or latency, causing instability.
+
 [#resolving-the-problem-cluster-status-changing-from-offline-to-available]
 == Resolving the problem: Cluster status changing from offline to available
 
-To attempt to resolve this issue, complete the following steps:
+The five validation attempts is default and cannot be changed, but you can change the `leaseDurationSeconds` interval. 
+
+Determine the amount of time (in minutes) that you want the cluster to be marked as `offline`, then multiply that value by 60 (to convert to seconds). Then divide by the default five number of attempts. The result is your `leaseDurationSeconds` value.
```
jc-berger: IBM SG encourages us to only use parenthesis for acronyms. I actually think it reads just as well with no parenthesis, so, I think you can remove them. 

Determine the amount of time, in minutes, that you want the cluster to marked as `offline`, then multiply that value by 60 to convert it to seconds. Then divide...

swopebe: all offline marked as it is part of code, since it is...and all () changed.

I don't think we need to explain offline because at this point they are in a troubleshooting file. The problem is that it switches between offline and available, so they know it is unstable -- we explain that early in the file.

jc-berger (approved):
/lgtm

PR #5686, opened 2023-10-23
Make subscription channel version variable (do not merge)

cwilkers:
The current (2.8, 2.9) documentation is still offering a subscription with 2.5 as the release channel.

Test render required! I don't know whether the `{product-version}` substitution will work in a code block.

If this works, backports would be recommended since the current 2.8 docs still say 2.5.


cwilkers:
Created ACM-8282 to track this bug

dockerymick:
Related issue: https://issues.redhat.com/browse/ACM-8282

@swopebe thanks again!

PR #5695, opened 2023-10-26
https://issues.redhat.com/browse/ACM-7819 CIM Add hosts

oafischer:
Does it make sense to put agent in backticks? -> `agent` resource

Should this go before or after the import topic?


jelkosz:
UI part looks good to me

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -0,0 +1,39 @@
+[#add-host-host-inventory]
+= Adding hosts to the host inventory by using the Discovery Image
+
+After creating your host inventory (infrastructure environment), you can discover your hosts and add them to your inventory. Download an ISO and attach it to each server with the method you choose, such as by using virtual media or writing the ISO to a USB drive.
```
jc-berger: I'm not sure what `infrastructure environment` is in parenthesis. If it's simply another name for `host inventory` I think there's a clearer way to convey that. 

For example:

After creating your host inventory, or infrastructure environment, you can discover your hosts and add them to your inventory.

jc-berger: I wasn't quite sure of the flow of these sentences. Am I correct with these transitions I've added? 

`
After creating your host inventory, or infrastructure environment, you can discover your hosts and add them to your inventory. To add hosts to your inventory, choose a method to download an ISO and attach it to each server. For example, you can download ISOs by using a virtual media or writing the ISO to a USB drive.
`

What I wrote is wordier, so we can define fine tune it. But I think it needs some transitions to help it flow. What do you think?

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+*Important:* Keep the Discovery ISO media connected to the device during the installation process and set each host to boot from the device one time.
```
jc-berger: Usually if we're adding an *Important* message, we'd want to give a `why`. 

Why is it important for users to keep the Discovery ISO media connected to the device?

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+[#add-host-prereqs]
+== Prerequisites
+
+- You must enable the central infrastructure management service. See _Enabling the central infrastructure management service_ for more information.
```
jc-berger, line 11: Optional:

For minimalism, you can eliminated `you must` and simply say:

-- Enable the central infrastructure...
-- Create a host inventory...

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+Download the ISO by completing the following steps:
+
+. Select *Infrastructure* > *Host inventory* in the console.
+
+. Select your infrastructure environment from the list.
+
+. Click *Add hosts* and select *With Discovery ISO*.
+
+You now see a URL to download the ISO. Booted hosts appear in the host inventory table. This might take a few minutes. You must approve each host before you can use it. You can select hosts from the inventory table by clicking *Actions* and selecting *Approve*.
```
jc-berger: Does `this` refer to the hosts appearing? 

If so, what do you think of:

`The host might take a few minutes to appear.`

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -0,0 +1,39 @@
...
+
+[#add-host-steps-cli]
+== Adding hosts by using the command line interface
+
+You can see the URL to download the ISO in the `isoDownloadURL` property in the status of your `InfraEnv` resource. See xref:cim_create_cli.adoc#create-host-inventory-cli[Creating a host inventory by using the command line interface] for more information about the `InfraEnv` resource.
```
jc-berger: Why aren't these links added under `Additional resources`? I think they're fine where they are. But since we have an `Additional resources` section, and you say to go to these links only if users want to learn more, I'm not sure why these links/xrefs aren't in the `Additional resources`? 

Also for clickability and findability, I think links should be at the end. So, 

For more information about the `InfraEnv` resource, see...

To learn more about the agent resource, see...

clusters/cluster_lifecycle/cim_add_host.adoc
oafischer: I like your version, I think I'll use that. Clarity is more important than conciseness I would say.

I see the problem with the parenthesis, though I think it might be better to say (also called infrastructure environment) instead, as the example you gave might sound like you might have created one or the other, which suggests they are different things. Does that work?

Ideally we'd only have one term to avoid confusion but that's just how it is at the moment.

clusters/cluster_lifecycle/cim_add_host.adoc
oafischer: @avishayt ?

clusters/cluster_lifecycle/cim_add_host.adoc
oafischer: I'll go with optional on this one to stay consistent with the rest of the CIM book but I like that our team is thinking along similar lines lol

clusters/cluster_lifecycle/cim_add_host.adoc
oafischer: Yep, will fix it, thanks!

clusters/cluster_lifecycle/cim_add_host.adoc
oafischer: Good point, I'll move them to the additional resources section.

clusters/cluster_lifecycle/cim_add_host.adoc
avishayt: If the ISO is disconnected then installation will fail

oafischer:
Thanks for the reviews, made some updates, let me know if there is anything else!

jc-berger (approved):
/lgtm

oafischer:
Does this look good? @avishayt @nmagnezi

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+Download the ISO by completing the following steps:
+
+. Select *Infrastructure* > *Host inventory* in the console.
+
+. Select your infrastructure environment from the list.
+
+. Click *Add hosts* and select *With Discovery ISO*.
+
+You now see a URL to download the ISO. Booted hosts appear in the host inventory table. Hosts might take a few minutes to appear.. You must approve each host before you can use it. You can select hosts from the inventory table by clicking *Actions* and selecting *Approve*.
```
avishayt:
```suggestion
You now see a URL to download the ISO. Booted hosts appear in the host inventory table. Hosts might take a few minutes to appear. You must approve each host before you can use it. You can select hosts from the inventory table by clicking *Actions* and selecting *Approve*.
```

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -0,0 +1,41 @@
...
+
+[#add-host-steps-cli]
+== Adding hosts by using the command line interface
+
+You can see the URL to download the ISO in the `isoDownloadURL` property in the status of your `InfraEnv` resource. See Creating a host inventory by using the command line interface for more information about the `InfraEnv` resource.
+
+Each booted host creates an agent resource in the same namespace. You must approve each host before you can use it. See to _Managing host inventories_ by using the agent resource learn more about the agent resource.
```
avishayt: > See to _Managing host inventories_ by using the agent resource learn more about the agent resource.

Looks like something got jumbled in this sentence.  Also, the agent resource will be documented as part of ACM-8234.  Maybe leave this sentence out now and add it back in the PR for ACM-8234?

oafischer: Fixed and I removed it for now as well. We'll just have to remember to add it back to this section when working on ACM-8234.

PR #5705, opened 2023-10-27
https://issues.redhat.com/browse/ACM-8258

jc-berger:
For 2.9. 

See Jira: https://issues.redhat.com/browse/ACM-8258

### DIFF
--- a/governance/policy_ctrl_adv_config.adoc
+++ b/governance/policy_ctrl_adv_config.adoc
@@ -59,11 +59,16 @@ spec:
 [#configure-debug-log] 
 == Configure debug log 
 
-When you configure and collect debug logs for each policy controller, the default log level might take up more space than you want. In {product-title-short} 2.9, you can reduce the volume of debug logs. You can reduce the volume below the default setting. 
+When you configure and collect debug logs for each policy controller, you can adjust the log level. 
 
-*Note:* Reducing the volume of debug logs can make issues more difficult to debug. 
+*Note:* Reducing the volume of debug logs means there is less information displayed from the logs.
 
-You can reduce the debug logs emitted by the policy controllers to be `error-only`. To reduce the debug logs, set the debug log value to `-1` in the annotation. 
+You can reduce the debug logs emitted by the policy controllers to be display error-only bugs in the logs. To reduce the debug logs, set the debug log value to `-1` in the annotation. See what each value represents:
+
+- `-1`: error logs only. 
+- `0`: informative logs. 
+- `1`: debug logs. 
+- `2`: verbose debugging logs. 
 
 To receive the second level of debugging information for the Kubernetes configuration controller, add the `log-level` annotation with the value of `2` to the `ManagedClusterAddOn` custom resource. By default, the `log-level` is set to `0`, which means you receive informative messages. View the following example:  
 
--- a/governance/policy_ctrl_adv_config.adoc
+++ b/governance/policy_ctrl_adv_config.adoc
@@ -59,7 +59,13 @@ spec:
 [#configure-debug-log] 
 == Configure debug log 
 
-Configure debug logs for each policy controller. To receive the second level of debugging information for the Kubernetes configuration controller, add the `log-level` annotation with the value of `2` to the `ManagedClusterAddOn` custom resource. By default, the `log-level` is set to `0`, which means you receive informative messages. View the following example:  
+When you configure and collect debug logs for each policy controller, the default log level might take up more space than you want. In {product-title-short} 2.9, you can reduce the volume of debug logs. You can reduce the volume below the default setting. 
+
+*Note:* Reducing the volume of debug logs can make issues more difficult to debug. 
+
+You can reduce the debug logs emitted by the policy controllers to be `error-only`. To reduce the debug logs, set the debug log value to `-1` in the annotation. 
+
+To receive the second level of debugging information for the Kubernetes configuration controller, add the `log-level` annotation with the value of `2` to the `ManagedClusterAddOn` custom resource. By default, the `log-level` is set to `0`, which means you receive informative messages. View the following example:  
 
 [source,yaml]
 ----
--- a/governance/policy_ctrl_adv_config.adoc
+++ b/governance/policy_ctrl_adv_config.adoc
@@ -65,10 +65,10 @@ When you configure and collect debug logs for each policy controller, you can ad
 
 You can reduce the debug logs emitted by the policy controllers to be display error-only bugs in the logs. To reduce the debug logs, set the debug log value to `-1` in the annotation. See what each value represents:
 
-- `-1`: error logs only. 
-- `0`: informative logs. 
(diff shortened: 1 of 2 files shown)


governance/policy_ctrl_adv_config.adoc
```diff
@@ -59,7 +59,13 @@ spec:
 [#configure-debug-log] 
 == Configure debug log 
 
-Configure debug logs for each policy controller. To receive the second level of debugging information for the Kubernetes configuration controller, add the `log-level` annotation with the value of `2` to the `ManagedClusterAddOn` custom resource. By default, the `log-level` is set to `0`, which means you receive informative messages. View the following example:  
+When you configure and collect debug logs for each policy controller, the default log level might take up more space than you want. In {product-title-short} 2.9, you can reduce the volume of debug logs. You can reduce the volume below the default setting. 
+
+*Note:* Reducing the volume of debug logs can make issues more difficult to debug.
```
dockerymick: I recommend that this note be removed. It seems like this should be known issue. What issues might the use encounter? @JustinKuli

dockerymick (changes requested):

governance/policy_ctrl_adv_config.adoc
JustinKuli: I wouldn't call it a "known issue" - it feels different to me than what we have in that section. This is just a risk that the user should be aware of, but it is a natural consequence of this configuration. When configuring the log level, the user is making a decision between being able to debug issues more accurately (higher level) or not having too many logs cluttering up their system (lower level).

In my mind, it fits in the same category of the user configuring the `evaluationInterval` for a policy; by increasing the time between evaluations, they may save some CPU cycles, but they should understand that their compliance history is less accurate (the cluster might be noncompliant for longer before the policy notices). Both are natural trade-offs.

I think usually the user will not be debugging problems themselves by digging through their logs. But if their log level is too low when they have a problem, they would just need to re-configure it, and recreate whatever problem they previously had.

governance/policy_ctrl_adv_config.adoc
```diff
@@ -59,7 +59,13 @@ spec:
 [#configure-debug-log] 
 == Configure debug log 
 
-Configure debug logs for each policy controller. To receive the second level of debugging information for the Kubernetes configuration controller, add the `log-level` annotation with the value of `2` to the `ManagedClusterAddOn` custom resource. By default, the `log-level` is set to `0`, which means you receive informative messages. View the following example:  
+When you configure and collect debug logs for each policy controller, the default log level might take up more space than you want. In {product-title-short} 2.9, you can reduce the volume of debug logs. You can reduce the volume below the default setting.
```
JustinKuli: This leads with the new feature, but I think it could be more general. Something like "The default log level might cause fewer or more logs than you want. In {product-title-short} 2.9, you can either increase or reduce the volume of debug logs from the default setting."

governance/policy_ctrl_adv_config.adoc
```diff
@@ -59,7 +59,13 @@ spec:
 [#configure-debug-log] 
 == Configure debug log 
 
-Configure debug logs for each policy controller. To receive the second level of debugging information for the Kubernetes configuration controller, add the `log-level` annotation with the value of `2` to the `ManagedClusterAddOn` custom resource. By default, the `log-level` is set to `0`, which means you receive informative messages. View the following example:  
+When you configure and collect debug logs for each policy controller, the default log level might take up more space than you want. In {product-title-short} 2.9, you can reduce the volume of debug logs. You can reduce the volume below the default setting. 
+
+*Note:* Reducing the volume of debug logs can make issues more difficult to debug. 
+
+You can reduce the debug logs emitted by the policy controllers to be `error-only`. To reduce the debug logs, set the debug log value to `-1` in the annotation.
```
JustinKuli: "error-only" isn't a name/code/setting thing, so I would prefer it isn't formatted like that.
```suggestion
You can reduce the debug logs emitted by the policy controllers to just be when an error occurs. To reduce the debug logs, set the debug log value to `-1` in the annotation. 
```

governance/policy_ctrl_adv_config.adoc
dockerymick: @JustinKuli thanks for explaining that. So it seems like users would only know if the log level is too low through trial and error? 

```suggestion
*Note:* Reducing the volume of debug logs means there is less information displayed from the logs.
```

governance/policy_ctrl_adv_config.adoc
dockerymick: Whats the log level range? so far i see [-1 to 2]

-1: error-only logs
0: informative messages
1: first level of debugging information
2: second level of debugging information

Would it be helpful to list what the values represent? 

```suggestion
You can reduce the debug logs emitted by the policy controllers to be display error-only bugs in the logs. To reduce the debug logs, set the debug log value to `-1` in the annotation. 
```

governance/policy_ctrl_adv_config.adoc
dockerymick: Remember to update the What's new - Governance topic: "You can now increase or reduce the volume of debug logs from the default settings. For more details, see Configure debug log"

```suggestion
When you configure and collect debug logs for each policy controller, you can adjust the log level. 
```

dockerymick (changes requested):

governance/policy_ctrl_adv_config.adoc
JustinKuli: Yes.

governance/policy_ctrl_adv_config.adoc
JustinKuli: I'd call them:
-1: error logs only
0: informative logs
1: debug logs
2: verbose debugging logs

I don't think we directly write any logs at a higher level than that, although the user could specify a higher value, and some libraries we use might emit more logs.

governance/policy_ctrl_adv_config.adoc
```diff
@@ -59,11 +59,16 @@ spec:
...
-*Note:* Reducing the volume of debug logs can make issues more difficult to debug. 
+*Note:* Reducing the volume of debug logs means there is less information displayed from the logs.
 
-You can reduce the debug logs emitted by the policy controllers to be `error-only`. To reduce the debug logs, set the debug log value to `-1` in the annotation. 
+You can reduce the debug logs emitted by the policy controllers to be display error-only bugs in the logs. To reduce the debug logs, set the debug log value to `-1` in the annotation. See what each value represents:
+
+- `-1`: error logs only. 
+- `0`: informative logs. 
+- `1`: debug logs. 
+- `2`: verbose debugging logs.
```
dockerymick: I like that you were consistent with the punctuation! However, I recommend that you remove periods since these are complete sentences.

dockerymick (approved):
One more suggestion. Otherwise, this looks good!

JustinKuli (approved):
LGTM

PR #5734, opened 2023-10-31
https://issues.redhat.com/browse/ACM-6906--removing content that is re…

swopebe:
…dundant, reducing files, adding links to support matrixes that may need to change


clusters/about/requirements.adoc
```diff
@@ -1,8 +0,0 @@
-[#requirements-and-recommendations]
-= Requirements and recommendations
```
swopebe, line 2: this file is silly :-) Moving this info to install overview.

install/requirements.adoc
```diff
@@ -1,33 +0,0 @@
-[#requirements-and-recommendations]
-= Requirements and recommendations
```
swopebe, line 2: https://redhat-internal.slack.com/archives/CUHCCV19P/p1695847285014559

swopebe: we already say this

cadawson (changes requested):
Looks good. Trouble is we will have to find out how to update the link on the product page to the latest versions of the support matrix if we want to use that link. Or we could use a conref for the link directly to the support matrix. Or we could hardcode it. We can figure something out.

PR #5741, opened 2023-11-02
[ACM-5456]: Updates to AWS docs in 2.8

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-5456

### DIFF
--- a/clusters/hosted_control_planes/configure_hosted_aws.adoc
+++ b/clusters/hosted_control_planes/configure_hosted_aws.adoc
@@ -373,12 +373,19 @@ oc label secret hypershift-operator-private-link-credentials -n local-cluster cl
 
 The hosted control planes feature is disabled by default. Enabling the feature automatically also enables the `hypershift-addon` managed cluster add-on. 
 
-. You can run the following command to enable the feature:
+. You can run the following commands to enable the feature:
 +
 ----
 oc patch mce multiclusterengine --type=merge -p '{"spec":{"overrides":{"components":[{"name":"hypershift-preview","enabled": true}]}}}' <1>
 ----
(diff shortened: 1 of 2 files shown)


clusters/hosted_control_planes/configure_hosted_aws.adoc
```diff
@@ -373,13 +373,20 @@ oc label secret hypershift-operator-private-link-credentials -n local-cluster cl
...
+. You can run the following commands to enable the feature:
 +
 ----
 oc patch mce multiclusterengine --type=merge -p '{"spec":{"overrides":{"components":[{"name":"hypershift-preview","enabled": true}]}}}' <1>
 ----
 <1> The default `MultiClusterEngine` resource instance name is `multiclusterengine`, but you can get the `MultiClusterEngine` name from your cluster by running the following command: `$ oc get mce`.
 
++
+----
+oc patch mce multiclusterengine --type=merge -p '{"spec":{"overrides":{"components":[{"name":"hypershift-local-hosting","enabled": true}]}}}' <1>
```
lahinson, line 395: I'm not sure if we need to include this command separately or simply modify the command on line 379. @bryan-cox When you can, let me know. Thanks!

bryan-cox: I thought we didn't need the command anymore on L379? Is that true @rokej?

rokej: This is 2.8 doc where this step is still necessary.

bryan-cox: Ok thanks Roke! I had my versions mixed up.

lgtm @lahinson

dockerymick (approved):
Similar to the PR for the 2.9, recommend adding the `source` syntax. It's not required tho, this will help with consistency. It's more important to have the `source` syntax for 2.9

PR #5753, opened 2023-11-03
[ACM-7575]: Adding sizing guidance for hosted control planes

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-7575


clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -28,24 +28,128 @@ Hosted control plane clusters offer several advantages:
 
 * Supports turn-key deployments or fully customized {ocp-short} provisioning
 
+[#hosted-sizing-guidance]
+== Hosted control plane sizing guidance
+
+Multiple factors, including hosted cluster workload and worker node count, come into play when you consider how many hosted clusters can fit within a certain number of control-plane nodes. Use this sizing guide to help with hosted cluster capacity planning. This guidance assumes a highly available hosted control plane topology. For more information about that topology, see _Distributing hosted cluster workloads_.
+
```
lahinson, line 35: @swopebe - I _think_ you added the following requirements (lines 36 - 41) to the 2.8 docs for HyperShift. Do you remember where you got that info from? The folks on the CNV team provided the additional guidance that I'm adding in this PR, and they wanted to know where the existing guidance was from.

clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -28,24 +28,122 @@ Hosted control plane clusters offer several advantages:
...
 See the following highly available hosted control plane requirements, which were tested with {ocp-short} version 4.12.9 and later:
 
 * 78 pods
 * Three 4 Gi PVs for etcd, three 1 Gi PVs for OVN
```
gqlo: By the way, these numbers are different in 4.14, we don't need pvs for OVN-IC. and etcd pv size has changed to be 8Gi, pod count is 75.

clusters/hosted_control_planes/hosted_intro.adoc
jeniferh: We are discussing if this section (L36-41) should be removed since some info is now outdated and we cover the total request sizing in a section below with slightly lower values based on what we measured in the latest version.

But one point I want to make sure is clear, this load based sizing was evaluated for on-prem since Cloud instance sizes can have much different limit behavior (i.e. typically memory constrained), in that case is it better suited for configure_hosted_bm.adoc?

lahinson: @jeniferh If some of the guidance is specific to hosted control planes on bare-metal platforms, then yes, we'd move that guidance to the bare metal intro, `configure_hosted_bm.adoc`. If it's a matter of on-prem vs. cloud, I think it still makes sense to leave the guidance here (`hosted_intro.adoc`) since this file is the overall intro to self-managed hosted control planes. cc @zanetworker Please correct me if I'm wrong 😅

lahinson:
Hi @rokej - When you can, PTAL at this PR, which adds additional sizing guidance to the hosted control planes intro. In particular, please see the discussion around lines 36 - 41. I'm not sure whether that guidance should stay in the 2.9 docs or if it can be replaced by the new, extended guidance. Feel free to reach out if you have any questions. Thank you!

clusters/hosted_control_planes/hosted_intro.adoc
zanetworker: @lahinson I do think this infromation is relevant for cloud environments but let's keep it in the on-prem section for now.

clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -28,24 +28,128 @@ Hosted control plane clusters offer several advantages:
...
 * Minimum CPU: approximately 5.5 cores
 * Minimum memory: approximately 19 GB
 
+[#hosted-sizing-guidance-pod-limit]
+=== Pod limits
+
+The `maxPods` setting for each node affects how many hosted clusters can fit in a control-plane node. It is important to note the `maxPods` value on all control-plane nodes. Plan for about 75 pods for each highly available hosted control plane.
+
+For bare-metal nodes, set `maxPods` to 500 by configuring the `KubeletConfig` values. For more information, see _Configuring the maximum number of pods per node_ in the {ocp-short} documentation.
```
zanetworker: @jeniferh @gqlo do we have data on the default use-case, 250 pods, otherwise this becomes an additional requirement pre-install.

jeniferh: with 250 pod limit it is almost always going to be the limiting factor, only about 3 HCPs per worker node regardless of resource size

gqlo: @zanetworker the hcp resource usage data stays the same when using 250 default limit. We only created a few guest clusters, it’s not reaching the default 250 limit.

clusters/hosted_control_planes/hosted_intro.adoc
lahinson: @zanetworker So should I move this info from the overall hosted control planes intro to the intro for bare metal? (I'm assuming by "on-prem" you mean "bare metal", but then again, assumptions are dangerous 😄)

jeniferh: Yes on-prem == bare metal :) we tend to use them interchangeably in the convos I have been in

jeniferh: So if all the new sizing doc text is moved to `configure_hosted_bm.adoc`, looks like the bullets covered in this intro section (currently L36-41) do need at least one update:
 * 78 pods --> this is probably fine, we measured 75 pods but this is just a bit more conservative and can apply to older versions
 * Three 4 Gi PVs for etcd, three 1 Gi PVs for OVN --> starting in 4.14 this should be "Three 8 Gi PVs for etcd", the size was bumped and there are no longer OVN PVCs w/ OVN-IC
 * The other two "Minimum" bullets are probably fine as is, they say "approximately" and are just slightly more conservative than what we measured.

zanetworker: on-prem is baretmetal or CNV (virt provider), AWS would then be cloud.

clusters/hosted_control_planes/hosted_intro.adoc
zanetworker: great, then should just hint that if we need more, then we need to reconfigure. The way I read it was that by default we need 500 max pods to consume this sizing guide.

clusters/hosted_control_planes/hosted_intro.adoc
lahinson: @zanetworker Thanks for the clarification about what exactly "on-prem" refers to. Since we're adding IBM Z and IBM Power as platforms in this release, would they be considered under the "on-prem" umbrella, too?

zanetworker: Correct :)

clusters/hosted_control_planes/hosted_intro.adoc
jeniferh: Would this change help? @zanetworker 

> For bare-metal nodes, the default `maxPods` setting of 250 is likely to be a limiting factor since roughly 3 hosted control planes would fit per node given the pod requirements, even if the machine had plenty of resources to spare. Setting `maxPods` to 500 by configuring the `KubeletConfig` value allows for greater hosted control plane density to take advantage of additional compute resources. For more information, see _Configuring the maximum number of pods per node_ in the {ocp-short} documentation.

clusters/hosted_control_planes/hosted_intro.adoc
lahinson: Thanks for the clarification! In that case, since this sizing guidance applies to all platforms except AWS, I will keep the guidance in the hosted_intro.adoc file but add a note to state that the guidance applies to on-premise deployments only.

jeniferh: Expanding on this sentence in the intro paragraph might help, a suggestion:

> This guidance assumes a highly available hosted control plane topology. The load-based sizing examples below were measured on a bare-metal cluster, note that Cloud-based instances may have different limiting factors (for instance, memory size).

clusters/hosted_control_planes/hosted_intro.adoc
zanetworker: @jeniferh lgtm :)

clusters/hosted_control_planes/hosted_intro.adoc
lahinson: @jeniferh Thanks! I added that wording to the intro.

clusters/hosted_control_planes/hosted_intro.adoc
lahinson: I updated the paragraph per @jeniferh's latest suggestion.

rokej:
@lahinson I think we can leave line 36 - 41 as a guideline.

lahinson:
Thanks for your feedback, @zanetworker and @jeniferh. When you can, let me know whether this ready to move along for peer review and merging.

jeniferh (approved):
Thanks for all the changes, looks great to me.

clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -28,26 +28,130 @@ Hosted control plane clusters offer several advantages:
 
 * Supports turn-key deployments or fully customized {ocp-short} provisioning
 
+[#hosted-sizing-guidance]
+== Hosted control plane sizing guidance
+
+Multiple factors, including hosted cluster workload and worker node count, come into play when you consider how many hosted clusters can fit within a certain number of control-plane nodes. Use this sizing guide to help with hosted cluster capacity planning. This guidance assumes a highly available hosted control plane topology. The load-based sizing examples were measured on a bare-metal cluster. Cloud-based instances might have different limiting factors, such as memory size. For more information about highly available hosted control plane topology, see _Distributing hosted cluster workloads_.
```
jc-berger: Optional

We're not obligated to list every possible factor. Therefore, to be concise and minimal, I think you can simply say something like: 

Hosted cluster workload and worker node count can impact how many hosted clusters you can fit within a certain number of control-plane nodes.

jc-berger: Optional: 

Links, even to other sections of our RHACM docs, can take users _away from the current doc_. Therefore, I prefer to use links at the end of the doc. 

At the end of this doc, can you have a section, `Additional resources`. 

In this Additional resources section, you can include this sentence: 

For more information about highly available hosted control plane topology, see _Distributing hosted cluster workloads_. 

However, I trust that you know the user experience for this task better than I. So, if you think it better serves the user to make them aware of the `Distributing hosted cluster workloads` section at the beginning of this doc, by all means, keep this sentence/reference where you currently have it :)

clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -28,26 +28,130 @@ Hosted control plane clusters offer several advantages:
...
+* Three 8 Gi PVs for etcd
 * Minimum CPU: approximately 5.5 cores
 * Minimum memory: approximately 19 GB
 
+[#hosted-sizing-guidance-pod-limit]
+=== Pod limits
+
+The `maxPods` setting for each node affects how many hosted clusters can fit in a control-plane node. It is important to note the `maxPods` value on all control-plane nodes. Plan for about 75 pods for each highly available hosted control plane. 
+
+For bare-metal nodes, the default `maxPods` setting of 250 is likely to be a limiting factor because roughly three hosted control planes fit per node given the pod requirements, even if the machine has plenty of resources to spare. Setting the `maxPods` value to 500 by configuring the `KubeletConfig` value allows for greater hosted control plane density to take advantage of additional compute resources. For more information, see _Configuring the maximum number of pods per node_ in the {ocp-short} documentation.
```
jc-berger: Minor, but IBM SG encourages us to avoid `per` so I would say, `three hosted control planes fit for each node given in the pod requirements` 

Note, I do think we are allowed to use Latin words like `per` on occasion, so you might not need to remove it here.

jc-berger: Minor, but you could make it more user focused here:

`Setting the `maxPods` value to 500 by configuring the `KubeletConfig` value allows for greater hosted control plane density, helping you to take  advantage of additional compute resources.`

jc-berger: Same note above with links. Again, optional :)

jc-berger (approved):
/lgtm

clusters/hosted_control_planes/hosted_intro.adoc
lahinson: Yep - I agree with you! That's why I added an "Additional resources" heading at the end of this section with the links. The reference in this paragraph is not a link, and I while I think it's worth mentioning at this point in the doc, I agree that we don't need a link here.

lahinson: I compromised between your suggestion and the dev's wording to say this:
>Many factors, including hosted cluster workload and worker node count, affect how many hosted clusters can fit within a certain number of control-plane nodes.

clusters/hosted_control_planes/hosted_intro.adoc
lahinson: I like your suggestion to make the maxPods sentence more user-focused. I adjusted that line as you suggested.

PR #5756, opened 2023-11-06
https://issues.redhat.com/browse/ACM-8396

oafischer:


clusters/release_notes/deprecate_remove.adoc
```diff
@@ -46,6 +46,12 @@ A _deprecated_ component, feature, or service is supported, but no longer recomm
 | Product or category | Affected item | Version | Recommended action | More details and links
 |===
 
+| Cluster lifecycle
+| Klusterlet OLM Operator
+| 2.9
```
showeimer: I think the Klusterlet OLM Operator is versioned the same as MCE, so this would be 2.4 - correct me if I'm wrong @elgnay

oafischer: Good catch, updated!

elgnay: Yes, the version aligns with the MCE version.

oafischer:
Updated, just waiting on peer review now.

cadawson (approved):
Looks good. Is there really no alternative that we can suggest there? Maybe the import is not the only place where it is used and we can't suggest somewhere else in the deprecation statement. If no alternative, so be it. Thanks!

PR #5761, opened 2023-11-06
Updates to HyperShift doc per chat with Brandi - Nov. 6

lahinson:
No Jira. These changes are based on conversations with Brandi on Nov. 6 about improving the HyperShift doc experience.


swopebe (approved):
thank you so much

PR #5775, opened 2023-11-07
No Jira: Fixing issues in the Clusters book - Nov. 7

lahinson:


lahinson:
FYI @oafischer - I commented out a few references to `scale-hosts-infrastructure-env` in this PR, as the file that contains that ID doesn't seem to be in the repo. Please feel free to add those references back whenever the file is in the repo. Let me know if you have any questions for me.

PR #5800, opened 2023-11-09
https://issues.redhat.com/browse/ACM-8497

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-8497

### DIFF
--- a/troubleshooting/trouble_cinder_csi_driver_volsync.adoc
+++ b/troubleshooting/trouble_cinder_csi_driver_volsync.adoc
@@ -6,14 +6,13 @@ If you use VolSync or use a default setting in a cinder Container Storage Interf
 [#symptom-volumesnapshot-error-state]
 == Symptom: `Volumesnapshot` error state
 
-You can configure a VolSync `ReplicationSource` or `ReplicationDestination` to use snapshots. Also, you can configure the `storageclass` and `volumesnapshotclass` in the `ReplicationSource` and `ReplicationDestination`. There is a parameter on the `volumesnapshotclass` called `force-create` with a default value of `false`. This `force-create` parameter on the `volumesnapshotclass` means cinder does not allow the `volumesnapshot` to be taken of a PVC in use. As a result, the `volumesnapshot` is in an error state. 
+You can configure a VolSync `ReplicationSource` or `ReplicationDestination` to use snapshots. Also, you can configure the `storageclass` and `volumesnapshotclass` in the `ReplicationSource` and `ReplicationDestination`. There is a parameter on the cinder  `volumesnapshotclass` called `force-create` with a default value of `false`. This `force-create` parameter on the `volumesnapshotclass` means cinder does not allow the `volumesnapshot` to be taken of a PVC in use. As a result, the `volumesnapshot` is in an error state. 
 
 [#resolving-setting-the-parameter-to-true]
 == Resolving the problem: Setting the parameter to `true`
 
 . Create a new `volumesnapshotclass` for the cinder CSI driver.
 . Change the paramater, `force-create`, to `true`. See the following sample YAML:
-
 
 +
 [source,yaml]
--- a/troubleshooting/trouble_cinder_csi_driver_volsync.adoc
+++ b/troubleshooting/trouble_cinder_csi_driver_volsync.adoc
@@ -1,0 +1,32 @@
+[#troubleshooting-the-cinder-csi-driver-for-volsync]
+= Troubleshooting the cinder Container Storage Interface (CSI) driver for VolSync
+
+If you use VolSync or use a default setting in a cinder Container Storage Interface (CSI) driver, you might encounter errors. 
+
+[#symptom-volumesnapshot-error-state]
+== Symptom: `Volumesnapshot` error state
+
+If you use the VolSync, it can create a `replicationsource/destination` that uses a cinder CSI driver. Within the cinder CSI driver, there is a parameter on the `volumesnapshotclass` called `force-create` with a default value of `false`. This `force-create` parameter on the `volumesnapshotclass` means cinder does not allow the `volumesnapshot` to be taken of a PVC in use. As a result, the `volumesnapshot` is in an error state. 
+
+[#resolving-setting-the-parameter-to-true]
+== Resolving the problem: Setting the parameter to `true`
+
+. Create a new `volumesnapshotclass`.
+.. *Optional*: If you prefer, update an existing `volumesnapshot`.
+. Change the paramater, `force-create`, to `true`. See the following sample YAML:
++
+[source,yaml]
+----
+apiVersion: snapshot.storage.k8s.io/v1
+deletionPolicy: Delete
+driver: cinder.csi.openstack.org
+kind: VolumeSnapshotClass
+metadata:
+  annotations:
+    snapshot.storage.kubernetes.io/is-default-class: 'true'
+  name: standard-csi
+parameters:
+  force-create: 'true'
+----
+
+For more information, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html-single/storage/index#persistent-storage-csi-snapshots[CSI Volume Snapshots]. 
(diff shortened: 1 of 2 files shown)


troubleshooting/trouble_cinder_csi_driver_volsync.adoc
```diff
@@ -0,0 +1,32 @@
+[#troubleshooting-the-cinder-csi-driver-for-volsync]
+= Troubleshooting the cinder Container Storage Interface (CSI) driver for VolSync
+
+If you use VolSync or use a default setting in a cinder Container Storage Interface (CSI) driver, you might encounter errors. 
+
+[#symptom-volumesnapshot-error-state]
+== Symptom: `Volumesnapshot` error state
+
+If you use the VolSync, it can create a `replicationsource/destination` that uses a cinder CSI driver. Within the cinder CSI driver, there is a parameter on the `volumesnapshotclass` called `force-create` with a default value of `false`. This `force-create` parameter on the `volumesnapshotclass` means cinder does not allow the `volumesnapshot` to be taken of a PVC in use. As a result, the `volumesnapshot` is in an error state.
```
jc-berger: In the Jira, I didn't see an exact correlation between the `replicationsource/destination` inside the CSI driver and the `volumesnapshotclass`. Therefore, it's my inference that this `volumesnapshotclass` is insie the CSI driver. 

Please correct me if I am incorrect.

troubleshooting/trouble_cinder_csi_driver_volsync.adoc
```diff
@@ -0,0 +1,32 @@
...
+apiVersion: snapshot.storage.k8s.io/v1
+deletionPolicy: Delete
+driver: cinder.csi.openstack.org
+kind: VolumeSnapshotClass
+metadata:
+  annotations:
+    snapshot.storage.kubernetes.io/is-default-class: 'true'
+  name: standard-csi
+parameters:
+  force-create: 'true'
```
jc-berger, line 29: In the Jira, the sample YAML I was given had the value set to `false`. 

However, I changed it to `true` because that's what we tell users to do as part of the troubleshooting. 

Please correct me if I am missing something.

troubleshooting/trouble_cinder_csi_driver_volsync.adoc
```diff
@@ -0,0 +1,32 @@
...
+----
+
+For more information, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html-single/storage/index#persistent-storage-csi-snapshots[CSI Volume Snapshots].
```
jc-berger: There was some back and forth about what link, if any, we'd provide users. Reviewers, please let me know what you think.

troubleshooting/trouble_cinder_csi_driver_volsync.adoc
```diff
@@ -0,0 +1,32 @@
...
+[#symptom-volumesnapshot-error-state]
+== Symptom: `Volumesnapshot` error state
+
+If you use the VolSync, it can create a `replicationsource/destination` that uses a cinder CSI driver. Within the cinder CSI driver, there is a parameter on the `volumesnapshotclass` called `force-create` with a default value of `false`. This `force-create` parameter on the `volumesnapshotclass` means cinder does not allow the `volumesnapshot` to be taken of a PVC in use. As a result, the `volumesnapshot` is in an error state. 
+
+[#resolving-setting-the-parameter-to-true]
+== Resolving the problem: Setting the parameter to `true`
+
+. Create a new `volumesnapshotclass`.
+.. *Optional*: If you prefer, update an existing `volumesnapshot`.
```
tesshuflower: I think we need to remove this "update an existing volumesnapshot" as nothing is done at the volumesnapshot level - this is a setting in the volumesnapshotclass - maybe instead we can say create a new `volumesnapshotclass` for cinder-csi or update the existing one.

troubleshooting/trouble_cinder_csi_driver_volsync.adoc
tesshuflower: You got it!

troubleshooting/trouble_cinder_csi_driver_volsync.adoc
tesshuflower: I don't think we should use this link, as this is just generic documentation about CSI.  This configuration setting is not a generic CSI thing, but something specific to the cinder implementation.  The only place I can find mention of cinder setting this is the original link upstream here: https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/cinder-csi-plugin/using-cinder-csi-plugin.md

troubleshooting/trouble_cinder_csi_driver_volsync.adoc
tesshuflower: I'm not sure the best way to explain it, but a volsync `ReplicationSource` or `ReplicationDestination` can be configured to use snapshots - you can configure the storageclass & volumesnapshotclass in the ReplicationSource/ReplicationDestination - or you can let the system decide (i.e. whatever is set as the system default).

Essentially it boils down to: if you're using VolSync to create volumesnapshots using a cinder-csi driver - you could encounter this error for PVCs that are in use.

Note that this can also happen without VolSync (if you just create the volumesnapshot yourself) using that same cinder-csi driver.

troubleshooting/trouble_cinder_csi_driver_volsync.adoc
swopebe: Pause on using this link...is this repo validated by QE? As technical writers, we cannot link to random open repos and such because they can change without notice...that is why it is better to link to official documentation. 

Let's get a QE-ack at minimum on this link so that is support opens an issue, we can at least say this was tested. 

This is RH standard.

troubleshooting/trouble_cinder_csi_driver_volsync.adoc
```diff
@@ -0,0 +1,32 @@
...
+. Change the paramater, `force-create`, to `true`. See the following sample YAML:
++
+[source,yaml]
+----
```
swopebe, line 19: @jc-berger this looks great but there is a format issue on this PR, line 18 and the rest of YAML should be grayed out. We usually need an extra space to fix that, so if you don't see it grayed out, just see if adding a space helps. I will do that now.

swopebe (changes requested):

troubleshooting/trouble_cinder_csi_driver_volsync.adoc
swopebe: yes, that was the fix. All good now.

troubleshooting/trouble_cinder_csi_driver_volsync.adoc
jc-berger: Please see the revised intro and let me know what you think.

troubleshooting/trouble_cinder_csi_driver_volsync.adoc
jc-berger: Okay, I removed the link entirely. I know a lot of RH writers, like my previous products, were pretty loath to include links to documentation outside our repo.

troubleshooting/trouble_cinder_csi_driver_volsync.adoc
jc-berger: Okay added an extra space before the + and yaml. 

Did you mean space in between the + and the yaml?

I wonder why the spacing was weird, because in our other files, it's just the one space before the yaml.

swopebe: well now you have two lol, I added one and it was fixed.

a break between a paragraph and this:
```
+
[source,yaml]
```

swopebe (approved):
looks great.

troubleshooting/trouble_cinder_csi_driver_volsync.adoc
```diff
@@ -0,0 +1,33 @@
+[#troubleshooting-the-cinder-csi-driver-for-volsync]
+= Troubleshooting the cinder Container Storage Interface (CSI) driver for VolSync
+
+If you use VolSync or use a default setting in a cinder Container Storage Interface (CSI) driver, you might encounter errors for the PVC that is in use. 
+
+[#symptom-volumesnapshot-error-state]
+== Symptom: `Volumesnapshot` error state
+
+You can configure a VolSync `ReplicationSource` or `ReplicationDestination` to use snapshots. Also, you can configure the `storageclass` and `volumesnapshotclass` in the `ReplicationSource` and `ReplicationDestination`. There is a parameter on the `volumesnapshotclass` called `force-create` with a default value of `false`. This `force-create` parameter on the `volumesnapshotclass` means cinder does not allow the `volumesnapshot` to be taken of a PVC in use. As a result, the `volumesnapshot` is in an error state.
```
tesshuflower: Just to be very clear, can we say "There is a parameter on the `cinder` volumesnapshotclass called force-create?"

PR #5801, opened 2023-11-09
https://issues.redhat.com/browse/ACM-7646

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-7646

Changes for 2.9 branch

### DIFF
--- a/governance/config_policy_ctrl.adoc
+++ b/governance/config_policy_ctrl.adoc
@@ -102,7 +102,9 @@ spec:
 
 | `spec.evaluationInterval.compliant`
 | Optional
-| Used to define how often the policy is evaluated when it is in the compliant state. The values must be in the format of a duration which is a sequence of numbers with time unit suffixes. For example, `12h30m5s` represents 12 hours, 30 minutes, and 5 seconds. It can also be set to `never` so that the policy is not reevaluated on the compliant cluster, unless the policy `spec` is updated. Configuration Policies are evaluated roughly every 10 seconds when the evaluation interval is not set. This can be longer if the Configuration Policy controller is saturated on the managed cluster. By default, the minimum time between evaluations for configuration policies is 10 seconds when the `evaluationInterval.compliant` is not set or empty.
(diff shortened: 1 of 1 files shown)


governance/config_policy_ctrl.adoc
```diff
@@ -102,7 +102,7 @@ spec:
 
 | `spec.evaluationInterval.compliant`
 | Optional
-| Used to define how often the policy is evaluated when it is in the compliant state. The values must be in the format of a duration which is a sequence of numbers with time unit suffixes. For example, `12h30m5s` represents 12 hours, 30 minutes, and 5 seconds. It can also be set to `never` so that the policy is not reevaluated on the compliant cluster, unless the policy `spec` is updated.
+| Used to define how often the policy is evaluated when it is in the compliant state. The values must be in the format of a duration which is a sequence of numbers with time unit suffixes. For example, `12h30m5s` represents 12 hours, 30 minutes, and 5 seconds. It can also be set to `never` so that the policy is not reevaluated on the compliant cluster, unless the policy `spec` is updated. Configuration Policies are evaluated roughly every 10 seconds when the evaluation interval is not set. This can be longer if the Configuration Policy controller is saturated on the managed cluster. By default, the minimum time between evaluations for configuration policies is 10 seconds when the `evaluationInterval.compliant` is not set or empty.
```
dockerymick: Instead of "roughly", let's use "approximately"

```suggestion
| Used to define how often the policy is evaluated when it is in the compliant state. The values must be in the format of a duration which is a sequence of numbers with time unit suffixes. For example, `12h30m5s` represents 12 hours, 30 minutes, and 5 seconds. It can also be set to `never` so that the policy is not reevaluated on the compliant cluster, unless the policy `spec` is updated. 

By default, the minimum time between evaluations for configuration policies is approximately 10 seconds when the `evaluationInterval.compliant` is not set or empty. This can be longer if the configuration policy controller is saturated on the managed cluster. 
```

dockerymick (approved):

dockerymick (approved):

PR #5808, opened 2023-11-10
https://issues.redhat.com/browse/ACM-8488 Metrics known issue

oafischer:
Also for 2.8

### DIFF
--- a/release_notes/known_issues_observability.adoc
+++ b/release_notes/known_issues_observability.adoc
@@ -173,7 +173,7 @@ Your query data appears from the Grafana dashboard.
 == Metrics collector does not detect proxy configuration
 //2.9:ACM-8488
 
-A proxy configuration in a managed cluster that you configure by using the `addonDeploymentConfig` is not detected by the metrics collector. As a workaround, you can enable the proxy by the removing the managed cluster `ManifestWork`. Removing the `ManifestWork` forces the changes in the `addonDeploymentConfig` to be applied.
+A proxy configuration in a managed cluster that you configure by using the `addonDeploymentConfig` is not detected by the metrics collector. As a workaround, you can enable the proxy by removing the managed cluster `ManifestWork`. Removing the `ManifestWork` forces the changes in the `addonDeploymentConfig` to be applied.
 
 [#proxy-ca-not-supported]
 == HTTPS proxy with a custom CA bundle is not supported
--- a/release_notes/known_issues_observability.adoc
+++ b/release_notes/known_issues_observability.adoc
@@ -171,7 +171,7 @@ Your query data appears from the Grafana dashboard.
 
 [#metrics-proxy-not-detected]
 == Metrics collector does not detect proxy configuration
-//2.9:1ACM-8488
+//2.9:ACM-8488
 
 A proxy configuration in a managed cluster that you configure by using the `addonDeploymentConfig` is not detected by the metrics collector. HTTPS proxy configuration with a custom CA bundle is not supported. HTTP proxy configuration is supported with manual configuration.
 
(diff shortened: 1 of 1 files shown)


release_notes/known_issues_observability.adoc
```diff
@@ -167,4 +167,12 @@ You must manually update the Thanos Query front-end data source configuration. C
 
 . To verify that the data is displayed, refresh your Grafana page. 
 
-Your query data appears from the Grafana dashboard.
\ No newline at end of file
+Your query data appears from the Grafana dashboard.
+
+[#metrics-proxy-not-detected]
+== Metrics collector does not detect proxy configuration
+//2.9:1ACM-8488
```
douglascamata: Is this a typo?

```suggestion
//2.9:ACM-8488
```

release_notes/known_issues_observability.adoc
```diff
@@ -167,4 +167,12 @@ You must manually update the Thanos Query front-end data source configuration. C
...
+
+A proxy configuration in a managed cluster that you configure by using the `addonDeploymentConfig` is not detected by the metrics collector. HTTPS proxy configuration with a custom CA bundle is not supported. HTTP proxy configuration is supported with manual configuration.
```
douglascamata: Could we split this paragraph into a few bullet points? I think it would be much easier to understand the different scenarios here.

douglascamata (changes requested):

release_notes/known_issues_observability.adoc
oafischer: Indeed, fixed, thanks!

release_notes/known_issues_observability.adoc
oafischer: If we add an explanatory sentence, sure. Maybe:

A proxy configuration in a managed cluster that you configure by using the `addonDeploymentConfig` is not detected by the metrics collector because of the following limitations:

- HTTPS proxy configuration with a custom CA bundle is not supported.
- HTTP proxy configuration is only supported with manual configuration.

Does that work?

douglascamata: The detection problem is separate of the lack of support for HTTPS + custom CA bundle. In this text they seem mixed up. Here's my rough suggestion:

```
==Metrics collector does not detect proxy configuration

A proxy configuration in a managed cluster that you configure by using the `addonDeploymentConfig` is not detected by the metrics collector. 

You can enable it by the removing the managed cluster `ManifestWork`. Removing the `ManifestWork` forces the changes in the `addonDeploymentConfig` to be applied.

==HTTPS proxy with custom CA bundle is not supported

A proxy configuration in a managed cluster will not work properly when a custom CA bundle is needed. Currently there is no workaround.
```

As a custom CA bundle would only make sense with HTTPS proxy and not HTTP proxy, I avoid mentioning HTTP proxy in the last point.

oafischer: I see, you are treating them as 2 issues. That makes sense. I updated the file @douglascamata, let me know if that looks good now. Thanks!

douglascamata (approved):
/lgtm

release_notes/known_issues_observability.adoc
```diff
@@ -167,4 +167,16 @@ You must manually update the Thanos Query front-end data source configuration. C
 
 . To verify that the data is displayed, refresh your Grafana page. 
 
-Your query data appears from the Grafana dashboard.
\ No newline at end of file
+Your query data appears from the Grafana dashboard.
```
cadawson, line 170: Appears _in_ the Grafana dashboard?

cadawson (approved):
Looks good. Thanks!

release_notes/known_issues_observability.adoc
oafischer: Not sure, didn't work on that section, I'd rather leave it for now. Thanks for the review!

PR #5819, opened 2023-11-12
https://issues.redhat.com/browse/ACM-6737 Add Nutanix to what's new

cadawson:


oafischer (approved):
Looks great! I don't think we need to duplicate the entry for console. I think it's not technically a new console feature, just something new that is accessed via the console.

PR #5832, opened 2023-11-13
https://issues.redhat.com/browse/ACM-8363 updates

dockerymick:
https://issues.redhat.com/browse/ACM-8363
Removing the line about RHACM. MCE can be installed w/o RHACM. It is not important for the MCE user to know that MCE is automatically installed with RHACM


dockerymick:
Screenshot 2023-11-13 at 12 33 23 PM

PR #5856, opened 2023-11-15
https://issues.redhat.com/browse/ACM-7793 Update docinfo.xml

swopebe:


clusters/docinfo.xml
```diff
@@ -2,7 +2,7 @@
 <productnumber>{product-version}</productnumber>
 <subtitle>Cluster management</subtitle>
 <abstract>
-    <para>With cluster lifecycle and multicluster engine, you can create and manage clusters. Cluster lifecycle is available through the {mce-short}.</para>
+    <para>With cluster lifecycle, also named the multicluster engine operator, you can create and manage clusters. From this guide, you can access cluster management tasks, release notes, and troubleshooting information.</para>
```
cadawson, line 5: You can create and manage clusters in {product-title-short} by using the functionality provided by the multicluster engine operator. From this guide, you can access cluster management tasks, release notes, and troubleshooting information.

cadawson (approved):
Yours looks good. I provided a thought on an alternative version, which I am sure it too wordy. Maybe you can see where I was going with it. Either way is approved, or if you come up with something better. Thanks!

clusters/docinfo.xml
swopebe: @cadawson clusters is MCE, we should not be using the RHACM conref here....we went through all that as a team. Can you explain why you are saying we should change the description to RHACM?

We had many meetings with PM about the language here, we need to present this guide as "cluster lifecycle" and "mce" -- not as RHACM doc bc the MCE user doesn't have to have RHACM.

I am now a bit concerned that we lost this understanding this release or that I missed something.

swopebe:
@cadawson I am going to merge this because it talks about mce/cluster lifecycle. Please ping if you have a reason to start incorporating RHACM into this doc that I missed this release. We set this up to be independent of RHACM, with only a few mentions of exceptions. Let me know if that is no longer clear to the doc team or something changed in strategy with PM.

PR #5877, opened 2023-11-21
https://issues.redhat.com/browse/ACM-8558--arm known issue

swopebe:


release_notes/known_issues_install.adoc
```diff
@@ -27,6 +27,12 @@ For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/o
 
 For more about deprecations and removals, see xref:../release_notes/deprecate_remove.adoc#deprecations-removals[Deprecations and removals].
 
+[#install-arm-diable]
+== Infrastructure operator error with ARM converged flow
```
swopebe, line 31: @cadawson do we use another term for "arm"?

release_notes/known_issues_install.adoc
```diff
@@ -27,6 +27,12 @@ For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/o
...
+//2.9: 8558
+ 
+When you install the `infrastructure-operator`, converged flow with ARM does not work. Set `ALLOW_CONVERGED_FLOW` to `false` to resolve this issue.
```
filanov, line 34: I'm not sure that a lot of users are familiar  with how to set the env var, the user will need to do that by defining a config map https://github.com/openshift/assisted-service/blob/master/docs/operator.md#specifying-environmental-variables-via-configmap

swopebe: @filanov since this is not official doc, we cannot link to it. We would need to place the step in the known issue, which I can do.

swopebe: Would like a cleaner version of this where we break the command from the file, as we don't skip that step in the docs. See the change that will go into the doc when complete, but still need a little info @filanov:

Run the following command to create a `ConfigMap`:
```
kubectl create -f -
```
See the following file with .... (this doesn't have a "converged flow" --do we need to add that spec here?)
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-assisted-service-config
  namespace: assisted-installer
data:
  LOG_LEVEL: "debug"
```

filanov: You don't need to link the unofficial link but explaining how to create the configmap can really help. 
@carbonin is there another way to avoid converged flow besides using the unsupported configmap?

carbonin: Not that I recall.

swopebe: If you guys are all ok with this draft, I will add it: https://github.com/stolostron/rhacm-docs/pull/5877#discussion_r1406633807

filanov: yes from my side, for the example i would suggest to put `ALLOW_CONVERGED_FLOW: false` for the example

swopebe: @filanov I just need to know where in the sample to add that:

apiVersion: v1
kind: ConfigMap
metadata:
  name: my-assisted-service-config
  namespace: assisted-installer
data:
  LOG_LEVEL: "debug"
  
  
  Can you place it?

carbonin:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-assisted-service-config
  namespace: assisted-installer
data:
  ALLOW_CONVERGED_FLOW: false
```

swopebe: That's in and ready for review. I should have caught this earlier...mostly we are using `oc` instead of `kubectl` but does that not apply here?

release_notes/known_issues_install.adoc
```diff
@@ -27,6 +27,33 @@ For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/o
...
+
+. Run the following command to create a `ConfigMap`:
```
carbonin: You also need to annotate the agentserviceconfig:

```
oc annotate --overwrite AgentServiceConfig agent unsupported.agent-install.openshift.io/assisted-service-configmap=my-assisted-service-config
```

swopebe: Ok, that would be step 3, I assume? I will add it and let me know. (Future: if we can get draft procedures in the template before we accept the issue, we can prevent some delay.)

swopebe: I went ahead and made both commands `oc`.

release_notes/known_issues_install.adoc
```diff
@@ -27,6 +27,40 @@ For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/o
 
 For more about deprecations and removals, see xref:../release_notes/deprecate_remove.adoc#deprecations-removals[Deprecations and removals].
 
+[#install-arm-diable]
+== Infrastructure operator error with ARM converged flow
```
dockerymick, line 31: There is a conref for Infrastructure Operator for Red Hat OpenShift ({assist-install}). is this another name for CIM?

release_notes/known_issues_install.adoc
```diff
@@ -27,6 +27,40 @@ For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/o
...
+//2.9: 8558
+ 
+When you install the `infrastructure-operator`, converged flow with ARM does not work. Set `ALLOW_CONVERGED_FLOW` to `false` to resolve this issue.
+
+. Run the following command to create a `ConfigMap`:
```
dockerymick: ConfigMap doesn't need to be monospaced

release_notes/known_issues_install.adoc
```diff
@@ -27,6 +27,40 @@ For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/o
...
+  ALLOW_CONVERGED_FLOW: false
+----
+
+. Annotate the `agentserviceconfig` with the following command:
+
++
+----
+oc annotate --overwrite AgentServiceConfig agent unsupported.agent-install.openshift.io/assisted-service-configmap=my-assisted-service-config
+----
+
```
dockerymick, line 63: How does the user know that the issue is resolved?

dockerymick (changes requested):

release_notes/known_issues_install.adoc
filanov: he will see the agent appear in the inventory

release_notes/known_issues_install.adoc
swopebe: I don't think so; it's worth checking in the near future. CIM is the _infrastructure operator_ but it is just called that in the doc; but we need to look at conrefs in general. I notice there are some confusing ones like "cincinatti" that need to be named better. We need to have conref additions peer reviewed like other repo additions.

release_notes/known_issues_install.adoc
swopebe: We don't always tell them that in a known issue, so if there is nothing for them to check, we keep it blank.

release_notes/known_issues_install.adoc
swopebe: @dockerymick this is correctly monospaced. It is referring to the `kind` that is named `ConfigMap`. I will ensure we revisit this in our doc meetings. Yes you can just call it a generic config map, but I coded it because we literally give them a `ConfigMap` sample. 

https://redhat-internal.slack.com/archives/CUHCCV19P/p1701112320733399

swopebe: Will add: The agent appears in the inventory when the issue is resolved.

release_notes/known_issues_install.adoc
swopebe: addressed in line 64.

release_notes/known_issues_install.adoc
```diff
@@ -27,6 +27,42 @@ For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/o
 
 For more about deprecations and removals, see xref:../release_notes/deprecate_remove.adoc#deprecations-removals[Deprecations and removals].
 
+[#install-arm-diable]
+== Infrastructure operator error with ARM converged flow
+//2.9: 8558
+ 
+When you install the `infrastructure-operator`, converged flow with ARM does not work. Set `ALLOW_CONVERGED_FLOW` to `false` to resolve this issue.
+
+. Run the following command to create a `ConfigMap` resource:
```
swopebe, line 36: added resource to align with what the step is for and the `kind`

release_notes/known_issues_install.adoc
```diff
@@ -27,6 +27,42 @@ For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/o
...
+
++
+----
+oc create -f
+----
+
+. See the following file sample with `ALLOW_CONVERGED_FLOW` set to `false`:
```
swopebe: We prob need to apply this with a command, right @filanov ?

dockerymick (approved):
Thanks Brandi, this looks good to me!

release_notes/known_issues_install.adoc
filanov: yes

filanov (approved):
LGTM from me

release_notes/known_issues_install.adoc
swopebe: @filanov can I assume the command is `oc apply -f` ?

carbonin: yes, that will work.

PR #5878, opened 2023-11-22
post-release corrections

krishnanshruthi:
@swopebrandi1 Pls review the changes that were suggested by the IBM Z team after going through the released documentation page.

cc: @SNiemann15

### DIFF
--- a/clusters/hosted_control_planes/config_dns_bm_ibmz.adoc
+++ b/clusters/hosted_control_planes/config_dns_bm_ibmz.adoc
@@ -29,7 +29,7 @@ api                   IN A 1xx.2x.2xx.1xx <1>
 api-int               IN A 1xx.2x.2xx.1xx
 ;
 ;
-*.apps ${HOSTED_CLUSTER_NAME}.${BASEDOMAIN}           IN A 1xx.2x.2xx.1xx
+*.apps        IN A 1xx.2x.2xx.1xx
 ;
 ;EOF
 ----
--- a/clusters/hosted_control_planes/config_dns_bm_ibmz.adoc
+++ b/clusters/hosted_control_planes/config_dns_bm_ibmz.adoc
@@ -29,7 +29,7 @@ api                   IN A 1xx.2x.2xx.1xx <1>
 api-int               IN A 1xx.2x.2xx.1xx
 ;
 ;
-*.apps.${HOSTED_CLUSTER_NAME}.${BASEDOMAIN}           IN A 1xx.2x.2xx.1xx
+*.apps ${HOSTED_CLUSTER_NAME}.${BASEDOMAIN}           IN A 1xx.2x.2xx.1xx
 ;
 ;EOF
 ----
--- a/clusters/hosted_control_planes/hosted_create_bare_metal.adoc
+++ b/clusters/hosted_control_planes/hosted_create_bare_metal.adoc
@@ -13,7 +13,8 @@ export HOSTED_CONTROL_PLANE_NAMESPACE="${CLUSTERS_NAMESPACE}-${HOSTED_CLUSTER_NA
 export BASEDOMAIN="krnl.es"
 export PULL_SECRET_FILE=$PWD/pull-secret
 export MACHINE_CIDR=192.168.122.0/24
-export OCP_RELEASE=4.14.0-ec.2-multi       
+export ETCD_STORAGE="lvm-storageclass"
+export OCP_RELEASE=4.14.0-multi       
 export SSH_KEY=${HOME}/.ssh/id_rsa.pub
 oc create ns ${HOSTED_CONTROL_PLANE_NAMESPACE}
 
@@ -22,6 +23,8 @@ hcp create cluster agent \
     --pull-secret=${PULL_SECRET_FILE} \
     --agent-namespace=${HOSTED_CONTROL_PLANE_NAMESPACE} \
     --base-domain=${BASEDOMAIN} \
+    --api-server-address=api.${HOSTED_CLUSTER_NAME}.${BASEDOMAIN} \
+    --etcd-storage-class=${ETCD_STORAGE} \
     --ssh-key  ${SSH_KEY} \
     --namespace ${CLUSTERS_NAMESPACE} \
     --control-plane-availability-policy SingleReplica \
--- a/clusters/hosted_control_planes/hosted_create_bare_metal.adoc
+++ b/clusters/hosted_control_planes/hosted_create_bare_metal.adoc
@@ -13,7 +13,8 @@ export HOSTED_CONTROL_PLANE_NAMESPACE="${CLUSTERS_NAMESPACE}-${HOSTED_CLUSTER_NA
 export BASEDOMAIN="krnl.es"
 export PULL_SECRET_FILE=$PWD/pull-secret
 export MACHINE_CIDR=192.168.122.0/24
-export ETCD_STORAGE="lvm-storageclass"
+export OCP_RELEASE=4.14.0-ec.2-multi       
+export SSH_KEY=${HOME}/.ssh/id_rsa.pub
 oc create ns ${HOSTED_CONTROL_PLANE_NAMESPACE}
 
 hcp create cluster agent \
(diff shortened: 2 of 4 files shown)


clusters/hosted_control_planes/config_dns_bm_ibmz.adoc
```diff
@@ -29,7 +29,7 @@ api                   IN A 1xx.2x.2xx.1xx <1>
 api-int               IN A 1xx.2x.2xx.1xx
 ;
 ;
-*.apps.${HOSTED_CLUSTER_NAME}.${BASEDOMAIN}           IN A 1xx.2x.2xx.1xx
+*.apps ${HOSTED_CLUSTER_NAME}.${BASEDOMAIN}           IN A 1xx.2x.2xx.1xx
```
veera-damisetti: this line should be **`*.apps        IN A 1xx.2x.2xx.1xx`**

clusters/hosted_control_planes/hosted_create_bare_metal.adoc
```diff
@@ -14,6 +14,8 @@ export BASEDOMAIN="krnl.es"
 export PULL_SECRET_FILE=$PWD/pull-secret
 export MACHINE_CIDR=192.168.122.0/24
 export ETCD_STORAGE="lvm-storageclass"
+export OCP_RELEASE=4.14.0-multi
```
veera-damisetti, line 17: As we are changing the version here , 
we need to change the version in rest of the doc also . 
4.14.0-ec.2 --> 4.14.0

clusters/hosted_control_planes/hosted_create_bare_metal.adoc
```diff
@@ -22,7 +24,12 @@ hcp create cluster agent \
     --agent-namespace=${HOSTED_CONTROL_PLANE_NAMESPACE} \
     --base-domain=${BASEDOMAIN} \
     --api-server-address=api.${HOSTED_CLUSTER_NAME}.${BASEDOMAIN} \
-    --etcd-storage-class=${ETCD_STORAGE}
+    --etcd-storage-class=${ETCD_STORAGE} \
+    --ssh-key  ${SSH_KEY} \
+    --namespace ${CLUSTERS_NAMESPACE} \
+    --control-plane-availability-policy SingleReplica \
+    --release-image=quay.io/openshift-release-dev/ocp-release:${OCP_RELEASE} \
+    --api-server-address=api.${HOSTED_CLUSTER_NAME}.${BASEDOMAIN}
```
SNiemann15, line 32: @swopebe This is Z specific (not sure about Power @alishaIBM? ) 

Should we add those lines with callouts? Add a second IBM Z only example? Or even have a new file `hosted_create_bare_metal_ibmz.adoc`

swopebe: These {ETCD_STORAGE} should be <ETCD_STORAGE> and yes, if you think the use needs more information about that variable, you can add callouts. It depends on if the variable you add, like <HOSTED_CLUSTER_NAME> (we normally lower case these) is self explanatory.

clusters/hosted_control_planes/scaling_nodepool_hosted_ibmz.adoc
```diff
@@ -105,16 +105,16 @@ See the following output:
...
 ----
 
 . Run the following command to check the `clusterversion` and cluster operator status:
 
 +
 [source,bash]
 ----
-oc --kubeconfig <hosted_control_plane_names>.kubeconfig get clusterversion,co
+oc --kubeconfig <hosted_cluster_name>.kubeconfig get clusterversion,co
 ----
```
swopebe, line 118: awesome change here on line 117

clusters/hosted_control_planes/scaling_nodepool_hosted_ibmz.adoc
```diff
@@ -124,27 +124,27 @@ See the following output:
 [source,bash]
 ----
 NAME                                         VERSION       AVAILABLE   PROGRESSING   SINCE   STATUS
-clusterversion.config.openshift.io/version   4.14.0-ec.2   True        False         40h     Cluster version is 4.14.0-ec.2
+clusterversion.config.openshift.io/version   4.14.0   True        False         40h     Cluster version is 4.14.0
 NAME                                                                           VERSION       AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
```
swopebe: This is where I fear we are documenting too much. I don't think we need all these lines....all these versions would have to change and we have to remember to go and change them. I think we need to instead give the user what to look for in the output instead of putting a whole output here that we will have to maintain. 

Is the purpose of the output to check for status? I feel we can say that without putting the whole table here, as we are trying to simplify the docs where we can.

Thoughts? @cadawson this is an example I mentioned.

swopebe (changes requested):
Comments for consideration

clusters/hosted_control_planes/scaling_nodepool_hosted_ibmz.adoc
cadawson: I agree that looks like a lot of lines. Can we summarize with something like: The resulting output shows that all of the operators are available and no longer progressing? That might not work if we are trying to list all of the required operators (which I am not sure we want to do here). Trimming this would be helpful if it is possible.

clusters/hosted_control_planes/hosted_create_bare_metal.adoc
veera-damisetti: {ETCD_STORAGE}  ... this we are not not updating anything related to that in this PR i think .

clusters/hosted_control_planes/scaling_nodepool_hosted_ibmz.adoc
SNiemann15: @swopebe @cadawson I would like to stay in sync with the x86 docs, where this and other examples are also present. For example [here](https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.9/html/clusters/cluster_mce_overview#hosted-control-planes-manage-bm). 
Shouldn't we leave it in the PR for the moment and get this one and Power merged asap. Next step find a solution for the entire doc set and change in sync across all architectures.

clusters/hosted_control_planes/hosted_create_bare_metal.adoc
swopebe: Right, for our doc, across the board, this is what needs to change.

swopebe: Please see the standards here: https://www.ibm.com/docs/en/ibm-style?topic=elements-commands

The entire RHACM doc needs to fit these standards, so that one guide doesn't look different from the other. IBM Tech writers and RH tech writers use this manual and these standards for many years now, but sometimes things slip. 

With contributions coming from different directions, we are working to ensure IBM and RH standards are met. These standards are set for UX, Accessibility, Translation, etc... 

See the overview and reasons for the manuals here: https://www.ibm.com/docs/en/ibm-style?topic=style-overview

I will remind my team, as well, though we just had a meeting and agreed this is what we should be doing so that we are consistent.

swopebe: If we want to merge, I will open an issue and this can be changed for another refresh?

swopebe (approved):
approving to merge

swopebe:
I cannot merge this because of the conflict. I cannot resolve in a fork.

clusters/hosted_control_planes/scaling_nodepool_hosted_ibmz.adoc
swopebe: The problem is that Hypershift has many contributors at this point and so we don't want to keep propagating things that will make life harder and are not beneficial. 

My goal is to always simplify the doc, given all our team is keeping up with, with PM and engineering help. I have more support now that might not have been there before to keep things simpler.

 We may have some instances where things get merged without question; we are thinking more critically about what we add to the doc more now. Even with outside contributions, our team has to maintain these documents, release notes, and everything that comes with it, so with outputs, I just asked the team to consider how much we need from the output and what does the user have to see.

PR #5918, opened 2023-12-04
[ACM-8471]: Adding disconnected HCP diagram and text

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-8471


clusters/hosted_control_planes/configure_hosted_disconnected.adoc
```diff
@@ -10,10 +10,35 @@ In the context of hosted control planes, a disconnected environment is an {ocp-s
 
 When you provision hosted control planes on bare metal, you use the Agent platform. The Agent platform and {mce-short} work together to enable disconnected deployments. The Agent platform uses the central infrastructure management service to add worker nodes to a hosted cluster. For an introduction to the central infrastructure management service, see xref:../cluster_lifecycle/cim_enable.adoc#enable-cim[Enabling the central infrastructure management service].
 
+The following diagram illustrates an example architecture of a disconnected environment:
+
+image:../images/489_RHACM_HyperShift_on_bare_metal_1223.png[Disconnected architecture diagram]
+
+. The user configures infrastructure services, including the Registry certificate, web server, and DNS, to ensure that the disconnected deployment works.
```
jparrill: I would change "including the Registry certificate" by "including the registry deployment with TLS support"

clusters/hosted_control_planes/configure_hosted_disconnected.adoc
```diff
@@ -10,10 +10,35 @@ In the context of hosted control planes, a disconnected environment is an {ocp-s
...
+. The user creates a config map in the `openshift-config` namespace. In this example, the config map is named `registry-config`. The content of the config map is the Registry CA certificate.
```
jparrill: I would add a tip here because if not it will not work, here is well described (https://hypershift-docs.netlify.app/labs/ipv4/tls-certificates/#adding-a-ca-to-the-management-cluster)

The ConfigMap data field should contain as a Key/Value: 

- **Key**: \<Registry DNS domain name\>..\<PORT\>, changing the typical ":" by ".."
- **Value**: The certificate content

```
registry.hypershiftbm.lab..5000: |
    -----BEGIN CERTIFICATE-----
    -----END CERTIFICATE-----
```

clusters/hosted_control_planes/configure_hosted_disconnected.adoc
```diff
@@ -10,10 +10,35 @@ In the context of hosted control planes, a disconnected environment is an {ocp-s
...
+. The user modifies the `images.config.openshift.io` custom resource (CR) specification and adds a new field named `additionalTrustedCA` with a value of `name: registry-config`.
+//where does the next step show up in the diagram? 
+. The user creates a config map that contains two data fields. One field contains the `registries.conf` file in `RAW` format, and the other field contains the Registry CA and is named `ca-bundle.crt`.
```
jparrill: It's not clear to me how it should looks like. Things to remark:

- It should belong to `multicluster-engine` namespace
- The ConfigMap Name is not set in stone, but you need to take note of it to put it as a reference in other objects (the next bullet point)

I would put a sample here:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: custom-registries
  namespace: multicluster-engine
  labels:
    app: assisted-service
data:
  ca-bundle.crt: |
    -----BEGIN CERTIFICATE-----
    ...
    ...
    ...
    -----END CERTIFICATE-----
  registries.conf: |
    unqualified-search-registries = ["registry.access.redhat.com", "docker.io"]

    [[registry]]
    prefix = ""
    location = "registry.redhat.io/openshift4"
    mirror-by-digest-only = true

    [[registry.mirror]]
      location = "registry.ocp-edge-cluster-0.qe.lab.redhat.com:5000/openshift4"

    [[registry]]
    prefix = ""
    location = "registry.redhat.io/rhacm2"
    mirror-by-digest-only = true
...
...
```

Feel free to change the domain names.

images/489_RHACM_HyperShift_on_bare_metal_1223.png
jparrill, line 1: The left side line it's a bit overlapped with the boundaries of the picture and it's not fully clear in the Github interface. If in the final doc looks good, it's fine :)

jparrill (changes requested):
Added some comments bellow. If the examples are not allowed here in this section, I would add a link or something to the right place of the doc.

images/489_RHACM_HyperShift_on_bare_metal_1223.png
lahinson: I'll be sure to check the rendered version on the RHACM docs staging site when that has been rebuilt.

clusters/hosted_control_planes/configure_hosted_disconnected.adoc
lahinson: @jparrill Should the other mentions of the Registry certificate (see lines 18 and 21) also be updated?

lahinson:
@jparrill Thanks for the feedback! PTAL at the latest changes and let me know what you think.

clusters/hosted_control_planes/configure_hosted_disconnected.adoc
jparrill: Sorry I was meant "Registry certificate with TLS support", updated the comment.

jparrill: > @jparrill Should the other mentions of the Registry certificate (see lines 18 and 21) also be updated?

it's not necessary but It's ok if you add it, up to you.

jparrill:
> @jparrill Thanks for the feedback! PTAL at the latest changes and let me know what you think.

Looks good now :)

clusters/hosted_control_planes/configure_hosted_disconnected.adoc
lahinson: Thanks for the clarification!

swopebe (changes requested):
This looks great, TY for doing this. Can we change "the user" content to match conversational tone, like we do in most of the doc?

 RH guidance:
[Conversational style](https://redhat-documentation.github.io/supplementary-style-guide/#conversational-style)
Follow the IBM Style guide advice of less conversational style in most cases.

So I pulled examples from _IBM Style_, where we mostly moved to "you" or "understood you" where we can just start with the verb.

https://www.ibm.com/docs/en/ibm-style?topic=medium-conversational-style

lahinson:
@swopebe Good feedback. I updated instances of "the user" with "you" to better engage the reader. Thanks!

swopebe (approved):
Thanks for changing the tone.

PR #5919, opened 2023-12-04
https://issues.redhat.com/browse/ACM-8832

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-8832

### DIFF
--- a/clusters/release_notes/known_issues.adoc
+++ b/clusters/release_notes/known_issues.adoc
@@ -35,7 +35,7 @@ Cluster lifecycle known issues and limitations are part of the Cluster lifecycle
 === A `StorageVersionMigration` error 
 //2.9:8832
 
-When you upgrade from {mce-short} to MCE 2.4.1, you might find a `StorageVersionMigration` error in the log of the cluster-manager pods under the namespace of {mce-short}. See the following sample of the error message: 
+When you upgrade from {mce-short} to MCE 2.4.1, you might find a `StorageVersionMigration` error in the log of the `cluster-manager` pods that run in the namespace of {mce-short}. See the following sample of the error message: 
 
 ----
 "CRDMigrationController" controller failed to sync "cluster-manager", err: ClusterManager.operator.open-cluster-management.io "cluster-manager" is invalid: status.conditions[5].reason: Invalid value: "StorageVersionMigration Failed. ": status.conditions[5].reason in body should match '^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$' 
--- a/clusters/release_notes/known_issues.adoc
+++ b/clusters/release_notes/known_issues.adoc
@@ -35,7 +35,9 @@ Cluster lifecycle known issues and limitations are part of the Cluster lifecycle
 === A `StorageVersionMigration` error 
 //2.9:8832
 
-When you upgrade from {mce-short} to MCE 2.4.1, you might find a `StorageVersionMigration` error in the log of the `cluster-manager` pods that run in the namespace of {mce-short}. See the following sample of the error message: 
+When you upgrade from {mce-short} to MCE 2.4.1, you might find a `StorageVersionMigration` error in the log of the `cluster-manager` pods that run in the namespace of {mce-short} to fit normal references of namespaces in Kubernetes products. These error messages have no impact on {mce-short}. Therefore, you do not need to take any action to resolve these errors. 
+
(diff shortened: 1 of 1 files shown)


clusters/release_notes/known_issues.adoc
```diff
@@ -31,6 +31,29 @@ Review the known issues for cluster lifecycle with {mce-short}. The following li
...
+----
+status:
+  conditions:
+  - lastUpdateTime: "2023-11-29T20:28:14Z"
+    message: 'failed to list resources: the server could not find the requested resource'
+    status: "True"
+    type: Failed 
+---- 
+
+These error messages have no impact on {mce-short} and {product-title-short}. Therefore, you do not need to take any action to resolve these errors.
```
jc-berger: Doc reviewers, please help me confirm I am using the proper prod references here for MCE and ACM.

clusters/release_notes/known_issues.adoc
```diff
@@ -31,6 +31,29 @@ Review the known issues for cluster lifecycle with {mce-short}. The following li
 
 Cluster lifecycle known issues and limitations are part of the Cluster lifecycle with {mce-short} documentation.
 
+[#a-storageversionmigration-error]
+=== A `StorageVersionMigration` error 
+//2.9:8832
+
+When you upgrade from {mce-short} to MCE 2.4.1, you might find a `StorageVersionMigration` error in the log of the cluster-manager pods under the namespace of {mce-short}. See the following sample of the error message:
```
jc-berger: Doc reviewers, I need to say, from MCE 2.4.0 to MCE 2.4.1. 

Would we have different product references for the different versions?

swopebe: You can say that, to ensure you don't mean from 2.3.0 or something, unless that is included. If it is specific to 2.4.0, then yes. If `cluster-manager` is a resource, it should be coded. We won't use directional language like under--best to always rephrase that as _from the namespace_, or _within the namespace_. I am not sure about that whole line:

namespace of {mce-short}    (What is the namespace, and let's write it here)

The `<instert-variable>` namespace 

Example from other docs: ...is available in the `openshift-monitoring` namespace.

swopebe (changes requested):

clusters/release_notes/known_issues.adoc
elgnay: This issue only happens when upgrading from 2.4.0 to 2.4.1. 

`cluster-manager` is the component name and it runs in `multicluster-engine` namespace.

clusters/release_notes/known_issues.adoc
swopebe: Yes, the references are fine, however line 55 mentions RHACM and this is MCE (cluster) doc where the user may not have RHACM yet. Not sure we need to have {product-title-short}. I would put early in the description that the error messages do not require resolution, instead of the bottom. I will approve with those suggestions.

swopebe (approved):

clusters/release_notes/known_issues.adoc
```diff
@@ -31,6 +31,29 @@ Review the known issues for cluster lifecycle with {mce-short}. The following li
 
 Cluster lifecycle known issues and limitations are part of the Cluster lifecycle with {mce-short} documentation.
 
+[#a-storageversionmigration-error]
+=== A `StorageVersionMigration` error 
+//2.9:8832
+
+When you upgrade from {mce-short} to MCE 2.4.1, you might find a `StorageVersionMigration` error in the log of the `cluster-manager` pods that run in the namespace of {mce-short}. See the following sample of the error message: 
+
```
swopebe, line 41: Wait, sorry, I missed this...we still need to change: that run in the namespace of {mce-short} to fit normal references of namespace in Kubernetes products.

swopebe (changes requested):
sorry, I see a change still needed that I mentioned earlier.

clusters/release_notes/known_issues.adoc
swopebe: @jc-berger see the statement above with the actual namespace. 

runs in `multicluster-engine` namespace (Namespace name is not the same as the product name {mce-short}).

clusters/release_notes/known_issues.adoc
swopebe: Ok, I think we should meet.

swopebe (approved):
chatted for a moment, approving.

PR #5936, opened 2023-12-06
[ACM-8923]: Update units of measurement in hosted sizing guidance

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-8923

### DIFF
--- a/clusters/hosted_control_planes/hosted_sizing_guidance.adoc
+++ b/clusters/hosted_control_planes/hosted_sizing_guidance.adoc
@@ -6,9 +6,9 @@ Many factors, including hosted cluster workload and worker node count, affect ho
 See the following highly available hosted control plane requirements, which were tested with {ocp-short} version 4.12.9 and later:
 
 * 78 pods
-* Three 8 Gi PVs for etcd
+* Three 8 GiB PVs for etcd
 * Minimum vCPU: approximately 5.5 cores
-* Minimum memory: approximately 19 GB
+* Minimum memory: approximately 19 GiB
 
 [#hosted-sizing-guidance-pod-limit]
 == Pod limits
@@ -23,7 +23,7 @@ For bare-metal nodes, the default `maxPods` setting of 250 is likely to be a lim
 To understand the request-based resource limit, consider the total request value of a hosted control plane. To calculate that value, add the request values of all highly available hosted control plane pods across the namespace. See the following examples:
 
 * Five vCPU requests for each highly available hosted control plane
-* 18 Gi memory requests for each highly available hosted control plane
+* 18 GiB memory requests for each highly available hosted control plane
 
 [#hosted-sizing-guidance-load-based-limit]
 == Load-based limit
@@ -32,7 +32,7 @@ Request-based sizing provides a maximum number of hosted control planes that can
 
 For example, to demonstrate resource scaling with API load, consider the following workload configuration:
 
-* One hosted cluster with 9 workers (8 vCPU, 32 Gi each), using the KubeVirt provider
+* One hosted cluster with 9 workers (8 vCPU, 32 GiB each), using the KubeVirt provider
 * Benchmark tool: link:https://github.com/cloud-bulldozer/kube-burner[Kube-burner]
 * The workload profile is a cluster density test that is configured to focus on API control-plane stress, based on this link:https://cloud-bulldozer.github.io/kube-burner/v1.7.9/ocp/#cluster-density-v2[definition]:
 
@@ -43,7 +43,7 @@ For example, to demonstrate resource scaling with API load, consider the followi
 Resource utilization is measured as the workload increased to the total namespace count. This data provided an estimation factor to increase the compute resource capacity based on the expected API load. Exact utilization rates can vary based on the type and pace of the cluster workload. 
 
 |===
-| Hosted control plane resource utilization scaling | vCPUs | Memory (GB)
+| Hosted control plane resource utilization scaling | vCPUs | Memory (GiB)
 
 | Default requests 
 | 5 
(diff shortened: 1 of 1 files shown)


jeniferh:
LGTM thanks!

gqlo (approved):

gqlo (approved):

clusters/hosted_control_planes/hosted_sizing_guidance.adoc
```diff
@@ -125,25 +125,25 @@ This example provides sizing guidance for the following scenario:
 
 | Calculated values based on worker node size and API rate | Server 1 | Server 2 | Calculation notes
 
-| Maximum hosted control planes per worker based on CPU requests 
+| Maximum hosted control planes per worker based on vCPU requests 
 | 12.8 
 | 25.6 
-| Number of worker CPUs / 5 total CPU requests per hosted control plane
+| Number of worker vCPUs / 5 total vCPU requests per hosted control plane
```
dockerymick, line 131: the backslash is a bit confusing for me. Based on the other rows it seems it seems like a math problem. I am taking a look at the [Mathematical equations](https://www.ibm.com/docs/en/ibm-style?topic=measurement-mathematical-equations) section from the Style guide
```suggestion
| Number of worker vCPUs divided by 5 total vCPU requests per hosted control plane
```

dockerymick: the backslash is a bit confusing for me. Based on the other rows it seems it seems like a math problem. I am taking a look at the [Mathematical equations](https://www.ibm.com/docs/en/ibm-style?topic=measurement-mathematical-equations) section from the Style guide
```suggestion
| Number of worker vCPUs divided by 5 total vCPU requests per hosted control plane
```

clusters/hosted_control_planes/hosted_sizing_guidance.adoc
```diff
@@ -125,25 +125,25 @@ This example provides sizing guidance for the following scenario:
...
 
-| Maximum hosted control planes per worker based on CPU usage 
+| Maximum hosted control planes per worker based on vCPU usage 
 | 5.4 
 | 10.7 
-| Number of CPUS / (2.9 measured idle CPU usage + (QPS target rate / 1000) * 9.0 measured CPU usage per 1000 QPS increase)
+| Number of vCPUS / (2.9 measured idle vCPU usage + (QPS target rate / 1000) * 9.0 measured vCPU usage per 1000 QPS increase)
```
dockerymick: is 2.9 the explicit number or are you referring to ACM 2.9 vCPU usage?
```suggestion
| Number of vCPUS/(2.9 measured idle vCPU usage + (QPS target rate/1000) * 9.0 measured vCPU usage per 1000 QPS increase)
```

dockerymick (changes requested):

clusters/hosted_control_planes/hosted_sizing_guidance.adoc
lahinson: Good point! Because we use mathematical symbols in other places in that table, I think we should replace the slash with the division sign, as it looks like the division sign is used in the Mathematical equations topic in IBM Style.

lahinson:
@dockerymick I replaced the slashes in the equations with division signs (÷), which aligns with what IBM Style uses. Thanks!

clusters/hosted_control_planes/hosted_sizing_guidance.adoc
```diff
@@ -125,38 +125,38 @@ This example provides sizing guidance for the following scenario:
 
 | Calculated values based on worker node size and API rate | Server 1 | Server 2 | Calculation notes
 
-| Maximum hosted control planes per worker based on CPU requests 
+| Maximum hosted control planes per worker based on vCPU requests 
 | 12.8 
 | 25.6 
-| Number of worker CPUs / 5 total CPU requests per hosted control plane
+| Number of worker vCPUs  5 total vCPU requests per hosted control plane
```
dockerymick: Should there be a comma or semicolon here?

clusters/hosted_control_planes/hosted_sizing_guidance.adoc
```diff
@@ -125,38 +125,38 @@ This example provides sizing guidance for the following scenario:
...
 
-| Maximum hosted control planes per worker based on CPU usage 
+| Maximum hosted control planes per worker based on vCPU usage 
 | 5.4 
 | 10.7 
-| Number of CPUS / (2.9 measured idle CPU usage + (QPS target rate / 1000) * 9.0 measured CPU usage per 1000 QPS increase)
+| Number of vCPUS ÷ (2.9 measured idle vCPU usage + (QPS target rate ÷ 1000) * 9.0 measured vCPU usage per 1000 QPS increase)
```
dockerymick: It seems like you should use the regular multiplication sign (×) 
I copied the symbol from the SG

What are your thoughts?

dockerymick (approved):
One more question. Otherwise, this looks good

clusters/hosted_control_planes/hosted_sizing_guidance.adoc
lahinson: oops! there was supposed to be a division sign there. I'll add that.

clusters/hosted_control_planes/hosted_sizing_guidance.adoc
lahinson: Yep, you're right. IBM Style says to use the multiplication sign (x). I'll make that update.

dockerymick: Instead of using the "x" i would use the Unicode character that i have in the suggestion. It can be confusing to see an "x" during translation 
```suggestion
| Number of vCPUS ÷ (2.9 measured idle vCPU usage + (QPS target rate ÷ 1000) × 9.0 measured vCPU usage per 1000 QPS increase)
```

lahinson: @dockerymick I replaced the asterisks (*) with the unicode for the multiplication sign. In GitHub, the unicode doesn't render, so we'll need to check on staging to make sure that everything looks ok.

dockerymick: Thank you!

PR #5307, opened 2023-07-24
https://issues.redhat.com/browse/ACM-6529 Updating Gatekeeper docs, removing upstream docs

dockerymick:
https://issues.redhat.com/browse/ACM-6529

PR #5943, opened 2023-12-07
[GRC] Remove policy generator version

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-8945

- Remove generator version (now that we have release branches in `stolostron`)
- Change URL to `stolostron` to direct RHACM users to the ACM generator rather than OCM

/cc @mprahl @JustinKuli @gparvin @yiraeChristineKim

### DIFF
--- a/governance/policy_generator.adoc
+++ b/governance/policy_generator.adoc
@@ -3,7 +3,7 @@
 
 The Policy Generator is a part of the {product-title} application lifecycle subscription GitOps workflow that generates {product-title-short} policies using Kustomize. The Policy Generator builds {product-title-short} policies from Kubernetes manifest YAML files, which are provided through a `PolicyGenerator` manifest YAML file that is used to configure it. The Policy Generator is implemented as a Kustomize generator plug-in. For more information on Kustomize, read the _Kustomize documentation_. 
 
-The Policy Generator version bundled in this version of {product-title-short} is v1.10.0. View the following topics to for more information:
+View the following topics to for more information:
 
 * <<policy-generator-capabilities,Policy Generator capabilities>>
 * <<policy-generator-configuration,Policy Generator configuration structure>>
@@ -377,7 +377,6 @@ Note that all the fields in the `policyDefaults` section except for `namespace` 
 == Additional resources
 
 * Read link:../gitops/gitops_policy_operator.adoc#gitops-policy-operator[Generating a policy to install GitOps Operator].
-* Read the link:https://github.com/open-cluster-management-io/policy-generator-plugin[`policy-generator-plugin`] repository for more details.
 * Read to xref:../governance/policy_set_ctrl.adoc#policy-set-controller[Policy set controller] for more details.
 * Read link:../applications/subscription_sample.adoc#applying-kustomize[Applying Kustomize] for more information.
 * Read the xref:../governance/grc_intro.adoc#governance[Governance] documentation for more topics.
@@ -386,4 +385,3 @@ Note that all the fields in the `policyDefaults` section except for `namespace` 
 * Refer link:https://open-policy-agent.github.io/gatekeeper/website/docs/[Gatekeeper] for more details.
 * Refer to the link:https://kustomize.io/[Kustomize documentation]. 
 * Return to the xref:../governance/third_party_policy.adoc#integrate-third-party-policy-controllers[Integrate third-party policy controllers] documentation.
-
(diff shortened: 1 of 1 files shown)


dhaiducek:
(Marked as draft for feedback)

governance/policy_generator.adoc
```diff
@@ -378,7 +378,7 @@ Note that all the fields in the `policyDefaults` section except for `namespace`
 == Additional resources
 
 * Read link:../gitops/gitops_policy_operator.adoc#gitops-policy-operator[Generating a policy to install GitOps Operator].
-* Read the link:https://github.com/open-cluster-management-io/policy-generator-plugin[`policy-generator-plugin`] repository for more details.
+* Read the link:https://github.com/stolostron/policy-generator-plugin[`policy-generator-plugin`] repository for more details.
```
dockerymick: @dhaiducek please create an issue related to this PR. We also shouldn't be pointing to stolostron

mprahl (approved):

governance/policy_generator.adoc
dhaiducek: I changed it because there are other places that point to `stolostron` and it's the most accurate representation for users of RHACM. But is the preference to point to `open-cluster-management-io`, then?

dhaiducek: For background, we have versions released in OCM-io, but we have release branches that are tied to RHACM versions in `stolostron`, so if consumers of the generator want to see behaviors in their current version of RHACM, they can go to the corresponding release branch in `stolostron` (which is also why I took out the version in this PR, since I had forgotten to update it for RHACM 2.9).

dhaiducek: (I've added a new Jira issue, https://issues.redhat.com/browse/ACM-8945, to the commit message and to the PR description here.)

dockerymick: To my understanding, open-cluster-management-io is our upstream project

dhaiducek: > To my understanding, open-cluster-management-io is our upstream project

True (and this may not be a valid point, so definitely correct me if I'm mistaken), but these docs also aren't upstream. For instance, Openshift doesn't point to the upstream `github.com/okd-project`--their links are to `github.com/openshift`. We created release branches in the generator in `stolostron` for maintainability, and I would like to extend that maintainability to the docs here also. However, if there's a doc policy to only point to upstream OCM-io repos, I'm fine leaving this particular link as-is and just removing the version reference.

dockerymick: Here is the link to the issue that was referenced in my PR where I replaced the stolostron link: https://issues.redhat.com/browse/ACM-6529

Here is the PR where the change was made: https://github.com/stolostron/rhacm-docs/pull/5307/files

I recall link change throughout the product doc where we replaced stolostron with open-cluster-management. We spoke with Stephanie to confirm that about 6 months ago. @swopebe will you help with confirming what links should be used in the product doc? As Dale mentioned this link update is more accurate for the user. 

For my own understanding, please correct if my understanding is incorrect:
upstream project == open-cluster-management
downstream project == stolostron

dhaiducek: So this particular link for the generator was updated here:
- https://github.com/stolostron/rhacm-docs/commit/4821c7132093610d0b60c376a0c05bffd57b9c09
- https://github.com/stolostron/rhacm-docs/pull/4787
- https://issues.redhat.com/browse/ACM-2598

We migrated the generator upstream to `open-cluster-management-io` in that Jira (ACM-2598), but the documentation here really should have stayed as `stolostron` because RHACM runs `stolostron` code, especially now that we created release branches in `stolostron` that align with RHACM.

swopebe: Sorry I am just getting to this. 

1. Yes, we have been guided not to point to internal downstream doc, but rather to document what is needed in the published docs. 

2. If we are pointing to something internally, can I ask why we are not documenting it?

3. The problem is this link can change without notice and if not verified regularly with QE, can create problems for support. In the rare case that we do it, we need to check with someone on the support side and get QE-ack. (this is what OCP does, guidance from OCP) 

But I am still leaning towards #2.

dhaiducek: Interesting. The generator is documented--this is a breadcrumb to the code (since the generator can technically be run independently from RHACM), but I suppose we could remove the link entirely.

mprahl: @swopebe could we think of the Policy Generator as a separate open source project that is included within ACM? If we do that, does that allow us to keep the link?

The goal is to make the Policy Generator GitHub repository discoverable so that users can run it on their laptop for easier ACM policy development and debugging.

swopebe: @mprahl I feel that is a larger question for the org about the use of your repo _within_ stolotron and would involve QE and Support, maybe? Stolostron as a whole is not considered the open source project from our perspective, though it is "open." Pointing to doc that is not agreed upon as an open source project by support/qe/dev etc...is not best practice. It's a valid question, it's just not one that the doc team would answer independently. I am happy to attend a meeting, discuss in an arch call or a Slack channel, or whatever is needed to follow up. @cadawson FYI.

swopebe (approved):

dockerymick:
@dhaiducek @mprahl @swopebrandi1 thanks everyone for the discussion and agreement. Thanks Brandi for your willingness to discuss with more of the team to make the best decision. And thanks for the edits!

dockerymick (approved):
Changes look good, thank you!

PR #5946, opened 2023-12-07
Fixing multiplication signs

lahinson:


clusters/hosted_control_planes/hosted_sizing_guidance.adoc
```diff
@@ -163,11 +163,11 @@ This example provides sizing guidance for the following scenario:
 | Maximum number of hosted control planes within a management cluster 
 | 16
 | 20 
-| Minimum of above maximums U+00D7 3 control-plane workers
+| Minimum of above maximums × 3 control-plane workers
```
dockerymick: My bad for missing this earlier. Please remove directional words like above, below, etc. I also noticed that the column names for Table 2 are not bold. the others are bold tho 

```suggestion
| Minimum of previously mentioned maximums × 3 control-plane workers
```

dockerymick (approved):
made a comment, something that i missed in earlier updates. My bad!

clusters/hosted_control_planes/hosted_sizing_guidance.adoc
lahinson: good catch! fixing now

dockerymick (approved):

PR #5948, opened 2023-12-08
https://issues.redhat.com/browse/ACM-8855

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-8855

PR #5949, opened 2023-12-08
doc secrets for 2.8

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-8855

See PR for 2.9: https://github.com/stolostron/rhacm-docs/pull/5948/files

### DIFF
--- a/business_continuity/backup_restore/backup_arch.adoc
+++ b/business_continuity/backup_restore/backup_arch.adoc
@@ -64,7 +64,7 @@ kind: MyResource
     cluster.open-cluster-management.io/backup: cluster-activation
 ----
 
-*Note:* Regardless of the `cluster.open-cluster-management.io/backup` label value, secrets and config maps are restored when passive resources are restored. Since the `cluster-activation` value is ignored for these types of resources, set the `cluster.open-cluster-management.io/backup: cluster-activation` value to the resources using these secrets. This way, the resources are not restored until the activation step.
+*Note:* Regardless of the `cluster.open-cluster-management.io/backup` label value, secrets and config maps are restored when passive resources are restored. Since the `cluster-activation` value is ignored in this case, set the `cluster.open-cluster-management.io/backup: cluster-activation` value to the resources using these secrets and config map instead. This way, they are not restored until the activation step.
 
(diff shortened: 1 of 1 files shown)


business_continuity/backup_restore/backup_arch.adoc
```diff
@@ -64,6 +64,8 @@ kind: MyResource
     cluster.open-cluster-management.io/backup: cluster-activation
 ----
 
+*Note:* Regardless of the `cluster.open-cluster-management.io/backup` label value, secrets and config maps are restored when passive resources are restored. Since the `cluster-activation` value is ignored for these types of resources, set the `cluster.open-cluster-management.io/backup: cluster-activation` value to the resources using these secrets. This way, the resources are not restored until the activation step.
```
birsanv: I feel resources word is overused and makes this confusing.. I had to get back to jirra to clarify

What about 

*Note:* Regardless of the `cluster.open-cluster-management.io/backup` label value, secrets and config maps are always restored when passive resources are restored. Since the `cluster-activation` value is ignored in this case, set the `cluster.open-cluster-management.io/backup: cluster-activation` value to the resources using these secrets and config map **instead**, so that they are not restored until the activation step.

birsanv (changes requested):

birsanv (approved):
lgtm ; possible change config map to config maps

birsanv (approved):
lgtm

business_continuity/backup_restore/backup_arch.adoc
```diff
@@ -64,6 +64,8 @@ kind: MyResource
     cluster.open-cluster-management.io/backup: cluster-activation
 ----
 
+*Note:* Regardless of the `cluster.open-cluster-management.io/backup` label value, secrets and config maps are restored when passive resources are restored. Since the `cluster-activation` value is ignored in this case, set the `cluster.open-cluster-management.io/backup: cluster-activation` value to the resources using these secrets and config maps instead. This way, they are not restored until the activation step.
```
oafischer: There seem to be a lot of "this/these/they" in this paragraph, makes it a little hard to follow. If possible I'd try to be more specific.

I'd also move the relative clause after the main clause for better clarity. -> Secrets and config maps are restored when passive resources are restored, regardless of...

I also recommend using "because" instead of "since" to make translation easier.

oafischer (approved):

PR #5963, opened 2023-12-12
https://issues.redhat.com/browse/ACM-8639 CIM update enabling

oafischer:
Also for 2.8 and 2.7.

### DIFF
--- a/clusters/cluster_lifecycle/cim_enable.adoc
+++ b/clusters/cluster_lifecycle/cim_enable.adoc
@@ -49,7 +49,11 @@ If your hub cluster is installed on a different platform, continue at <<enable-c
 
 If you need to modify the `Provisioning` resource, complete the following steps:
 
-. Check if you have the `Provisioning` resource by running the following command.
+. Check if you have the `Provisioning` resource by running the following command:
++
+----
+oc get provisioning
+----
 
 . If you receive a `No resources found` error, create the `Provisioning` resource by adding the following YAML content:
 +
--- a/clusters/cluster_lifecycle/cim_enable.adoc
+++ b/clusters/cluster_lifecycle/cim_enable.adoc
@@ -64,7 +64,7 @@ If you already have a `Provisioning` resource, you must modify the resource if y
 - VMware vSphere
 - User-provisioned infrastructure (UPI) method and the platform is `None`
 
-If your hub cluster is installed on a different platform, continue at Enabling central infrastructure management in disconnected environments_ or _Enabling central infrastructure management in connected environments_.
+If your hub cluster is installed on a different platform, continue at _Enabling central infrastructure management in disconnected environments_ or _Enabling central infrastructure management in connected environments_.
 
 . Modify the `Provisioning` resource to allow the Bare Metal Operator to watch all namespaces by running the following command:
 
--- a/clusters/cluster_lifecycle/cim_enable.adoc
+++ b/clusters/cluster_lifecycle/cim_enable.adoc
@@ -36,26 +36,31 @@ oc apply -f
 ----
 
 [#enable-cim-provision]
-== Modifying the _Provisioning_ resource
+== Creating or modifying the _Provisioning_ resource
 
-Modify the `Provisioning` resource if your hub cluster is installed on one of the following platforms:
+If you do not have a `Provisioning` resource, you must create the resource. If you already have a `Provisioning` resource, you must modify the resource if your hub cluster is installed on one of the following platforms:
 
 - Bare metal
 - Red Hat OpenStack Platform
 - VMware vSphere
 - User-provisioned infrastructure (UPI) method and the platform is `None`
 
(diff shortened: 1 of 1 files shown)


clusters/cluster_lifecycle/cim_enable.adoc
```diff
@@ -28,8 +47,35 @@ Modify the `Provisioning` resource if your hub cluster is installed on one of th
...
+  watchAllNamespaces: true
+----
++
+.. Apply the content by running the following command:
++
+----
+oc apply -f
+----
+
+. Modify the `Provisioning` resource to allow the Bare Metal Operator to watch all namespaces by running the following command:
```
avishayt: This is only needed if the provisioning resource was found.

So:
1. Check if the provisioning resource exists
2. If not, create it
3. If yes, modify it

avishayt (commented):
One small comment, otherwise looks good!

clusters/cluster_lifecycle/cim_enable.adoc
oafischer: Ah I see, I'll restructure this section then.

oafischer:
Updated, this one is a little convoluted, if someone sees a way to simplify all the checks the user has to do feel free to make suggestions.

clusters/cluster_lifecycle/cim_enable.adoc
```diff
@@ -6,8 +6,6 @@ The central infrastructure management service is provided with the {mce-short} a
 [#enable-cim-prerequisites]
 == Prerequisites
```
swopebe, line 8: There should always be a phrase between a header and the content.

clusters/cluster_lifecycle/cim_enable.adoc
```diff
@@ -16,19 +14,77 @@ See the following prerequisites before creating an host inventory:
 - You need a configured default storage class.
 - For disconnected environments only, complete the procedure for link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/scalability_and_performance/clusters-at-the-network-far-edge#ztp-acm-preparing-to-install-disconnected-acm_ztp-deploying-disconnected[Clusters at the network far edge] in the {ocp-short} documentation.
 
+[#enable-cim-bmh-crd]
+== Creating a bare metal host custom resource definition
+
+You can check if you already have a bare metal host custom resource definition by running the following command:
```
swopebe: I feel these are steps, and should reflect a task, even though they are technical contingent on the environment. You get around that by using a description between the header and the task.

clusters/cluster_lifecycle/cim_enable.adoc
```diff
@@ -16,19 +14,77 @@ See the following prerequisites before creating an host inventory:
...
+----
+oc apply -f
+----
+
 [#enable-cim-provision]
-== Modifying the _Provisioning_ resource
+== Creating or modifying the _Provisioning_ resource
 
-Modify the `Provisioning` resource if your hub cluster is installed on one of the following platforms:
+If you do not have a `Provisioning` resource, you must create the resource. If you already have a `Provisioning` resource, you must modify the resource if your hub cluster is installed on one of the following platforms:
```
swopebe: Same here.

clusters/cluster_lifecycle/cim_enable.adoc
```diff
@@ -16,19 +14,77 @@ See the following prerequisites before creating an host inventory:
...
+oc apply -f
+----
+
+You can now continue with <<enable-cim-disconnected,Enabling central infrastructure management in disconnected environments>> or <<enable-cim-connected,Enabling central infrastructure management in connected environments>>.
+
+[#enable-cim-mod-provision]
+=== Modifying the _Provisioning_ resource
 
-Modify the `Provisioning` resource to allow the Bare Metal Operator to watch all namespaces by running the following command:
+If you have the `Provisioning` resource and your hub cluster is installed on one of the previously listed platforms, modify the `Provisioning` resource by running the following command. Running the command allows the Bare Metal Operator to watch all namespaces:
```
swopebe: same here.

swopebe: See this for conditional steps.

swopebe (changes requested):
I think the wording is great, the format should probably be more like a task and you can get some ideas here: https://www.ibm.com/docs/en/ibm-style?topic=format-procedures#indicating-optional-and-conditional-steps

oafischer:
Updated again, thanks for the link to that IBM Style topic @swopebe, it was helpful but I had to do some rearranging to conform with it. Let me know how it looks now.

swopebe (approved):
Works for me, the only thing I see is there is a section with one step...now with the new format, that can stand on it's own with the bulleted options--if it works. If not, I am not gonna fuss about it because it is tricky.

avishayt:
LGTM

PR #6005, opened 2024-01-09
https://issues.redhat.com/browse/ACM-7900 Improving backup and restore

dockerymick:
https://issues.redhat.com/browse/ACM-7900

Current updates are related to Performance and scalability. Pulling in @akrzos and @berenss to help with the review

### DIFF
--- a/install/perform_scale.adoc
+++ b/install/perform_scale.adoc
@@ -9,7 +9,8 @@
 
 * <<maximum-number-of-managed-clusters,Maximum number of managed clusters>>
 * <<search-scalability,Search scalability>>
-* <<scaling-for-observability,Scaling for observability>>
+* <<scaling-for-observability,Observability scalability>>
+* <<backup-restore-scalability,Backup and restore scalability>>
 
 [#maximum-number-of-managed-clusters]
 == Maximum number of managed clusters
@@ -45,8 +46,7 @@ The managed clusters are {sno} virtual machines hosted on Red Hat Enterprise Lin
 [#search-scalability]
 == Search scalability
 
-The scalability of the Search component depends on the performance of the data store.
-The query run time is an important variable when analyzing the search performance.
+The scalability of the Search component depends on the performance of the data store. The query run time is an important variable when analyzing the search performance.
 
 [#query-execution-considerations]
 === Query run time considerations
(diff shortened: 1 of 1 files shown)


dockerymick:
@jc-berger [Slack convo](https://redhat-internal.slack.com/archives/C0282HW2YHZ/p1705271507623289) that @birsanv tagged us in. 

It appears that #2 of the request list is needed: 

"2. Can we rollback from the passive hub to the initial primary ?

It is indicated to always use a clean hub cluster for the passive hub. If the passive hub cluster has user data created on it prior to the restore operation, this hub could not be reliably used as a passive configuration: the data on this hub is not reflective of the data available with the backed up resources and after the restore is completed, there could be resources on this hub that were not part of the backup.  During the hub restore operation, you could require for existing resources to be cleaned up if they are not part of the backup data being restored, but this only affects resource types known to the backup; any resources with a CRD not known to the hub backup resource will be ignored.

There could be one situation where is safe to use an existing hub cluster for a restore operation, and this is during a disaster recovery test. In this case you are just testing the hub backup scenario so you want to roll back to the primary hub after the backup data was restored on the passive hub. In this situation, the initial primary has not been used to create new resources, the backup data has just temporarily changed sides, from the primary hub to the passive. "

Will you help figure where rolling back backups should be added?

birsanv:
Thank you Mikela

A suggestion ..

Create a new link under https://access.redhat.com/documentation/en-us/red_hat_advanced_cluster_management_for_kubernetes/2.9/html/business_continuity/business-cont-overview#restore-backup
 
Titled 
Restoring data back to the initial primary hub
 
Content 
It is indicated to always use a clean hub cluster for the passive hub. If the passive hub cluster has user data created on it prior to the restore operation, this hub could not be reliably used as a passive configuration: the data on this hub is not reflective of the data available with the backed up resources and after the restore is completed, there could be resources on this hub that were not part of the backup.  During the hub restore operation, you could require for existing resources to be cleaned up if they are not part of the backup data being restored, but this only affects resource types known to the backup; any resources with a CRD not known to the hub backup resource will be ignored. The restore cleans up resources created by a previous backup but will not clean up user resources.
 
There could be one situation where is safe to use an existing hub cluster for a restore operation, and this is during a disaster recovery test. In this case you are just testing the hub backup scenario so you want to roll back to the primary hub after the backup data was restored on the passive hub. In this situation, the initial primary has not been used to create new resources, the backup data has just temporarily changed sides, from the primary hub to the passive.  

Thank you,
Valentina Birsan
Principal Software Engineer - Advanced Cluster Management for Kubernetes
Red Hat Canada

berenss (approved):
this is great information thank you! lgtm

dockerymick:
@berenss thanks for your approval ! @jc-berger will you provide a peer review for this? I see you are addressing [Valentina's suggestion](https://github.com/stolostron/rhacm-docs/pull/6005#issuecomment-1895973533) in your [PR #6051](https://github.com/stolostron/rhacm-docs/pull/6051)

After approval, I will merge this before we refresh

jc-berger (approved):
/lgtm

PR #6007, opened 2024-01-09
added known issue for nmstate limitation

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-9128

### DIFF
--- a/clusters/release_notes/known_issues.adoc
+++ b/clusters/release_notes/known_issues.adoc
@@ -32,15 +32,15 @@ Review the known issues for cluster lifecycle with {mce-short}. The following li
 Cluster lifecycle known issues and limitations are part of the Cluster lifecycle with {mce-short} documentation.
 
 [#limitation-with-nmstate]
-=== Limitation with `nmstate`
+=== Limitation with _nmstate_
 //2.9:9128
 
-Make your development quicker by configuring copy and paste features. To configure the `copy-from-mac` feature in the the `assisted-installer`, complete the following steps: 
+Develop quicker by configuring copy and paste features. To configure the `copy-from-mac` feature in the the `assisted-installer`, complete the following steps: 
 
 . Go to the `nmstate` definition interface. 
 . Copy the `mac-address` in the `nmstate` definition interface. 
 
-If the `mac-address` is provided outside the `nmstate` definition, the `nmstate` cannot implement the `copy-from-mac` feature. This issue is the `nmstate` limitation. 
+If the `mac-address` is provided outside the `nmstate` definition, the `nmstate` cannot implement the `copy-from-mac` feature.  
 
 [#a-storageversionmigration-error]
 === A `StorageVersionMigration` error 
--- a/clusters/release_notes/known_issues.adoc
+++ b/clusters/release_notes/known_issues.adoc
@@ -31,6 +31,17 @@ Review the known issues for cluster lifecycle with {mce-short}. The following li
 
 Cluster lifecycle known issues and limitations are part of the Cluster lifecycle with {mce-short} documentation.
 
+[#limitation-with-nmstate]
+=== Limitation with `nmstate`
+//2.9:9128
+
+Make your development quicker by configuring copy and paste features. To configure the `copy-from-mac` feature in the the `assisted-installer`, complete the following steps: 
+
+. Go to the `nmstate` definition interface. 
+. Copy the `mac-address` in the `nmstate` definition interface. 
+
+If the `mac-address` is provided outside the `nmstate` definition, the `nmstate` cannot implement the `copy-from-mac` feature. This issue is the `nmstate` limitation. 
+
 [#a-storageversionmigration-error]
 === A `StorageVersionMigration` error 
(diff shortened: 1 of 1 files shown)


clusters/release_notes/known_issues.adoc
```diff
@@ -31,6 +31,17 @@ Review the known issues for cluster lifecycle with {mce-short}. The following li
 
 Cluster lifecycle known issues and limitations are part of the Cluster lifecycle with {mce-short} documentation.
 
+[#limitation-with-nmstate]
+=== Limitation with `nmstate`
```
swopebe: I am making an update to our internal docs to remind us all that instead of codeblock here in headers, we use _italics_. (It's because of the rendering, which we can revisit, but that was decided as a team a while ago.)

clusters/release_notes/known_issues.adoc
```diff
@@ -31,6 +31,17 @@ Review the known issues for cluster lifecycle with {mce-short}. The following li
...
+//2.9:9128
+
+Make your development quicker by configuring copy and paste features. To configure the `copy-from-mac` feature in the the `assisted-installer`, complete the following steps:
```
swopebe: Make your development quicker 

to 

Develop quicker

Some examples of voice and verb usage here: https://www.ibm.com/docs/en/ibm-style?topic=grammar-verbs

clusters/release_notes/known_issues.adoc
```diff
@@ -31,6 +31,17 @@ Review the known issues for cluster lifecycle with {mce-short}. The following li
...
+=== Limitation with `nmstate`
+//2.9:9128
+
+Make your development quicker by configuring copy and paste features. To configure the `copy-from-mac` feature in the the `assisted-installer`, complete the following steps: 
+
+. Go to the `nmstate` definition interface. 
+. Copy the `mac-address` in the `nmstate` definition interface. 
+
+If the `mac-address` is provided outside the `nmstate` definition, the `nmstate` cannot implement the `copy-from-mac` feature. This issue is the `nmstate` limitation. 
+
```
swopebe, line 39: This issue is the `nmstate` limitation. (line 43) Is that needed?

??? 

Is this referring to the entry in the doc, which is a "limitation" in the doc? I think this is clear already and is redundant. It seems like ending a task with: This is the procedure. (?)

swopebe (changes requested):
See comments.

clusters/release_notes/known_issues.adoc
filanov: it's not really an nmstate limitation, it's the way assisted is using nmstate, we are using configuration generation  that doesn't support this option, but there is a workaround that is worth mentioning. 
Ori will provide the details

ori-amizur: It is nmstate limitation.  It needs the mac-address in the interface to copy the mac-address from.

clusters/release_notes/known_issues.adoc
```diff
@@ -31,6 +31,17 @@ Review the known issues for cluster lifecycle with {mce-short}. The following li
 
 Cluster lifecycle known issues and limitations are part of the Cluster lifecycle with {mce-short} documentation.
 
+[#limitation-with-nmstate]
+=== Limitation with _nmstate_
+//2.9:9128
+
+Develop quicker by configuring copy and paste features. To configure the `copy-from-mac` feature in the the `assisted-installer`, complete the following steps: 
+
+. Go to the `nmstate` definition interface.
```
jc-berger: So, am I right these are steps to configure `copy-from-mac`? 

Or, are these steps the workaround for the `nmstate` limitation?

clusters/release_notes/known_issues.adoc
```diff
@@ -31,6 +31,17 @@ Review the known issues for cluster lifecycle with {mce-short}. The following li
...
+[#limitation-with-nmstate]
+=== Limitation with _nmstate_
+//2.9:9128
+
+Develop quicker by configuring copy and paste features. To configure the `copy-from-mac` feature in the the `assisted-installer`, complete the following steps: 
+
+. Go to the `nmstate` definition interface. 
+. Copy the `mac-address` in the `nmstate` definition interface. 
+
+If the `mac-address` is provided outside the `nmstate` definition, the `nmstate` cannot implement the `copy-from-mac` feature.
```
ori-amizur: In order to use `copy-from-mac` the mac-address must be part of the nmstate definition.
The same mac-address must be part of the interface-mac mapping which is provided outside of nmstate definition.
It means that if we want to use `copy-from-mac` we need to provide the mac-address twice.

clusters/release_notes/known_issues.adoc
ori-amizur: The mac-address today doesn't exist in the nmstate definition.  It needs to be copied from the interface-mac mapping.

oafischer (approved):
lgtm

Optional: Use "As a result" instead of "therefore". Translation of "therefore" can be a little weird/have different meanings in some languages.

PR #6008, opened 2024-01-10
https://issues.redhat.com/browse/ACM-8893 CLC clean up (WIP)

oafischer:
Release images clean up


clusters/cluster_lifecycle/release_image_intro.adoc
```diff
@@ -1,61 +1,10 @@
...
-  labels:
-    channel: fast
-    visible: 'true'
-  name: img4.13.8-x86-64-appsub
-spec:
-  releaseImage: quay.io/openshift-release-dev/ocp-release:4.12.8-x86_64
-----
- 
-Additional release images are stored, but are not visible in the console. To view all of the available release images, run `kubectl get clusterimageset`.
+The release image specifies the version of {ocp} to use for building the cluster. By default, {ocp-short} uses the `clusterImageSets` resources to get the list of supported release images.
```
jc-berger: Not a big deal to decrease anthropomorphism and to increase user focus, you could rephrase to something like:

When you build your cluster, use the version of {ocp} that the release image specifies.

clusters/cluster_lifecycle/release_images_specify.adoc
```diff
@@ -1,25 +1,24 @@
 [#release-images-specify]
 = Specifying release images
 
-When you create a cluster on a provider by using {mce-short}, you must specify a release image to use for the new cluster. The release image specifies which version of {ocp} is used to build the cluster. By default, the `clusterImageSets` resources are used by {ocp-short} to get the list of supported release images.
+When you create a cluster on a provider by using {mce}, you must specify a release image to use for the new cluster. See the following topics to learn how to specify a release image.
```
jc-berger: Small changes for minimalism. Also, since `see the following topics` introduces the list, I think the sentence should end on that statement. So something like:

When you create a cluster on a provider by using {mce}, specify a release image to use for your new cluster. To specify a release image, see the following topics:

clusters/cluster_lifecycle/release_images_specify.adoc
```diff
@@ -1,25 +1,24 @@
...
+- <<cluster-image-set,Locating _ClusterImageSets_>>
+- <<config-cluster-image-set,Configuring _ClusterImageSets_>>
+- <<manual-release-image-cross-arch,Creating a release image to deploy a cluster on a different architecture>>
 
 [#cluster-image-set]
 == Locating _ClusterImageSets_
 
-The files that reference the release images are YAML files that are maintained in the link:https://github.com/stolostron/acm-hive-openshift-releases[acm-hive-openshift-releases GitHub repository] GitHub repository.
-Those files are used create the list of the available release images in the console. This includes the latest fast channel images from {ocp-short}. 
+The YAML files referencing the release images are maintained in the _acm-hive-openshift-releases_ GitHub repository. The files are used to create the list of the available release images in the console. This includes the latest fast channel images from {ocp-short}.
```
jc-berger, line 13: Is the user using the files to create the list of available release images? If so, let's make it more user focused (see rewrite below). If the files automatically creates these lists, then leaving the sentence as-is in passive voice is fine :)
 
Use the files to create a lists of available release images in the console. The files also include the latest fast channel images from {ocp-short}.

clusters/cluster_lifecycle/release_images_specify.adoc
```diff
@@ -34,42 +33,47 @@ spec:
...
 - Releases: Contains files that reference all of the release images for each {ocp-short} version (stable, fast, and candidate channels)
++
 *Note:* These releases have not all been tested and determined to be stable.
 
-* Stable: Contains files that reference the latest two stable versions of the release images for each {ocp-short} version that is supported.
-
-**Note:** By default, the current list of release images is updated one time an hour. After upgrading the product, it may take up to an hour for the list to reflect the recommended release image versions for the new version of the product.
+- Stable: Contains files that reference the latest two stable versions of the release images for each supported {ocp-short} version..
++
+*Note:* By default, the current list of release images updates one time every hour. After upgrading the product, it might take up to one hour for the list to reflect the recommended release image versions for the new version of the product.
```
jc-berger, line 52: Do we want to hold ourselves accountable to a strict one hour timeframe? Or to err on the side of caution, can we keep it broad? 

After the upgrading the product, it can take time to reflect he recommended release image version for the new version of the product.

clusters/cluster_lifecycle/release_images_specify.adoc
```diff
@@ -34,42 +33,47 @@ spec:
...
 
 [#config-cluster-image-set]
 == Configuring _ClusterImageSets_
 
-You can configure your `ClusterImageSets` with the following options. 
+You can configure your `ClusterImageSets` with the following options:
 
-- *Option 1:* Specify the image reference for the specific `ClusterImageSet` that you want to use in the console when creating a cluster. Each new entry you specify persists and is available for all future cluster provisions. See the following example entry: 
+- Option 1: Specify the image reference for the specific `ClusterImageSet` that you want to use in the console when creating a cluster. Each new entry you specify persists and is available for all future cluster provisions See the following example entry:
```
jc-berger: Not the biggest deal, but IBM SG does encourage us to lead with reason/purpose, then give action. So it would be:

To create a cluster in the console, specify the image reference for the specific `CluterImageSet` that you want to use.

clusters/cluster_lifecycle/release_images_specify.adoc
```diff
@@ -34,42 +33,47 @@ spec:
...
 +
 ----
 quay.io/openshift-release-dev/ocp-release:4.6.8-x86_64
 ----
 
-- *Option 2:* Manually create and apply a `ClusterImageSets` YAML file from the `acm-hive-openshift-releases` GitHub repository.
+- Option 2: Manually create and apply a `ClusterImageSets` YAML file from the `acm-hive-openshift-releases` GitHub repository.
 
-- *Option 3:* Follow the `README.md` in the link:https://github.com/stolostron/cluster-image-set-controller/blob/main/README.md[cluster-image-set-controller GitHub repository] to enable automatic updates of `ClusterImageSets` from a forked GitHub repository.
+- Option 3: Follow the `README.md` in the _cluster-image-set-controller_ GitHub repository to enable automatic updates of `ClusterImageSets` from a forked GitHub repository.
```
jc-berger: Same reason for suggested change for option 1. Again, this feedback is optional, as it's not a strict rule in the IBM SG. 

To enable automatic of `ClusterImageSets` from a forked GitHub repository, follow the `README.md` in the _cluster-image-set-controller_ GitHub repository.

clusters/cluster_lifecycle/release_images_specify.adoc
```diff
@@ -88,7 +92,7 @@ To create the release image for {ocp-short} images that do not support multiple
 
 . From the link:https://quay.io/repository/openshift-release-dev/ocp-release[{ocp-short} release registry], create a link:https://docs.docker.com/registry/spec/manifest-v2-2/[manifest list] that includes `x86_64`, `s390x`, `aarch64`, and `ppc64le` release images.
 
-.. Pull the manifest lists for both architectures in your environment from the link:https://quay.io/repository/openshift-release-dev/ocp-release?tab=tags[Quay repository] using the following example commands, replacing `4.x.1` with the current version of the product:
+.. Pull the manifest lists for both architectures in your environment from the link:https://quay.io/repository/openshift-release-dev/ocp-release?tab=tags[Quay repository] by running the following example commands. Replace `4.x.1` with the current version of the product:
```
jc-berger, line 95: Think it would read better if the `replace` action was a separate step? So:

.. Pull the manifest lists...
.. Replace `4.x.1`...

clusters/cluster_lifecycle/release_images_specify.adoc
```diff
@@ -97,32 +101,29 @@ podman pull quay.io/openshift-release-dev/ocp-release:4.x.1-s390x
 podman pull quay.io/openshift-release-dev/ocp-release:4.x.1-aarch64
 ----
 
-.. Log in to your private repository where you maintain your images:
+.. Log in to your private repository where you maintain your images by running the following command. Replace `<private-repo>` with the path to your repository.:
```
jc-berger, line 104: Same question in previous comment. What do you think of: 

.. Log in to your private repository...
.. Replace `<private-repo>`...

clusters/cluster_lifecycle/release_images_specify.adoc
```diff
@@ -97,32 +101,29 @@ podman pull quay.io/openshift-release-dev/ocp-release:4.x.1-s390x
...
 +
 ----
 podman login <private-repo>
 ----
-+
-Replace `private-repo` with the path to your repository.
 
-.. Add the release image manifest to your private repository by running the following commands that apply to your environment, replacing `4.x.1` with the current version of the product:
+
+.. Add the release image manifest to your private repository by running the following commands that apply to your environment. Replace `4.x.1` with the current version of the product. Replace `<private-repo>` with the path to your repository:
```
jc-berger, line 111: Same question as above.

clusters/cluster_lifecycle/release_images_specify.adoc
jc-berger: Ah, reviewing the whole doc, it looks like this `replace` is a recurring pattern. 

In which case, that would require a lot of tedious work to make every `replace` action a substep. And I don't even know if it would enhance and improve the user experience. 

With that said, don't feel obliged to make the change :) It can stay as-is!

clusters/cluster_lifecycle/release_images_specify.adoc
oafischer: I assume ACM or some part of ACM or OCP uses the files to create the list. It's definitely not the user creating the list.

clusters/cluster_lifecycle/release_images_specify.adoc
oafischer: This would be a technical change, I think it's safer to stick with the existing info since it was dev reviewed at some point in the past.

clusters/cluster_lifecycle/release_images_specify.adoc
oafischer: Fixed, thanks!

clusters/cluster_lifecycle/release_image_intro.adoc
oafischer: Updated

clusters/cluster_lifecycle/release_images_specify.adoc
oafischer: Updated

clusters/cluster_lifecycle/release_images_specify.adoc
oafischer: Updated

clusters/cluster_lifecycle/release_images_specify.adoc
oafischer: In the past we used to list out each variable replacement step after the command in a list, similar to what you are suggesting.

We decided a while back that we wanted to move those variable replacement steps to before the command (which makes more sense from a chronological point) and simply make it part of the prompt, without a list, unless there are several steps, in which case we might want to consider a general statement like "replace variables where needed".

jc-berger (approved):
/lgtm

PR #6018, opened 2024-01-16
OCPBUGS#25308: Updated the commands in "Configuring MetalLB" proc 

sr1kar99:


clusters/hosted_control_planes/config_metallb_bm.adoc
```diff
@@ -7,34 +7,36 @@ You must use a load balancer, such as MetalLB. The following example shows the s
...
 ----
-oc create -f - 
+cat << EOF| oc create -f -
 apiVersion: metallb.io/v1beta1
 kind: IPAddressPool
 metadata:
   name: metallb
   namespace: metallb-system
-spec:
+ spec:
```
jeniferh: Looks like my example mistakenly added a space before spec, this is incorrect and spec should be aligned w/ metadata (no space). Otherwise the examples all work correctly now, thanks!

jeniferh (approved):
One space issue, otherwise LGTM

clusters/hosted_control_planes/config_metallb_bm.adoc
sr1kar99: Done.
Removed the extra space.

PR #6040, opened 2024-01-18
https://issues.redhat.com/browse/ACM-8899 - sync (Required for 2.10 only, do not merge)

swopebe:


governance/create_gatekeeper.adoc
```diff
@@ -7,6 +7,8 @@ Use the Gatekeeper operator policy to install the Gatekeeper operator and Gateke
 
 - <<install-gatekeeper-operator-policy,Installing Gatekeeper using a Gatekeeper operator policy>>
 - <<creating-a-gatekeeper-policy-from-the-console,Creating a Gatekeeper policy from the console>>
+* <<gatekeeper-audit-sync,Setting _auditFromCache_ for sync details>>
+* <<gatekeeper-operator-sample,Gatekeeper operator custom resource>>
```
dockerymick: The references to subsections should be removed from the TOC, it'll appear in the TOC of the product doc

governance/create_gatekeeper.adoc
```diff
@@ -88,6 +90,55 @@ spec:
     other-annotation: "another test"
 ----
 
+[#gatekeeper-audit-sync]
+=== Setting _auditFromCache_ for sync details
+
+The gatekeeper operator exposes a setting in the CRD within the `auditFromCache` audit, which is disabled by default. If you set it to `Enabled`, you break your constraints because the cache requires additional settings. You need to set `config.gatekeeper.sh` for the sync details.
```
dockerymick: change gatekeeper operator to Gatekeeper Operator

https://github.com/gatekeeper/gatekeeper-operator

@mprahl would you help confirm? I can also add this task to my clean up issue @swopebe

governance/create_gatekeeper.adoc
```diff
@@ -88,6 +90,55 @@ spec:
...
+
+. Set `auditFromCache` to `Automatic` in the `operator.gatekeeper.sh/v1alpha1` API, as displayed in the following example:
```
dockerymick:
```suggestion
. Set `auditFromCache` to `Automatic` in the `Gatekeeper` resource, as displayed in the following example:
```

governance/create_gatekeeper.adoc
```diff
@@ -88,6 +90,55 @@ spec:
...
+
++
+[source,yaml]
+----
+apiVersion: operator.gatekeeper.sh/v1alpha1
+kind: Gatekeeper metadata:
```
dockerymick: This example seems a little off. metadata should be in a new line. @yiraeChristineKim will you confirm?

governance/create_gatekeeper.adoc
```diff
@@ -88,6 +90,55 @@ spec:
...
+kind: Gatekeeper metadata: 
+name: gatekeeper 
+spec: 
+  audit: replicas: 2 
+    logLevel: DEBUG 
+    auditFromCache: Automatic
+----
+
++
+See that the Gatekeeper operator adds `syncOnlys` to the config file in the following example:
```
dockerymick:
```suggestion
See that the Gatekeeper operator adds `syncOnlys` parameter section to the config file in the following example:
```

governance/create_gatekeeper.adoc
```diff
@@ -88,6 +90,55 @@ spec:
...
+----
+
+. Get the explanation of the `sync` setting, run the following command:
+
++
+[source,bash]
+----
+oc explain gatekeeper.spec.audit.auditFromCache
+----
+
```
dockerymick, line 146: I assume that the explanation of the setting is displayed from the users screen? Is there a verification step?

dockerymick (changes requested):

governance/create_gatekeeper.adoc
mprahl: I think at the very least that Gatekeeper should be capitalized. I'm not sure about "operator" though.

governance/create_gatekeeper.adoc
swopebe: I agree, but there already was one here. We decided not to add them, yes?

governance/create_gatekeeper.adoc
swopebe: operator is not caps, correct. Actually the other operators that we conref are also not capped...(MCE, GH)

swopebe: This doc linked doesn't follow the IBM or RH format, so I don't think that is what we should use for standards. I can see if it is officially in a source for naming guidance, however.

governance/create_gatekeeper.adoc
swopebe: @dockerymick actually if we are going to say Gatekeeper resource, I would rather write it as it is in the yaml: name: gatekeeper.

swopebe: I was not given verification, actually I wasn't given this as a task. This is all the template provided. @yiraeChristineKim can you help here?

governance/create_gatekeeper.adoc
```diff
@@ -24,11 +26,17 @@ Gatekeeper operator policy is monitored by the {product-title-short} configurati
 [#creating-a-gatekeeper-policy-from-the-console]
 == Creating a Gatekeeper policy from the console
 
-Install the Gatekeeper policy by creating a Gatekeeper policy from the console. Alternatively, navigate to the _Additional resources_ section for a reference to the sample YAML to deploy `policy-gatekeeper-operator.yaml`.
+See the following procedure to create the policy from the console:
```
swopebe, line 27: I also changed this section because it was a task, but presented in paragraph form and needed a short description.

swopebe: feel free to change short description on line 29.

governance/create_gatekeeper.adoc
```diff
@@ -88,6 +96,56 @@ spec:
...
++
+See that the Gatekeeper operator adds `syncOnlys` parameter section to the config file in the following example:
+
++
+[source,yaml]
+----
+apiVersion: config.gatekeeper.sh/v1alpha1
+kind: Config
+metadata:
+ name: config
```
swopebe, line 126: please check the YAML here-- placement of name on both samples.

governance/create_gatekeeper.adoc
dockerymick: Correct, we did agree to not list subsections in the TOC of the file

governance/create_gatekeeper.adoc
dockerymick: Sounds good, thanks for confirming ! We are on the same page to use "Gatekeeper operator".

governance/create_gatekeeper.adoc
yiraeChristineKim: User type `oc explain gatekeeper.spec.audit.auditFromCache` on own terminal

governance/create_gatekeeper.adoc
```diff
@@ -88,6 +96,56 @@ spec:
...
+metadata: 
+  name: gatekeeper 
+spec: 
+  audit: replicas: 2 
+    logLevel: DEBUG 
+    auditFromCache: Automatic
+----
+
++
+See that the Gatekeeper operator adds `syncOnlys` parameter section to the config file in the following example:
```
yiraeChristineKim, line 118: `The gatekeeper operator collects the resources from Constraints that users installed and then inserts them into the Config resource. The gatekeeper operator will create config resources if it does not exist.
`
Should we add this?  I guess users might get confused about the config resource. I think we need to explain where syncOnlys are from.

governance/create_gatekeeper.adoc
```diff
@@ -88,6 +96,56 @@ spec:
...
+
++
+[source,yaml]
+----
+apiVersion: config.gatekeeper.sh/v1alpha1
+kind: Config
+metadata:
+ name: config
+ namespace: "gatekeeper-system"
```
yiraeChristineKim, line 127: Could you change this to `openshift-gatekeeper-system`

governance/create_gatekeeper.adoc
yiraeChristineKim: @dockerymick right. `metadata` should be in the new line Thanks!

PR #6043, opened 2024-01-19
Quick update

dockerymick:
Honestly, I saw a red x and thought the book was broken. However, I remember that it is because of how multicluster global hub referenced. 
Screenshot 2024-01-18 at 11 04 50 PM

@oafischer will you help with getting that updated? I will create a separate issue for the Pantheon book investigation 

However, I did notice extra brackets in link references


oafischer:
I can take a look

PR #6061, opened 2024-01-24
https://issues.redhat.com/browse/ACM-8962

dockerymick:
https://issues.redhat.com/browse/ACM-8962

### DIFF
--- a/governance/cert_policy_ctrl.adoc
+++ b/governance/cert_policy_ctrl.adoc
@@ -133,7 +133,7 @@ When your certificate policy controller is created on your hub cluster, a replic
 
 - Learn how to manage a certificate policy, see xref:../governance/create_policy.adoc#managing-security-policies[Managing security policies] for more details. 
 
-- Refer to xref:../governance/policy_controllers.adoc#policy-controllers[Policy controllers] for more topics.
+- Refer to xref:../governance/policy_controllers_intro.adoc#policy-controllers[Policy controllers introduction] for more topics.
 
-- Return to the xref:../governance/ert_intro.adoc#cert-intro[Certificates introduction].
+- Return to the xref:../governance/cert_intro.adoc#cert-intro[Certificates introduction].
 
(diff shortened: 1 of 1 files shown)


governance/cert_policy_ctrl.adoc
```diff
@@ -21,6 +21,8 @@ The certificate policy controller is created on your managed cluster. The contro
 
 Certificate policy controller does not support the `enforce` feature.
 
+*Note:* The certificate policy controller expects to find a certificate in a secret in only the `tls.crt` key. If a secret is stored under a different key, add a label named `certificate_key_name` with a value set to the key to let the certificate policy controller know to look in a different key. For example, if a secret contains a certificate stored in the key named `sensor-cert.pem`, add the following label to the secret: `certificate_key_name: sensor-cert.pem`.
```
jc-berger: Is "expects to find" the most accurate verb choice? 

If the certificate policy controller _automatically_ looks for a certificate in a secret in the `tls.crt` then I think we could be more direct in our word choice. Something like:

The certificate policy controller automatically looks for a certificate in a secret in only the `tls.crt` key.

jc-berger (approved):
/lgtm

governance/cert_policy_ctrl.adoc
gparvin: Thanks, I agree with the update recommended  here.

gparvin:
This looks good once the suggestion provided is handled.  Thanks @dockerymick

governance/cert_policy_ctrl.adoc
dockerymick: Update after peer review
```suggestion
*Note:* The certificate policy controller automatically looks for a certificate in a secret in only the `tls.crt` key. If a secret is stored under a different key, add a label named `certificate_key_name` with a value set to the key to let the certificate policy controller know to look in a different key. For example, if a secret contains a certificate stored in the key named `sensor-cert.pem`, add the following label to the secret: `certificate_key_name: sensor-cert.pem`. 
```

dockerymick:
Thanks for the reviews, feedback, and approvals! Merging

PR #6082, opened 2024-01-31
https://issues.redhat.com/browse/ACM-9161 Expand globalnet docs

oafischer:
Also for 2.9 and 2.8.

### DIFF
--- a/networking/submariner/install_subctl.adoc
+++ b/networking/submariner/install_subctl.adoc
@@ -1,7 +1,7 @@
 [#installing-subctl-command-utility]
 = Installing the subctl command utility
 
-The `subctl` utility ships in a container image. Complete the following steps to install the `subctl` utility locally:
+The `subctl` utility ships in a container image. To install the `subctl` utility locally, complete the following steps:
 
 . Log in to the registry by running the following command and entering your credentials when prompted:
 +
--- a/networking/submariner/install_subctl.adoc
+++ b/networking/submariner/install_subctl.adoc
@@ -1,7 +1,7 @@
 [#installing-subctl-command-utility]
 = Installing the subctl command utility
 
-The `subctl` utility is shipped in a container image. Complete the following steps to install the `subctl` utility locally:
+The `subctl` utility ships in a container image. Complete the following steps to install the `subctl` utility locally:
 
 . Log in to the registry by running the following command and entering your credentials when prompted:
 +
@@ -26,6 +26,11 @@ tar -C /tmp/ -xf /tmp/subctl-v0.16*-linux-amd64.tar.xz
 ----
 install -m744 /tmp/subctl-v0.16*/subctl-v0.16*-linux-amd64 /$HOME/.local/bin/subctl
 ----
+
+*Notes:*
+
+- Make sure that the `subctl` and Submariner versions match.
+- For disconnected environments only, make sure to mirror the `submariner-nettest` image.
 
 [#using-subctl-commands]
 == Using the subctl commands
--- a/networking/submariner/subm_configure.adoc
+++ b/networking/submariner/subm_configure.adoc
@@ -102,7 +102,7 @@ spec:
 oc apply -f submariner-broker.yaml
 ----
 
-[#submariner-globalnet-enable]
+[#subm-globalnet-numperip]
 === Configuring the number of global IPs
 
 You can assign a configurable number of global IPs by changing the value of the `numberOfIPs` field in the `ClusterGlobalEgressIP` resource. The default value is 8. See the following example:
--- a/networking/submariner/subm_configure.adoc
+++ b/networking/submariner/subm_configure.adoc
@@ -1,7 +1,11 @@
 [#configuring-submariner]
 = Configuring Submariner
 
-{product-title} provides Submariner as an add-on for your hub cluster. You can find more information about Submariner in the link:https://submariner.io/[Submariner open source project documentation].
+{product-title} provides Submariner as an add-on for your hub cluster. Read the following topics to learn how to configure Submariner:
+
+- <<submariner-prereqs,Prerequisites>>
+- <<submariner-ports,Submariner ports table>>
+- <<submariner-globalnet,Globalnet>>
 
 [#submariner-prereqs]
 == Prerequisites
@@ -9,7 +13,7 @@
 Ensure that you have the following prerequisites before using Submariner:
 
 * A credential to access the hub cluster with `cluster-admin` permissions.
-* IP connectivity must be configured between the gateway nodes. When connecting two clusters, at least one of the clusters must be accessible to the gateway node using its public or private IP address designated to the gateway node. See link:https://submariner.io/operations/nat-traversal[Submariner NAT Traversal] for more information.
+* IP connectivity must be configured between the gateway nodes. When connecting two clusters, at least one of the clusters must be accessible to the gateway node by using its public or private IP address designated to the gateway node. _Submariner NAT Traversal_ for more information.
 * If you are using OVN Kubernetes, clusters must be at {ocp} version 4.12 or later. 
 * If your {ocp} clusters use OpenShift SDN CNI, the firewall configuration across all nodes in each of the managed clusters must allow 4800/UDP in both directions.
 * The firewall configuration must allow 4500/UDP and 4490/UDP on the gateway nodes for establishing tunnels between the managed clusters.
@@ -26,7 +30,7 @@ Ensure that you have the following prerequisites before using Submariner:
 [#submariner-ports]
 == Submariner ports table
 
-View the following table to see which Submariner ports you need to enable:
+View the following table to see the Submariner ports that you must enable:
 
 |===
 | Name | Default value | Customizable | Optional or required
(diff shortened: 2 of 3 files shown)


networking/submariner/subm_configure.adoc
```diff
@@ -77,22 +86,41 @@ oc get ManagedClusterSet <cluster-set-name> -o jsonpath="{.metadata.annotations[
...
 oc apply -f submariner-broker.yaml
-----  
+----
+
+[#subm-globalnet-numperip]
+=== Configuring the number of global IPs
+
+You can assign a configurable number of global IPs by changing the value of the `numberOfIPs` field in the `ClusterGlobalEgressIP` resource. The default value is 8. See the following example:
+
+[source,yaml]
```
yboaron, line 110: Not 100% sure about this, 
@sridhargaddam  WDYT?

networking/submariner/subm_configure.adoc
```diff
@@ -77,22 +86,41 @@ oc get ManagedClusterSet <cluster-set-name> -o jsonpath="{.metadata.annotations[
...
 kind: Broker
 metadata:
-  name: submariner-broker
-  namespace: <broker-namespace>
+  name: submariner-broker <1>
+  namespace: broker-namespace <2>
 spec:
-  globalnetEnabled: <true-or-false>
+  globalnetEnabled: true-or-false <3>
 ----
```
yboaron, line 93: Maybe we can add here details about how to change the default GlobalNet CIDR,
Submariner uses "242.0.0.0/8" as default GlobalNet CIDR, to change this value , 'globalnetCIDRRange' field should be set (in submariner-broker spec), the CIDR should be /8.

cc @sridhargaddam

oafischer: That's a good idea, we can remove the other section that we are not sure about if we do this.

networking/submariner/subm_configure.adoc
```diff
@@ -47,22 +51,27 @@ View the following table to see which Submariner ports you need to enable:
...
 
-See the link:https://submariner.io/getting-started/#prerequisites[Submariner upstream prerequisites documentation] for more detailed information about the prerequisites.
-
 [#submariner-globalnet]
 == Globalnet
 
-Globalnet is a feature included with the Submariner add-on that supports connectivity between clusters with overlapping CIDRs. Globalnet is a cluster set wide configuration, and can be selected when the first managed cluster is added to the cluster set. When Globalnet is enabled, each managed cluster is allocated a global CIDR from the virtual Global Private Network. The global CIDR is used for supporting inter-cluster communication.
+Globalnet is a Submariner add-on feature that allows you to connect clusters with overlapping Classless Inter-Domain Routings (CIDRs), without changing the CIDRs on existing clusters. Globalnet is a cluster set-wide configuration that you can select when you add the first managed cluster to a cluster set.
+
+If you enable Globalnet, the virtual Global Private Network assigns a subnet to each managed cluster that you can configure during deployment.
```
sridhargaddam: How about the following?

```suggestion
If you enable Globalnet, every managed cluster receives a global CIDR from the virtual Global Private Network, which is used to facilitate inter-cluster communication.
```

networking/submariner/subm_configure.adoc
```diff
@@ -47,22 +51,27 @@ View the following table to see which Submariner ports you need to enable:
 | Required
 |===
 
-See the link:https://submariner.io/getting-started/#prerequisites[Submariner upstream prerequisites documentation] for more detailed information about the prerequisites.
-
 [#submariner-globalnet]
 == Globalnet
 
-Globalnet is a feature included with the Submariner add-on that supports connectivity between clusters with overlapping CIDRs. Globalnet is a cluster set wide configuration, and can be selected when the first managed cluster is added to the cluster set. When Globalnet is enabled, each managed cluster is allocated a global CIDR from the virtual Global Private Network. The global CIDR is used for supporting inter-cluster communication.
+Globalnet is a Submariner add-on feature that allows you to connect clusters with overlapping Classless Inter-Domain Routings (CIDRs), without changing the CIDRs on existing clusters. Globalnet is a cluster set-wide configuration that you can select when you add the first managed cluster to a cluster set.
```
sridhargaddam, line 57: Should "cluster set" be a single word - i.e., "clusterSet"?

networking/submariner/subm_configure.adoc
```diff
@@ -47,22 +51,27 @@ View the following table to see which Submariner ports you need to enable:
...
+
+If you enable Globalnet, the virtual Global Private Network assigns a subnet to each managed cluster that you can configure during deployment.
 
 *Important:* You must enable Globalnet when clusters in a cluster set might have overlapping CIDRs.
 
-After Submariner add-ons are installed on any clusters in the set, you cannot change the Globalnet setting. When using the console, the `ClusterAdmin` can enable Globalnet for a cluster set by selecting the option *Enable Globalnet* when enabling the Submariner add-on for clusters in the cluster set. After enabling Globalnet you cannot disable it, unless you remove Submariner.
+After you install Submariner add-ons on any clusters in the set, you cannot change the Globalnet setting. When using the console, the `ClusterAdmin` can enable Globalnet for a cluster set by selecting the option *Enable Globalnet* when enabling the Submariner add-on for clusters in the cluster set. After enabling Globalnet, you cannot disable it, unless you remove Submariner.
```
sridhargaddam: I think we now support modifying the state of `Globalnet` when the clusterSet is empty. So users need not remove Submariner to change the status of Globalnet, they just have to remove all the managed clusters from the clusterSet. @tpantelis @vthapar to confirm.

networking/submariner/subm_configure.adoc
oafischer: In this case we are just referring to a "set of clusters" in general, but we can also refer to `clusterSet`, I'm OK with either.

networking/submariner/subm_configure.adoc
sridhargaddam: Normally, the default GlobalnetCIDR should be fine for majority of the users. We also have many other params which we can include if required, but it would be too verbose and can even lead to confusion. IMO the purpose of this section is to document the minimum steps on how to enable Globanet manually.

> the CIDR should be /8.

JFYI, the CIDR need not be /8

networking/submariner/subm_configure.adoc
sridhargaddam: Its correct. Please see - https://github.com/submariner-io/submariner/blob/384d06730181423df4ef503f69dd39f062b44350/pkg/globalnet/controllers/types.go#L77

networking/submariner/subm_configure.adoc
tpantelis: I think in end-user docs it should be "cluster set". "clusterSet" isn't a word.

networking/submariner/subm_configure.adoc
tpantelis: That sounds correct.

networking/submariner/subm_configure.adoc
oafischer: I agree that we want to keep it simple. Since the new topic that we weren't sure about is correct, and it discusses how to change the CIDR, I think we can leave this section as it is.

networking/submariner/subm_configure.adoc
oafischer: Updated, thanks!

oafischer:
Thanks for the reviews @sridhargaddam @tpantelis @yboaron! I made some of the suggested changes. Feel free to take a look at it one last time.

networking/submariner/install_subctl.adoc
```diff
@@ -27,6 +27,11 @@ tar -C /tmp/ -xf /tmp/subctl-v0.16*-linux-amd64.tar.xz
 install -m744 /tmp/subctl-v0.16*/subctl-v0.16*-linux-amd64 /$HOME/.local/bin/subctl
 ----
 
+*Notes:*
```
jc-berger, line 30: Notes or prereq? 

Notes are important info the user should know to make the steps easier and to improve the user experience. 

required action or steps the user must take should be documented as a prereq or as steps part of the procedure/task

networking/submariner/install_subctl.adoc
```diff
@@ -1,7 +1,7 @@
 [#installing-subctl-command-utility]
 = Installing the subctl command utility
 
-The `subctl` utility is shipped in a container image. Complete the following steps to install the `subctl` utility locally:
+The `subctl` utility ships in a container image. Complete the following steps to install the `subctl` utility locally:
```
jc-berger: IBM SG says to lead with reason, then action. 

So: 

To install the `subctl` utility locally, complete the following steps:

FWIW, I think that re-writes a bit better and most straightforward for "why" the using is completing these steps we're giving them.

networking/submariner/subm_configure.adoc
```diff
@@ -1,15 +1,19 @@
 [#configuring-submariner]
 = Configuring Submariner
 
-{product-title} provides Submariner as an add-on for your hub cluster. You can find more information about Submariner in the link:https://submariner.io/[Submariner open source project documentation].
+{product-title} provides Submariner as an add-on for your hub cluster. Read the following topics to learn how to configure Submariner:
```
jc-berger: Some as before. I think the "why am I doing this?" is clearer if we start with reason.

To learnhow to configure Submariner, read the following topics:

networking/submariner/subm_configure.adoc
```diff
@@ -1,15 +1,19 @@
...
+- <<submariner-globalnet,Globalnet>>
 
 [#submariner-prereqs]
 == Prerequisites
 
 Ensure that you have the following prerequisites before using Submariner:
 
 * A credential to access the hub cluster with `cluster-admin` permissions.
-* IP connectivity must be configured between the gateway nodes. When connecting two clusters, at least one of the clusters must be accessible to the gateway node using its public or private IP address designated to the gateway node. See link:https://submariner.io/operations/nat-traversal[Submariner NAT Traversal] for more information.
+* IP connectivity must be configured between the gateway nodes. When connecting two clusters, at least one of the clusters must be accessible to the gateway node by using its public or private IP address designated to the gateway node. _Submariner NAT Traversal_ for more information.
```
jc-berger: At the end, shouldn't we document it as an action for the user to take? 

For more information, see _Submariner NAT Traversal_.

networking/submariner/install_subctl.adoc
oafischer: These are notes.

networking/submariner/subm_configure.adoc
oafischer: Yes I must have deleted the word "see" by accident, good catch! Should be fixed now.

jc-berger (approved):
/lgtm

networking/submariner/subm_configure.adoc
oafischer: Just to be clear, is this also true for 2.9 and 2.8?

PR #6100, opened 2024-02-02
ACM#9464: Add instructions to download  by using the CLI

xenolinux:
Jira: https://issues.redhat.com/browse/ACM-9464

This PR adds the instructions to install the `hcp` binary by using the CLI.

### DIFF
--- a/clusters/hosted_control_planes/install_hcp_cli.adoc
+++ b/clusters/hosted_control_planes/install_hcp_cli.adoc
@@ -32,3 +32,48 @@ hcp create cluster <platform> --help <1>
 ----
 
 <1> The supported platforms are `aws`, `agent`, and `kubevirt`.
+
+[#hosted-install-console]
+== Installing the hosted control plane command line interface by using the CLI
+
+You can install the hosted control plane command line interface, `hcp`, by using the CLI, by completing the following steps:
+
+. Get the URL to download the `hcp` binary by running the following command:
++
+----
+oc get ConsoleCLIDownload hcp-cli-download -o json | jq -r ".spec"
+----
+
+. Download the `hcp` binary by running the following command:
++
+----
+wget <hcp_cli_download_url> <1>
+----
++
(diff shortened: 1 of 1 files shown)


clusters/hosted_control_planes/install_hcp_cli.adoc
```diff
@@ -31,4 +31,49 @@ You can now use the `hcp create cluster` command to create and manage hosted clu
 hcp create cluster <platform> --help <1>
 ----
 
-<1> The supported platforms are `aws`, `agent`, and `kubevirt`.
\ No newline at end of file
+<1> The supported platforms are `aws`, `agent`, and `kubevirt`.
+
+[#hosted-install-console]
```
elsapassaro, line 36: Just a minor comment, wouldn't it be better to switch the anchors of the two section for consistency? 
I.e. use `[#hosted-install-cli]` for this section to download from CLI and  `[#hosted-install-console]` for the previous section (line 1) to download it from console

xenolinux: oops yes makes sense; updated the PR

elsapassaro (approved):
/lgtm

clusters/hosted_control_planes/install_hcp_cli.adoc
```diff
@@ -1,4 +1,4 @@
-[#hosted-install-cli]
+[#hosted-install-console]
```
dockerymick: I'm not sure why the anchor is being changed. It would be inaccurate to change the anchor to console. Also keep in mind that if the anchor is changed here, we have to search for other instances and also update them. The anchor should remain the same
```suggestion
[#hosted-install-cli]
```

dockerymick (changes requested):

clusters/hosted_control_planes/install_hcp_cli.adoc
xenolinux: Addressed

PR #6104, opened 2024-02-05
IBM Power documentation updates for HCP

alishaIBM:
Issue: https://issues.redhat.com/browse/ACM-9778

### DIFF
--- a/clusters/hosted_control_planes/scale_nodepool_hosted_ibmpower.adoc
+++ b/clusters/hosted_control_planes/scale_nodepool_hosted_ibmpower.adoc
@@ -101,8 +101,8 @@ See the following output:
 [source,bash]
 ----
 NAME                                CLUSTER  NODENAME PROVIDERID     PHASE     AGE   VERSION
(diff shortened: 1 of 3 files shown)


clusters/hosted_control_planes/scale_nodepool_hosted_ibmpower.adoc
```diff
@@ -120,27 +120,27 @@ See the following output:
 [source,bash]
```
swopebe, line 120: I really would rather we remove these long outputs this release, if we can. We discussed calling out specifically what the user needs to know or see from the output, which in this case is `true`, `false` and maybe other info, but if the user sees this in the terminal, why do we need to repeat it in the doc.

We are trying to simplify to doc for less maintenance.

swopebe (approved):
Everything looks good but I do have the same comment as last release about the output.

swopebe:
I am waiting on a peer review from my team and then this is ready to merge.

jc-berger (approved):
/lgtm

PR #6105, opened 2024-02-06
https://issues.redhat.com/browse/ACM-8899 cache setting

swopebe:
from other PR:

The gatekeeper operator collects the resources from Constraints that users installed and then inserts them into the Config resource. The gatekeeper operator will create config resources if it does not exist.
Should we add this? I guess users might get confused about the config resource. I think we need to explain where syncOnlys are from.


yiraeChristineKim:
I agree to add that "The gatekeeper operator collects the resources from Constraints that users installed and then inserts them into the Config resource. The gatekeeper operator will create config resources if it does not exist."

governance/create_gatekeeper.adoc
```diff
@@ -88,6 +94,58 @@ spec:
     other-annotation: "another test"
 ----
 
+[#gatekeeper-audit-sync]
+=== Configuring _auditFromCache_ for sync details
+
+The Gatekeeper operator exposes a setting in the CRD within the `auditFromCache` audit, which is disabled by default. If you set it to `Enabled`, you break your constraints because the cache requires additional settings. You need to set `config.gatekeeper.sh` for the sync details.
```
yiraeChristineKim: I don't think it breaks constraints when it is enabled. This is from Gatekeeper doc 
```
By default, audit will request each resource from the Kubernetes API during each audit cycle. To rely on the audit informer cache instead, use the flag --audit-from-cache=true. Note that this requires replication of Kubernetes resources into the audit cache before they can be evaluated against the enforced policies. Refer to the [Replicating data](https://open-policy-agent.github.io/gatekeeper/website/docs/sync) section for more information.
```

`audit-from-cache` same as `auditFromCache`

yiraeChristineKim: `Replicating data` is same as setting syncOnly in `config.gatekeeper.sh`

governance/create_gatekeeper.adoc
```diff
@@ -88,6 +94,58 @@ spec:
...
+
+The Gatekeeper operator collects resources from constraints that users install and then inserts those resources into the config resource. If resources do not exist, the operator creates config resources.
+
```
yiraeChristineKim, line 103: To me this is very clear Thx

governance/create_gatekeeper.adoc
swopebe: @yiraeChristineKim I am not sure what you mean by this: audit-from-cache same as auditFromCache

Are you asking about the header? We don't `code` the headers because they build our funny, we italicize instead, but yes this is the same (if that is the question).

Can you please revise for whatever is needed for the following comments, which I gathered from what was given to me at some point:

`I don't think it breaks constraints when it is enabled. This is from Gatekeeper doc`
`Replicating data is same as setting syncOnly in config.gatekeeper.sh`

What needs to change, for certain?

yiraeChristineKim: Sorry for the unclear, "you break your constraints because the cache requires additional settings.". Actually, it doesn't break if a user doesn't set `conifg.spec.sync`. How about removing "you break your constraints because the cache requires additional settings" and add "for more detail click this link"(https://open-policy-agent.github.io/gatekeeper/website/docs/audit/#configuring-audit)

yiraeChristineKim (approved):

yiraeChristineKim (approved):

governance/create_gatekeeper.adoc
```diff
@@ -88,6 +94,58 @@ spec:
     other-annotation: "another test"
 ----
 
+[#gatekeeper-audit-sync]
+=== Configuring _auditFromCache_ for sync details
+
+The Gatekeeper operator exposes a setting in the CRD within the `auditFromCache` audit, which is disabled by default. If you enable `auditFromCache`, then you need to set `config.gatekeeper.sh` for the sync details. See link:https://open-policy-agent.github.io/gatekeeper/website/docs/audit/#configuring-audit[Configuring Audit] in the Gatekeeper documentation.
```
dockerymick:
```suggestion
The Gatekeeper operator exposes a setting in the custom resource definition within the `auditFromCache` audit, which is disabled by default. If you enable `auditFromCache`, then you need to set `config.gatekeeper.sh` for the sync details. See link:https://open-policy-agent.github.io/gatekeeper/website/docs/audit/#configuring-audit[Configuring Audit] in the Gatekeeper documentation.
```

governance/create_gatekeeper.adoc
```diff
@@ -88,6 +94,58 @@ spec:
...
+
+The Gatekeeper operator collects resources from constraints that users install and then inserts those resources into the config resource. If resources do not exist, the operator creates config resources.
```
dockerymick, line 102: I recommend writing out configuration instead of using "config"

dockerymick (approved):
Left some suggestions and checked links. Links are good!

PR #6110, opened 2024-02-06
https://issues.redhat.com/browse/ACM-9681 known issue with Kube/API

swopebe:
https://issues.redhat.com/browse/ACM-9681

### DIFF
--- a/release_notes/known_issues_observability.adoc
+++ b/release_notes/known_issues_observability.adoc
@@ -26,8 +26,8 @@ Review the known issues for {product-title}. The following list contains known i
 For your {ocp} cluster, see https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/release_notes/ocp-4-12-release-notes#ocp-4-12-known-issues. 
 
 For more about deprecations and removals, see xref:../release_notes/deprecate_remove.adoc#deprecations-removals[Deprecations and removals].
-[#permission-to-managed-cluster-monitoring-denied]
 
+[#permission-to-managed-cluster-monitoring-denied]
 == Observatorium API gateway pods in a restored hub cluster might have stale tenant data
 //2.9:9681
 
--- a/release_notes/known_issues_observability.adoc
+++ b/release_notes/known_issues_observability.adoc
@@ -49,8 +49,10 @@ caller=handler.go:551 level=error component=receive component=receive-handler te
 
 See the following procedure to resolve this issue:
 
-. Scale the Observatorium API gateway deployment down to zero.
-. Scale the Observatorium API back to 2 or N if working with a custom deployment.
+. Scale down the `observability-observatorium-api` deployment instances from `N` to `0`.
+. Scale up the `observability-observatorium-api` deployment instances from `0` to `N`. if working with a custom deployment.
+
+*Note:* `N` = `2` by default, but might be greater than `2` in some custom configuration environments.
 
 This restarts all Observatorium API gateway pods with the correct tenant information, and the data from collectors start displaying in Grafana in between 5-10 minutes.
 
--- a/release_notes/known_issues_observability.adoc
+++ b/release_notes/known_issues_observability.adoc
@@ -50,7 +50,7 @@ caller=handler.go:551 level=error component=receive component=receive-handler te
 See the following procedure to resolve this issue:
 
 . Scale down the `observability-observatorium-api` deployment instances from `N` to `0`.
-. Scale up the `observability-observatorium-api` deployment instances from `0` to `N`. if working with a custom deployment.
+. Scale up the `observability-observatorium-api` deployment instances from `0` to `N`. 
 
 *Note:* `N` = `2` by default, but might be greater than `2` in some custom configuration environments.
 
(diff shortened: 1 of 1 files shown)


release_notes/known_issues_observability.adoc
```diff
@@ -26,6 +26,33 @@ Review the known issues for {product-title}. The following list contains known i
...
+----
+
+Thanos receives pods logs with the following errors:
+
+----
+caller=handler.go:551 level=error component=receive component=receive-handler tenant=xxxx err="no matching hashring to handle tenant" msg="internal server error"
+----
+
+See the following procedure to resolve this issue:
+
```
swopebe, line 51: these are confusing to me, we are telling the user to go through 1 step to 0 and one back to 2 or N? In that order? Or are these two different methods?

subbarao-meduri: @swopebe
The deployment name is `observability-observatorium-api`. By default, they are deployed as 2 instances. However, customer could have changed number of instances to more than 2 if scaling is needed. The fix is
1. Scale down  number of instances from N -> 0 
2. Scale up number of instances from 0 -> N 
(N = 2 by default, but could be > 2 in some environments, if custom configuration is used)

subbarao-meduri (approved):
/lgtm

release_notes/known_issues_observability.adoc
```diff
@@ -26,6 +26,35 @@ Review the known issues for {product-title}. The following list contains known i
 For your {ocp} cluster, see https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/release_notes/ocp-4-12-release-notes#ocp-4-12-known-issues. 
 
 For more about deprecations and removals, see xref:../release_notes/deprecate_remove.adoc#deprecations-removals[Deprecations and removals].
+[#permission-to-managed-cluster-monitoring-denied]
+
+== Observatorium API gateway pods in a restored hub cluster might have stale tenant data
```
dockerymick:
```suggestion

[#permission-to-managed-cluster-monitoring-denied]
== Observatorium API gateway pods in a restored hub cluster might have stale tenant data
```

release_notes/known_issues_observability.adoc
```diff
@@ -26,6 +26,35 @@ Review the known issues for {product-title}. The following list contains known i
...
+//2.9:9681
+
+The Observatorium API gateway pods in a restored hub cluster might contain stale tenant data after a backup and restore procedure because of a Kubernetes limitation. See link:https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically[Mounted ConfigMaps are updated automatically] for more about the limitation.
```
dockerymick, line 34: I see that "ConfigMap" is used. If I am recalling correctly, I believe that is why we had instances of `ConfigMap`. We wanted to be consistent with Kubernetes. With seeing this, we should be aligned with Kubernetes. Let's discuss with the team

release_notes/known_issues_observability.adoc
```diff
@@ -26,6 +26,35 @@ Review the known issues for {product-title}. The following list contains known i
...
+
+As a result, the Observatorium API and Thanos gateway rejects metrics from collectors, and the {product-title-short} Grafana dashboards do not display data.
+
+See the following errors from the Observatorium API gateway pod logs:
+
+----
```
dockerymick, line 40:
```suggestion
[source,bash]
----
```

release_notes/known_issues_observability.adoc
```diff
@@ -26,6 +26,35 @@ Review the known issues for {product-title}. The following list contains known i
...
+level=error name=observatorium caller=logchannel.go:129 msg="failed to forward metrics" returncode="500 Internal Server Error" response="no matching hashring to handle tenant\n"
+----
+
+Thanos receives pods logs with the following errors:
+
+----
```
dockerymick, line 46:
```suggestion
[source,bash]
----
```

dockerymick (changes requested):

release_notes/known_issues_observability.adoc
swopebe: @dockerymick  ConfigMap can be used, but what we don't want is configmap or configMap or anything else we try here.

swopebe: Here is the guidance: yes[ config map (noun)](https://redhat-documentation.github.io/supplementary-style-guide/#config-map)
Description: In Red Hat OpenShift, a config map holds configuration data for pods to consume. The API object for a config map is ConfigMap.

So you are right, we need to code it or write it out.

release_notes/known_issues_observability.adoc
swopebe, line 40: Not so sure about this one, we actually decided not to do this but I see it here and there in the doc now; I think some folks started doing on on their own. I think this is valid for a bash command. This is an error, however. I am going to reject this and bring it up again tomorrow.

release_notes/known_issues_observability.adoc
swopebe, line 46: same, it is an error and not a bash command. (???) Interested in the reason for _source, bash_ here, but we have not discussed using this across the team. Let's chat about it tomorrow.

release_notes/known_issues_observability.adoc
swopebe: @dockerymick I am going to leave the title the way it is here, but normally, yes we would code. Here is why:

This is the exact title of the third part documentation, it isn't our content.

release_notes/known_issues_observability.adoc
dockerymick: My reason is because it is displayed from the terminal. I considered CLI views as "bash"

dockerymick (approved):

release_notes/known_issues_observability.adoc
swopebe: Ok, I understand, but want to see what others use for errors. These would potentially build with a copy block but and error would not need that.

PR #6113, opened 2024-02-07
IBM Z documentation updates for HCP

krishnanshruthi:
Issue: https://issues.redhat.com/browse/ACM-9727

### DIFF
--- a/clusters/hosted_control_planes/scale_nodepool_hosted_ibmz.adoc
+++ b/clusters/hosted_control_planes/scale_nodepool_hosted_ibmz.adoc
@@ -137,4 +137,6 @@ clusterversion.config.openshift.io/version   4.15.0-ec.2   True        False    
 oc --kubeconfig <hosted_cluster_name>.kubeconfig get clusteroperators
 ----
 
-For each component of your cluster, the output shows: VERSION, AVAILABLE, PROGRESSING, DEGRADED, SINCE, and MESSAGE. For an output example, see the link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.14/html/installing/installing-on-ibm-z-and-ibm-linuxone#installation-operators-config_installing-ibm-z[Initial Operator configuration] section in the {ocp-short} documentation.
+For each component of your cluster, the output shows the following cluster operator statuses: `VERSION`, `AVAILABLE`, `PROGRESSING`, `DEGRADED`, `SINCE`, and `MESSAGE`. 
+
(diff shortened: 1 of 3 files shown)


clusters/hosted_control_planes/scale_nodepool_hosted_ibmz.adoc
```diff
@@ -144,25 +144,25 @@ See the following output:
 [source,bash]
```
swopebe: I really would rather we remove these long outputs this release, if we can. We discussed calling out specifically what the user needs to know or see from the output, which in this case is true, false and maybe other info, but if the user sees this in the terminal, why do we need to repeat it in the doc.

We are trying to simplify to doc for less maintenance.

swopebe (approved):
See comment about output that I would like to continue to discuss, if possible.

clusters/hosted_control_planes/scale_nodepool_hosted_ibmz.adoc
SNiemann15: @swopebe How about removing the entire table and simply writing something like: 

For each component of your cluster, the output shows:  VERSION, AVAILABLE, PROGRESSING, DEGRADED, SINCE, and MESSAGE if any. For an output example see the  Initial Operator configuration of your provider in [Installing and configuring OpenShift Container Platform clusters](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.14/html/installing/index).

swopebe: @SNiemann15 I love that idea if you all agree to it!

clusters/hosted_control_planes/scale_nodepool_hosted_ibmz.adoc
```diff
@@ -137,32 +137,4 @@ clusterversion.config.openshift.io/version   4.14.0-ec.2   True        False
...
-openshift-controller-manager               4.14.2    True        False         False      5h47m   
-openshift-samples                          4.14.2    True        False         False      5h34m   
-operator-lifecycle-manager                 4.14.2    True        False         False      5h47m   
-operator-lifecycle-manager-catalog         4.14.2    True        False         False      5h47m   
-operator-lifecycle-manager-packageserver   4.14.2    True        False         False      5h47m   
-service-ca                                 4.14.2    True        False         False      5h35m   
-storage                                    4.14.2    True        False         False      5h47m   
-----
+For each component of your cluster, the output shows: VERSION, AVAILABLE, PROGRESSING, DEGRADED, SINCE, and MESSAGE. For an output example, see the link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.14/html/installing/installing-on-ibm-z-and-ibm-linuxone#installation-operators-config_installing-ibm-z[Initial Operator configuration] section in the {ocp-short} documentation.
```
swopebe: coding these output, but other than that, this works I think

jc-berger (approved):
/lgtm

PR #6115, opened 2024-02-07
added subscription status

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-9822

### DIFF
--- a/applications/subscription_report.adoc
+++ b/applications/subscription_report.adoc
@@ -21,6 +21,8 @@ Detailed application status is available on the managed clusters, while the `sub
 
 The package-level managed cluster status is located in `<namespace:<your-appsub-namespace>` on the managed cluster and contains detailed status for all the resources that are deployed by the application. For every `appsub` that is deployed to a managed cluster, there is a `SubscriptionStatus` CR created in the `appsub` namespace on the managed cluster. Every resource is reported with detailed errors if errors exist. 
 
+The package status only indicates the status of an individual package. You can view the overall subscription status by referencing the field, `.status.subscription`. 
+
 See the following `SubscriptionStatus` sample YAML file:
 
 [source,yaml]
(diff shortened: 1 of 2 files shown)


applications/subscription_report.adoc
```diff
@@ -74,6 +74,22 @@ statuses:
     phase: Deployed
 ----
 
+The new field, `.status.subscription`, indicates the overall subscription status. Previously, you only had the package status that indicated the status of an individual package.
```
swopebe: Let's not use "new" in regular tech doc because it creates a maintenance issue. Next release, it is not "new," but where we can use that language is in "What's new" and other release notes because they are recreated each release. 

You can change to something like:

The `.status.subscription` field indicates...

Remove the _previously_ statement for the same reasons; an actual task does not need to reflect the conditions before. That is what we do with release notes.

 Then ensure it goes in to _What's new_, where you _can_ also use the previously statement if you choose.
 
 Yes, you get this stuff in the draft sometimes, but we have to separate "release note" language from procedures that will likely span multiple releases.

swopebe: Everything else looks great.

applications/subscription_report.adoc
```diff
@@ -74,6 +74,22 @@ statuses:
...
+----
+  apiVersion: v1
+    kind: Service
+    lastUpdateTime: "2021-09-13T20:12:34Z"
+    name: redis-master
+    namespace: test-ns-2
+  subscription:
+    lastUpdateTime: "2021-09-13T20:12:34Z"
+    phase: Deployed 
+----
```
swopebe, line 80: This YAML seems off, so seek developer to help if that is the case. Otherwise, looks good. Thank you.

swopebe (approved):

applications/subscription_report.adoc
```diff
@@ -74,6 +74,22 @@ statuses:
...
+See the following `.status.subscription` sample YAML file:
+
+[source,yaml]
+----
+  apiVersion: v1
+    kind: Service
+    lastUpdateTime: "2021-09-13T20:12:34Z"
+    name: redis-master
+    namespace: test-ns-2
+  subscription:
```
philipwu08: Add this "subscription" section to the end of the previous YAML. 

I was trying to show that this section goes after the last deployment "redis-slave"

PR #6125, opened 2024-02-12
Updated namespace metadata in example

ngraham20:
https://issues.redhat.com/browse/ACM-4594

### DIFF
--- a/clusters/install_upgrade/install_connected.adoc
+++ b/clusters/install_upgrade/install_connected.adoc
@@ -162,6 +162,7 @@ apiVersion: operators.coreos.com/v1
 kind: OperatorGroup
 metadata:
   name: <default>
+  namespace: <namespace>
 spec:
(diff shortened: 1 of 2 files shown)


swopebe (approved):

swopebe:
@ngraham20 since this is work that is labeled by a train, it's not really a "bug" -- is this something we need to mention in the What's new as "new function" (that is how we treat Train doc work) or is it an internal fix that we don't need to mention for GA?

ngraham20:
> @ngraham20 since this is work that is labeled by a train, it's not really a "bug" -- is this something we need to mention in the What's new as "new function" (that is how we treat Train doc work) or is it an internal fix that we don't need to mention for GA?

Hey @swopebe The train label just describes when the work is intended to be completed. It's a bug, and not "new function", so however you need to treat that on your end is how we should move forward with this

PR #6131, opened 2024-02-13
ACM-9804 Remove old known issues for search

jlpadilla:
Jira issue: https://issues.redhat.com/browse/ACM-9804

## Changes
Removes 2 troubleshoot topics that are no longer valid for search.
1. Redisgraph was removed in 2.7 so troubleshoot topic is no longer valid.
2. The multicloud-console route was also removed several releases back so this troubleshoot topic is no longer valid.


swopebe:
Thank you so much @jlpadilla for helping us keep the repo updated.

swopebe (approved):

PR #6157, opened 2024-02-21
(2.10) Added Component Table for MultiClusterEngine

dislbenn:
Currently, we don't provide a table of components for MCE that describes the components that are being deployed by our operator; therefore, this PR will add it.


swopebe:
Pause on this.

PR #6158, opened 2024-02-21
(2.9) Added Component Table for MultiClusterEngine 

dislbenn:
Currently, we don't provide a table of components for MCE that describes the components that are being deployed by our operator; therefore, this PR will add it.

### DIFF
--- a/clusters/install_upgrade/adv_config_install.adoc
+++ b/clusters/install_upgrade/adv_config_install.adoc
@@ -1,7 +1,12 @@
 [#advanced-config-engine]
 = Advanced configuration 
 
-The {mce-short} is installed using an operator that deploys all of the required components. The {mce-short} can be further configured during or after installation by adding one or more of the following attributes to the `MultiClusterEngine` custom resource:
+The {mce-short} is installed using an operator that deploys all of the required components. The {mce-short} can be further configured during or after installation. Learn more about the advanced configuration options.
+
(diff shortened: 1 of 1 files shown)


swopebe:
I have to pump the breaks here for a sec.
We took our components a while ago, for good reason:

The teams could not keep up with the tables
The tables were becoming out of date quickly

This was raised on arch calls etc many times last year with debate about adding them back in. Additionally, a change like this needs to have architecture doc review. I have not been pulled in to a discussion about components since last year. I need to chat about this.

clusters/install_upgrade/adv_config_install.adoc
```diff
@@ -1,7 +1,58 @@
...
+kind: MultiClusterEngine
+metadata:
+  name: multiclusterengine
+spec:
+  overrides:
+    components:
+    - name: <name> <1>
+      enabled: true
+----
+1. Replace `name` with the name of the component.
```
oafischer:
```suggestion
<1> Replace `name` with the name of the component.
```

We could also move this step into the prompt since it's the only one

swopebe: yes, I was going to but it's old, so I left it....thanks for doing it for me!

oafischer (approved):

swopebe (approved):

PR #6173, opened 2024-02-27
https://issues.redhat.com/browse/ACM-10095 Operator policy controller docs

oafischer:
2.10 only


governance/policy_controllers_intro.adoc
```diff
@@ -9,5 +9,6 @@ Policy controllers check for policy violations, and can make the cluster status
 * xref:../governance/cert_policy_ctrl.adoc#certificate-policy-controller[Certificate policy controller]
 * xref:../governance/iam_policy_ctrl.adoc#iam-policy-controller[IAM policy controller (Deprecated)]
 * xref:../governance/policy_set_ctrl.adoc#policy-set-controller[Policy set controller]
+* xref:../governance/policy_operator.adoc#policy-operator#policy-operator[Operator policy controller]
```
JustinKuli: I don't know how these details work on the page, but my spidey senses are tingling about the doubled `#policy-operator`. Does the link work as expected? I don't know where to check it.

governance/policy_operator.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+- `channel`
+- `name`
+- `namespace`
+- `source`
+- `sourceNamespace` 
+
+For more information, see the link:https://docs.openshift.com/container-platform/4.12/operators/understanding/olm/olm-understanding-olm.html#olm-subscription_olm-understanding-olm[Subscription] topic in the {ocp} documentation.
+
+You can use the `spec.versions` field in an operator policy to declare which versions of the operator are compliant. If the field is empty, any version running on the cluster is considered compliant. If you specify any of the fields, the version on the managed cluster must match one of the versions in the list for the policy to be compliant. If the policy is in `enforce` mode, it only approves updates to versions that are compliant.
```
JustinKuli:
```suggestion
You can use the `spec.versions` field in an operator policy to declare which versions of the operator are compliant. If the field is empty, any version running on the cluster is considered compliant. If the field is not empty, the version on the managed cluster must match one of the versions in the list for the policy to be compliant. If the policy is in `enforce` mode, it only approves updates to versions that are compliant.
```

Saying "any of the fields" feels too vague, but I'm open to other options.

JustinKuli (changes requested):
Sounds good to me overall! I'm glad that the description & example is sufficient - the tables feel hard to maintain and kind of noisy.

governance/policy_operator.adoc
oafischer: That works too, thanks!

governance/policy_controllers_intro.adoc
oafischer: Good catch, classic copy paste error. Fixed now. Thanks!

JustinKuli (approved):
LGTM, thanks!

governance/policy_operator.adoc
```diff
@@ -0,0 +1,48 @@
+[#policy-operator]
+= Operator policy controller
+
+The operator policy controller allows you to monitor or install Operator Lifecycle Manager (OLM) operators across your clusters. You can monitor the health of various pieces of the operator and you can specify how to automatically handle updates to the operator. You can also distribute an operator policy to managed clusters by using the governance framework and adding the policy to the `policy-templates` field of a policy on the hub cluster.
```
jc-berger: Is it "or" or "and"? That is, do users need to choose which action or can they use both? If the latter, I would change to "and" : 

```suggestion
The operator policy controller allows you to monitor and install Operator Lifecycle Manager (OLM) operators across your clusters. You can monitor the health of various pieces of the operator and you can specify how to automatically handle updates to the operator. You can also distribute an operator policy to managed clusters by using the governance framework and adding the policy to the `policy-templates` field of a policy on the hub cluster.
```

jc-berger: "You can" can be redundant. What do you think of this rewrite that's a little more user action focused too : 

```suggestion
The operator policy controller allows you to monitor or install Operator Lifecycle Manager (OLM) operators across your clusters. Use the operator policy controller to monitor the health of various pieces of the operator and to specify how you want to automatically handle updates to the operator. You can also distribute an operator policy to managed clusters by using the governance framework and adding the policy to the `policy-templates` field of a policy on the hub cluster.
```

jc-berger (approved):
/lgtm

governance/policy_operator.adoc
```diff
@@ -0,0 +1,48 @@
...
+  subscription:
+    channel: stable-3.10
+    name: quay-operator
+    namespace: quay-opns
+    installPlanApproval: Automatic
+    source: redhat-operators
+    sourceNamespace: openshift-marketplace
+  versions:
+    - quay-operator.v3.10.3
+----
```
dockerymick, line 48: I recommend considering the additional resources section, if it is needed

oafischer: Added!

dockerymick: Thanks Oliver! i would also take the user back to the Policy controller intro since they were directed from there.

oafischer: I believe we agreed to not add the _return to_ links in additional resources. But we can bring this up to the team again

dockerymick: My understanding that we agreed to not return to the beginning of the same file. It's fine tho, not a big deal at all. If someone shouts about it we can easily add the link whenever. Thanks! I'm good with your current changes !

oafischer: OK, let's clarify in our channel though, maybe I misunderstood. I'll merge this for now, thanks!

PR #6174, opened 2024-02-27
https://issues.redhat.com/browse/ACM-9761 Add guide to use api-resources

oafischer:
Also for 2.9


dockerymick (approved):
@oafischer this looks good to me, i can see how the sample is confusing tho. Maybe we can consider removing the asterisks that are in the `data:
 AllowedResources` and `DeniedResources` parameter section and replace it with `<list-your-apigroups>`, `<api-group-resources>`, etc.

jlpadilla (approved):
/lgtm

PR #6191, opened 2024-03-04
https://issues.redhat.com/browse/ACM-10154 Known issue ClusterCurator

oafischer:
2.10 only

### DIFF
--- a/clusters/release_notes/known_issues.adoc
+++ b/clusters/release_notes/known_issues.adoc
@@ -517,10 +517,10 @@ The `cluster-proxy-addon` does not work when there is another proxy configuratio
 * There is a proxy configuration in the agent `AddOnDeploymentConfig` resource.
 
 [#ocp-ded-support-curator]
-=== _ClusterCurator_ does not support OpenShift Dedicated clusters
+=== The _ClusterCurator_ does not support {ocp-short} Dedicated clusters
 //2.10:ACM-10154
 
-When you upgrade an OpenShift Dedicated cluster by using the `ClusterCurator`, the upgrade fails because the `ClusterCurator` does not support OpenShift Dedicated clusters.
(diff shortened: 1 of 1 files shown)


fxiang1 (approved):
LGTM! Thanks Oliver!

clusters/release_notes/known_issues.adoc
```diff
@@ -516,6 +516,12 @@ The `cluster-proxy-addon` does not work when there is another proxy configuratio
 
 * There is a proxy configuration in the agent `AddOnDeploymentConfig` resource.
 
+[#ocp-ded-support-curator]
+=== The _ClusterCurator_ does not support {ocp-short} Dedicated clusters
+//2.10:ACM-10154
+
+When you upgrade an {ocp-short} Dedicated cluster by using the `ClusterCurator`, the upgrade fails because the `ClusterCurator` does not support {ocp-short} Dedicated clusters.
```
dockerymick: Does Cluster Curator have to be monospaced? 
```suggestion
When you upgrade an {ocp-short} Dedicated cluster by using the `ClusterCurator` resource, the upgrade fails because the Cluster curator does not support {ocp-short} Dedicated clusters.
```

dockerymick (approved):
Question and suggestion, other than that this looks good to me

clusters/release_notes/known_issues.adoc
oafischer: I believe it is referring to the kind but maybe @fxiang1 can confirm?

fxiang1: @oafischer The first ClusterCurator instance refers to the kind but the second ClusterCurator instance refers to itself and not the kind. I guess in this case we need to separate it if not referring to the kind?

oafischer: Thanks for the clarification, yes if it's not the kind then we don't use monospace. I'll update it.

PR #6205, opened 2024-03-06
https://issues.redhat.com/browse/ACM-10247 InfraEnv no proxy docs

oafischer:
2.10 only

### DIFF
--- a/clusters/cluster_lifecycle/create_cluster_on_prem.adoc
+++ b/clusters/cluster_lifecycle/create_cluster_on_prem.adoc
@@ -325,10 +325,6 @@ spec:
 | Optional
 | Consolidates advanced network configuration such as static IPs, bridges, and bonds for the hosts. The host network configuration is specified in one or more `NMStateConfig` resources with labels you choose. The `nmStateConfigLabelSelector` property is a Kubernetes label selector that matches your chosen labels. The network configuration for all `NMStateConfig` labels that match this label selector is included in the Discovery Image.  When you boot, each host compares each configuration to its network interfaces and applies the appropriate configuration. 
 
-| `demo-nmstate-label`
-| Optional
-| The label name and the label value must match the values in the `label` section of the `NMStateConfig` custom resource that you created in <<on-prem-creating-your-cluster-with-the-cli-nmstateconfig,Optional: Create the NMStateConfig custom resource>>.
-
 | `proxy`
 | Optional
 | You can specify proxy settings required by the host during discovery in the proxy section.
--- a/clusters/cluster_lifecycle/create_cluster_on_prem.adoc
+++ b/clusters/cluster_lifecycle/create_cluster_on_prem.adoc
@@ -321,6 +321,9 @@ spec:
 | Optional
 | You can specify your SSH public key, which enables you to access the host when it is booted from the discovery ISO image. 
 
+| `nmStateConfigLabelSelector`
+| Optional
+| Consolidates advanced network configuration such as static IPs, bridges, and bonds for the hosts. The host network configuration is specified in one or more `NMStateConfig` resources with labels you choose. The `nmStateConfigLabelSelector` property is a Kubernetes label selector that matches your chosen labels. The network configuration for all `NMStateConfig` labels that match this label selector is included in the Discovery Image.  When you boot, each host compares each configuration to its network interfaces and applies the appropriate configuration. 
 
 | `demo-nmstate-label`
 | Optional
@@ -329,10 +332,6 @@ spec:
 | `proxy`
 | Optional
 | You can specify proxy settings required by the host during discovery in the proxy section.
-
-| `nmStateConfigLabelSelector`
-| Optional
-| Consolidates advanced network configuration such as static IPs, bridges, and bonds for the hosts. The host network configuration is specified in one or more `NMStateConfig` resources with labels you choose. The `nmStateConfigLabelSelector` property is a Kubernetes label selector that matches your chosen labels. The network configuration for all `NMStateConfig` labels that match this label selector is included in the Discovery Image.  When you boot, each host compares each configuration to its network interfaces and applies the appropriate configuration. 
 |===
 
 *Note:* When provisioning with IPv6, you cannot define a CIDR address block in the `noProxy` settings. You must define each address separately.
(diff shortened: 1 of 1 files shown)


clusters/cluster_lifecycle/create_cluster_on_prem.adoc
```diff
@@ -298,16 +298,40 @@ spec:
...
+|===
+| Field | Optional or required | Description
+
+| `sshAuthorizedKey`
+| Optional
+| You can specify your SSH public key, which enables you to access the host when it is booted from the discovery ISO image. 
+
+
+| `demo-nmstate-label`
+| Required
```
carbonin: Creating nmstate is optional so I don't feel like this should be required.

carbonin: Maybe this should also describe `nmStateConfigLabelSelector` since that's the infraenv field where `demo-nmstate-label` is an arbitrary label name that can be changed by the user.

carbonin (commented):
Proxy stuff looks good, just some comments about the nmstate selector.

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
oafischer: Updated to optional.

We recently discussed this on the doc team and we want to avoid describing/documenting Kubernetes fields such as `nmStateConfigLabelSelector`. That being said, this is not the first time I've gotten the request to describe this field, so we do actually have it described in the _Creating a host inventory by using the command line interface_ topic. Are we OK with leaving it at that?

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
```diff
@@ -298,16 +298,40 @@ spec:
     namespace: demo-worker4 <2>
   pullSecretRef: 
     name: pull-secret
-    sshAuthorizedKey: <your_public_key_here> <3>
+    sshAuthorizedKey: <your_public_key_here>
```
carbonin: This indentation looks off for some reason. Can you ensure it's not nested under `pullSecretRef`.

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
carbonin: I'm not sure what you mean by "Kubernetes fields".
`nmStateConfigLabelSelector` is a part of the CR spec. It's no different from `sshAuthorizedKey` or `proxy`

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
oafischer: Does it appear correct in the published docs? I haven't changed the indentation so it might just look off.

carbonin: Nope, it also looks wrong in the docs 
image
Can you fix it in this PR or would you rather do that elsewhere since that definitely won't affect only 2.10?

oafischer: I'll fix it here and update it in older versions as well. Thanks for catching that

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
oafischer: I discussed again with our architect and we can add this description. I'll use the same as what I added in the CIM section.

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
oafischer: Updated

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
```diff
@@ -298,16 +298,43 @@ spec:
...
+
+| `sshAuthorizedKey`
+| Optional
+| You can specify your SSH public key, which enables you to access the host when it is booted from the discovery ISO image. 
+
+| `nmStateConfigLabelSelector`
+| Optional
+| Consolidates advanced network configuration such as static IPs, bridges, and bonds for the hosts. The host network configuration is specified in one or more `NMStateConfig` resources with labels you choose. The `nmStateConfigLabelSelector` property is a Kubernetes label selector that matches your chosen labels. The network configuration for all `NMStateConfig` labels that match this label selector is included in the Discovery Image.  When you boot, each host compares each configuration to its network interfaces and applies the appropriate configuration. 
+
+| `demo-nmstate-label`
```
carbonin: Do you want to also keep this?

oafischer: The description? I assumed that's what you were referring to in a previous comment but if it's not needed I can remove it.

carbonin: If we are explaining the use of `nmStateConfigLabelSelector` do you think we still need to explain this?

I'm okay with it either way, but I had thought that the `nmStateConfigLabelSelector` would replace this.

oafischer: Oh I see what you mean now. Always better to keep it simple. So I'll remove the `demo-nmstate-label` entry, and leave the `nmStateConfigLabelSelector` entry in the table, correct?

carbonin: Yes, I think that would be most clear.

oafischer: Done, sorry for misunderstanding at first.

carbonin (approved):

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
```diff
@@ -345,4 +368,4 @@ You can confirm that it was approved when the output of the previous command inc
 
 - See xref:../cluster_lifecycle/release_images.adoc#release-images-intro[Release images]
```
dockerymick: broken link according to the tool
Screenshot 2024-03-07 at 10 12 40 AM

dockerymick (approved):
Left a comment related to a link. Otherwise, this looks good!

PR #6229, opened 2024-03-13
https://issues.redhat.com/browse/ACM-10223--removing content

swopebe:
Not all of the {mce-short} components have infrastructure node support, so some worker nodes are still required when installing {mce-short} on infrastructure nodes.

REMOVED


jc-berger (approved):
/lgtm

PR #6264, opened 2024-03-19
Add note about eventing change (GRC)

dhaiducek:
Adding to "What's new" because the Bug issue is internal and won't be contained in the Errata.

ref: https://issues.redhat.com/browse/ACM-10527

### DIFF
--- a/governance/grc_arch.adoc
+++ b/governance/grc_arch.adoc
@@ -15,7 +15,7 @@ View the following component descriptions for the governance architecture diagra
 +
 * Changes to policies in the watched cluster namespace on the managed cluster.
 * Generated events on policies in the watched cluster namespace on the managed cluster.
-* During each reconcile, the controller creates and updates the policy status for the hub cluster and managed cluster in the cluster namespace, based on the policy changes or generated events.
+* During each reconcile, the controller creates and updates the policy status for the hub cluster and managed cluster in the cluster namespace, based on the policy changes or generated events. The status only contains updates that are relevant to the current policy and does not consider past statuses if the policy was deleted and recreated.
 
 - *Template sync controller:* Runs on your managed clusters and watches for template function changes for policies in the cluster namespace on the managed cluster. The changes start the reconcile process, and for each reconcile, it creates,updates, and deletes objects defined in the `spec.policy-templates` of the policies.
 


release_notes/whats_new.adoc
```diff
@@ -70,6 +70,8 @@ See link:../observability/observe_environments_intro.adoc#observing-environments
 
 * Enable the `auditEventsInvolvedNamespace` to manage which namespace audit event to create, and the `admissionEventsInvolvedNamespace` to to manage which namespace admission event to create. See link:../governance/create_gatekeeper.adoc#managing-gatekeeper-operator-policies[Managing Gatekeeper operator policies].
 
+* Previously if a policy were deleted and recreated with the same name, it would populate its status with the events from the deleted policy. Now, the status will only contain updates relevant to the current policy since its creation and won't consider the old events.
```
dockerymick: I recommend we take another look at the known issues. Does this currently exist in the doc? If yes, we should only remove the known issue. We do not announce fixes unless it's a part of the errata. Alternatively, we can add a note in the Governance doc that states: "The status only contain updates relevant to the current policy."

dockerymick (changes requested):

release_notes/whats_new.adoc
dhaiducek: It does not exist in the doc and wasn't listed as a known issue. But it also arose from an internal Errata, so it won't be visible to customers. I'll look for another spot we might put it.

dockerymick: @dhaiducek what are your thoughts about adding more details in the ["Status sync controller" description](https://github.com/stolostron/rhacm-docs/blob/2.10_stage/governance/grc_arch.adoc), specifically the third bullet point? 

"During each reconcile, the controller creates and updates the policy status for the hub cluster and managed cluster in the cluster namespace, based on the policy changes or generated events. The status only contain updates that are relevant to the policy."

dhaiducek: @dockerymick Thanks for the review--I've moved the note and tweaked your wording a little bit. Please take another look when you have a moment.

dockerymick (approved):
Looks good to me! @dhaiducek thanks for accepting my suggestion!

governance/grc_arch.adoc
```diff
@@ -15,7 +15,7 @@ View the following component descriptions for the governance architecture diagra
 +
 * Changes to policies in the watched cluster namespace on the managed cluster.
 * Generated events on policies in the watched cluster namespace on the managed cluster.
-* During each reconcile, the controller creates and updates the policy status for the hub cluster and managed cluster in the cluster namespace, based on the policy changes or generated events.
+* During each reconcile, the controller creates and updates the policy status for the hub cluster and managed cluster in the cluster namespace, based on the policy changes or generated events. The status only contains updates that are relevant to the current policy and does not consider past statuses if the policy was deleted and then recreated.
```
cadawson:
```suggestion
* During each reconcile, the controller creates and updates the policy status for the hub cluster and managed cluster in the cluster namespace, based on the policy changes or generated events. The status only contains updates that are relevant to the current policy and does not consider past statuses if the policy was deleted and recreated.
```

Interpretation of what I think I am reading to make sure: It will show you a status that only applies to the current policy if the policy was deleted and recreated. It doesn't retain the history of the past statuses from before it was blown away and recreated. Is that it? If so, looks good!

dhaiducek: Yes, that's correct @cadawson

cadawson (approved):
I think this looks good. Thanks. I included a suggestion and a clarification of what I think this means. If I understood it right, then you are good to go. Thanks!

jc-berger (approved):
/lgtm

dockerymick:
@cadawson @jc-berger thanks for the reviews and approvals! @dhaiducek thanks for accepting my suggestion and creating the PR! Merging this PR

PR #6350, opened 2024-04-15
https://issues.redhat.com/browse/ACM-10953--adding known issue

swopebe:


release_notes/known_issues_install.adoc
```diff
@@ -21,12 +21,20 @@ Known issue process and when to write:
...
 
 For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/release_notes/ocp-4-12-release-notes#ocp-4-12-known-issues[{ocp-short} known issues]. 
 
 For more about deprecations and removals, see xref:../release_notes/deprecate_remove.adoc#deprecations-removals[Deprecations and removals].
 
+[#uninstall-manual-remove-cr]
+== 
+//2.10:10953
+
+When you uninstall {product-title-short} from {ocp-short}, but then install that same version of {product-title-short} and upgrade the version, the upgrade might fail because the `StorageVersionMigration` custom resources were not removed.
```
xuezhaojun: Not the "same version of ACM" but any version before that version.

For example, when QE test upgrade path, they follow the path of 2.8 -> 2.9 -> 2.10.  

And then, QE uninstall 2.10 from OCP, and want to test upgrade path 2.8->2.9->2.10 again. This time, the upgrade will fail because of the remained `StorageVersionMigration` CRs make the migration controller think the migration is already done.

release_notes/known_issues_install.adoc
```diff
@@ -21,12 +21,22 @@ Known issue process and when to write:
 Or consider a troubleshooting topic.
 ////
 
-Review the known issues for installation. The following list contains known issues for this release, or known issues that continued from the previous release. 
+Review the known issues for installation and upgrading. The following list contains known issues for this release, or known issues that continued from the previous release.
```
jc-berger: For consistency, should it be: 
```suggestion
Review the known issues for installing and upgrading. The following list contains known issues for this release, or known issues that continued from the previous release. 
```

jc-berger: or:
```suggestion
Review the known issues for installation and upgrade. The following list contains known issues for this release, or known issues that continued from the previous release. 
```

release_notes/known_issues_install.adoc
```diff
@@ -21,12 +21,22 @@ Known issue process and when to write:
...
 
 For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/release_notes/ocp-4-12-release-notes#ocp-4-12-known-issues[{ocp-short} known issues]. 
 
 For more about deprecations and removals, see xref:../release_notes/deprecate_remove.adoc#deprecations-removals[Deprecations and removals].
 
+[#uninstall-manual-remove-cr]
+== 
+//2.10:10953
+
+When you uninstall {product-title-short} from {ocp-short}, but then install an earlier version of {product-title-short} and upgrade the version, the upgrade might fail because the `StorageVersionMigration` custom resources were not removed.
```
jc-berger: I don't think it's too confusing (like you said was a concern). But perhaps breaking it up into two sentences might prevent potential confusion. What do you think of the following?:

```suggestion
When you uninstall {product-title-short} from {ocp-short}, but then install an earlier version of {product-title-short} and upgrade the version, the upgrade might fail. The upgrade fails because the `StorageVersionMigration` custom resources were not removed.
```

jc-berger: Also, although I think the original sentence if fine, if you concerned about possible confusion, perhaps omitting the first part would make it clearer. What do you think of this revision?:

```suggestion
If you install an earlier version of {product-title-short} on {ocp-short}, then upgrade the version, the upgrade might fail because the `StorageVersionMigration` custom resources were not removed.
```

jc-berger (approved):
/lgtm

release_notes/known_issues_install.adoc
swopebe: This change is not related to the uninstall>upgrade scenario-- not sure it applies, but please check for understanding on the original line 34.

swopebe: Changing to upgrade fails because of the CRs exist, the upgrade fails.

release_notes/known_issues_install.adoc
```diff
@@ -21,12 +21,22 @@ Known issue process and when to write:
...
 
 For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/release_notes/ocp-4-12-release-notes#ocp-4-12-known-issues[{ocp-short} known issues]. 
 
 For more about deprecations and removals, see xref:../release_notes/deprecate_remove.adoc#deprecations-removals[Deprecations and removals].
 
+[#uninstall-manual-remove-cr]
+== 
+//2.10:10953
+
+When you uninstall {product-title-short} from {ocp-short}, but then install an earlier version of {product-title-short} and upgrade the version, the upgrade might fail. The upgrade fails because the `StorageVersionMigration` custom resources were not removed.
```
jc-berger:
```suggestion
Uninstalling {product-title-short} from {ocp-short} can cause issues if you later want to install earlier versions and then upgrade. 

For instance, when you uninstall {product-title-short} from {ocp-short}, then install an earlier version of {product-title-short} and upgrade that version, the upgrade might fail. The upgrade fails because the `StorageVersionMigration` custom resources were not removed.
```

jc-berger (approved):
/lgtm

release_notes/known_issues_install.adoc
jc-berger: Thanks for clarifying! I thought the issue was by installing an older version then upgrading it. I didn't know the issue was hinged upon having uninstalled the current version. 

Perhaps just a simple intro sentence specifying that uninstalling can cause issues would help? See my suggestion in my follow up review and let me know if you think that helps :)

swopebe: Good idea @jc-berger -- it is hard to follow, I agree.

PR #6374, opened 2024-04-29
ACM#9532: Document hosting cluster topologies

xenolinux:
https://issues.redhat.com/browse/ACM-9532


clusters/hosted_control_planes/mce_hc.adoc
```diff
@@ -0,0 +1,45 @@
+[#mce-hosting-cluster]
+= Using the {mce-short} hub cluster as a hosting cluster
+
+By default, the {mce-short} hub cluster is configured as the only hosting cluster for all hosted control planes.
+
+After installing {mce-short} and creating the `MultiClusterEngine` instance, {product-title-short} enables the following components automatically:
+
```
swopebe, line 7: @xenolinux and @rokej I have an issue for this as well for the integration doc. I did not see this line but we should not be discussing ACM in this section of the documentation. I should have peer reviewed this and caught it. IF ACM is a prereq for this information, let's discuss if this is the best place for this.

PR #6396, opened 2024-05-03
https://issues.redhat.com/browse/ACM-11433 Remove unused html files

oafischer:
We can probably backport this too


dockerymick (approved):
Approving, let's run a build immediately after to verify unless you were able to check the build with your local branch.

oafischer:
I ran a build before (should still be the one that is up right now) and everything is green.

dhaiducek (approved):
I was hoping I could get credit for the PR with 32,000 line changes, but I guess I'll approve it. 😉 

Thanks for considering, investigating, and updating it, @dockerymick and @oafischer! 🥳

PR #6418, opened 2024-05-10
added workaround section for troubleshooting

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-11539


add-ons/configure_nodeselector_tolerations_addons.adoc
```diff
@@ -98,3 +98,59 @@ oc -n <managed-cluster> patch managedclusteraddons <addon-name> --type='json' -p
 * Replace `config-name` with the name of the `AddonDeploymentConfig` that you just created
 
 The new configuration that you referenced in the add-on `ManagedClusterAddon` overrides the global default configuration that you previously defined in the `ClusterManagementAddon` add-on.
+
+After you configure the add-on tolerations, then the cluster-proxy add-ons should run on your managed clusters and use the new add-on tolerations. If the cluster-proxy add-ons do not use your add-on tolerations, or specifically, they do not run in the infra node, then you receive errors. To workaround these errors, complete the following steps: 
+
+. To ensure your configuration for the add-on tolerations get pushed to your managed cluster, create a  `cluster-proxy add-on supportconfig` by using the following template:
```
hchenxa: what does the `cluster-proxy addon supportconfig` mean here?

jc-berger: @FDewaleyne I believe this is the type of config that users need to create, yes? 

If I'm incorrect, I'll make any changes you recommend.

FDewaleyne: this is a config that should be there by default but could be abscent for unknown reasons.
It does not cause errors, but instead part of the cluster-proxy addon will ignore the change in toleration and keep running on the compute nodes. no error.

add-ons/configure_nodeselector_tolerations_addons.adoc
```diff
@@ -98,3 +98,59 @@ oc -n <managed-cluster> patch managedclusteraddons <addon-name> --type='json' -p
 * Replace `config-name` with the name of the `AddonDeploymentConfig` that you just created
 
 The new configuration that you referenced in the add-on `ManagedClusterAddon` overrides the global default configuration that you previously defined in the `ClusterManagementAddon` add-on.
+
+After you configure the add-on tolerations, then the cluster-proxy add-ons should run on your managed clusters and use the new add-on tolerations. If the cluster-proxy add-ons do not use your add-on tolerations, then they run on a compute node. Running on a compute node could affects costs and add loads on the infra and master nodes, making these nodes less charged. To prevent the add-ons from running on compute nodes, complete the following steps: 
+
+. To ensure your configuration for the add-on tolerations get pushed to your managed cluster, create a  `cluster-proxy add-on supportconfig` by using the following template: 
++
+[source,yaml]
```
swopebe: formatting issue here.

add-ons/configure_nodeselector_tolerations_addons.adoc
swopebe: Formatting issues are still there in this PR, but also since this appears to be only related to a certain use-case, we should treat it as _troubleshooting._ The main doc is for all users and t-shooting is for these types of cases.

swopebe (changes requested):
I think we need to look at this as troubleshooting.

add-ons/configure_nodeselector_tolerations_addons.adoc
swopebe: The way this is written, it is a workaround that goes into troubleshooting, that is how the official doc would categorize workarounds.

That said, if this needs to be part of the process, then we build it in to the task, not as a workaround, but a required step.

troubleshooting/trouble_cluster_proxy_addon_ignoring_addon_toleration.adoc
```diff
@@ -0,0 +1,57 @@
+[#troubleshooting-cluster-proxy-addon-ignoring-addon-toleration]
+= Troubleshooting the cluster proxy add-on ignoring the add-on toleration
+
+After you configure the add-on toleration, then by default, the cluster-proxy add-ons should run on your managed clusters and use the new add-on toleration. For unknown reasons, the cluster-proxy add-ons might not use your add-on toleration. In this case, they ignore the toleration changes and run on a compute node. Running on a compute node could affect costs and add loads on the infra and master nodes, making these nodes less charged. If your cluster-proxy add-ons do not automatically use your add-on toleration, complete the following steps: 
+
+. To ensure your configuration for the add-on toleration gets pushed to your managed cluster, create a  `cluster-proxy add-on supportconfig` by using the following template: 
++
```
swopebe: please create a space between paragraph and file sample to fix the formatting. If the code is not grayed out, that shows there is a formatting issue.

troubleshooting/trouble_cluster_proxy_addon_ignoring_addon_toleration.adoc
```diff
@@ -0,0 +1,57 @@
+[#troubleshooting-cluster-proxy-addon-ignoring-addon-toleration]
+= Troubleshooting the cluster proxy add-on ignoring the add-on toleration
+
+After you configure the add-on toleration, then by default, the cluster-proxy add-ons should run on your managed clusters and use the new add-on toleration. For unknown reasons, the cluster-proxy add-ons might not use your add-on toleration. In this case, they ignore the toleration changes and run on a compute node. Running on a compute node could affect costs and add loads on the infra and master nodes, making these nodes less charged. If your cluster-proxy add-ons do not automatically use your add-on toleration, complete the following steps:
```
swopebe: This is not necessarily a symptom, see how the other files hit the user with a symptom right away. SO read through this and sift through the symptom, since we are not giving a description like a main topic. Start with the symptom.

troubleshooting/trouble_cluster_proxy_addon_ignoring_addon_toleration.adoc
```diff
@@ -0,0 +1,57 @@
...
+  - defaultConfig:
+      name: klusterlet-addon-deploy-config
+      namespace: open-cluster-management-hub
+    group: addon.open-cluster-management.io
+    resource: addondeploymentconfigs
+----
+
+
+. To ensure the pods, such as the cluster-proxy-proxy-agent, on your managed cluster gets configured with the add-on toleration, verify that your `ClusterManagementAddon` looks like the following sample: 
++
```
swopebe: Please do not use: looks like

swopebe: Same format issue here.

swopebe (changes requested):
1. Make clear the symptom and get developers to review it. See the other files to see what I mean.
2. Formatting issue.
3. Rewrite "looks like"

Other than this, it looks good.

troubleshooting/trouble_cluster_proxy_addon_ignoring_addon_toleration.adoc
swopebe: Look at other topics for how we introduce samples.

troubleshooting/trouble_cluster_proxy_addon_ignoring_addon_toleration.adoc
swopebe: So the symptom is something that the user sees right away, like a fail status or increased loads. We need a symptom. 

This is the result of the problem: affecting costs, adding loads on the infra and master nodes, and making these nodes less charged.

How does the user known there is a problem, do they have some kind of alert and error message. That is what we mean by symptom. I will approve but we need a clear symptom, then you can let that other content after the symptom.

swopebe (approved):

troubleshooting/trouble_cluster_proxy_addon_ignoring_addon_toleration.adoc
```diff
@@ -0,0 +1,59 @@
+[#troubleshooting-cluster-proxy-addon-ignoring-addon-toleration]
+= Troubleshooting the cluster proxy add-on ignoring the add-on toleration
+
+For unknown reasons, the cluster-proxy add-ons might not automatically use your add-on toleration. In this case, they ignore the toleration changes and run on a compute node. Running on a compute node can result in the following symptoms: affecting costs, adding loads on the infra and master nodes, and making these nodes less charged. If your cluster-proxy add-ons do not automatically use your add-on toleration, complete the following steps: 
+
+. To ensure your configuration for the add-on toleration gets pushed to your managed cluster, create a  `cluster-proxy add-on supportconfig` by using the following template: 
+
++
+[source,bash]
```
oafischer: This doesn't look like a terminal command, not sure if bash is the right choice

troubleshooting/trouble_cluster_proxy_addon_ignoring_addon_toleration.adoc
swopebe: Need to add headers here, I didn't see that until now.

swopebe: Also, please cut down this filename, also just noticed that. We have shown examples in meetings about how we need to reduce filenames, this one is too long. We don't need the title in the filename, look at other topics for examples.

troubleshooting/trouble_cluster_proxy_addon_ignoring_addon_toleration.adoc
```diff
@@ -0,0 +1,58 @@
+[#troubleshooting-cluster-proxy-addon-ignoring-addon-toleration]
```
swopebe: Please cut down this anchor. I just posted an example of reducing anchor words on Slack for an example, and we have discussed this in training. I will share the checklist that we came up with that may help.

swopebe (changes requested):

swopebe (changes requested):
I thought I submitted this is a past review and on Slack but maybe not. Please change filenames and anchors to more concise and usable lengths.

add-ons/configure_nodeselector_tolerations_addons.adoc
```diff
@@ -97,4 +97,4 @@ oc -n <managed-cluster> patch managedclusteraddons <addon-name> --type='json' -p
 * Replace `config-namespace` with the namespace of the `AddonDeploymentConfig` that you just created
 * Replace `config-name` with the name of the `AddonDeploymentConfig` that you just created
 
-The new configuration that you referenced in the add-on `ManagedClusterAddon` overrides the global default configuration that you previously defined in the `ClusterManagementAddon` add-on.
+The new configuration that you referenced in the add-on `ManagedClusterAddon` overrides the global default configuration that you previously defined in the `ClusterManagementAddon` add-on.
```
swopebe, line 100: this is OK.

troubleshooting/trouble_cluster_ignoring_addon_toleration.adoc
```diff
@@ -0,0 +1,58 @@
...
+  - group: proxy.open-cluster-management.io
+    resource: managedproxyconfigurations
+    defaultConfig:
+      name: cluster-proxy
+  - defaultConfig:
+      name: klusterlet-addon-deploy-config
+      namespace: open-cluster-management-hub
+    group: addon.open-cluster-management.io
+    resource: addondeploymentconfigs
+----
```
swopebe: content looks good

troubleshooting/troubleshooting_intro.adoc
```diff
@@ -44,6 +44,7 @@ To view the main documentation about managing your clusters, see link:../cluster
 - xref:../troubleshooting/trouble_cluster_remove_namespace.adoc#trouble-cluster-remove-namespace[Namespace remains after deleting a cluster]
 - xref:../troubleshooting/trouble_auto_import_secret_exists.adoc#trouble-auto-import-secret-exists[Auto-import-secret-exists error when importing a cluster]
 - xref:../troubleshooting/trouble_cinder_csi_driver_volsync.adoc#troubleshooting-the-cinder-csi-driver-for-volsync[Troubleshooting the cinder Container Storage Interface (CSI) driver for VolSync]
+- xref:../troubleshooting/trouble_cluster_ignoring_addon_toleration.adoc#troubleshooting-cluster-ignoring-addon-toleration[Troubleshooting the cluster proxy add-on ignoring the add-on toleration]
```
swopebe: anchor and filename should be more concise

swopebe (approved):

PR #6424, opened 2024-05-13
[2.11_stage] ACM#11578: Update the services exposed on management control plane

xenolinux:
https://issues.redhat.com/browse/ACM-11578

### DIFF
--- a/clusters/hosted_control_planes/bm_intro.adoc
+++ b/clusters/hosted_control_planes/bm_intro.adoc
@@ -53,26 +53,42 @@ oc get managedclusters local-cluster
 * You need to xref:../hosted_control_planes/install_hcp_cli.adoc#hosted-install-cli[install the hosted control plane command line interface].
 
 [#firewall-port-reqs-bare-metal]
-== Bare metal firewall and port requirements
+== Bare metal firewall, port, and service requirements
 
-Ensure that you meet the firewall and port requirements so that ports can communicate between the management cluster, the control plane, and hosted clusters:
+You must meet the firewall, port, and service requirements so that ports can communicate between the management cluster, the control plane, and hosted clusters.
 
-* The `kube-apiserver` service runs on port 6443 by default and requires ingress access for communication between the control plane components.
+*Note:* Services run on their default ports. However, if you use the `NodePort` publishing strategy, services run on the port that is assigned by the `NodePort` service.
 
-** If you use the `NodePort` publishing strategy, ensure that the node port that is assigned to the `kube-apiserver` service is exposed.
+Use firewall rules, security groups, or other access controls to restrict access to only required sources. Avoid exposing ports publicly unless necessary. For production deployments, use a load balancer to simplify access through a single IP address.
+
+A hosted control plane exposes the following services on bare metal:
+
+* `APIServer`
+
+** The `APIServer` service runs on port 6443 by default and requires ingress access for communication between the control plane components.
 ** If you use MetalLB load balancing, allow ingress access to the IP range that is used for load balancer IP addresses.
 
-* If you use the `NodePort` publishing strategy, use a firewall rule for the `ignition-server` and `Oauth-server` settings.
+* `OAuthServer`
 
-* The `konnectivity` agent, which establishes a reverse tunnel to allow bi-directional communication on the hosted cluster, requires egress access to the cluster API server address on port 6443. With that egress access, the agent can reach the `kube-apiserver` service.
+** The `OAuthServer` service runs on port 443 by default when you use the route and ingress to expose the service.
+** If you use the `NodePort` publishing strategy, use a firewall rule for the `OAuthServer` service.
 
+* `Konnectivity`
+
+** The `Konnectivity` service runs on port 443 by default when you use the route and ingress to expose the service.
+** The `Konnectivity` agent, which establishes a reverse tunnel to allow bi-directional communication on the hosted cluster, requires egress access to the cluster API server address on port 6443. With that egress access, the agent can reach the `APIServer` service.
 ** If the cluster API server address is an internal IP address, allow access from the workload subnets to the IP address on port 6443.
 ** If the address is an external IP address, allow egress on port 6443 to that external IP address from the nodes.
 
-* If you change the default port of 6443, adjust the rules to reflect that change.
-* Ensure that you open any ports that are required by the workloads that run in the clusters.
-* Use firewall rules, security groups, or other access controls to restrict access to only required sources. Avoid exposing ports publicly unless necessary.
-* For production deployments, use a load balancer to simplify access through a single IP address.
+* `Ignition`
+
+** The `Ignition` service runs on port 443 by default when you use the route and ingress to expose the service.
+** If you use the `NodePort` publishing strategy, use a firewall rule for the `Ignition` service.
+
+You do not need the following services on bare metal:
+
+* `OVNSbDb`
+* `OIDC`
 
 [#infrastructure-reqs-bare-metal]
 == Bare metal infrastructure requirements
@@ -95,4 +111,4 @@ For additional resources about hosted control planes on bare metal, see the foll
 
 * To deploy the SR-IOV Operator, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.14/html/networking/hardware-networks#sriov-operator-hosted-control-planes_configuring-sriov-operator[Deploying the SR-IOV Operator for hosted control planes].
 
-* If you want to disable the automatic import feature, see xref:../hosted_control_planes/hosted_disable_auto_import.adoc#hosted-disable-auto-import[Disabling the automatic import of hosted clusters into {mce-short}].
+* If you want to disable the automatic import feature, see xref:../hosted_control_planes/disable_auto_import.adoc#hosted-disable-auto-import[Disabling the automatic import of hosted clusters into {mce-short}].
--- a/clusters/hosted_control_planes/enable_ext_dns_aws.adoc
+++ b/clusters/hosted_control_planes/enable_ext_dns_aws.adoc
@@ -1,15 +1,15 @@
 [#hosted-enable-ext-dns-aws]
 = Enabling external DNS
 
-Because the control plane and the data plane are separate in hosted control planes, you can configure DNS in two independent areas:
+The control plane and the data plane are separate in hosted control planes. You can configure DNS in two independent areas:
 
 - Ingress for workloads within the hosted cluster, such as the following domain: `*.apps.service-consumer-domain.com`
 
 - Ingress for service endpoints within the management cluster, such as API or OAUTH endpoints through the service provider domain: `*.service-provider-domain.com`
 
-The input for the `hostedCluster.spec.dns` dictates the Ingress for workloads within the hosted cluster. The input for `hostedCluster.spec.services.servicePublishingStrategy.route.hostname` dictates the Ingress for service endpoints within the management cluster.
+The input for the `hostedCluster.spec.dns` manages the ingress for workloads within the hosted cluster. The input for `hostedCluster.spec.services.servicePublishingStrategy.route.hostname` manages the ingress for service endpoints within the management cluster.
 
-External DNS creates name records for hosted cluster `Services` that specify a publishing type of `LoadBalancer` or `Route` and provide a hostname for that publishing type. For hosted clusters with `Private` or `PublicAndPrivate` endpoint access types, only the `APIServer` and `OAuth` services support hostnames. For `Private` hosted clusters, the DNS record resolves to a private IP of a Virtual Private Cloud (VPC) Endpoint in your VPC.
+External DNS creates name records for hosted cluster `Services` that specify a publishing type of `LoadBalancer` or `Route` and provide a hostname for that publishing type. For hosted clusters with `Private` or `PublicAndPrivate` endpoint access types, only the `APIServer` and `OAuth` services support hostnames. For `Private` hosted clusters, the DNS record resolves to a private IP address of a Virtual Private Cloud (VPC) endpoint in your VPC.
 
 A hosted control plane exposes the following services:
 
@@ -20,30 +20,34 @@ A hosted control plane exposes the following services:
 * `OVNSbDb`
 * `OIDC`
 
-Each of those services is exposed by using `servicePublishingStrategy` in the `HostedCluster` specification. By default, for the `LoadBalancer` and `Route` types of `servicePublishingStrategy`, you publish the service in one of two ways:
+You can expose these services by using the `servicePublishingStrategy` field in the `HostedCluster` specification. By default, for the `LoadBalancer` and `Route` types of `servicePublishingStrategy`, you can publish the service in one of the following ways:
 
 * By using the hostname of the load balancer that is in the status of the `Service` with the `LoadBalancer` type
-* In the `status.host` field of the `Route`
+* By using the `status.host` field of the `Route` resource
 
-However, when you deploy hosted control planes in a managed service context, those methods can expose the Ingress subdomain of the underlying management cluster and limit options for the management cluster lifecycle and disaster recovery.
+However, when you deploy hosted control planes in a managed service context, those methods can expose the ingress subdomain of the underlying management cluster and limit options for the management cluster lifecycle and disaster recovery.
 
-When a DNS indirection is layered on the `LoadBalancer` and `Route` publishing types, a managed service operator can publish all public hosted cluster services by using a service-level domain. This architecture allows remapping on the DNS name to a new `LoadBalancer` or `Route` and does not expose the Ingress domain of the management cluster. Hosted control planes uses external DNS to achieve that indirection layer.
+When a DNS indirection is layered on the `LoadBalancer` and `Route` publishing types, a managed service operator can publish all public hosted cluster services by using a service-level domain. This architecture allows remapping on the DNS name to a new `LoadBalancer` or `Route` and does not expose the ingress domain of the management cluster. Hosted control planes uses external DNS to achieve that indirection layer.
 
-You can deploy `external-dns` alongside the `hypershift` Operator in the `hypershift` namespace of the management cluster. The external DNS watches for `Services` or `Routes` that have the `external-dns.alpha.kubernetes.io/hostname` annotation. That annotation is used to create a DNS record that points to the `Service`, such as a record, or the `Route`, such as a CNAME record.
+You can deploy `external-dns` alongside the HyperShift Operator in the `hypershift` namespace of the management cluster. External DNS watches for `Services` or `Routes` that have the `external-dns.alpha.kubernetes.io/hostname` annotation. That annotation is used to create a DNS record that points to the `Service`, such as a record, or the `Route`, such as a CNAME record.
+
+You can use external DNS on cloud environments only. For the other environments, you need to manually configure DNS and services.
+
+For more information about external DNS, see link:https://github.com/kubernetes-sigs/external-dns[external DNS].
 
 [#external-dns-prereqs-aws]
 == Prerequisites
 
 Before you can set up external DNS for hosted control planes, you must meet the following prerequisites:
 
-* An external public domain that you can point to
+* You created an external public domain
 
-* Access to the AWS Route53 Management console
+* You have access to the AWS Route53 Management console
 
 [#set-up-external-dns-aws]
 == Setting up external DNS for hosted control planes
 
-If you plan to provision hosted control plane clusters with service-level DNS (external DNS), complete the following steps:
+To provision hosted control plane clusters with service-level DNS (external DNS), complete the following steps:
 
 . Create an AWS credential secret for the HyperShift Operator and name it `hypershift-operator-external-dns-credentials` in the `local-cluster` namespace.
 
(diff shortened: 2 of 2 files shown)


clusters/hosted_control_planes/enable_ext_dns_aws.adoc
```diff
@@ -1,47 +1,49 @@
...
+When a DNS indirection is layered on the `LoadBalancer` and `Route` publishing types, a managed service operator can publish all public hosted cluster services by using a service-level domain. This architecture allows remapping on the DNS name to a new `LoadBalancer` or `Route` and does not expose the ingress domain of the management cluster. Hosted control planes uses external DNS to achieve that indirection layer.
 
-You can deploy `external-dns` alongside the `hypershift` Operator in the `hypershift` namespace of the management cluster. The external DNS watches for `Services` or `Routes` that have the `external-dns.alpha.kubernetes.io/hostname` annotation. That annotation is used to create a DNS record that points to the `Service`, such as a record, or the `Route`, such as a CNAME record.
+You can deploy `external-dns` alongside the HyperShift Operator in the `hypershift` namespace of the management cluster. External DNS watches for `Services` or `Routes` that have the `external-dns.alpha.kubernetes.io/hostname` annotation. That annotation is used to create a DNS record that points to the `Service`, such as a record, or the `Route`, such as a CNAME record.
```
jparrill, line 32: I would mention that this External DNS is only suitable on cloud environments, in other cases you will need to handle the DNS manually and add the proper entries for the relevant services.

jparrill (commented):
small nit, otherwise LGTM

clusters/hosted_control_planes/bm_intro.adoc
```diff
@@ -53,26 +53,43 @@ oc get managedclusters local-cluster
...
+== Bare metal firewall, port, and service requirements
 
-Ensure that you meet the firewall and port requirements so that ports can communicate between the management cluster, the control plane, and hosted clusters:
+You must meet the firewall, port, and service requirements so that ports can communicate between the management cluster, the control plane, and hosted clusters.
 
-* The `kube-apiserver` service runs on port 6443 by default and requires ingress access for communication between the control plane components.
+*Note:* Services run on their default ports. However, if you use the `NodePort` publishing strategy, services run on the port that is assigned by the `NodePort` service. Ensure that the node port that is assigned to the service is exposed.
 
-** If you use the `NodePort` publishing strategy, ensure that the node port that is assigned to the `kube-apiserver` service is exposed.
+* Use firewall rules, security groups, or other access controls to restrict access to only required sources. Avoid exposing ports publicly unless necessary.
```
lahinson:
```suggestion
Use firewall rules, security groups, or other access controls to restrict access to only required sources. Avoid exposing ports publicly unless necessary.
```
Because line 58 was rewritten to no longer function as a lead-in sentence to a list, lines 62 and 63 can stand alone as sentences rather than be formatted as list items.

clusters/hosted_control_planes/bm_intro.adoc
```diff
@@ -53,26 +53,43 @@ oc get managedclusters local-cluster
...
+* For production deployments, use a load balancer to simplify access through a single IP address.
```
lahinson:
```suggestion
For production deployments, use a load balancer to simplify access through a single IP address.
```

clusters/hosted_control_planes/bm_intro.adoc
```diff
@@ -53,26 +53,43 @@ oc get managedclusters local-cluster
...
+
+* `APIServer`
+
+** The `APIServer` service runs on port 6443 by default and requires ingress access for communication between the control plane components.
 ** If you use MetalLB load balancing, allow ingress access to the IP range that is used for load balancer IP addresses.
 
-* If you use the `NodePort` publishing strategy, use a firewall rule for the `ignition-server` and `Oauth-server` settings.
+* `OAuthServer`
+
+** The `OAuthServer` service runs on port 443 by default when exposed by using the route and ingress.
```
lahinson: Line 74 is a bit unclear. Does it mean that the service is exposed by using the route and ingress, or does it mean that the service runs on port 443 by using the route and ingress? cc @jparrill

clusters/hosted_control_planes/bm_intro.adoc
```diff
@@ -53,26 +53,43 @@ oc get managedclusters local-cluster
...
+** If you use the `NodePort` publishing strategy, use a firewall rule for the `OAuthServer` service.
 
-* The `konnectivity` agent, which establishes a reverse tunnel to allow bi-directional communication on the hosted cluster, requires egress access to the cluster API server address on port 6443. With that egress access, the agent can reach the `kube-apiserver` service.
+* `Konnectivity`
 
+** The `Konnectivity` service runs on port 443 by default when exposed by using the route and ingress.
```
lahinson: Similar question here as I have on line 74. Does this sentence mean that the service is exposed by using the route and ingress, or does it mean that the service runs on port 443 by using the route and ingress? cc @jparrill

clusters/hosted_control_planes/bm_intro.adoc
```diff
@@ -53,26 +53,43 @@ oc get managedclusters local-cluster
...
 ** If the cluster API server address is an internal IP address, allow access from the workload subnets to the IP address on port 6443.
 ** If the address is an external IP address, allow egress on port 6443 to that external IP address from the nodes.
 
-* If you change the default port of 6443, adjust the rules to reflect that change.
-* Ensure that you open any ports that are required by the workloads that run in the clusters.
-* Use firewall rules, security groups, or other access controls to restrict access to only required sources. Avoid exposing ports publicly unless necessary.
-* For production deployments, use a load balancer to simplify access through a single IP address.
+* `Ignition`
+
+** The `Ignition` service runs on port 443 by default when exposed by using the route and ingress.
```
lahinson: Like lines 74 and 79, the statement on line 86 could be clearer.

clusters/hosted_control_planes/bm_intro.adoc
```diff
@@ -53,26 +53,43 @@ oc get managedclusters local-cluster
...
+** If you use the `NodePort` publishing strategy, use a firewall rule for the `Ignition` service.
+
+You do not require the following services on bare metal:
```
lahinson:
```suggestion
You do not need the following services on bare metal:
```

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
```diff
@@ -1,47 +1,51 @@
 [#hosted-enable-ext-dns-aws]
 = Enabling external DNS
 
-Because the control plane and the data plane are separate in hosted control planes, you can configure DNS in two independent areas:
+The control plane and the data plane are separate in hosted control planes. You can configure DNS in two independent areas:
```
lahinson, line 4: Nice revision!

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
```diff
@@ -1,47 +1,51 @@
...
+The input for the `hostedCluster.spec.dns` manages the ingress for workloads within the hosted cluster. The input for `hostedCluster.spec.services.servicePublishingStrategy.route.hostname` manages the ingress for service endpoints within the management cluster.
 
-External DNS creates name records for hosted cluster `Services` that specify a publishing type of `LoadBalancer` or `Route` and provide a hostname for that publishing type. For hosted clusters with `Private` or `PublicAndPrivate` endpoint access types, only the `APIServer` and `OAuth` services support hostnames. For `Private` hosted clusters, the DNS record resolves to a private IP of a Virtual Private Cloud (VPC) Endpoint in your VPC.
+External DNS creates name records for hosted cluster `Services` that specify a publishing type of `LoadBalancer` or `Route` and provide a hostname for that publishing type. For hosted clusters with `Private` or `PublicAndPrivate` endpoint access types, only the `APIServer` and `OAuth` services support hostnames. For `Private` hosted clusters, the DNS record resolves to a private IP of a Virtual Private Cloud (VPC) endpoint in your VPC.
```
lahinson:
```suggestion
External DNS creates name records for hosted cluster `Services` that specify a publishing type of `LoadBalancer` or `Route` and provide a hostname for that publishing type. For hosted clusters with `Private` or `PublicAndPrivate` endpoint access types, only the `APIServer` and `OAuth` services support hostnames. For `Private` hosted clusters, the DNS record resolves to a private IP address of a Virtual Private Cloud (VPC) endpoint in your VPC.
```

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
```diff
@@ -1,47 +1,51 @@
...
 
 * `APIServer`
 * `OAuthServer`
 * `Konnectivity`
 * `Ignition`
+* `OVNSbDb`
+* `OIDC`
 
-Each of those services is exposed by using `servicePublishingStrategy` in the `HostedCluster` specification. By default, for the `LoadBalancer` and `Route` types of `servicePublishingStrategy`, you publish the service in one of two ways: 
+Each of those services is exposed by using `servicePublishingStrategy` in the `HostedCluster` specification. By default, for the `LoadBalancer` and `Route` types of `servicePublishingStrategy`, you can publish the service by using one of the following ways:
```
lahinson:
```suggestion
Each of those services is exposed by using `servicePublishingStrategy` in the `HostedCluster` specification. By default, for the `LoadBalancer` and `Route` types of `servicePublishingStrategy`, you can publish the service in one of the following ways:
```

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
```diff
@@ -1,47 +1,51 @@
...
 
-You can deploy `external-dns` alongside the `hypershift` Operator in the `hypershift` namespace of the management cluster. The external DNS watches for `Services` or `Routes` that have the `external-dns.alpha.kubernetes.io/hostname` annotation. That annotation is used to create a DNS record that points to the `Service`, such as a record, or the `Route`, such as a CNAME record.
+You can deploy `external-dns` alongside the HyperShift Operator in the `hypershift` namespace of the management cluster. External DNS watches for `Services` or `Routes` that have the `external-dns.alpha.kubernetes.io/hostname` annotation. That annotation is used to create a DNS record that points to the `Service`, such as a record, or the `Route`, such as a CNAME record.
+
+You can use External DNS on cloud environments only. For the other environments, you need to configure DNS and services manually.
```
lahinson:
```suggestion
You can use external DNS on cloud environments only. For the other environments, you need to manually configure DNS and services.
```
Minor edits:
- External DNS doesn't need to be capitalized unless it's at the beginning of a sentence
- Per IBM Style, try to place adverbs as close to the word that they modify as possible

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
```diff
@@ -1,47 +1,51 @@
...
 
 [#external-dns-prereqs-aws]
 == Prerequisites
 
 Before you can set up external DNS for hosted control planes, you must meet the following prerequisites:
 
-* An external public domain that you can point to
+* You have created an external public domain
```
lahinson:
```suggestion
* You created an external public domain
```

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
```diff
@@ -147,7 +156,7 @@ This example shows the resulting `services` block for the hosted cluster:
 
 When the Control Plane Operator creates the `Services` and `Routes`, it annotates them with the `external-dns.alpha.kubernetes.io/hostname` annotation. The value is the `hostname` field in the `servicePublishingStrategy` for that type. The Control Plane Operator uses that name for the service endpoints, and it expects that if the hostname is set, a mechanism exists, such as external-dns or otherwise, which can create the DNS records.
```
lahinson:
```suggestion
When the Control Plane Operator creates the `Services` and `Routes`, it annotates them with the `external-dns.alpha.kubernetes.io/hostname` annotation. The value is the `hostname` field in the `servicePublishingStrategy` for that type. The Control Plane Operator uses that name for the service endpoints, and it expects that if the hostname is set, a mechanism exists, such as external DNS or otherwise, which can create the DNS records.
```
I'm not sure why "external DNS" is all lowercase and hyphenated. I suspect it should be "external DNS". Maybe check with @jparrill?

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
```diff
@@ -178,29 +187,43 @@ The following table notes when it is valid to set `hostname` for a service and e
 [#deploy-cluster-cli-external-dns-aws]
 == Deploying a cluster by using the command line interface and external DNS
 
-You need to deploy the `hypershift` and `external-dns` Operators when the external public hosted zone already exists. Ensure that the `external-dns` Operator is running and that the internal flags point to the public hosted zone by entering the following commands:
+You need to deploy the HyperShift Operator and the External DNS Operator when the external public hosted zone already exists.
```
lahinson: A question for @jparrill: Does the external public hosted zone usually already exist when users reach this task? I ask because the way this line is currently written, it almost sounds like a conditional statement. Would it be correct to revise it to something like this? "If the external public hosted zone already exists, you need to deploy the HyperShift Operator and the External DNS Operator."

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
```diff
@@ -178,29 +187,43 @@ The following table notes when it is valid to set `hostname` for a service and e
...
-    --name <cluster_name> \
+    --name <hosted_cluster_name> \ <3>
     --namespace clusters \
-    --base-domain service-consumer-domain.com \ <1>
-    --node-pool-replicas 2 \
-    --pull-secret ${HOME}/pull_secret.json \
-    --release-image quay.io/openshift-release-dev/ocp-release:4.13.0-ec.3-x86_64 \
-    --external-dns-domain=service-provider-domain.com \ <2>
-    --endpoint-access=PublicAndPrivate <3>
+    --base-domain service-consumer-domain.com \ <4>
```
lahinson: Is `service-consumer-doman.com` a variable? Should it be `<service_provider_domain>`?

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
```diff
@@ -178,29 +187,43 @@ The following table notes when it is valid to set `hostname` for a service and e
...
 ----
 
-<1> Points to the public hosted zone, `service-consumer-domain.com`, which is typically in an AWS account that the service consumer owns.
-<2> Points to the public external DNS hosted zone, `service-provider-domain.com`, which is typically in an AWS account that the service provider owns.
-<3> Set as `PublicAndPrivate.` The external DNS can be used with only `Public` or `PublicAndPrivate` configurations.
\ No newline at end of file
+<1> Specify the instance type, for example, `m6i.xlarge`.
+<2> Specify the AWS region, for example, `us-east-1`.
+<3> Specify your hosted cluster name, for example, `my-external-aws`.
+<4> Points to the public hosted zone that the service consumer owns, for example, `service-consumer-domain.com`.
```
lahinson: Should this be revised as follows? `Specify the public hosted zone that the service consumer owns, for example, `service-consumer-domain.com`.`

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
```diff
@@ -178,29 +187,43 @@ The following table notes when it is valid to set `hostname` for a service and e
...
+<5> Specify the node replica count, for example, `2`.
+<6> Specify the path to your pull secret file.
+<7> Specify the supported {ocp-short} version that you want to use, for example, `4.14.0-x86_64`.
+<8> Points to the public external DNS hosted zone that the service provider owns, for example, `service-provider-domain.com`.
```
lahinson: See comment about line 224. Should line 228 be revised to be an imperative statement?

lahinson (changes requested):
@xenolinux Thanks so much for improving these docs! I left a few suggestions as well as some questions that I'm hoping @jparrill can answer.

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
xenolinux: In the older version `--base-domain` is using the consumer domain. So I think this should be `<service_consumer_domain>`

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
xenolinux: Revised.

clusters/hosted_control_planes/bm_intro.adoc
xenolinux: per slack discussion, imo, it means that "the service is exposed by using the route and ingress" updated the sentence accordingly. @jparrill can confirm though.

clusters/hosted_control_planes/bm_intro.adoc
xenolinux: per slack discussion, imo, it means that "the service is exposed by using the route and ingress" updated the sentence accordingly. @jparrill can confirm though.

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
xenolinux: If it should be "external-dns" then it should have been within the back ticks in the first place `external-dns`.  I feel that [1] is confusing, for eg, we are unable to say if it's some kind of an API object or just external DNS entity.

[1] The value is the `hostname` field in the `servicePublishingStrategy` for that type. The Control Plane Operator uses that name for the service endpoints, and it expects that if the hostname is set, a mechanism exists, such as external-dns or otherwise, which can create the DNS records.

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
xenolinux: Per the "Creating the public DNS hosted zone" section that is listed just above this section, the public hosted zone already exists when users reach this task. 
See https://github.com/stolostron/rhacm-docs/pull/6424/files#diff-802df7b14d46e2d93353517b7ceb4c69bf22144b4147bbe2107eb42df1de05d0L92

xenolinux: I believe creating public zone is not optional so users must already have this public zone, and hence must install the HyperShift Operator and the External DNS Operator. -- confirming with Juan on the channel.

jparrill: Creating a Public HostedZone in Route53 is optional as long as customers do not use the `PublicAndPrivate` or `Public` configurations for their HostedClusters.

xenolinux: Per the overall discussion, adding the following clarification sentences:
- "The External DNS Operator uses the public DNS hosted zone to create your public hosted cluster."
- "To create a hosted cluster by using the `PublicAndPrivate` or `Public` publishing strategy, you must have the following artifacts configured in your management cluster:

* The public DNS hosted zone
* The External DNS Operator
* The HyperShift Operator"
- "For more information about external DNS, see link:https://github.com/kubernetes-sigs/external-dns[external DNS]."
- "Verify that the External DNS Operator is running by entering the following command:
oc get pod -n hypershift -lapp=external-dns
See the following example output:
NAME                            READY   STATUS    RESTARTS   AGE
external-dns-7c89788c69-rn8gp   1/1     Running   0          40s"
=====
Edit: I re-updated these sentences after confirming on slack.

clusters/hosted_control_planes/enable_ext_dns_aws.adoc
xenolinux: per the slack discussion - Updating to 

"The Control Plane Operator creates the `Services` and `Routes` resources and annotates them with the `external-dns.alpha.kubernetes.io/hostname` annotation. For `Services` and `Routes`, the Control Plane Operator uses a value of the `hostname` parameter in the `servicePublishingStrategy` field for the service endpoints. You can use a mechanism that creates the DNS records, such as the `external-dns` deployment."

xenolinux:
While updating, I realised the command to verify the status of EDNS Operator is missing. 

https://github.com/stolostron/rhacm-docs/pull/6424/files#diff-802df7b14d46e2d93353517b7ceb4c69bf22144b4147bbe2107eb42df1de05d0R204

xenolinux:
> While updating, I realised the command to verify the status of EDNS Operator is missing.
> 
> https://github.com/stolostron/rhacm-docs/pull/6424/files#diff-802df7b14d46e2d93353517b7ceb4c69bf22144b4147bbe2107eb42df1de05d0R204

Done. Added the command.

oafischer (approved):
Thanks for these updates and revisions! Approving, but take note that there is a broken link not related to your changes.
Screenshot 2024-05-27 at 14-40-02 rhacm-docs_clusters_hosted_control_planes_bm_intro adoc at 9969a64ca3ff728ce95f0e4c596e1ab037185528 · stolostron_rhacm-docs

xenolinux:
> Thanks for these updates and revisions! Approving, but take note that there is a broken link not related to your changes. Screenshot 2024-05-27 at 14-40-02 rhacm-docs_clusters_hosted_control_planes_bm_intro adoc at 9969a64ca3ff728ce95f0e4c596e1ab037185528 · stolostron_rhacm-docs

Fixed the link.

PR #6437, opened 2024-05-15
Add URL checking to the GitHub Actions linter

dhaiducek:
Adds a link checker to the GitHub Action linting.

Also resolves the broken links that were reported (since otherwise this PR would merge with a failing workflow).

ref: https://issues.redhat.com/browse/ACM-11442

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -6,15 +6,22 @@ on:
 
 jobs:
   asciidoc:
-    name: 'Build AsciiDocs'
+    name: Build AsciiDocs
     runs-on: ubuntu-latest
     steps:
       - name: Check out the repository
-        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
+        uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b
+
+      - name: Install asciidoctor
+        run: |
+          sudo apt-get install -y asciidoctor
+
       - name: Build each .adoc with Asciidoctor, fail on WARN or above
         run: |
-          FAILED=0
-          if ! docker run --rm -v $(pwd):/rhacm-docs/ asciidoctor/docker-asciidoctor@sha256:59a7e165ed6375dd5dfb930cb90019f1ccaee39787dc97a3918d7022a7503543 find /rhacm-docs/ -name \*.adoc -exec asciidoctor --failure-level WARN --trace --verbose --warnings {} +; then
-            echo "Build errors were detected, see per-file details above."
-            exit 1
-          fi
+          find . -name \*.adoc -exec asciidoctor --failure-level WARN --trace --verbose --warnings {} +
+
+      - name: Link Check
+        uses: lycheeverse/lychee-action@2b973e86fc7b1f6b36a93795fe2c9c6ae1118621
+        with:
+          args: --no-progress --timeout 30 --scheme http --scheme https --require-https "./**/main.html"
+          fail: true
--- a/gitops/gitops_registering.adoc
+++ b/gitops/gitops_registering.adoc
@@ -82,8 +82,8 @@ rules:
 
 [#additional-resources-gitops]
 == Additional resources
-//OF | fix first link | Aug 28 23
-- Refer to xref:../applications/gitops_config.adoc#tolerations-config[Configuring placement tolerations for {product-title-short} and OpenShift GitOps] for more details.
+
+- Refer to xref:../gitops/gitops_tolerations_config.adoc#tolerations-config[Configuring application placement tolerations for GitOps] for more details.
 
 - See the link:https://github.com/stolostron/multicloud-integrations/blob/main/examples/managedclustersetbinding.yaml[`multicloud-integrations` managed cluster set binding] example.
 
--- a/gitops/gitops_registering.adoc
+++ b/gitops/gitops_registering.adoc
@@ -15,8 +15,6 @@ To configure GitOps with the Push model, you can register a set of one or more {
 
 Complete the following steps to register managed clusters to GitOps:
 
-. Create managed cluster sets and add managed clusters to those managed cluster sets. See the example for managed cluster sets in the link:https://github.com/stolostron/multicloud-integrations/blob/main/examples/openshift-gitops/managedclusterset.yaml[`multicloud-integrations managedclusterset`].
-
 +
 See the link:../clusters/cluster_lifecycle/create_clusterset.adoc#creating-a-managedclusterset[Creating a _ManagedClusterSet_] documentation for more information.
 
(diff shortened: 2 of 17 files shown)


dhaiducek:
/cc @skitt @dfarrell07 

Please take a look at the link checker update, since you all authored the original workflow.

governance/iam_policy_ctrl.adoc
```diff
@@ -77,4 +77,4 @@ Enter `inform`. The IAM policy controller only supports the `inform` feature.
 [#iam-policy-sample]
 == IAM policy sample
 
-See link:https://github.com/open-cluster-management/policy-collection/blob/main/stable/AC-Access-Control/policy-limitclusteradmin.yaml[`policy-limitclusteradmin.yaml`] to view the IAM policy sample. See xref:../governance/create_policy.adoc#managing-security-policies[Managing security policies] for more information. Refer to xref:../governance/policy_controllers.adoc#policy-controllers[Policy controllers] for more topics.
+See xref:../governance/create_policy.adoc#managing-security-policies[Managing security policies] for more information. Refer to xref:../governance/policy_controllers.adoc#policy-controllers[Policy controllers] for more topics.
```
dockerymick, line 80: The link for policy controllers is broken but the IAM Policy is being removed, so no need to fix it

dockerymick (approved):
I made one update that was related to a hidden comment. I also left a comment about the IAM Policy controller being removed. Otherwise, this looks good to me. Approving this

gitops/gitops_registering.adoc
```diff
@@ -84,10 +82,10 @@ rules:
 
 [#additional-resources-gitops]
 == Additional resources
-//OF | fix first link | Aug 28 23
-- Refer to xref:../applications/gitops_config.adoc#tolerations-config[Configuring placement tolerations for {product-title-short} and OpenShift GitOps] for more details.
 
-- See the link:https://github.com/open-cluster-management/multicloud-integrations/blob/main/examples/managedclusterset.yaml[`multicloud-integrations` managed cluster set] example.
+- Refer to xref:../gitops/gitops_tolerations_config.adoc#tolerations-config[Configuring application placement tolerations for GitOps] for more details.
+
+- See the link:https://github.com/stolostron/multicloud-integrations/blob/main/examples/managedclustersetbinding.yaml[`multicloud-integrations` managed cluster set binding] example.
```
dockerymick: Is there a different example that we can point to? I recall Brandi mentioned that we shouldn't point to stolostron. Preferably one from the open-cluster-management repo.

governance/iam_policy_ctrl.adoc
dhaiducek: > The link for policy controllers is broken

I'm a little surprised the asciidoc run didn't pick up on that...I thought it checked those. I might have to dig at that.

gitops/gitops_registering.adoc
dhaiducek: The original ManagedClusterSet file got deleted, but it looks like when they deleted it they were already wanting it to be skipped and just use the ManagedClusterSetBinding:

```
#            Use the existing 'global' or 'default' ManagedClusterSet instead.
```
https://github.com/stolostron/multicloud-integrations/commit/dd2d25c0390a4e287e7419d4d83ae7bd0a0e8ced#diff-63080df67636f451472893dfb67012fffaca6ad61ca1a6ddc42f45da437092acL3

If needed, someone from App Lifecycle ought to be able to confirm what direction we should go on it.

governance/iam_policy_ctrl.adoc
dhaiducek: Actually, it seems at quick glance, there actually are about 248 internal cross references like it that are broken. We can merge this PR as-is for the external link checking and then I can create a new issue and PR for the internal links that are broken. Apparently the original linter wasn't quite as thorough as I thought it was...

For reference, here's the list it found: https://github.com/dhaiducek/rhacm-docs/actions/runs/9102688813#summary-25022880725

dhaiducek: Sorry for my back-and-forth here. It seems the link checker doesn't handle the nested directories quite as well, so there are actually a lot of false positives in there and I'll have to poke at it some more before opening any issues on it.

gitops/gitops_registering.adoc
dockerymick: @xiangjingli will you help provide a link to the sample from the open-cluster-management repo?

.github/workflows/linting.yml
```diff
@@ -6,15 +6,22 @@ on:
 
 jobs:
   asciidoc:
-    name: 'Build AsciiDocs'
+    name: Build AsciiDocs
     runs-on: ubuntu-latest
     steps:
       - name: Check out the repository
-        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
+        uses: actions/checkout@v4
```
skitt: Tags can be moved, which causes silent updates to actions; it’s best to refer to specific commits (and ideally, review the changes in actions). dependabot knows how to handle this.

```suggestion
        uses: actions/checkout@44c2b7a8a4ea60a981eaca3cf939b5f4305c123b
```

.github/workflows/linting.yml
```diff
@@ -6,15 +6,22 @@ on:
...
         run: |
-          FAILED=0
-          if ! docker run --rm -v $(pwd):/rhacm-docs/ asciidoctor/docker-asciidoctor@sha256:59a7e165ed6375dd5dfb930cb90019f1ccaee39787dc97a3918d7022a7503543 find /rhacm-docs/ -name \*.adoc -exec asciidoctor --failure-level WARN --trace --verbose --warnings {} +; then
-            echo "Build errors were detected, see per-file details above."
-            exit 1
-          fi
+          find . -name \*.adoc -exec asciidoctor --failure-level WARN --trace --verbose --warnings {} +
+
+      - name: Link Check
+        uses: lycheeverse/lychee-action@v1
```
skitt: Same here:

```suggestion
        uses: lycheeverse/lychee-action@2b973e86fc7b1f6b36a93795fe2c9c6ae1118621
```

.github/workflows/linting.yml
```diff
@@ -6,15 +6,22 @@ on:
...
+          sudo apt-get install -y asciidoctor
+
       - name: Build each .adoc with Asciidoctor, fail on WARN or above
         run: |
-          FAILED=0
-          if ! docker run --rm -v $(pwd):/rhacm-docs/ asciidoctor/docker-asciidoctor@sha256:59a7e165ed6375dd5dfb930cb90019f1ccaee39787dc97a3918d7022a7503543 find /rhacm-docs/ -name \*.adoc -exec asciidoctor --failure-level WARN --trace --verbose --warnings {} +; then
-            echo "Build errors were detected, see per-file details above."
-            exit 1
-          fi
+          find . -name \*.adoc -exec asciidoctor --failure-level WARN --trace --verbose --warnings {} +
```
skitt, line 21: This is fine as far as I’m concerned.

dhaiducek:
Thanks for the review, @skitt! I've also added a Dependabot configuration to keep the action up-to-date. (@dockerymick for reference, Dependabot will automatically open a PR to keep the linter up-to-date if there are any updates required.)

Surprisingly, the HCP links came up broken again (it looks like they updated some of the "hcp" to be "hosted-control-planes"), but I've gone through and fixed the ones it found.

gitops/gitops_registering.adoc
xiangjingli: @dockerymick   Here is the link for the managedClusterSetBinding yaml sample

https://github.com/open-cluster-management-io/multicloud-integrations/blob/main/examples/managedclustersetbinding.yaml

dockerymick:
```suggestion
- See the link:https://github.com/open-cluster-management-io/multicloud-integrations/blob/main/examples/managedclustersetbinding.yaml[`multicloud-integrations` managed cluster set binding] example.
```

dockerymick (approved):
Thanks for the reviews and patience. I am approving this and also mentioning that there are discussion about the content team moving away from Ascii.

dhaiducek:
The HCP links seem to be changing a bit more frequently than I'd think they would, but I double checked and fixed this second round of broken links.

PR #6444, opened 2024-05-20
https://issues.redhat.com/browse/ACM-11746

haoqing0110:
- Klusterlet OLM operator references are removed from documentation
- Release Note included indicating that the Klusterlet OLM operator has been completely removed and not supported

### DIFF
--- a/clusters/release_notes/deprecate_remove.adoc
+++ b/clusters/release_notes/deprecate_remove.adoc
@@ -64,4 +64,10 @@ A _removed_ item is typically function that was deprecated in previous releases 
 
 |===
 |Product or category | Affected item | Version | Recommended action | More details and links
+
+| Cluster lifecycle
+| Klusterlet OLM Operator
+| 2.6
+| None
+| None
 |===
--- a/clusters/release_notes/deprecate_remove.adoc
+++ b/clusters/release_notes/deprecate_remove.adoc
@@ -51,7 +51,7 @@ A _deprecated_ component, feature, or service is supported, but no longer recomm
 | None
 
 | Cluster lifecycle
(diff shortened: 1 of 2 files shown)


oafischer (approved):
Looks great, thanks @haoqing0110!

showeimer (approved):

clusters/release_notes/deprecate_remove.adoc
```diff
@@ -64,4 +64,10 @@ A _removed_ item is typically function that was deprecated in previous releases
 
 |===
 |Product or category | Affected item | Version | Recommended action | More details and links
+
+| Cluster lifecycle
+| Klusterlet OLM Operator
+| 2.6
```
swopebe, line 70: 2.6 is EOL and as I mentioned on scrum today, I just removed 2.7 and earlier from this same doc. It's in a PR. Remember we only go N-2 here.

clusters/release_notes/deprecate_remove.adoc
```diff
@@ -64,4 +64,10 @@ A _removed_ item is typically function that was deprecated in previous releases
 
 |===
 |Product or category | Affected item | Version | Recommended action | More details and links
+
+| Cluster lifecycle
+| Klusterlet OLM Operator
```
swopebe: We do have a content attribute for OLM, as I am pretty sure OLM itself is not approved. That is the only thing that may change here. Approving otherwise.

swopebe (approved):

clusters/release_notes/deprecate_remove.adoc
oafischer: Updated

PR #6457, opened 2024-05-23
https://issues.redhat.com/browse/ACM-11353 adding info to policy table-using attribute for proper OLM name

swopebe:


governance/policy_operator.adoc
```diff
@@ -42,13 +44,18 @@ If the `remediationAction` set to `inform`, the controller only reports the stat
 
 | `spec.subscription`
 | Required
-a| Define the configurations to create an operator subscription. You must add information in the following fields to create an operator subscription:
```
mprahl, line 45: The `a|` is required for it to be treated as a bulleted list. Was it intentional to remove this?

swopebe: Oh, OK, no I thought this was a typo and didn't see why it was used in VS Code. I will add it @mprahl . Anything else?

swopebe: @mprahl I don't see the difference here, as the list within the table is bulleted, I just coded it as normal. What else are we trying to bullet here? We don't do this in tables typically, unless someone goes rogue for some reason. 
Here is 2.11 as it is now: https://github.com/stolostron/rhacm-docs/edit/2.11_stage/governance/policy_operator.adoc

swopebe: The way I have it built shows a bulleted list. @dockerymick what does the a do here that is not already done?

mprahl: Here is what is rendered without the "a" and with newlines between the list items. As you can see, it's just seen as plaintext:
image

When the "a" is present and there are no newlines between the list items, ASCIIDoc sees it as unordered list and renders it as such:
image

mprahl: The rest of the text looks good to me.

swopebe: it just seems hacky to me, but let me see how it builds with a change in the bullet structure. I would rather keep the tables clean if we can, especially since the other books are not doing this. I will check with the build.

swopebe: Ok, I tried various bulleted list techniques that we use regularly and this `a` codes it best, which means I will show the team to ensure we are creating bulleted lists in tables the same way and see if this is in guidance anywhere. If we don't, someone else will just see it as a typo and remove it again.

dockerymick: @mprahl @swopebe thanks to you both for editing the file and having discussion. Here is the link that describes the `a` operator: https://docs.asciidoctor.org/asciidoc/latest/tables/format-cell-content/

As Brandi mentioned, we should mention this to the doc team and potentially add it to our internal docs so that others do not delete the `a` operator.

governance/policy_operator.adoc
```diff
@@ -42,13 +44,18 @@ If the `remediationAction` set to `inform`, the controller only reports the stat
...
+| Define the configurations to create an operator subscription. You must add information in the following fields to create an operator subscription:
```
dockerymick:
```suggestion
a| Define the configurations to create an operator subscription. You must add information in the following fields to create an operator subscription:
```

governance/policy_operator.adoc
swopebe: thanks for the resource. This is ready for final review when you are ready.

dockerymick (approved):
Thank you!

PR #6484, opened 2024-06-03
https://issues.redhat.com/browse/ACM-11257

oafischer:
Train 14, also for 2.10.


dockerymick (approved):
This looks good to me! Thank you!

governance/policy_generator.adoc
```diff
@@ -65,6 +65,26 @@ data:
   key2: value2
 ----
 
+You can also use the `object-templates-raw` manifest to automatically generate a `ConfigurationPolicy` with the content you add. See the following example:
```
JeffeyL, line 68: Since the generated Policy example below does not use object-templates-raw, maybe it would be better for this section to be located after showing the generated Policy?

oafischer: Can you explain more @JeffeyL? The policy below uses object-templates-raw as far as I can tell.

PR #6508, opened 2024-06-06
[ACM-12049]: Update hcp requirements, clean up intro

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-12049

This PR adds information to the note in the Requirements section to clarify that the management cluster and workers must use the same kind of infrastructure.

I noticed that the hosted-intro.adoc file as a whole could be cleaned up, so I removed some of the long lists of links that were acting as a sort of "next steps". When you're looking at the docs on the portal or on docs.redhat.com, the next steps are obvious and don't need to be called out explicitly. I kept the links to OpenShift and the "Configuring additional networks.." docs as additional resources. FYI @xenolinux @swopebe


clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -31,29 +15,27 @@ Hosted control plane clusters offer several advantages:
...
+| Bare metal | 4.14 - 4.15 | 4.14 - 4.15 (only)
+| {ocp-virt} | 4.14 - 4.15 | 4.14 - 4.15 (only)
+| AWS (Technology Preview) | 4.11 - 4.15 | 4.14 - 4.15 (only)
+|===
 
-To configure additional networks, guaranteed CPUs, and VM scheduling for node pools, see the following documentation:
+**Note:** The hub cluster and workers must run on the same platform for hosted control planes. In addition, the hosting cluster and workers must run on the same infrastructure. For example, you cannot run your hosting cluster on bare metal and your workers on the cloud.
 
-* xref:../hosted_control_planes/managing_nodepools_kubevirt.adoc#managing-nodepools-hosted-cluster-kubevirt[Configuring additional networks, guaranteed CPUs, and VM scheduling for node pools]
+The control plane runs as pods that are contained in a single namespace and is associated with the hosted control plane cluster. When {ocp-short} creates this type of hosted cluster, it creates a worker node that is independent of the control plane.
```
zanetworker:
```suggestion
The control plane runs as pods that are contained in a single namespace and is associated with the hosted cluster. When {ocp-short} creates this type of hosted cluster, it creates a worker node that is independent of the control plane.
```

zanetworker: also was a bit confused about this statement "creates this type of hosted cluster". Which type do you mean?

lahinson: Good question. I think we can just say "When {ocp-short} creates a hosted cluster..."

lahinson: FYI @xenolinux Please see Adel's question and my suggested revision and let me know if anything should change. Thanks!

xenolinux: Thanks. Nicely done! 
Can we simplify this sentence? "The control plane runs as pods that are contained in a single namespace and is associated with the hosted cluster."
Something like -- feel free to tweak the sentence 
"The control plane pods associated with the hosted cluster run in a single namespace"

clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -31,29 +15,27 @@ Hosted control plane clusters offer several advantages:
...
-* xref:../hosted_control_planes/disable_hosted.adoc#disable-hosted-control-planes[Disabling the hosted control feature]
+|===
+| Platform | Hosting {ocp-short} version | Hosted {ocp-short} version
+| Bare metal | 4.14 - 4.15 | 4.14 - 4.15 (only)
+| {ocp-virt} | 4.14 - 4.15 | 4.14 - 4.15 (only)
+| AWS (Technology Preview) | 4.11 - 4.15 | 4.14 - 4.15 (only)
+|===
 
-To configure additional networks, guaranteed CPUs, and VM scheduling for node pools, see the following documentation:
+**Note:** The hub cluster and workers must run on the same platform for hosted control planes. In addition, the hosting cluster and workers must run on the same infrastructure. For example, you cannot run your hosting cluster on bare metal and your workers on the cloud.
```
zanetworker: >The hub cluster and workers must run on the same platform 
I think we can relax this, but keep the "infrastructure environment" piece. What matters is either cp + workers are all on-prem or all in cloud. On prem could mean that my hub / management clusters run on vsphere for example, and the workers baremetal. It could mean that my hub is baremetal and my workers are openshift virtualization.

clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -31,29 +15,27 @@ Hosted control plane clusters offer several advantages:
...
 
-* xref:../hosted_control_planes/managing_nodepools_kubevirt.adoc#managing-nodepools-hosted-cluster-kubevirt[Configuring additional networks, guaranteed CPUs, and VM scheduling for node pools]
+The control plane runs as pods that are contained in a single namespace and is associated with the hosted cluster. When {ocp-short} creates a hosted cluster, it creates a worker node that is independent of the control plane.
```
zanetworker: What do we want to say here? 

Is "ocp_short" replaced by "OpenShift"? Should we use personas? For example, when the cluster service consumer creates a hosted cluster, the HyperShift operator will reconcile the control plane as yet another workload in a dedicated project/namespace on that management cluster.

clusters/hosted_control_planes/hosted_intro.adoc
lahinson: @xenolinux I like where you're headed. How's this? `The control plane is associated with a hosted cluster and runs as pods in a single namespace.`

xenolinux: Nice! Thanks

clusters/hosted_control_planes/hosted_intro.adoc
lahinson: @zanetworker Yep, {ocp-short} is a variable that is rendered as "OpenShift Container Platform" in the docs. I know it's hard to decipher without a preview. So this is how the current revision reads:

`When OpenShift Container Platform creates a hosted cluster, it creates a worker node that is independent of the control plane.`

Do our customers typically think in terms of "the cluster service consumer"? If so, I'm happy to use that terminology. That would look something like this:

`When the cluster service consumer creates a hosted cluster, the HyperShift Operator creates the control plane as a workload in a dedicated namespace on a hosting cluster.` (I used "hosting cluster" here instead of "management cluster" because in the MCE docs, "hosting cluster" is the preferred term.)

zanetworker: Makes sense to me Laura. I can't say we are consistent, but IMO describing from the user's / persona perspective helps.

lahinson: Thanks for your feedback! I can word things from the user/persona's perspective.

clusters/hosted_control_planes/hosted_intro.adoc
zanetworker: Here I think we should put the emphasis on the infrastructure and relax the platform constraints (e.g., running management cluster on VMWare but using Agent to provision BM workers).

clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -27,9 +27,9 @@ The following table indicates which {ocp-short} versions are supported for each
 | AWS (Technology Preview) | 4.11 - 4.15 | 4.14 - 4.15 (only)
 |===
 
-**Note:** The hub cluster and workers must run on the same platform for hosted control planes. In addition, the hosting cluster and workers must run on the same infrastructure. For example, you cannot run your hosting cluster on bare metal and your workers on the cloud.
+**Note:** The hosting cluster and workers must run on the same infrastructure. For example, you cannot run your hosting cluster on bare metal and your workers on the cloud. However, the hub cluster and workers do not need to run on the same platform. For example, you might run your hosting cluster on OpenShift Virtualization and workers on bare metal.
```
zanetworker: Can we reverse it? I.e., Hosting clustero on baremetal, workers on OpenShift virtualization. Both are on-prem providers.

lahinson: Sure - done!

zanetworker (approved):
lgtm

clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -31,29 +15,27 @@ Hosted control plane clusters offer several advantages:
...
-To use hosted control planes, start with the following documentation:
-
-. xref:../hosted_control_planes/hosted_sizing_guidance.adoc#hosted-sizing-guidance[Hosted control plane sizing guidance]
-. xref:../hosted_control_planes/install_hcp_cli.adoc#hosted-install-cli[Installing the hosted control plane command line interface]
-. xref:../hosted_control_planes/distribute_cluster_workloads.adoc#hosted-cluster-workload-distributing[Distributing hosted cluster workloads]
+[#hosted-control-requirements]
+== Requirements
 
-Then, see the documentation related to the platform that you plan to use:
+The following table indicates which {ocp-short} versions are supported for each platform. In the table, _Hosting {ocp-short} version_ refers to the {ocp-short} version where {mce-short} is enabled:
```
jc-berger, line 25: Optional:

The note you had after the table seems to include pertinent information for users to understand the table and what the table does for them. 

Therefore, I'm wondering if it would serve the user experience more efficiently if this content came _before_ the table. What do you think? 

```suggestion
The hosting cluster and workers must run on the same infrastructure. For example, you cannot run your hosting cluster on bare metal and your workers on the cloud. However, the hub cluster and workers do not need to run on the same platform. For example, you might run your hosting cluster on bare metal and workers on OpenShift Virtualization.

The control plane is associated with a hosted cluster and runs as pods in a single namespace. When the cluster service consumer creates a hosted cluster, it creates a worker node that is independent of the control plane.

The following table indicates which {ocp-short} versions are supported for each platform. In the table, _Hosting {ocp-short} version_ refers to the {ocp-short} version where {mce-short} is enabled:
```

clusters/hosted_control_planes/hosted_intro.adoc
```diff
@@ -31,29 +15,27 @@ Hosted control plane clusters offer several advantages:
...
-* xref:../hosted_control_planes/disable_hosted.adoc#disable-hosted-control-planes[Disabling the hosted control feature]
+|===
+| Platform | Hosting {ocp-short} version | Hosted {ocp-short} version
+| Bare metal | 4.14 - 4.15 | 4.14 - 4.15 (only)
+| {ocp-virt} | 4.14 - 4.15 | 4.14 - 4.15 (only)
+| AWS (Technology Preview) | 4.11 - 4.15 | 4.14 - 4.15 (only)
+|===
 
-To configure additional networks, guaranteed CPUs, and VM scheduling for node pools, see the following documentation:
+**Note:** The hosting cluster and workers must run on the same infrastructure. For example, you cannot run your hosting cluster on bare metal and your workers on the cloud. However, the hub cluster and workers do not need to run on the same platform. For example, you might run your hosting cluster on bare metal and workers on OpenShift Virtualization.
```
jc-berger: If you liked my earlier comment, delete the note here.

jc-berger (approved):
/lgtm

clusters/hosted_control_planes/hosted_intro.adoc
lahinson: Your suggestions makes good sense to me, @jc-berger. I'll incorporate your revision. Thank you!

PR #6002, opened 2024-01-06
added note about clusters being removed

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-9065

PR #6518, opened 2024-06-10
expanded observability note with config settings getting replaced

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-11724

See earlier PR for expanded notes for observability: https://github.com/stolostron/rhacm-docs/pull/6002/files

### DIFF
--- a/observability/observability_enable.adoc
+++ b/observability/observability_enable.adoc
@@ -41,7 +41,7 @@ Enable the observability service by creating a `MultiClusterObservability` custo
 
 - When observability is enabled or disabled on {ocp-short} managed clusters that are managed by {product-title-short}, the observability endpoint operator updates the `cluster-monitoring-config` config map by adding additional `alertmanager` configuration that automatically restarts the local Prometheus.
 - The observability endpoint operator updates the `cluster-monitoring-config` config map by adding additional `alertmanager` configurations that automatically restart the local Prometheus. Therefore, when you insert the `alertmanager` configuration in the {ocp-short} managed cluster, the configuration removes the settings that relate to the retention of the Prometheus metrics. 
-- When you import an {ocp-short} managed cluster onto the hub cluster, {product-title-short} overwrites the `retentionSize` configuration under the `prometheusK8s` section of the config map. 
+- When you import an {ocp-short} managed cluster onto the hub cluster, {product-title-short} removes all the configuration settings from the `cluster-monitoring-config` config map of the {ocp-short} managed cluster. When the configuration settings are removed, {product-title-short} replaces it with the settings from the hub cluster.  
 
 Complete the following steps to enable the observability service: 
  
--- a/observability/observability_enable.adoc
+++ b/observability/observability_enable.adoc
@@ -41,7 +41,6 @@ Enable the observability service by creating a `MultiClusterObservability` custo
 
 - When observability is enabled or disabled on {ocp-short} managed clusters that are managed by {product-title-short}, the observability endpoint operator updates the `cluster-monitoring-config` config map by adding additional `alertmanager` configuration that automatically restarts the local Prometheus.
 - The observability endpoint operator updates the `cluster-monitoring-config` config map by adding additional `alertmanager` configurations that automatically restart the local Prometheus. When you insert the `alertmanager` configuration in the {ocp-short} managed cluster, the configuration removes the settings that relate to the retention field of the Prometheus metrics. 
-- When you configure the `MultiClusterObservability` to the `cluster-monitoring-config`, then the only fields that persist are the fields inside the `cluster-monitoring-config`. You lose all other fields. 
 
 Complete the following steps to enable the observability service: 
  
--- a/observability/observability_enable.adoc
+++ b/observability/observability_enable.adoc
@@ -40,8 +40,8 @@ Enable the observability service by creating a `MultiClusterObservability` custo
 *Note:* 
 
 - When observability is enabled or disabled on {ocp-short} managed clusters that are managed by {product-title-short}, the observability endpoint operator updates the `cluster-monitoring-config` config map by adding additional `alertmanager` configuration that automatically restarts the local Prometheus.
-- The observability endpoint operator updates the `cluster-monitoring-config` config map by adding additional `alertmanager` configurations that automatically restart the local Prometheus. Therefore, when you insert the `alertmanager` configuration in the {ocp-short} managed cluster, the configuration removes the settings that relate to the retention of the Prometheus metrics. 
-- When you import an {ocp-short} managed cluster onto the hub cluster, {product-title-short} removes all the configuration settings from the `cluster-monitoring-config` config map of the {ocp-short} managed cluster. When the configuration settings are removed, {product-title-short} replaces it with the settings from the hub cluster.  
+- The observability endpoint operator updates the `cluster-monitoring-config` config map by adding additional `alertmanager` configurations that automatically restart the local Prometheus. When you insert the `alertmanager` configuration in the {ocp-short} managed cluster, the configuration removes the settings that relate to the retention field of the Prometheus metrics. 
+- When you configure the `MultiClusterObservability` to the `cluster-monitoring-config`, then the only fields that persist are the fields inside the `cluster-monitoring-config`. You lose all other fields. 
 
 Complete the following steps to enable the observability service: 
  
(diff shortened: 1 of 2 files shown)


observability/observability_enable.adoc
```diff
@@ -41,6 +41,7 @@ Enable the observability service by creating a `MultiClusterObservability` custo
 
 - When observability is enabled or disabled on {ocp-short} managed clusters that are managed by {product-title-short}, the observability endpoint operator updates the `cluster-monitoring-config` config map by adding additional `alertmanager` configuration that automatically restarts the local Prometheus.
 - The observability endpoint operator updates the `cluster-monitoring-config` config map by adding additional `alertmanager` configurations that automatically restart the local Prometheus. Therefore, when you insert the `alertmanager` configuration in the {ocp-short} managed cluster, the configuration removes the settings that relate to the retention of the Prometheus metrics. 
+- When you import an {ocp-short} managed cluster into the hub cluster, {product-title-short} removes all the configuration settings in that {ocp-short} managed cluster `cluster-monitoring-config` configmap. After removing the configuration settings, {product-title-short} replaces it with the settings passed down from the hub cluster.
```
dockerymick: Can we use something different from "passed down". Does the console take the action of replacing the configuration settings or does the user have to remove it? Removing anthropomorphism and removed directional phrase

```suggestion
- When you import an {ocp-short} managed cluster onto the hub cluster, {product-title-short} removes all the configuration settings from the `cluster-monitoring-config` config map of the {ocp-short} managed cluster. When the configuration settings are removed, {product-title-short} replaces it with the settings from the hub cluster.  
```

dockerymick (changes requested):

dockerymick (approved):

dockerymick:
We still need development to review. Added @subbarao-meduri and @douglascamata as reviewers

observability/observability_enable.adoc
```diff
@@ -41,6 +41,7 @@ Enable the observability service by creating a `MultiClusterObservability` custo
 
 - When observability is enabled or disabled on {ocp-short} managed clusters that are managed by {product-title-short}, the observability endpoint operator updates the `cluster-monitoring-config` config map by adding additional `alertmanager` configuration that automatically restarts the local Prometheus.
 - The observability endpoint operator updates the `cluster-monitoring-config` config map by adding additional `alertmanager` configurations that automatically restart the local Prometheus. Therefore, when you insert the `alertmanager` configuration in the {ocp-short} managed cluster, the configuration removes the settings that relate to the retention of the Prometheus metrics. 
+- When you import an {ocp-short} managed cluster onto the hub cluster, {product-title-short} removes all the configuration settings from the `cluster-monitoring-config` config map of the {ocp-short} managed cluster. When the configuration settings are removed, {product-title-short} replaces it with the settings from the hub cluster.
```
douglascamata: This is not correct. The only thing being overwritten is the `retentionSize` configuration under the `prometheusK8s` section of the config map. 

We should probably document this as a known issue.

douglascamata (changes requested):

douglascamata (approved):
/lgtm

observability/observability_enable.adoc
```diff
@@ -41,6 +41,7 @@ Enable the observability service by creating a `MultiClusterObservability` custo
 
 - When observability is enabled or disabled on {ocp-short} managed clusters that are managed by {product-title-short}, the observability endpoint operator updates the `cluster-monitoring-config` config map by adding additional `alertmanager` configuration that automatically restarts the local Prometheus.
 - The observability endpoint operator updates the `cluster-monitoring-config` config map by adding additional `alertmanager` configurations that automatically restart the local Prometheus. Therefore, when you insert the `alertmanager` configuration in the {ocp-short} managed cluster, the configuration removes the settings that relate to the retention of the Prometheus metrics. 
+- When you import an {ocp-short} managed cluster onto the hub cluster, {product-title-short} overwrites the `retentionSize` configuration under the `prometheusK8s` section of the config map.
```
swopebe: Let's not use "under" --- reason:

Let's say we were to run an accessibility test on content, this is a word that could get flagged as a directional word that could confuse a user with a screen reader...then one would have to go justify the use of the word. It's kinder to the user not to use it and it reduces overhead to change or explain.

swopebe: I sometimes change to "from the ___section" "Within the section" something that is not directional.

release_notes/known_issues_observability.adoc
```diff
@@ -27,6 +27,12 @@ For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/o
 
 For more about deprecations and removals, see xref:../release_notes/deprecate_remove.adoc#deprecations-removals[Deprecations and removals].
 
+[#configurations-getting-overwritten]
+== Configurations getting overwritten
+//2.11:11724
+
+When you import an {ocp-short} managed cluster onto the hub cluster, {product-title-short} overwrites the `retentionSize` configuration under the `prometheusK8s` section of the config map. There is currently no workaround for this issue.
```
swopebe: Same with under but now I have to ask, why are we documenting a known issue with the same content that we put in the task? (Usually it is one or the other, the task is more official and widely used, the known issue is something that will be fixed in an upcoming release (typically).

swopebe (changes requested):

release_notes/known_issues_observability.adoc
jc-berger: @douglascamata you suggested including it as a known issue as well. Do you think this content should _only_ exist as a known issue? 

If you want it as a note in the observability doc and as a known issue, do we have workaround steps?

observability/observability_enable.adoc
coffeegoesincodecomesout: @douglascamata I have a case attached to the following knowledge base article [1]
that details an issue a CU had with additional alertmanager secrets [2]being removed when importing a managed cluster into ACM - hence the ask to expand this statement in the documentation 


[1] - https://access.redhat.com/solutions/7070731 
[2] - https://access.redhat.com/documentation/de-de/openshift_container_platform/4.14/html-single/monitoring/index#monitoring-adding-a-secret-to-the-alertmanager-configuration_configuring-the-monitoring-stack

release_notes/known_issues_observability.adoc
douglascamata: We can have this content as a known issue only. There's no workaround or any action from the user that will prevent the configuration from being overwritten.

observability/observability_enable.adoc
douglascamata: @coffeegoesincodecomesout let me double check what's the situation for that alertmanager configuration. The root cause is more generic and can involve other fields of the config map being reset to their default values.

douglascamata: @coffeegoesincodecomesout in the `alertmanagerMain` section of the config map, these are the fields that **will not** be overwritten:

* `enabled`
* `enableUserAlertmanagerConfig`
* `logLevel`
* `nodeSelector`
* `tolerations`
* `resources`
* `volumeClaimTemplate`

Everything else will be gone. As you can see, the `secrets` key is not one of the configurations that will stay.

douglascamata: @coffeegoesincodecomesout it's hard to document everything that will or will not be overwritten from this config map, as it's quite extensive. 

Basically the MCO is locked to the configuration from the commit `20fcb9f6ef4e00ccde05a13347bddfabe24e8c4c` of the CMO: https://github.com/openshift/cluster-monitoring-operator/blob/20fcb9f6ef4e00ccde05a13347bddfabe24e8c4c/pkg/manifests/config.go#L95. Only fields present in that version of the `ClusterMonitoringConfiguration` will persist.

jc-berger:
@douglascamata and @coffeegoesincodecomesout thank you both for devoting time to this PR! As you said, Douglas, it's tedious to document all the fields that do and do not get overwritten, and that can be problematic to maintain. 

Therefore, please let me know what you think of my newest commit, where I specified only the retention field in Prometheus is guaranteed to be overwritten, and how only fields in the cluster monitoring config are guaranteed to persist.

douglascamata:
@jc-berger honestly the new text reads more confusing to me than the previous, even though the previous one is incomplete. :/

jc-berger:
@douglascamata and @coffeegoesincodecomesout Nigel suggested over Slack that we focus on the known issue. For the observability notes here, he said we could keep the original content, but I did keep my small change to specify it's the retention field getting removed in the Prometheus matrix. 

What do y'all think?

coffeegoesincodecomesout:
@jc-berger /lgtm

dockerymick (approved):

PR #6520, opened 2024-06-11
added another note about backup collision

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-12065


birsanv (approved):
lgtm ; thanks !

business_continuity/backup_restore/backup_schedule.adoc
```diff
@@ -162,6 +162,11 @@ Since the `BackupSchedule.cluster.open-cluster-management.io` resource is still
 +
 Since the `BackupSchedule.cluster.open-cluster-management.io` resource is still enabled on Hub2, it writes backups at the same storage location that corrupts the backup data. Any hub cluster restoring the latest backups from this location might use Hub2 data instead of Hub1 data. In this scenario, stopping Hub2 first or deleting the `BackupSchedule.cluster.open-cluster-management.io` resource on Hub2 before starting Hub1 fixes the backup collision issue.
 
+. The administrator tests a disaster scenario by making Hub2 a primary hub cluster, without stopping Hub1 first, causing the following conditions:
```
dockerymick: Can we use something different from Hub1 and Hub2? Can Hub1 be considered the initial hub cluster?

dockerymick (changes requested):

business_continuity/backup_restore/backup_schedule.adoc
birsanv: Hub1 is primary hub, Hub2 is passive hub

dockerymick: @birsanv thanks for clarifying, i recommend that we use those terms instead of Hub1 and Hub2

birsanv: thank you @dockerymick ; just note this is a part of a bullet section and Hub1, Hub2 has been used in the other bullets too. If we change this here we need to change in the other bullets I think

dockerymick: @birsanv I agree with changing the instances from Hub1 to **primary hub cluster** and Hub2 to **passive hub cluster**

dockerymick:
```suggestion
. The administrator tests a disaster scenario by making the passive hub cluster a primary hub cluster, without stopping the primary hub first, causing the following conditions: 
```

business_continuity/backup_restore/backup_schedule.adoc
```diff
@@ -162,6 +162,11 @@ Since the `BackupSchedule.cluster.open-cluster-management.io` resource is still
...
 
+. The administrator tests a disaster scenario by making Hub2 a primary hub cluster, without stopping Hub1 first, causing the following conditions: 
+- Hub1 is still running.
+- Hub1 backup data is restored on Hub2, including managed clusters backup. Therefore, Hub2 is now the active cluster.
+- Since the `BackupSchedule.cluster.open-cluster-management.io` resource is still enabled on Hub1, it writes backups at the same storage location which corrupts the backup data. For example, any hub cluster restoring the latest backups from this location might use Hub1 data instead of Hub2 data. To avoid data corruption, Hub1 `BackupSchedule` resource status automatically changes to `BackupCollision`. In this scenario, to avoid getting into this backup collision state, stop Hub1 first or delete the `BackupSchedule.cluster.open-cluster-management.io` resource on Hub1 before restoring managed clusters data on  Hub2.
```
dockerymick:
```suggestion
- Primary hub cluster is still running.
- Primary hub cluster backup data is restored on the passive hub cluster, including managed clusters backup. Therefore, the passive hub cluster is now the active cluster.
- Since the `BackupSchedule.cluster.open-cluster-management.io` resource is still enabled on the primary hub cluster, it writes backups at the same storage location which corrupts the backup data. For example, any hub cluster restoring the latest backups from this location might use the primary hub cluster data instead of the passive hub cluster data. To avoid data corruption, the primary hub cluster `BackupSchedule` resource status automatically changes to `BackupCollision`. In this scenario, to avoid getting into this backup collision state, stop the primary hub cluster first or delete the `BackupSchedule.cluster.open-cluster-management.io` resource on the primary hub cluster before restoring managed clusters data on the passive hub cluster.
```

business_continuity/backup_restore/backup_schedule.adoc
```diff
@@ -146,22 +146,27 @@ As a result, the latest backups are generated by a hub cluster that is no longer
...
 
 . The primary hub cluster fails unexpectedly, which is caused by the following conditions:
-- Communication from the primary hub cluster to Hub1 fails.
-- The Hub1 backup data is restored on a secondary hub cluster, called Hub2.
-- The administrator creates the `BackupSchedule.cluster.open-cluster-management.io` resource on Hub2, which is now the primary hub cluster and generates backup data to the common storage location.
-- Hub1 unexpectedly starts working again.
+- Communication from the primary hub cluster fails.
+- The primary hub cluster backup data is restored on a secondary hub cluster, called the passive hub cluster.
+- The administrator creates the `BackupSchedule.cluster.open-cluster-management.io` resource on the passive hub cluster, which is now the primary hub cluster and generates backup data to the common storage location.
+- The primary hub cluster unexpectedly starts working again.
```
birsanv: We are overloading the primary and passive cluster naming .. 
@dockerymick the reason we got with using names for the hubs like Hub1 , Hub2 was that they would change states, from primary becoming secondary and back again. So a hub can be primary and then becomes passive. So it's confusing to just call it primary or passive

If Hub1, Hub2 is not pallatable, can we use some other naming convention, but stay away from just saying primary or passive hub ?

dockerymick: Sure, what naming do you have in mind?

birsanv (approved):
lgtm

dockerymick (approved):

PR #6542, opened 2024-06-18
added known issue for MCO and CMO

jc-berger:
see Jira for this known issue: https://issues.redhat.com/browse/ACM-12213 

see Jira for related doc task: https://issues.redhat.com/browse/ACM-11724

### DIFF
--- a/release_notes/known_issues_observability.adoc
+++ b/release_notes/known_issues_observability.adoc
@@ -27,11 +27,11 @@ For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/o
 
 For more about deprecations and removals, see xref:../release_notes/deprecate_remove.adoc#deprecations-removals[Deprecations and removals].
 
-[#configurations-losing-value]
-== Configurations losing value
+[#configurations-default-value]
+== Configurations default value
 //2.11:12213
 
-The `MultiClusterObservability` (MCO) uses an earlier version of the Cluster Monitoring Operator (CMO). When MCO uses an old version of CMO, it can cause the following symptom: it prevents the `cluster-monitoring-config` config map from updating to the other configurations within {product-title-short} observability. The configurations then get reset to their default values in the CMO. 
+The `MultiClusterObservability` uses an earlier version of the Cluster Monitoring Operator. The earlier version prevents the `cluster-monitoring-config` config map from updating to the other configurations within observability. The configurations then get reset to their default values.  
 
 There is currently no workaround for this issue. 
 
--- a/release_notes/known_issues_observability.adoc
+++ b/release_notes/known_issues_observability.adoc
@@ -31,13 +31,9 @@ For more about deprecations and removals, see xref:../release_notes/deprecate_re
 == Configurations losing value
 //2.11:12213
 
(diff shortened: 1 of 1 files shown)


release_notes/known_issues_observability.adoc
```diff
@@ -27,6 +27,18 @@ For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/o
 
 For more about deprecations and removals, see xref:../release_notes/deprecate_remove.adoc#deprecations-removals[Deprecations and removals].
 
+[#configurations-losing-value]
+== Configurations losing value
+//2.11:12213
+
+The `MultiClusterObservability` (MCO) uses an old version of the Cluster Monitoring Operator (CMO). When MCO uses an old version of CMO, it can cause the following symptom: it prevents the `cluster-monitoring-config` config map from updating to the other configurations within {product-title-short} observability. When the CMO tries to turn the configurations into text, it causes the following symptom: the `cluster-monitoring-config` config maps might get reset to their type's value of `0`. For example, the `retentionSize` configuration gets a value of `0` and no longer exists in the current version of MCO.
+
```
swopebe, line 35: We don't typically do this for the resources in the doc: (MCO) 

You can remove those, use the resource name enough for the user but you don't have to repeat it a lot.

Please remove` "it causes the following symptom,` it's extra here. 

Just write concisely (example):

_The previous_ (take out the word old, can use _earlier_ as well) _version prevents the `cluster-monitoring-config` config map from updating to the other configurations within observability.

_ (You do not need to put RHACM here because that is understood...observability is a RHACM feature.)

swopebe: When the CMO tries to turn the configurations into text, it causes the following symptom: the `cluster-monitoring-config` config maps might get reset to their type's value of `0`.

Recommend:

When `MultiClusterObservability` configures to text, the `cluster-monitoring-config` config maps might reset `type` to a `0` value.

_Please remove possessives and as much anthropomorphism as possible. We don't use pronouns for the function, for example._

release_notes/known_issues_observability.adoc
```diff
@@ -27,6 +27,18 @@ For your {ocp} cluster, see link:https://access.redhat.com/documentation/en-us/o
...
+To workaround these symptoms, complete the following steps:
+
+. Create a local cluster within your {product-title-short} deployment.
+. Go to your CMO stack inside your `cluster-monitoring-configuration` config map.
```
swopebe: Don't need to say RHACM deployment, that is understood. Observability is a RHACM feature.

swopebe: Again, we normally don't abbreviate the resources, but if you feel strongly about this, can you show me where it is done within the text of our doc (not the code text, but the text) and we can discuss?

douglascamata (approved):
/lgtm

swopebe (changes requested):

release_notes/known_issues_observability.adoc
swopebe: please link the doc task to the PR. I think this is in our internal docs, as well, but you can do 

`git cm -m "url_for_doc_jira -- quick description" `before you `git push` to add your Doc Jira or do it after in the comments, but we link doc PRs to doc issues.

swopebe (approved):

PR #6569, opened 2024-06-25
added error message

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-12263


jc-berger:
Closing as we don't need to include the error message.

PR #6608, opened 2024-07-01
Swope acm mce discovery policy

swopebe:


clusters/acm_integration/integration_import_rosa.adoc
```diff
@@ -0,0 +1,193 @@
...
+
+. Install {product-title-short}. See the {product-title-short} link:../../install/install_overview.adoc#installing[Installing and upgrading] documentation.
+. Learn about Policies from the introduction to link:../../governance/grc_intro.adoc#governance[Governance].
+
+[#creating-rosa-policy]
+== Creating the automatic import policy
+
+The following policy and procedure is an example of how to import all your discovered {rosa} clusters automatically. 
+
+//To create the policy in the console, see Creating a cluster security policy from the console (link). By default this policy does not make changes to your cluster.
```
swopebe: @gparvin we commented this out and we didn't come back to it. In the meeting with Randy, you mentioned that we may want to lead the user to the console at some point without documenting the console. Do you still think we should add anything else at the beginning (or the end) for a console-only procedure?

gparvin: Let's stick with the CLI method we went through today.  My main concern was just that Randy's link (in the console) will point to the doc which describes the steps for the CLI.  While the same steps would work in the console, there would be a lot of effort at this point to get that in and the CLI doc is consistent with other similar procedures.   Thanks

clusters/acm_integration/integration_import_rosa.adoc
```diff
@@ -0,0 +1,193 @@
...
+. Learn about _Policies_ from the introduction to link:../../governance/grc_intro.adoc#governance[Governance] in the {product-title-short} documentation.
+
+[#creating-rosa-policy]
+== Creating the automatic import policy
+
+The following policy and procedure is an example of how to import all your discovered {rosa} clusters automatically. 
+
+//To create the policy in the console, see Creating a cluster security policy from the console (link). By default this policy does not make changes to your cluster. 
+
+Complete the following procedure by using the terminal:
```
gparvin: If there is a standard way to describe starting a CLI session with the ACM hub cluster then use that wording.  If this is the standard way then that's fine too.  I just wasn't sure if we clarified this is on the hub or if that's the assumed cluster.

clusters/acm_integration/integration_import_rosa.adoc
```diff
@@ -0,0 +1,193 @@
...
+subjects:
+- apiGroup: policy.open-cluster-management.io
+  kind: Policy
+  name: policy-rosa-autoimport
+----
+
+. To verify, run the following command:
+
+----
+oc get policy policy-rosa-autoimport
```
gparvin:
```suggestion
oc get policy policy-rosa-autoimport -n <namespace>
```

clusters/acm_integration/integration_import_rosa.adoc
swopebe: I will remove that comment, then. Thank you!

clusters/acm_integration/integration_import_rosa.adoc
swopebe: Can write: `Log in to to your hub cluster from the CLI` -- that is what I would normally add.

gparvin (approved):

clusters/acm_integration/acm_integration_intro.adoc
```diff
@@ -0,0 +1,12 @@
+[#acm-integration]
+= {product-title-short} integration
+
+If you are using {mce-short} with {product-title-short} installed, you can access more multicluster management features, such as _Observability_ and _Policy_.
+
+. Install {product-title-short}. See the {product-title-short} link:../../install/install_overview.adoc#installing[Installing and upgrading] documentation.
+. See link:../../install/adv_config_install.adoc#advanced-config-hub[MultiClusterHub advanced configuration] for details about {product-title-short} after you install.
```
oafischer, line 7: Are these prerequisites?

clusters/acm_integration/integration_import_rosa.adoc
```diff
@@ -0,0 +1,196 @@
...
+  kind: Placement
+  name: placement-policy-rosa-autoimport
+subjects:
+- apiGroup: policy.open-cluster-management.io
+  kind: Policy
+  name: policy-rosa-autoimport
+----
+
+. To verify, run the following command:
+
```
oafischer, line 193:
```suggestion

+
```

clusters/acm_integration/acm_integration_intro.adoc
oafischer, line 1: Even though this is already short, I would prefer if the name were the same as the other files. So just `integration_intro`

modules/common-attributes.adoc
```diff
@@ -13,6 +13,7 @@
 :ocp-virt-short: OpenShift Virtualization
 :olm: Operator Lifecycle Manager
 :ocm: OpenShift Cluster Manager
+:rosa: OpenShift Service on AWS
```
oafischer, line 16: Let's make the team aware of this as well, I know I have used this in some of my doc before so it will be nice to have a conref now. We might need a tech debt issue to go through the doc and replace spelled out versions with the conref.

oafischer (approved):
Some minor comments/questions/formatting suggestions. Otherwise looks great, thanks!

clusters/acm_integration/acm_integration_intro.adoc
swopebe: @oafischer this is the intro and we need to say this for every topic. So it's here because they need to know they cannot enter any topic without this knowledge. The install will be a prereq for all the topics in case they do not land on the intro.

modules/common-attributes.adoc
swopebe: @oafischer I was going to, for sure, but I didn't want anyone worried about find and replace right now; will add this to a tech debt issue, there are a couple I need to add/change.

clusters/acm_integration/acm_integration_intro.adoc
swopebe: Needs to have ACM integration because it is entered from MCE, so leaving this one as is. Within MCE, integration can mean anything and we want it clear to all that these are ACM integration files from MCE.

clusters/acm_integration/acm_integration_intro.adoc
oafischer: Got it, I haven't seen these listed as steps before so I was a little confused.

modules/common-attributes.adoc
dockerymick: Based on the [product list](https://docs.google.com/spreadsheets/d/1DLS_lS3VKidgZIvcLmLp9BoiqptkvqHWfe1D5FD2kfk/edit?gid=1259317633#gid=1259317633), the conref that you have is the approved short form. Should we also consider adding the full name (Red Hat OpenShift Service on AWS)? Then the short conref could be {rosa-short}: OpenShift Service on AWS

PR #6657, opened 2024-07-11
added section for migrating storage

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-12482

### DIFF
--- a/observability/observability_alerts.adoc
+++ b/observability/observability_alerts.adoc
@@ -152,6 +152,23 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
 amtool silence expire --alertmanager.url="http://localhost:9093" $(amtool silence query --alertmanager.url="http://localhost:9093" -q) 
 ----
 
+[#migrating-observability-storage]
+=== Migrating observability storage
+
+If you are using alert silencers, but you want to migrate your observability storage, complete the following steps:
+
+. In the `MultiClusterObservability`, set the field, `.spec.storageConfig.storageClass` to the new storage class that you chose. 
+. To ensure the data of the old `PersistentVolumes` is retained even when you delete the `PersistentVolumeClaim`, go to all your existing `PersistentVolumes` and change the `reclaimPolicy` to `"Retain": `oc patch pv <your-pv-name> -p '{"spec":{"persistentVolumeReclaimPolicy":"Retain"}}'`. 
++
+*Note:* Continuing to the next step, you have a risk of loosing ephemeral data. If you want to ensure you do not loose any data, see link:https://access.redhat.com/solutions/6922821[Migrate persistent data to another Storage Class in DG 8 Operator in OCP 4]. If you are unconcerned of losing ephemeral data, you can continue to the next step. 
+. Delete both the `StatefulSet` and the `PersistentVolumeClaim` in the following `StatefulSet` cases:
+.. `alertmanager-db-observability-alertmanager`
+.. `data-observability-thanos-compact`
+.. `data-observability-thanos-receive-default`
+.. `data-observability-thanos-store-shard`
+.. *Note:* You might need to delete then recreate the `MultiClusterObservability` Operator Pod so that you can create the new `StatefulSet`. 
+. Verify that the new `StatefulSet` and `PersistentVolumes` use the new `StorageClass` that you chose.  
+
 [#supress-alerts]
 == Suppressing alerts
 
--- a/observability/observability_alerts.adoc
+++ b/observability/observability_alerts.adoc
@@ -170,7 +170,7 @@ To migrate your PV storage, complete the following steps:
 .. `data-observability-thanos-compact`
 .. `data-observability-thanos-receive-default`
 .. `data-observability-thanos-store-shard`
-.. *Note:* You might need to delete then re-create the `MultiClusterObservability` operator pod so that you can create the new `StatefulSet`. 
+.. *Note:* You might need to delete, then re-create, the `MultiClusterObservability` operator pod so that you can create the new `StatefulSet`. 
 . Verify that the new `StatefulSet` and `PersistentVolumes` use the new `StorageClass` that you chose.  
 
 [#supress-alerts]
--- a/observability/observability_alerts.adoc
+++ b/observability/observability_alerts.adoc
@@ -152,21 +152,25 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
 amtool silence expire --alertmanager.url="http://localhost:9093" $(amtool silence query --alertmanager.url="http://localhost:9093" -q) 
 ----
 
-[#migrating-observability-storage]
-=== Migrating observability storage
+[#migrating-storage-pv]
+=== Migrating storage for Persistent Volumes used by observability pods
 
-If you are using alert silencers, but you want to migrate your observability storage, complete the following steps:
+When you use alert silencers, you can also migrate storage for the Persistent Volumes (PVs) used by the observability pods. 
 
-. In the `MultiClusterObservability`, set the field, `.spec.storageConfig.storageClass` to the new storage class that you chose. 
-. To ensure the data of the old `PersistentVolumes` is retained even when you delete the `PersistentVolumeClaim`, go to all your existing `PersistentVolumes` and change the `reclaimPolicy` to `"Retain": `oc patch pv <your-pv-name> -p '{"spec":{"persistentVolumeReclaimPolicy":"Retain"}}'`. 
-+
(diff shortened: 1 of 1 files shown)


observability/observability_alerts.adoc
```diff
@@ -152,6 +152,23 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
...
+. In the `MultiClusterObservability`, set the field, `.spec.storageConfig.storageClass` to the new storage class that you chose. 
+. To ensure the data of the old `PersistentVolumes` is retained even when you delete the `PersistentVolumeClaim`, go to all your existing `PersistentVolumes` and change the `reclaimPolicy` to `"Retain": `oc patch pv <your-pv-name> -p '{"spec":{"persistentVolumeReclaimPolicy":"Retain"}}'`. 
++
+*Note:* Continuing to the next step, you have a risk of loosing ephemeral data. If you want to ensure you do not loose any data, see link:https://access.redhat.com/solutions/6922821[Migrate persistent data to another Storage Class in DG 8 Operator in OCP 4]. If you are unconcerned of losing ephemeral data, you can continue to the next step. 
+. Delete both the `StatefulSet` and the `PersistentVolumeClaim` in the following `StatefulSet` cases:
+.. `alertmanager-db-observability-alertmanager`
+.. `data-observability-thanos-compact`
+.. `data-observability-thanos-receive-default`
+.. `data-observability-thanos-store-shard`
+.. *Note:* You might need to delete then recreate the `MultiClusterObservability` Operator Pod so that you can create the new `StatefulSet`.
```
swopebe: missing punctuation line 169

observability/observability_alerts.adoc
```diff
@@ -152,6 +152,23 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
...
 
+[#migrating-observability-storage]
+=== Migrating observability storage
+
+If you are using alert silencers, but you want to migrate your observability storage, complete the following steps:
+
+. In the `MultiClusterObservability`, set the field, `.spec.storageConfig.storageClass` to the new storage class that you chose. 
+. To ensure the data of the old `PersistentVolumes` is retained even when you delete the `PersistentVolumeClaim`, go to all your existing `PersistentVolumes` and change the `reclaimPolicy` to `"Retain": `oc patch pv <your-pv-name> -p '{"spec":{"persistentVolumeReclaimPolicy":"Retain"}}'`. 
++
+*Note:* Continuing to the next step, you have a risk of loosing ephemeral data. If you want to ensure you do not loose any data, see link:https://access.redhat.com/solutions/6922821[Migrate persistent data to another Storage Class in DG 8 Operator in OCP 4]. If you are unconcerned of losing ephemeral data, you can continue to the next step.
```
swopebe: losing is misspelled

swopebe: A better word here for simplicity, replace ephemeral?

Please write these, just in general, in the clearer format:

**set the `.spec.storageConfig.storageClass` field.** line 160. 

Remove the need for comma by just getting straight to the point. This is a general comment for all the doc.

Line 161 seems off, like a longer command that we should separate?

observability/observability_alerts.adoc
```diff
@@ -152,6 +152,23 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
...
++
+*Note:* Continuing to the next step, you have a risk of loosing ephemeral data. If you want to ensure you do not loose any data, see link:https://access.redhat.com/solutions/6922821[Migrate persistent data to another Storage Class in DG 8 Operator in OCP 4]. If you are unconcerned of losing ephemeral data, you can continue to the next step. 
+. Delete both the `StatefulSet` and the `PersistentVolumeClaim` in the following `StatefulSet` cases:
+.. `alertmanager-db-observability-alertmanager`
+.. `data-observability-thanos-compact`
+.. `data-observability-thanos-receive-default`
+.. `data-observability-thanos-store-shard`
+.. *Note:* You might need to delete then recreate the `MultiClusterObservability` Operator Pod so that you can create the new `StatefulSet`. 
+. Verify that the new `StatefulSet` and `PersistentVolumes` use the new `StorageClass` that you chose.  
+
```
swopebe, line 175: double check capitalization here with **pod**, I am not sure off the top of my head how we treat that in the doc.

observability/observability_alerts.adoc
swopebe: Notes are not steps, but they can append a step if needed. Since you have this built as a step, either can make it imperative or make it a note.

IMO: deleting a pod to recreate a stateful set seems to me like more than a note...how do they know when or if they need to do this?

swopebe (changes requested):

observability/observability_alerts.adoc
swopebe: loose? (I realize this may be in the drafts from dev, but that is why we have to rewrite our drafts almost all the time -- especially since they don't always have an English background.)

Please rewrite this for conciseness:

*Note:* Continuing to the next step, you have a risk of loosing ephemeral data. If you want to ensure you do not loose any data, see....

This should go _with_ the next step, we don't need to preannounce that they are continuing. Simply add it as part of the next step and remove filler words, there are a bunch here:

Example rewrite: 

**To reduce the risk of losing data, see ....**

observability/observability_alerts.adoc
swopebe: Other things: We should be capitalizing Observability as a major component of our product. We need to NOT use the word old or older, remember this sends negative connotation to customers about the version they are on or were on. (Previous, earlier)

swopebe (changes requested):

douglascamata (commented):
In general I think this documentation has to be very clear in stating that it talks about the storage for the PVs (Persistent Volumes) used by the Observability pods.

It cannot be confused with the object storage that is used to store the metrics collected from the clusters.

douglascamata (approved):
/lgtm

observability/observability_alerts.adoc
```diff
@@ -152,6 +152,27 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
...
+. In the `MultiClusterObservability`, set the `.spec.storageConfig.storageClass` field to the new storage class. 
+. To ensure the data of the earlier `PersistentVolumes` is retained even when you delete the `PersistentVolumeClaim`, go to all your existing `PersistentVolumes`.
+. Change the `reclaimPolicy` to `"Retain": `oc patch pv <your-pv-name> -p '{"spec":{"persistentVolumeReclaimPolicy":"Retain"}}'`. 
+. *Optional:* To avoid losing data, see link:https://access.redhat.com/solutions/6922821[Migrate persistent data to another Storage Class in DG 8 Operator in OCP 4].
+. Delete both the `StatefulSet` and the `PersistentVolumeClaim` in the following `StatefulSet` cases:
+.. `alertmanager-db-observability-alertmanager`
+.. `data-observability-thanos-compact`
+.. `data-observability-thanos-receive-default`
+.. `data-observability-thanos-store-shard`
+.. *Note:* You might need to delete then re-create the `MultiClusterObservability` operator pod so that you can create the new `StatefulSet`.
```
oafischer:
```suggestion
.. *Note:* You might need to delete, then re-create, the `MultiClusterObservability` operator pod so that you can create the new `StatefulSet`. 
```

oafischer (approved):
One small suggestion, otherwise looks good

observability/observability_alerts.adoc
```diff
@@ -152,6 +152,27 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
...
+
+When you use alert silencers, you can also migrate storage for the Persistent Volumes (PVs) used by the observability pods. 
+
+*Note:* The storage for PVs is different from the object storage used to store the metrics collected from your clusters.  
+
+To migrate your PV storage, complete the following steps:
+
+. In the `MultiClusterObservability`, set the `.spec.storageConfig.storageClass` field to the new storage class. 
+. To ensure the data of the earlier `PersistentVolumes` is retained even when you delete the `PersistentVolumeClaim`, go to all your existing `PersistentVolumes`.
+. Change the `reclaimPolicy` to `"Retain": `oc patch pv <your-pv-name> -p '{"spec":{"persistentVolumeReclaimPolicy":"Retain"}}'`.
```
cadawson, line 166: @jc-berger Can you take a quick look at the tics in line 165 and make sure that they look ok? Looks like you might have one extra, but I might be wrong. Thanks!

jc-berger: @cadawson thanks, Chris! I think we're good. I'm looking at the file in VS Code, as that's a bit easier to see the tics, and it looks good to me: 

Screenshot 2024-07-24 at 10 54 29 AM

PR #6661, opened 2024-07-12
https://issues.redhat.com/browse/ACM-9830--editing versions or removing

swopebe:


clusters/cluster_lifecycle/addon_managed_service.adoc
```diff
@@ -16,7 +16,7 @@ A `TokenRequest` is made with the `ServiceAccount` on the managed cluster to the
 [#serviceaccount_prereqs]
 == Prerequisites
 
-- {ocp} version 4.13 or later must be deployed in your environment, and you must be logged in with the command line interface (CLI).
+- You need a supported {ocp} environment.
 - You need the {mce-short} installed.
```
swopebe, line 20: We don't have to say the CLI unless the user needs a different CLI than what the product or doc uses: `oc`

They are running oc commands, it is obvious. 

We also don't need to say 4.13 or later. OCP supports (as of now) N-2. When we do this, someone has to stop and think: Does this change this release? 

Only in certain cases, like there is a problem with the current released version, do we need to say this.

clusters/cluster_lifecycle/ansible_config_hosted_cluster.adoc
```diff
@@ -15,7 +15,7 @@
 
 You must meet the following prerequisites to run Automation templates on your clusters:
 
-* {ocp-short} 4.14 or later
+* A supported version of {ocp-short}
```
swopebe, line 18: 4.14 and later IS what is supported, per the support matrix, which reduces the requirement of us having to say this.

oafischer (approved):

PR #6680, opened 2024-07-16
https://issues.redhat.com/browse/ACM-11809 Unbind non bm troubleshoot

oafischer:
2.11 and 2.10


clusters/support_troubleshooting/trouble_unbind_non_bm.adoc
```diff
@@ -0,0 +1,20 @@
...
+For managed clusters without a `BareMetalHost`, removing cluster information for a late binding cluster does not unbind bare metal nodes with late binding.
+
+[#resolving-unbind-non-bm-late-binding]
+== Resolving the problem: Unbinding non bare metal nodes with late binding
+
+To unbind the non bare metal nodes with late binding on Nutanix or VMware vSphere, complete the following steps:
+
+. Remove the cluster information. See xref:../cluster_lifecycle/remove_managed_cluster.adoc#remove-managed-cluster[Removing a cluster from management] to learn more.
+
+. Delete the root disks.
```
trewest: I think `Clean the root disk` is more appropriate

trewest:
FYI @carbonin

clusters/support_troubleshooting/trouble_unbind_non_bm.adoc
```diff
@@ -0,0 +1,20 @@
+[#troubleshooting-unbind-non-bm-late-binding]
+= Troubleshooting: Unbinding non bare metal nodes with late binding
+
+If you are using the Nutanix or VMware vSphere platform, you must complete additional manual steps to destroy a late binding cluster and return the nodes back to the Discovery ISO. If you are using Zero Touch Provisioning (ZTP), the nodes are automatically returned to the Discovery ISO.
+
+[#symptom-unbind-non-bm-late-binding]
+== Symptom: Unbinding non bare metal nodes with late binding
+
+For managed clusters without a `BareMetalHost`, removing cluster information for a late binding cluster does not unbind bare metal nodes with late binding.
```
trewest: > removing cluster information for a late binding cluster does not unbind bare metal nodes with late binding

Can you elaborate on what this means a bit more? I would think it should say something like
'''
For late binding managed clusters without `BareMetalHosts`, removing cluster information does not automatically return all nodes to the Discovery ISO. 
'''

clusters/support_troubleshooting/trouble_unbind_non_bm.adoc
```diff
@@ -0,0 +1,20 @@
+[#troubleshooting-unbind-non-bm-late-binding]
+= Troubleshooting: Unbinding non bare metal nodes with late binding
+
+If you are using the Nutanix or VMware vSphere platform, you must complete additional manual steps to destroy a late binding cluster and return the nodes back to the Discovery ISO. If you are using Zero Touch Provisioning (ZTP), the nodes are automatically returned to the Discovery ISO.
```
trewest:
```suggestion
If you are using the Nutanix or VMware vSphere platforms or you are booting baremetal nodes without `BareMetalHosts`, you must complete additional manual steps to destroy a late binding cluster and return the nodes back to the Discovery ISO. If you are using Zero Touch Provisioning (ZTP), the nodes are automatically returned to the Discovery ISO.
```

clusters/support_troubleshooting/trouble_unbind_non_bm.adoc
```diff
@@ -0,0 +1,20 @@
+[#troubleshooting-unbind-non-bm-late-binding]
+= Troubleshooting: Unbinding non bare metal nodes with late binding
+
```
trewest, line 3: I just want to point out that we should be explicit that this documentation is only relevant to scenarios where an end-user want to completely destroy their cluster and return their nodes to a late-binding pool. There are still supported scenarios for removing individual workers while the cluster is still operational.

clusters/support_troubleshooting/trouble_unbind_non_bm.adoc
carbonin: I don't think this is specific to vsphere at all.

Any situation where you unbind an entire cluster at once and you don't have BMHs will run into this scenario.

carbonin: >If you are using Zero Touch Provisioning (ZTP)

This is also not true.
ZTP never uses late binding so the concept of "unbinding" doesn't really apply.
Also ZTP always uses BMH, so it's probably not worth calling out here at all.

oafischer:
Made some updates @carbonin @trewest. Let me know if we need any other changes.

clusters/support_troubleshooting/trouble_unbind_non_bm.adoc
```diff
@@ -0,0 +1,20 @@
+[#troubleshooting-unbind-non-bm-late-binding]
+= Troubleshooting: Unbinding non bare metal nodes with late binding
```
trewest: I think  `Unbinding non bare metal nodes` could be a little bit misleading. I think something like "Returning non-bm cluster to late-binding pool" might be more appropriate.  @carbonin Does that sound more accurate to you?

carbonin: Yeah because this is specific to removing the entire cluster (as you mentioned in the next comment)

clusters/support_troubleshooting/trouble_unbind_non_bm.adoc
```diff
@@ -0,0 +1,20 @@
...
+
+[#symptom-unbind-non-bm-late-binding]
+== Symptom: Unbinding non bare metal nodes with late binding
+
+For late binding managed clusters without `BareMetalHosts`, removing cluster information does not automatically return all nodes to the Discovery ISO.
+
+[#resolving-unbind-non-bm-late-binding]
+== Resolving the problem: Unbinding non bare metal nodes with late binding
+
+To unbind the non bare metal nodes with late binding on Nutanix or VMware vSphere, complete the following steps:
```
carbonin: I'm not sure it's worth calling out the platforms here.

oafischer:
Updated, ready for another review @trewest @carbonin . Thanks!

carbonin (approved):

trewest (approved):
/lgtm

swopebe (approved):
looks great

PR #6693, opened 2024-07-17
making another attempt to fix Clusters book

dockerymick:


cadawson (approved):
Looks good. Let's hope it works.

PR #6714, opened 2024-07-24
added context and extra steps

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-12876

### DIFF
--- a/observability/observability_alerts.adoc
+++ b/observability/observability_alerts.adoc
@@ -155,7 +155,7 @@ amtool silence expire --alertmanager.url="http://localhost:9093" $(amtool silenc
 [#migrating-observability-storage]
 === Migrating observability storage
 
-If you use alert silencers, you can migrate observability storage while retaining the silencers from its earlier state. To do this, migrate your {product-title-short} observability storage by creating new `StatefulSets` and `Persistent Volumes` (PV)s that use your chosen `StorageClass`. This way, you have more control over where to store your data and how to use it during storage. 
+If you use alert silencers, you can migrate observability storage while retaining the silencers from its earlier state. To do this, migrate your {product-title-short} observability storage by creating new `StatefulSets` and `Persistent Volumes` (PV)s that use your chosen `StorageClass`. 
 
 *Note:* The storage for PVs is different from the object storage used to store the metrics collected from your clusters.  
 
--- a/observability/observability_alerts.adoc
+++ b/observability/observability_alerts.adoc
@@ -161,7 +161,7 @@ If you use alert silences, you should migrate your {product-title-short} observa
 
 When you use `StatefulSets` and PVs to migrate your observability storage, it stores the following data components:
 
-- *Observatorium:* Uploads its data to the object storage. Other component areas do not store data that can be lost; instead, they automatically regenerate what they need during a reboot. 
+- *Observatorium/Thanos:* Uploads the data it ingests to object storage. Some of its components do store data in PVs, but this data is usually automatically regenerated on startup, and thus can be lost without harm. 
 - *Alertmanager:* Only stores silenced alerts. If you want to keep these silenced alerts, you must migrate that data to the new PV. 
 
 To migrate your observability storage, complete the following steps:
--- a/observability/observability_alerts.adoc
+++ b/observability/observability_alerts.adoc
@@ -170,7 +170,7 @@ To migrate your observability storage, complete the following steps:
 . To ensure the data of the earlier `PersistentVolumes` is retained even when you delete the `PersistentVolumeClaim`, go to all your existing `PersistentVolumes`.
 . Change the `reclaimPolicy` to `"Retain": `oc patch pv <your-pv-name> -p '{"spec":{"persistentVolumeReclaimPolicy":"Retain"}}'`. 
 . *Optional:* To avoid losing data, see link:https://access.redhat.com/solutions/6922821[Migrate persistent data to another Storage Class in DG 8 Operator in OCP 4].
-. Delete both the `StatefulSet` and the `PersistentVolumeClaim` in the following `StatefulSet` cases:
+. Delete both the `StatefulSet` and the `PersistentVolumeClaim` in the following `StatefulSet` resources:
 .. `alertmanager-db-observability-alertmanager`
 .. `data-observability-thanos-compact`
 .. `data-observability-thanos-receive-default`
--- a/observability/observability_alerts.adoc
+++ b/observability/observability_alerts.adoc
@@ -155,7 +155,7 @@ amtool silence expire --alertmanager.url="http://localhost:9093" $(amtool silenc
 [#migrating-observability-storage]
 === Migrating observability storage
 
-When you use alert silencers, you can migrate your {product-title-short} observability storage by creating new `StatefulSets` and `Persistent Volumes` to use your chosen `StorageClass`. By creating new `StatefulSets` and PVs, you have more control over where your data is stored and how it is used in storage.
+If you use alert silences, you should migrate your {product-title-short} observability storage by creating new `StatefulSets` and `Persistent Volumes` that use your chosen `StorageClass`, to retain these silences. By creating new `StatefulSets` and PVs, you have more control over where your data is stored and how it is used in storage.
 
 *Note:* The storage for PVs is different from the object storage used to store the metrics collected from your clusters.  
 
--- a/observability/observability_alerts.adoc
+++ b/observability/observability_alerts.adoc
@@ -155,7 +155,7 @@ amtool silence expire --alertmanager.url="http://localhost:9093" $(amtool silenc
 [#migrating-observability-storage]
 === Migrating observability storage
 
-If you use alert silencers, you can migrate observability storage while retaining the silencers from its earlier state. To do this, migrate your {product-title-short} observability storage by creating new `StatefulSets` and `Persistent Volumes` (PV)s that use your chosen `StorageClass`. 
+If you use alert silencers, you can migrate observability storage while retaining the silencers from its earlier state. To do this, migrate your {product-title-short} observability storage by creating new `StatefulSets` and `PersistentVolumes` (PV) resources that use your chosen `StorageClass` resource. 
 
 *Note:* The storage for PVs is different from the object storage used to store the metrics collected from your clusters.  
 
--- a/observability/observability_alerts.adoc
+++ b/observability/observability_alerts.adoc
@@ -152,14 +152,20 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
 amtool silence expire --alertmanager.url="http://localhost:9093" $(amtool silence query --alertmanager.url="http://localhost:9093" -q) 
 ----
 
-[#migrating-storage-pv]
-=== Migrating storage for Persistent Volumes used by observability pods
+[#migrating-observability-storage]
+=== Migrating observability storage
 
-When you use alert silencers, you can also migrate storage for the Persistent Volumes (PVs) used by the observability pods. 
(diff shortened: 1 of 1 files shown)


observability/observability_alerts.adoc
```diff
@@ -152,14 +152,20 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
 amtool silence expire --alertmanager.url="http://localhost:9093" $(amtool silence query --alertmanager.url="http://localhost:9093" -q) 
 ----
 
-[#migrating-storage-pv]
-=== Migrating storage for Persistent Volumes used by observability pods
+[#migrating-observability-storage]
+=== Migrating observability storage
 
-When you use alert silencers, you can also migrate storage for the Persistent Volumes (PVs) used by the observability pods. 
+When you use alert silencers, you can migrate your {product-title-short} observability storage by creating new `StatefulSets` and `Persistent Volumes` to use your chosen `StorageClass`. By creating new `StatefulSets` and PVs, you have more control over where your data is stored and how it is used in storage.
```
jc-berger: SME/Dev Reviewers: 

Looking at the Jira, there seemed to be context and content regarding StatefulSets and PVs that we didn't originally have in the intro. 

I tried to capture the relationship between PVs and StatefulSets, and how them helps users migrate their storage. 

I'm happy to keep this PR simple, and simply include some extra steps, like the Jira asked. 

With that said, SME reviewers, please let me know what you think of this expanded introduction and if it helps users determine why they would want to migrate observability storage and complete these procedure steps....especially since the procedure is kind of long, I think some context like "why this helps you" would be beneficial to the user.

saswatamcode: So I think the meaning here is slightly different
```suggestion
If you use alert silences, you should migrate your {product-title-short} observability storage by creating new `StatefulSets` and `Persistent Volumes` that use your chosen `StorageClass`, to retain these silences. By creating new `StatefulSets` and PVs, you have more control over where your data is stored and how it is used in storage.
```

saswatamcode: @jc-berger I'm not sure if the "By creating new `StatefulSets` and PVs, you have more control over where your data is stored and how it is used in storage." line is really relevant in this context?

observability/observability_alerts.adoc
```diff
@@ -152,14 +152,20 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
...
 
 *Note:* The storage for PVs is different from the object storage used to store the metrics collected from your clusters.  
 
-To migrate your PV storage, complete the following steps:
+When you use `StatefulSets` and PVs to migrate your observability storage, it stores the following data components:
+
+- *Observatorium:* Uploads its data to the object storage. Other component areas do not store data that can be lost; instead, they automatically regenerate what they need during a reboot. 
+- *Alertmanager:* Only stores silenced alerts. If you want to keep these silenced alerts, you must migrate that data to the new PV. 
+- *Metrics collector:* Deploys as deployments on the managed clusters.
```
saswatamcode: I don't think metrics collectors really use PV?
```suggestion
```

observability/observability_alerts.adoc
```diff
@@ -152,14 +152,20 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
...
 
-When you use alert silencers, you can also migrate storage for the Persistent Volumes (PVs) used by the observability pods. 
+When you use alert silencers, you can migrate your {product-title-short} observability storage by creating new `StatefulSets` and `Persistent Volumes` to use your chosen `StorageClass`. By creating new `StatefulSets` and PVs, you have more control over where your data is stored and how it is used in storage.
 
 *Note:* The storage for PVs is different from the object storage used to store the metrics collected from your clusters.  
 
-To migrate your PV storage, complete the following steps:
+When you use `StatefulSets` and PVs to migrate your observability storage, it stores the following data components:
+
+- *Observatorium:* Uploads its data to the object storage. Other component areas do not store data that can be lost; instead, they automatically regenerate what they need during a reboot.
```
saswatamcode: Feel free to change my writing style, but the meaning is slightly different here.
```suggestion
- *Observatorium/Thanos:* Uploads the data it ingests to object storage. Some of its components do store data in PVs, but this data is usually automatically regenerated on startup, and thus can be lost without harm. 
```

observability/observability_alerts.adoc
```diff
@@ -152,14 +152,20 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
...
+[#migrating-observability-storage]
+=== Migrating observability storage
 
-When you use alert silencers, you can also migrate storage for the Persistent Volumes (PVs) used by the observability pods. 
+When you use alert silencers, you can migrate your {product-title-short} observability storage by creating new `StatefulSets` and `Persistent Volumes` to use your chosen `StorageClass`. By creating new `StatefulSets` and PVs, you have more control over where your data is stored and how it is used in storage.
 
 *Note:* The storage for PVs is different from the object storage used to store the metrics collected from your clusters.  
 
-To migrate your PV storage, complete the following steps:
+When you use `StatefulSets` and PVs to migrate your observability storage, it stores the following data components:
```
saswatamcode: I don't think we are really using these to migrate storage, but rather migrating to new storage.

observability/observability_alerts.adoc
saswatamcode, line 1: Is this the correct doc to include information on migration of PVs for MCO? I think it would make sense to add this to some storage doc instead of alerts?

observability/observability_alerts.adoc
```diff
@@ -171,6 +177,8 @@ To migrate your PV storage, complete the following steps:
 .. `data-observability-thanos-receive-default`
 .. `data-observability-thanos-store-shard`
 .. *Note:* You might need to delete, then re-create, the `MultiClusterObservability` operator pod so that you can create the new `StatefulSet`. 
+. Re-create a new `PersistentVolumeClaim` with the same name but the correct `StorageClass`. 
+. Create a new `PersistentVolumeClaim` referring to the old `PersistentVolume`.
```
saswatamcode, line 180: Not entirely sure if this is correct. Maybe @thibaultmg you remember this support case?

saswatamcode (commented):
Hi! Thanks for doing this some comments.

observability/observability_alerts.adoc
jc-berger: @saswatamcode thanks for the awesome feedback! So, the original Jira that wanted to include these steps asked for them to be included in this doc: https://issues.redhat.com/browse/ACM-12482 

Before we find a new doc to move this content to, are you sure the MCO storage content is out of place here under the Alert Silencers section?

observability/observability_alerts.adoc
jc-berger: Yes, happy to make any changes to the steps!

observability/observability_alerts.adoc
jc-berger: @saswatamcode I'm happy to remove it if it's not helpful to the user. 

With that said, we do want to include a statement including the "benefit" or "why would I do this" "what's in it for me" before we give the user a procedure to complete. Especially since this procedure is kind of long and complicated. 

Am I incorrect that migrating the observability storage gives users more control over where the data is stored and how it's used in storage?

observability/observability_alerts.adoc
saswatamcode: Yup, I think this deserves it's own section, I wouldn't really look for how to migrate storage when reading up on alerting 🙂

observability/observability_alerts.adoc
saswatamcode: Hmm, so what are the potential reasons for migrating PV storage?
The most common one IMO is moving to a larger PV. A less common one would be that I need to change the storage backend of my PVs. 
However in any case, it won't really give them more control on where it is stored or how it is used.

Thanos will use PVs in the way we intended upstream. And the PVs for ACM are always owned by customer, we don't own or manage their PVs, so they have control over it.

This is just a migration.

observability/observability_alerts.adoc
jc-berger: Gellert approved the PR offline, so these steps are good!

observability/observability_alerts.adoc
```diff
@@ -152,14 +152,19 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
 amtool silence expire --alertmanager.url="http://localhost:9093" $(amtool silence query --alertmanager.url="http://localhost:9093" -q) 
 ----
 
-[#migrating-storage-pv]
-=== Migrating storage for Persistent Volumes used by observability pods
+[#migrating-observability-storage]
+=== Migrating observability storage
 
-When you use alert silencers, you can also migrate storage for the Persistent Volumes (PVs) used by the observability pods. 
+If you use alert silencers, you can migrate observability storage while retaining the silencers from its earlier state. To do this, migrate your {product-title-short} observability storage by creating new `StatefulSets` and `Persistent Volumes` (PV)s that use your chosen `StorageClass`.
```
dockerymick:
```suggestion
If you use alert silencers, you can migrate observability storage while retaining the silencers from its earlier state. To do this, migrate your {product-title-short} observability storage by creating new `StatefulSets` and `PersistentVolumes` (PV) resources that use your chosen `StorageClass` resource. 
```

observability/observability_alerts.adoc
```diff
@@ -152,14 +152,19 @@ amtool silence expire --alertmanager.url="http://localhost:9093" "d839aca9-ed46-
...
 
-When you use alert silencers, you can also migrate storage for the Persistent Volumes (PVs) used by the observability pods. 
+If you use alert silencers, you can migrate observability storage while retaining the silencers from its earlier state. To do this, migrate your {product-title-short} observability storage by creating new `StatefulSets` and `Persistent Volumes` (PV)s that use your chosen `StorageClass`. 
 
 *Note:* The storage for PVs is different from the object storage used to store the metrics collected from your clusters.  
 
-To migrate your PV storage, complete the following steps:
+When you use `StatefulSets` and PVs to migrate your observability data to new storage, it stores the following data components:
+
+- *Observatorium or Thanos:* Receives data then uploads it to object storage. Some of its components store data in PVs. For this data, it usually automatically regenerates it on a startup, so there is no consequence if you lose this data.
```
dockerymick: it is not clear to me what "it" is. Is it the component of the object storage? When you say "usually" is that default behavior or are there situations where the object storage will not automatically regenerate 

```suggestion
- *Observatorium or Thanos:* Receives data then uploads it to object storage. Some of its components store data in PVs. For this data, the Observatorium or Thanos automatically regenerates the object storage on a startup, so there is no consequence if you lose this data.  
```

observability/observability_alerts.adoc
```diff
@@ -171,6 +176,8 @@ To migrate your PV storage, complete the following steps:
 .. `data-observability-thanos-receive-default`
 .. `data-observability-thanos-store-shard`
 .. *Note:* You might need to delete, then re-create, the `MultiClusterObservability` operator pod so that you can create the new `StatefulSet`.
```
dockerymick: Even though this is not a part of your changes, I would change this to Important instead of Note. Especially if it is a recommended step. Throughout the doc, we don't write "Note" before a step. This is the only instance. Alternatively, we can change it to "Optional"
```suggestion
.. *Important:* You might need to delete, then re-create, the `MultiClusterObservability` operator pod so that you can create the new `StatefulSet`. 
```

observability/observability_alerts.adoc
```diff
@@ -171,6 +176,8 @@ To migrate your PV storage, complete the following steps:
 .. `data-observability-thanos-receive-default`
```
dockerymick, line 176: We list say that these are "cases" but I think its best to say "resources". So should we say something like:

 Delete both the `StatefulSet` and the `PersistentVolumeClaim` in the following `StatefulSet` resources:

dockerymick (changes requested):

dockerymick (approved):

PR #6759, opened 2024-08-06
https://issues.redhat.com/browse/ACM-12873

swopebe:


modules/common-attributes.adoc
```diff
@@ -1,9 +1,7 @@
-:addons: klusterlet add-ons
 :aap: Red Hat Ansible Automation Platform
 :aap-short: Ansible Automation Platform
 :assist-install: Infrastructure Operator for Red Hat OpenShift
-:cincinnati: Red Hat OpenShift Update Service
-:cincinnati-short: OpenShift Update Service
+:update-service: OpenShift Update Service
```
swopebe, line 4: This is not in the product naming guidance, and only used in short from what I can tell in OCP docs; let's just use the one. Further, Cincinnati is the protocol for the Service, so we should use a conref that represents the service name. I see no valid abbreviation so going with this.

oafischer (approved):

PR #6763, opened 2024-08-07
https://issues.redhat.com/browse/ACM-12870 Fix nmstate example

oafischer:
Also for 2.12 and 2.10

### DIFF
--- a/clusters/cluster_lifecycle/cim_network.adoc
+++ b/clusters/cluster_lifecycle/cim_network.adoc
@@ -30,7 +30,7 @@ metadata:
   labels:
     some-key: <some-value>
 spec:
-  config:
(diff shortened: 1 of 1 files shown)


carbonin:
This doesn't make sense to me.
This doesn't match the API in https://github.com/openshift/assisted-service/blob/master/api/v1beta1/nmstate_config_types.go#L42-L56

I don't think this is a valid change.

carbonin (changes requested):
This doesn't make sense to me.
This doesn't match the API in https://github.com/openshift/assisted-service/blob/master/api/v1beta1/nmstate_config_types.go#L42-L56

I don't think this is a valid change.

PR #6783, opened 2024-08-13
https://issues.redhat.com/browse/ACM-13074

dockerymick:


console/console.adoc
```diff
@@ -21,9 +21,9 @@ image:../images/nav-2.7.png[Navigation]
 
 From the {acm} _Home_ page, you can access more information and you can search across the product. 
 
-Click *Welcome* for introductory information about each product function. Click *Overview* to see updates, such as _Application_ status, or to access clickable _Cluster_ percentage values for policy violations and more.
+Click *Welcome* for introductory information about each product function. Click *Overview* (Deprecated) to see updates, such as _Application_ status, or to access clickable _Cluster_ percentage values for policy violations and more.
```
zlayne: The classic overview layout is deprecated. I want to be clear to the user that the Overview page is not going away, rather the classic view is.
To me, the label here makes me think the entire overview page is going away? 
Same with the below - https://github.com/stolostron/rhacm-docs/pull/6783/files#diff-3e0b11e3da84fe002d57a0370949fd7aa09374fb658699ef3e5ad5d044278b72R47

release_notes/deprecate_remove.adoc
```diff
@@ -46,6 +46,12 @@ A _deprecated_ component, feature, or service is supported, but no longer recomm
 |===
 | Product or category | Affected item | Version | Recommended action | More details and links
 
+| _Overview_ page
+| {acm} search 
+| 2.12
+| Enable the _Fleet view_ switch to view the new default overview page.
+| The layout of the {acm-short} search _Overview_ page is deprecated.
```
dockerymick:
```suggestion
| The previous layout of the {acm-short} _Overview_ page is deprecated.
```

console/console.adoc
dockerymick: Gotchu, so what is now on the Fleet view page?

zlayne: Right what is now on Fleet view page is what will be the default Overview page in 2.13 - and it will no longer include a toggle switch.

release_notes/deprecate_remove.adoc
zlayne: I think just need to remove the reference to search here.
`The previous layout of the {acm-short} _Overview_ page is deprecated.`

console/console.adoc
dockerymick: @zlayne what updates will the user be able to see from the Fleet view? Will it be different from what they see in the Overview page? If it will be default in 2.13, using the deprecated label will be best bc the Overview will still be available. Deprecated means that it is available but no longer supported

zlayne: Nothing will change on the Fleet view layout. The classic layout will have the following
Screenshot 2024-08-13 at 2 30 17 PM

swopebe: "classic" and  "fleet view" are not used in the doc that I know of. What we need to know is what do we need to flag in the doc if not this?

swopebe: https://github.com/stolostron/rhacm-docs/blob/2.12_stage/console/console.adoc#console-overview (this is the description of the console)

zlayne: The purpose of this change is to deprecate the "classic/old overview layout". The Overview page itself is not being deprecated. Therefore I don't think it is correct to add the deprecated label next to the "Overview" text here or in `insights_intro` doc. 

If "classic/old layout" is not mentioned anywhere in the docs, perhaps no deprecation label is needed and we only need to add the entry in the `deprecate_remove.adoc` doc?

dockerymick: We might not need to add it to deprecations at all. It seems like we just need to explain the new view in the docs, instead of adding a feature deprecation notice. We can still mention it in what's new, but if it's not being deprecated then maybe the flag from the console is not necessary

zlayne: I will keep the deprecation label in the console itself, so users are made aware the old overview layout will no longer be available in 2.13. 
I'll let you decide on what documentation is best (what's new v. release note deprecation), but the `(Deprecated)` text should be removed from Overview page references as the Overview page itself is not going away.

observability/insights_intro.adoc
```diff
@@ -44,7 +44,7 @@ oc get policyreport --all-namespaces
 You can view the identified issues on a specific cluster. Complete the following steps:
 
 . Log in to your {acm-short} cluster.
-. Select *Overview* from the navigation menu. 
+. Select *Overview* (Deprecated) from the navigation menu.
```
dockerymick:
```suggestion
. Select *Overview* from the navigation menu. 
```

console/console.adoc
```diff
@@ -21,9 +21,9 @@ image:../images/nav-2.7.png[Navigation]
...
 
-From the _Overview_ page, you can view the following information:
+From the _Overview_ page (Deprecated), you can view the following information:
```
dockerymick:
```suggestion
From the _Overview_ page, you can view the following information:
```

console/console.adoc
dockerymick:
```suggestion
Click *Welcome* for introductory information about each product function. Click *Overview* to see updates such as, _Application_ status or to access clickable _Cluster_ percentage values for policy violations and more.
```

zlayne:
@dockerymick Hopefully this image isn't too zoomed out - was trying to capture all page content for fleet view. Let me know if you have any other questions.
Screenshot 2024-08-27 at 10 40 42 AM

dockerymick:
@zlayne thanks! This works!

console/console.adoc
```diff
@@ -33,7 +33,25 @@ From the _Overview_ page, you can view the following information:
...
+- Number of worker cores
+
+The following information from Red Hat Insights is displayed:
+
+- Cluster recommendations
+- Number of risk predictions
+- Cluster health which includes the status and violations
+- A view of your resources based on your custom query.
+
+If observability is enabled, alert and failing operator metrics from across your fleet are also displayed.
```
dockerymick, line 54: @zlayne thanks for adding the screen capture! It was helpful with adding more details about what the user could view from the Fleet view switch. Looking at the list, is there any information that was not available to the user in 2.11? I recall that Fleet view was added in 2.9. Did users always have access to the information that I listed?

zlayne: This looks great - thanks.

Update risks were added in 2.10, and the top summary cards (cluster, app types, policies etc.) have been redesigned for 2.12. In previous releases those cards looked as follows:
Screenshot 2024-08-27 at 3 29 16 PM

But otherwise, everything else should have been available since 2.9.

release_notes/deprecate_remove.adoc
```diff
@@ -46,6 +46,12 @@ A _deprecated_ component, feature, or service is supported, but no longer recomm
 |===
 | Product or category | Affected item | Version | Recommended action | More details and links
 
+| _Overview_ page
+| {acm} search 
+| 2.12
+| Enable the _Fleet view_ switch to view the new default overview page.
```
dockerymick:
```suggestion
| Enable the _Fleet view_ switch to view the new default _Overview_ page.
```

release_notes/deprecate_remove.adoc
```diff
@@ -46,6 +46,12 @@ A _deprecated_ component, feature, or service is supported, but no longer recomm
 |===
 | Product or category | Affected item | Version | Recommended action | More details and links
 
+| _Overview_ page
+| {acm} search 
+| 2.12
+| Enable the _Fleet view_ switch to view the new default _Overview_ page.
+| The previous layout of the {acm-short} _Overview_ page is deprecated.
```
swopebe, line 53: We should indicate in the doc issue, maybe even change the title, that this also leads to a what's new entry about new fleet view....OR open a new issue for that.

swopebe (approved):
Looks great, thanks for thinking analytically about how this change impacts the doc.

dockerymick:
@zlayne thanks for your patience! When you have a chance will you take another look?

zlayne (approved):

Xiang-Yin-ACM:
Looks good for me. Great work!

Xiang-Yin-ACM (commented):
Looks great

PR #6788, opened 2024-08-14
https://issues.redhat.com/browse/ACM-12340--announce EOL

swopebe:


oafischer (approved):

about/welcome.adoc
```diff
@@ -3,6 +3,8 @@
 
 Kubernetes provides a platform for deploying and managing containers in a standard, consistent control plane. However, as application workloads move from development to production, they often require multiple fit-for-purpose Kubernetes clusters to support DevOps pipelines.
 
+*Deprecated:* The 2.7 and earlier versions of {product-title-short} are no longer supported and as a result, documentation for versions 2.7 and earlier are not updated. The documentation might remain available, but without any Errata or other updates available.
```
jc-berger: What do you think about these minor adjustments to make the content a bit more declarative? 

Although I don't think `as a result` is a big deal, I do think Vale encourages to use `so` or `therefore` instead. 

```suggestion
*Deprecated:* The 2.7 and earlier versions of {product-title-short} are no longer supported, so documentation for versions 2.7 and earlier are not updated. Any remaining documentation for these versions does not have any Errata or available updates. 
```

jc-berger (approved):
/lgtm

about/welcome.adoc
swopebe: @jc-berger this doesn't make sense because it is one version:

**Any remaining documentation for these versions does not have any Errata or available updates.**

swopebe: I see, for the "earlier" part.

swopebe: Any remaining documentation 

Actually would be "Any documentation that remains"

swopebe: I will remove _as a result_; the rest I think is fine as is.

The 2.7 and earlier versions of {product-title-short} are no longer supported. Documentation for versions 2.7 and earlier are not updated. The documentation might remain available, but without any Errata or other updates available.

jc-berger (approved):
/lgtm

PR #6805, opened 2024-08-20
https://issues.redhat.com/browse/ACM-13348

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-13348

### DIFF
--- a/release_notes/known_issues_global.adoc
+++ b/release_notes/known_issues_global.adoc
@@ -23,6 +23,30 @@ Or consider a troubleshooting topic.
 
 Review the known issues for the {global-hub} Operator. The following list contains known issues for this release, or known issues that continued from the previous release. For your {ocp-short} cluster, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/release_notes/ocp-4-12-release-notes#ocp-4-12-known-issues[{ocp-short} known issues].
 
+[#kafka-topics-after-upgrade]
+== The {global-hub} Operator does not use Kafka topics
+
+There are three topics named `spec`, `status` and `event` in your Kafka cluster. You can configure these Kafka topics after you upgrade Kafka to {global-hub} 1.3 or later. After this upgrade, you only have two topics in the system: one for `spec` path and one for `status` path. The default external topic is `gh-spec` and `gh-status`. Align these topics between your Kafka clusters and the {global-hub} Operator. You can either create these two topics in your Kafka cluster, or you can modify the {global-hub} Operator to use the existing topics in your Kafka cluster. For example, you can update the {global-hub} Operator by using the existing `spec` and `event` topics.
+
+Use the following YAML sample to update your {global-hub} Operator to use the Kafka topics:
+
+[source,yaml]
+----
+apiVersion: operator.open-cluster-management.io/v1alpha4
+kind: MulticlusterGlobalHub
+...
+spec:
+  availabilityConfig: Basic
+  dataLayer:
+    kafka:
+      topics:
+        specTopic: spec
+        statusTopic: event
+    postgres:
+      retention: 18m
+  enableMetrics: false
+----
+
 [#kafka-operator-keeps-restarting]
 == Kafka operator keeps restarting 
 
--- a/release_notes/known_issues_global.adoc
+++ b/release_notes/known_issues_global.adoc
@@ -23,20 +23,19 @@ Or consider a troubleshooting topic.
 
 Review the known issues for the {global-hub} Operator. The following list contains known issues for this release, or known issues that continued from the previous release. For your {ocp-short} cluster, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/release_notes/ocp-4-12-release-notes#ocp-4-12-known-issues[{ocp-short} known issues].
 
-[#byo-kafka-topics-after-upgrade]
-== The {global-hub} operand does not use BYO Kafka topics
+[#kafka-topics-after-upgrade]
+== The {global-hub} operand does not use your Kafka topics
 
-There are three topics named `spec`, `status` and `event` in your "bring your own" (BYO) Kafka cluster. You can configure these BYO Kafka topics after you upgrade your BYO Kafka to {global-hub} 1.3 or later. After this upgrade, you only have two topics in the system: one for `spec` path and one for `status` path. The default external topic is `gh-status`. 
+There are three topics named `spec`, `status` and `event` in your own Kafka that you brought. You can configure your own Kafka topics after you upgrade your Kafka to {global-hub} 1.3 or later. Then, you only have two topics in the system: one for `spec` path and one for `status` path. The default external topic is `gh-status`. 
 
-The {global-hub} operand is the API or resource of your system, and the {global-hub} operator is the controller. After upgrading {global-hub}, align the topics between your BYO Kafka clusters and the {global-hub} operand. You can either create the `gh-status` topic in your BYO Kafka cluster, or you can change the {global-hub} operand to use the existing topics, such as: `spec` and `status`. 
+The {global-hub} operand is the API or resource of your system, and the {global-hub} operator is the controller. After upgrading {global-hub}, align the topics between your specific Kafka clusters and the {global-hub} operand. You can either create the `gh-status` topic in your Kafka cluster, or you can change the {global-hub} operand to use the existing topics, such as: `spec` and `status`. 
 
-Use the following YAML sample to update your {global-hub} operand to use your BYO Kafka topics:
+Use the following YAML sample to update your {global-hub} operand to use your own Kafka topics:
 
 [source,yaml]
 ----
 apiVersion: operator.open-cluster-management.io/v1alpha4
 kind: MulticlusterGlobalHub
-...
 spec:
   availabilityConfig: Basic
   dataLayer:
--- a/release_notes/known_issues_global.adoc
+++ b/release_notes/known_issues_global.adoc
@@ -26,9 +26,9 @@ Review the known issues for the {global-hub} Operator. The following list contai
 [#kafka-topics-after-upgrade]
 == The {global-hub} operand does not use your Kafka topics
 
-There are three topics named `spec`, `status` and `event` in your own Kafka that you brought. You can configure your own Kafka topics after you upgrade your Kafka to {global-hub} 1.3 or later. Then, you only have two topics in the system: one for `spec` path and one for `status` path. The default external topic is `gh-status`. 
+There are three topics named `spec`, `status` and `event` in your own Kafka that you brought. After upgrading {global-hub} to 1.3 or later, only the `spec` and `status` topics remain. Configure these topics through the {global-hub} operand. The default topis for your transport that your brought are `gh-spec` and `gh-status`. 
 
-The {global-hub} operand is the API or resource of your system, and the {global-hub} operator is the controller. After upgrading {global-hub}, align the topics between your specific Kafka clusters and the {global-hub} operand. You can either create the `gh-status` topic in your Kafka cluster, or you can change the {global-hub} operand to use the existing topics, such as: `spec` and `status`. 
+The {global-hub} operand is the API or resource of your system, and the {global-hub} operator is the controller. After upgrading {global-hub}, align the topics between your own Kafka clusters and the {global-hub} operand. You can either create the `gh-spec` and `gh-status` topic in your Kafka cluster, or you can change the {global-hub} operand to use the existing `spec` and `status` topics.
 
 Use the following YAML sample to update your {global-hub} operand to use your own Kafka topics:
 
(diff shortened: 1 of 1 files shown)


release_notes/known_issues_global.adoc
```diff
@@ -23,6 +23,30 @@ Or consider a troubleshooting topic.
 
 Review the known issues for the {global-hub} Operator. The following list contains known issues for this release, or known issues that continued from the previous release. For your {ocp-short} cluster, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/release_notes/ocp-4-12-release-notes#ocp-4-12-known-issues[{ocp-short} known issues].
 
+[#kafka-topics-after-upgrade]
+== The {global-hub} Operator does not use Kafka topics
+
+There are three topics named `spec`, `status` and `event` in your Kafka cluster. You can configure these Kafka topics after you upgrade Kafka to {global-hub} 1.3 or later. After this upgrade, you only have two topics in the system: one for `spec` path and one for `status` path. The default external topic is `gh-spec` and `gh-status`. Align these topics between your Kafka clusters and the {global-hub} Operator. You can either create these two topics in your Kafka cluster, or you can modify the {global-hub} Operator to use the existing topics in your Kafka cluster. For example, you can update the {global-hub} Operator by using the existing `spec` and `event` topics.
```
jc-berger: Dev reviewers:

The file in the multicluster-global-hub repo says `Global Hub operand` Am I incorrect to user `Global Hub Operator` here? 

If `Global Hub operand` is a separate thing, may we do a brief definition or explanation of it here?

release_notes/known_issues_global.adoc
```diff
@@ -23,6 +23,30 @@ Or consider a troubleshooting topic.
 
 Review the known issues for the {global-hub} Operator. The following list contains known issues for this release, or known issues that continued from the previous release. For your {ocp-short} cluster, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/release_notes/ocp-4-12-release-notes#ocp-4-12-known-issues[{ocp-short} known issues].
 
+[#kafka-topics-after-upgrade]
+== The {global-hub} Operator does not use Kafka topics
```
jc-berger: Dev reviewers:

It seems to me that the known issue is that users have to manually configure the Global Hub operator (or operand) to use the Kafka topics. That is, after the upgrade to Global Hub 1.3 or later, users lose some Kafka topics, and Global Hub does not automatically use it. 

If I am incorrect, please correct me, and I'llmake changes to this content. Thanks :)

release_notes/known_issues_global.adoc
jc-berger: Update: 

After doing some reading, I see the difference between `operand` and `operator`. thank you for your patience :) 

Since I didn't know the distinction, I did add a sentence explaining it. If you think this sentence is unneeded because our users would know the difference, let me know, and I'll remove it.

release_notes/known_issues_global.adoc
```diff
@@ -23,6 +23,32 @@ Or consider a troubleshooting topic.
...
+
+The {global-hub} operand provides the data and values for the {global-hub} operator to work. After upgrading {global-hub}, align the topics between your Kafka clusters and the {global-hub} operand. You can either create the  `gh-spec` and `gh-status` topics in your Kafka cluster, or you can update the {global-hub} operand to use the existing topics, such as: `spec` and `event`. 
+
+Use the following YAML sample to update your {global-hub} operand to use the Kafka topics:
+
+[source,yaml]
+----
+apiVersion: operator.open-cluster-management.io/v1alpha4
+kind: MulticlusterGlobalHub
+...
```
jc-berger: Dev reviewers:

Do we need the `... ` here?

release_notes/known_issues_global.adoc
```diff
@@ -23,6 +23,32 @@ Or consider a troubleshooting topic.
 
 Review the known issues for the {global-hub} Operator. The following list contains known issues for this release, or known issues that continued from the previous release. For your {ocp-short} cluster, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/release_notes/ocp-4-12-release-notes#ocp-4-12-known-issues[{ocp-short} known issues].
 
+[#kafka-topics-after-upgrade]
+== The {global-hub} operand does not use Kafka topics
+
+There are three topics named `spec`, `status` and `event` in your Kafka cluster. You can configure these Kafka topics after you upgrade Kafka to {global-hub} 1.3 or later. After this upgrade, you only have two topics in the system: one for `spec` path and one for `status` path. The default external topic is `gh-spec` and `gh-status`.
```
yanmxa: I think `gh-status` and `gh-event` both make sense. But @clyang82 prefers to use `gh-event` and @hchenxa [seems like](https://github.com/stolostron/multicluster-global-hub/pull/1053#discussion_r1719197679) to use `gh-status`.

/cc @clyang82 @hchenxa, We need to agree on the terms~

release_notes/known_issues_global.adoc
```diff
@@ -23,6 +23,32 @@ Or consider a troubleshooting topic.
 
 Review the known issues for the {global-hub} Operator. The following list contains known issues for this release, or known issues that continued from the previous release. For your {ocp-short} cluster, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/release_notes/ocp-4-12-release-notes#ocp-4-12-known-issues[{ocp-short} known issues].
 
+[#kafka-topics-after-upgrade]
+== The {global-hub} operand does not use Kafka topics
+
+There are three topics named `spec`, `status` and `event` in your Kafka cluster. You can configure these Kafka topics after you upgrade Kafka to {global-hub} 1.3 or later. After this upgrade, you only have two topics in the system: one for `spec` path and one for `status` path. The default external topic is `gh-spec` and `gh-status`. 
+
+The {global-hub} operand provides the data and values for the {global-hub} operator to work. After upgrading {global-hub}, align the topics between your Kafka clusters and the {global-hub} operand. You can either create the  `gh-spec` and `gh-status` topics in your Kafka cluster, or you can update the {global-hub} operand to use the existing topics, such as: `spec` and `event`.
```
yanmxa: As mentioned above, it needs @clyang82 and @hchenxa to be consistent with the item `gh-status` or `gh-event`?

release_notes/known_issues_global.adoc
yanmxa: Since there are other fields behind `...`. I think we can keep it :)

release_notes/known_issues_global.adoc
yanmxa: > Global Hub operand

The `Global Hub operand` is the API or resources of the system. but the `Global Hub Operator` is the controller. So we can't replace the it with the later

yanmxa: `Align these topics between your Kafka clusters and the {global-hub} Operator` -> `Align these topics between your Kafka clusters and the {global-hub} Operand`
`or you can modify the {global-hub} Operator` -> `or you can modify the {global-hub} Operand`
`For example, you can update the {global-hub} Operator` -> `For example, you can update the {global-hub} Operand`

release_notes/known_issues_global.adoc
yanmxa: ` Global Hub operator` -> ` Global Hub operand` 

You are right about the upgrade case!

release_notes/known_issues_global.adoc
hchenxa: yes, since the spec name is `statusTopic`, so I prefer to use `gh-status` as the default value.

release_notes/known_issues_global.adoc
```diff
@@ -23,6 +23,32 @@ Or consider a troubleshooting topic.
 
 Review the known issues for the {global-hub} Operator. The following list contains known issues for this release, or known issues that continued from the previous release. For your {ocp-short} cluster, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/release_notes/ocp-4-12-release-notes#ocp-4-12-known-issues[{ocp-short} known issues].
 
+[#kafka-topics-after-upgrade]
+== The {global-hub} operand does not use Kafka topics
```
hchenxa: I think the issue only happen when use the BYO kafka.

yanmxa: Yes. It only appears in the BYO case.

release_notes/known_issues_global.adoc
yanmxa: @jc-berger After discussing with @clyang82 and @hchenxa. Our final conclusion is to use `gh-status`.

release_notes/known_issues_global.adoc
```diff
@@ -23,6 +23,32 @@ Or consider a troubleshooting topic.
...
+
+There are three topics named `spec`, `status` and `event` in your Kafka cluster. You can configure these Kafka topics after you upgrade Kafka to {global-hub} 1.3 or later. After this upgrade, you only have two topics in the system: one for `spec` path and one for `status` path. The default external topic is `gh-spec` and `gh-status`. 
+
+The {global-hub} operand is the API or resource of your system, and the {global-hub} operator is the controller. After upgrading {global-hub}, align the topics between your Kafka clusters and the {global-hub} operand. You can either create the  `gh-spec` and `gh-status` topics in your Kafka cluster, or you can modify the {global-hub} operand to use the existing topics, like: `spec` and `event`. 
+
```
yanmxa, line 32: need change `event` to `status`

release_notes/known_issues_global.adoc
```diff
@@ -23,6 +23,32 @@ Or consider a troubleshooting topic.
...
+apiVersion: operator.open-cluster-management.io/v1alpha4
+kind: MulticlusterGlobalHub
+...
+spec:
+  availabilityConfig: Basic
+  dataLayer:
+    kafka:
+      topics:
+        specTopic: spec
+        statusTopic: event
```
yanmxa: also need change the `event` to `status`

release_notes/known_issues_global.adoc
swopebe: From what I can tell BYO is not part of a product or technology name. It could be seen as colloquial. Please check on that.

Also, we want to avoid abbreviations unless they are necessary (translation issue)

"bring your own" -- quotation marks used as emphasis
https://www.ibm.com/docs/en/ibm-style?topic=punctuation-quotation-marks

Please do not ever use quotation marks in technical documentation unless we are citing a document, which we never do (I have not seen in my 9 years). This is an accessibility issue.

swopebe: **these** BYO Kafka topics
**this** upgrade

Self-referential, these can be rewritten.

Simple change: After you upgrade. (this also should replace _After upgrading global-hub_)

But honestly, you don't even need to say that; if you look at the line before it, you already say "after you upgrade"

You can scan through to see where that can be more concise.

release_notes/known_issues_global.adoc
swopebe: We should not have these in here; @jc-berger we need clean YAML file samples, you are correct! TY.

We do not add these to the samples in the doc.

swopebe (changes requested):
See comments.

release_notes/known_issues_global.adoc
```diff
@@ -23,6 +23,31 @@ Or consider a troubleshooting topic.
 
 Review the known issues for the {global-hub} Operator. The following list contains known issues for this release, or known issues that continued from the previous release. For your {ocp-short} cluster, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.12/html/release_notes/ocp-4-12-release-notes#ocp-4-12-known-issues[{ocp-short} known issues].
 
+[#kafka-topics-after-upgrade]
+== The {global-hub} operand does not use your Kafka topics
+
+There are three topics named `spec`, `status` and `event` in your own Kafka that you brought. You can configure your own Kafka topics after you upgrade your Kafka to {global-hub} 1.3 or later. Then, you only have two topics in the system: one for `spec` path and one for `status` path. The default external topic is `gh-status`.
```
yanmxa: "You can configure your own Kafka topics after you upgrade your Kafka to {global-hub} 1.3 or later. Then, you only have two topics in the system: one for `spec` path and one for `status` path. The default external topic is `gh-status`. " 
-> 
"After upgrading {global-hub} to 1.3 or later, only two topics remain: one for spec and one for status. You can configure these through the {global-hub} operand. The default topics for BYO transport are `gh-spec` and `gh-status`."

release_notes/known_issues_global.adoc
```diff
@@ -23,6 +23,31 @@ Or consider a troubleshooting topic.
...
 
+[#kafka-topics-after-upgrade]
+== The {global-hub} operand does not use your Kafka topics
+
+There are three topics named `spec`, `status` and `event` in your own Kafka that you brought. You can configure your own Kafka topics after you upgrade your Kafka to {global-hub} 1.3 or later. Then, you only have two topics in the system: one for `spec` path and one for `status` path. The default external topic is `gh-status`. 
+
+The {global-hub} operand is the API or resource of your system, and the {global-hub} operator is the controller. After upgrading {global-hub}, align the topics between your specific Kafka clusters and the {global-hub} operand. You can either create the `gh-status` topic in your Kafka cluster, or you can change the {global-hub} operand to use the existing topics, such as: `spec` and `status`.
```
yanmxa: suggestion1: "your specific Kafka clusters" -> "your own kafka clusters"
suggestion2: "You can either create the `gh-status` topic in your Kafka cluster" -> "You can either create the `gh-spec` and `gh-status` topics in your Kafka cluster"
suggestion3: "or you can change the {global-hub} operand to use the existing topics, such as: `spec` and `status`." -> "or you can change the {global-hub} operand to use the existing `spec` and `status` topics"

swopebe (commented):
Looks great, thank you!

swopebe (approved):

yanmxa:
Thanks @jc-berger !

PR #6819, opened 2024-08-26
Add concept modules for SiteConfig

amolnar-gh:
Issues:
* https://issues.redhat.com/browse/TELCODOCS-1875
* https://issues.redhat.com/browse/ACM-14071


modules/common-attributes.adoc
```diff
@@ -22,3 +22,4 @@
 :quay-short: Quay
 :imagesdir: ../images
 :sno: single-node OpenShift
```
swopebe, line 24: We need to double check the naming guidance here, most RH operators are lower case "operator"

siteconfig/cluster-templates.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+* Cluster-level templates that must reference only cluster-specific fields.
+* Node-level templates that can reference both cluster-specific and node-specific fields.
+
+You can customize the templated fields as the {sco} supports all link:https://masterminds.github.io/sprig/[sprig library functions].
+Additionally, the `ClusterInstance` API provides a new function that you can use while creating your custom manifests:
+
+`toYaml`:: The `toYaml` function encodes an item into a YAML string. If the item cannot be converted to YAML the function returns an empty string.
+
+.Using .toYaml in the ClusterInstance.Spec.Proxy field
```
swopebe: We don't use modules, I think that is what line 19 is?

siteconfig/cluster-templates.adoc
```diff
@@ -0,0 +1,111 @@
...
+`CurrentNode`:: The {sco} explicitly controls the iteration of the node objects and exposes this variable to access all the content for the current node being handled in templating.
+`InstallConfigOverrides`:: Contains the merged `networkType`, `cpuPartitioningMode` and `installConfigOverrides` content.
+`ControlPlaneAgents`:: Consists of the number of control plane agents and it is automatically derived from the `ClusterInstance` node objects.
+`WorkerAgents`:: Consists of the number of worker agents and it is automatically derived from the `ClusterInstance` node objects.
+
+You can create a custom templated field by capitalizing the field name in the text template.
+For example, the `ClusterInstance` `spec` fields is referenced with the `.Spec` prefix.
+However, special variable fields must be referenced with the `.SpecialVars` prefix.
+
+[IMPORTANT]
```
swopebe: We don't use all caps for special tagging. This is an accessibility issue. See https://stolostron.github.io/rhacm-docs/special_tag_guidance. Additionally, the interruption that RH docs creates has been cause for concern based on some user feedback, so we don't add the boxes to the doc in this manner.

modules/common-attributes.adoc
swopebe: For diagrams, see our internal doc for that procedure if this did not go through the design team, it will need to go through that process.

swopebe:
This is a large PR that will need some more review. Let's pause here if possible and make the changes that I have mentioned. I will schedule a meeting before we add more to the PR or add other PRs.

siteconfig/cluster-templates.adoc
swopebe: RH is behind on accessibility, but I have taken this to ContentX to update the guidance based on the following:

https://www.bdadyslexia.org.uk/advice/employers/creating-a-dyslexia-friendly-workplace/dyslexia-friendly-style-guide

https://accessibility.huit.harvard.edu/design-readability

Here we decided not to do this for these reasons.

siteconfig/clusterinstance-cr-reference.adoc
```diff
@@ -0,0 +1,7 @@
+[#siteconfig-clusterinstance-cr]
```
swopebe: We are trying to simplify our filenames to make like a little easier when building links.

`clusterinstance_cr.adoc` should suffice and notice the underscore. Then you can also shorten the anchors. 
[#clusterinstance-cr] or something similar. We have done a lot of work to simplify these pieces of the doc to make building links and such easier.

siteconfig/ref-clusterinstance-cr.adoc
```diff
@@ -0,0 +1,56 @@
+.Reference ClusterInstance CR
```
swopebe: This is an API doc, see how ours are built.

siteconfig/siteconfig_intro.adoc
```diff
@@ -1,6 +1,59 @@
...
+
+////
+include::siteconfig/cnf-deploying-siteconfig-operator.adoc[leveloffset=+1]
+
+include::siteconfig/cnf-installing-clusters.adoc[leveloffset=+1]
+
+[role="_additional-resources"]
+.Additional resources
+
+[role="_additional-resources"]
```
swopebe: all of these are just headers for our doc, see the repo for examples since we do not use modules. (I think that is what this is, anyway)

swopebe (changes requested):
I am still working on this, but have some comments that can be addressed in the meantime. Thank you.

modules/common-attributes.adoc
swopebe: If this went through design, that's fine, we just have to make sure it appears as our other diagrams:

https://stolostron.github.io/rhacm-docs/insert_images

siteconfig/cluster-templates.adoc
swopebe: https://accessibility.huit.harvard.edu/design-readability

siteconfig/ref-clusterinstance-cr.adoc
swopebe: https://github.com/stolostron/rhacm-docs/tree/2.12_stage/apis (There is history here, since we came from markdown and had a tool to use at IBM, but we should make it the same. Recently Oliver took this on so he may be able to help. There is a VS Code plugin that we used.)

modules/common-attributes.adoc
amolnar-gh: Re Operator: In OCP, we use capitalized Operator in the docs. I wanted to make sure that it's not just OCP, and according to the style guide, it needs to be capitalized: https://redhat-documentation.github.io/supplementary-style-guide/#operator

Re diagram: I opened a [ticket](https://issues.redhat.com/browse/DIAGRAMS-454) when the feature went dev preview and set myself a reminder that this must be updated once it's done :) There's a possibility that it'll be done after GA, unfortunately. The current diagram is just a placeholder.

siteconfig/cluster-templates.adoc
amolnar-gh: For me, modules are separate files that are referenced in an assembly so I'm not sure. However, this is supposed to be a title for an example codeblock. This is how it looks like in the KB article:
Screenshot from 2024-08-27 08-47-48

Now looking at this, if you do use titles in example code snippets, I should rephrase the title, something like Example use of...

modules/common-attributes.adoc
swopebe: Ok, I can tell by the font @amolnar-rh that this is likely going to be an accessibility issue so if we don't have a new diagram for GA, I don't want to publish this one. 

I will open an issue for the design team while you are out.

amolnar-gh: I know, the font is not final! I'll remove the diagram for now to avoid accidentally including it.

Please, use the open ticket for any communication with the design team: https://issues.redhat.com/browse/DIAGRAMS-454
I already nicely asked for an earlier completion date but the backlog didn't allow for it :(

swopebe: @oafischer is following up on this design for completion for GA, but we may need to add it in a refresh.

oafischer: The design team has a long backlog, might be a few weeks before we get the diagram.

siteconfig/siteconfig_intro.adoc
```diff
@@ -1,6 +1,44 @@
...
+
+* xref:../siteconfig/siteconfig_flow#operator-flow[The {sco} flow]
+* xref:../siteconfig/cluster-templates.adoc#cluster-templates[Cluster templates overview]
+
+////
+Temporarily leaving this for my reference for the structure and links.
+
+include::../siteconfig/siteconfig-operator-flow.adoc
+
+include::../siteconfig/clusterinstance-cr-reference.adoc              |
```
swopebe: We use underscore for naming files, so wanted to mention that, but I also a not sure we need these, since we can just point to the API doc.

siteconfig/siteconfig_intro.adoc
```diff
@@ -1,6 +1,44 @@
...
+
+include::../siteconfig/cluster-templates.adoc
+
+include::siteconfig/create-custom-cluster-templates.adoc
+
+include::siteconfig/cnf-deploying-siteconfig-operator.adoc[leveloffset=+1]
+
+include::siteconfig/cnf-installing-clusters.adoc[leveloffset=+1]
+
+* xref:../../installing/installing_on_prem_assisted/installing-on-prem-assisted.adoc#installing-on-prem-assisted[Installing an on-premise cluster using the {ai-full}]
```
swopebe, line 43: We can work with @oafischer for the conref bc he has an issue to fix ours in our doc, previously set by original author.

{ai-full}

modules/common-attributes.adoc
swopebe: If that is the case, then we need to remove the image and wait for it. Our GA is likely moving up, so we will keep the image issue open. Thank you @oafischer

siteconfig/siteconfig_intro.adoc
amolnar-gh: Thanks! These are removed but I checked all the remaining file names to only include underscores, should be all correct now.

siteconfig/siteconfig_intro.adoc
oafischer: {ai} would make the most sense if we follow the other conrefs that we already have.

swopebe (approved):
We still have some things to address, indicated in comments. Let's merge, place, then keep working on it.

mpmaruthu:
Telco RAN QE: Looks good to me.
/Peri M

PR #6854, opened 2024-08-30
Remove hosted KI/Trouble files

oafischer:
2.12 only

Includes only release notes/KI/troublehooting etc.


dockerymick (approved):
Thanks Oliver!

PR #6872, opened 2024-09-03
2.10 refresh for EOL announcement

swopebe:


dockerymick (approved):
Similar comment from the previous PR. Some commands aren't greyed out, but that can be handled in a different PR. Approving!

PR #6873, opened 2024-09-03
2.11 refresh MCE correction, Errata 2.6.2 and more

swopebe:


dockerymick (approved):
I noticed some of the commands were not greyed out, but that is small and can be fixed later in a different PR. Approving!

PR #6891, opened 2024-09-05
https://issues.redhat.com/browse/ACM-13434--making necessary changes

swopebe:
1. This file had a citation not properly italicized.
2. We were calling it "an ISO" and ISO is the type of file or type of media, so we need to be specific.
3. Approve each host is a step in the procedure, but buried.

I am not sure what this is, because there are no steps but this begins with a gerund:

Adding...

You can....

See this...

Approve...

Is the only thing to do here is approve. I would go here looking for a procedure. Seems we are missing some info:

```
== Adding hosts by using the command line interface

You can see the URL to download the ISO file in the `isoDownloadURL` property in the status of your `InfraEnv` resource. See _Creating a host inventory by using the command line interface_ for more information about the `InfraEnv` resource.

Each booted host creates an `Agent` resource in the same namespace. 

Approve each host so that you can use it.
```
Then we need to address the issue:

```
The documentation is not thorough around the possible options for the discovery ISO.

The UI download link automatically encodes the minimal.iso in the URL, so this is the one the user gets by default.

The minimal.iso does not contain the full rootfs.  This makes it easier and faster to transport across a network.  However, as a result of the changes applied to make it lightweight, it is less compatible with systems, especially older ones.  It strips the UEFI secondary partition that some older bootloaders are looking for.  Older systems may still make use of the full.iso file, but we need to let users know it is possible. 
```

### DIFF
--- a/clusters/cluster_lifecycle/cim_add_host.adoc
+++ b/clusters/cluster_lifecycle/cim_add_host.adoc
@@ -55,7 +55,6 @@ oc get infraenv -n <infra env namespace> <infra env name> -o jsonpath='{.status.
 See the following output:
 
 +
-[source,bash]
 ----
 https://assisted-image-service-assisted-installer.apps.example-acm-hub.com/byapikey/eyJhbGciOiJFUzI1NiIsInC93XVCJ9.eyJpbmZyYV9lbnZfaWQcTA0Y38sWVjYi02MTA0LTQ4NDMtODasdkOGIxYTZkZGM5ZTUifQ.3ydTpHaXJmTasd7uDp2NvGUFRKin3Z9Qct3lvDky1N-5zj3KsRePhAM48aUccBqmucGt3g/4.16/x86_64/minimal.iso
 ----
@@ -78,7 +77,6 @@ oc get agent -n <infra env namespace>
 You get an output that is similar to the following output:
 
 +
-[source,bash]
 ----
 NAME                                   CLUSTER   APPROVED   ROLE          STAGE
 24a92a6f-ea35-4d6f-9579-8f04c0d3591e             false      auto-assign   
@@ -104,7 +102,6 @@ oc get agent -n <infra env namespace>
 You get an output that is similar to the following output with a `true` value:
 
 +
-[source,bash]
 ----
 NAME                                   CLUSTER   APPROVED   ROLE          STAGE
 173e3a84-88e2-4fe1-967f-1a9242503bec             true       auto-assign    
--- a/clusters/cluster_lifecycle/cim_add_host.adoc
+++ b/clusters/cluster_lifecycle/cim_add_host.adoc
@@ -1,9 +1,15 @@
 [#add-host-host-inventory]
 = Adding hosts to the host inventory by using the Discovery Image
 
-After creating your host inventory (infrastructure environment), you can discover your hosts and add them to your inventory. To add hosts to your inventory, choose a method to download an ISO file and attach it to each server. For example, you can download ISO files by using a virtual media, or by writing the ISO to a USB drive.
+After you create your host inventory (infrastructure environment), you can discover your hosts and add them to your inventory. 
 
-*Important:* To prevent the installation from failing, keep the Discovery ISO media connected to the device during the installation process and set each host to boot from the device one time.
+To add hosts to your inventory, choose a method to download an ISO file and attach it to each server. For example, you can download ISO files by using a virtual media, or by writing the ISO to a USB drive.
+
+*Important:* To prevent the installation from failing, keep the Discovery ISO media connected to the device during the installation process, and set each host to boot from the device one time.
+
+* <<add-host-prereqs,Prerequisites>>
+* <<add-host-steps-console,Adding hosts by using the console>>
+* <<add-host-steps-cli,Adding hosts by using the command line interface>>
 
 [#add-host-prereqs]
 == Prerequisites
@@ -37,7 +43,7 @@ The URL to download the ISO file in the `isoDownloadURL` property is in the stat
 
 Each booted host creates an `Agent` resource in the same namespace. 
 
-1. Run the following command to view the download URL in the `InfraEnv` custom resource:
+. Run the following command to view the download URL in the `InfraEnv` custom resource:
 
 +
 [source,bash]
@@ -66,6 +72,7 @@ Next, you need to approve each host. See the following procedure:
 oc get agent -n <infra env namespace>
 ----
 
++
 You get an output that is similar to the following output:
 
 +
@@ -91,6 +98,7 @@ oc patch agent -n <infra env namespace> <agent name> -p '{"spec":{"approved":tru
 oc get agent -n <infra env namespace>
 ----
 
++
 You get an output that is similar to the following output with a `true` value:
 
 +
--- a/clusters/cluster_lifecycle/cim_add_host.adoc
+++ b/clusters/cluster_lifecycle/cim_add_host.adoc
@@ -30,9 +30,29 @@ You now see a URL to download the ISO file. Booted hosts appear in the host inve
 [#add-host-steps-cli]
 == Adding hosts by using the command line interface
 
-You can see the URL to download the ISO file in the `isoDownloadURL` property in the status of your `InfraEnv` resource. See _Creating a host inventory by using the command line interface_ for more information about the `InfraEnv` resource.
+The URL to download the ISO file in the `isoDownloadURL` property is in the status of your `InfraEnv` resource. See _Creating a host inventory by using the command line interface_ for more information about the `InfraEnv` resource.
 
-Each booted host creates an `Agent` resource in the same namespace. Approve each host so that you can use it.
+Each booted host creates an `Agent` resource in the same namespace. 
+
+1. Run the following command to view the download URL in the `InfraEnv` custom resource:
+
++
+[source,bash]
+----
+oc get infraenv -n <infra env namespace> <infra env name> -o jsonpath='{.status.isoDownloadURL}'
+----
+
+See the following output:
+
++
+[source,bash]
+----
+https://assisted-image-service-assisted-installer.apps.example-acm-hub.com/byapikey/eyJhbGciOiJFUzI1NiIsInC93XVCJ9.eyJpbmZyYV9lbnZfaWQcTA0Y38sWVjYi02MTA0LTQ4NDMtODasdkOGIxYTZkZGM5ZTUifQ.3ydTpHaXJmTasd7uDp2NvGUFRKin3Z9Qct3lvDky1N-5zj3KsRePhAM48aUccBqmucGt3g/4.16/x86_64/minimal.iso
+----
+
+. Use the URL to download the ISO file and boot your hosts with the ISO file.
+
+. Approve each host.
 
 [#additional-resources-add-host]
 == Additional resources
--- a/clusters/cluster_lifecycle/cim_add_host.adoc
+++ b/clusters/cluster_lifecycle/cim_add_host.adoc
@@ -1,7 +1,7 @@
 [#add-host-host-inventory]
 = Adding hosts to the host inventory by using the Discovery Image
 
-After creating your host inventory (infrastructure environment) you can discover your hosts and add them to your inventory. To add hosts to your inventory, choose a method to download an ISO and attach it to each server. For example, you can download ISOs by using a virtual media or writing the ISO to a USB drive.
+After creating your host inventory (infrastructure environment), you can discover your hosts and add them to your inventory. To add hosts to your inventory, choose a method to download an ISO file and attach it to each server. For example, you can download ISO files by using a virtual media, or by writing the ISO to a USB drive.
 
 *Important:* To prevent the installation from failing, keep the Discovery ISO media connected to the device during the installation process and set each host to boot from the device one time.
 
@@ -14,7 +14,7 @@ After creating your host inventory (infrastructure environment) you can discover
 [#add-host-steps-console]
 == Adding hosts by using the console
 
-Download the ISO by completing the following steps:
+Download the ISO file by completing the following steps:
 
 . Select *Infrastructure* > *Host inventory* in the console.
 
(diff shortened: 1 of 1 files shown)


clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -1,7 +1,7 @@
 [#add-host-host-inventory]
 = Adding hosts to the host inventory by using the Discovery Image
 
-After creating your host inventory (infrastructure environment) you can discover your hosts and add them to your inventory. To add hosts to your inventory, choose a method to download an ISO and attach it to each server. For example, you can download ISOs by using a virtual media or writing the ISO to a USB drive.
+After creating your host inventory (infrastructure environment), you can discover your hosts and add them to your inventory. To add hosts to your inventory, choose a method to download an ISO file and attach it to each server. For example, you can download ISO files by using a virtual media, or by writing the ISO to a USB drive.
```
CrystalChun, line 5: Maybe additional docs here: 
By default, the ISO provided is a minimal ISO. The minimal ISO does not contain the RootFS and will download the RootFS at a later stage.

If a full ISO is desired, follow the steps 
1. Create a config map in the `multicluster-engine` namespace to override the `assisted-service` environment variable
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: assisted-service-config-override
  namespace: multicluster-engine
data:
  ISO_IMAGE_TYPE: "full-iso"
```
2. Annotate the `AgentServiceConfig` to reference that config map
```bash
oc annotate --overwrite AgentServiceConfig agent unsupported.agent-install.openshift.io/assisted-service-configmap=assisted-service-config-override
```
Verify the assisted-service pod has restarted
```
oc get po -n multicluster-engine | grep assisted
```

swopebe: Thank you! Will do. Can you look at:

[#add-host-steps-cli]
== Adding hosts by using the command line interface @CrystalChun ? It is written as a task but I am not sure why the writer set it up that way with no steps. Are we missing steps here? If not, it needs to be written differently, which I can handle.

CrystalChun: There are two steps for this task. I've added them in another comment below.

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -14,22 +14,25 @@ After creating your host inventory (infrastructure environment) you can discover
...
++
+You now see a URL to download the ISO file. Booted hosts appear in the host inventory table. Hosts might take a few minutes to appear. 
+
+. Approve each host so that you can use it. You can select hosts from the inventory table by clicking *Actions* and selecting *Approve*.
 
 [#add-host-steps-cli]
 == Adding hosts by using the command line interface
 
-You can see the URL to download the ISO in the `isoDownloadURL` property in the status of your `InfraEnv` resource. See Creating a host inventory by using the command line interface for more information about the `InfraEnv` resource.
+You can see the URL to download the ISO file in the `isoDownloadURL` property in the status of your `InfraEnv` resource. See _Creating a host inventory by using the command line interface_ for more information about the `InfraEnv` resource.
```
CrystalChun: Download the ISO URL by doing the following:

1. View the download URL in the `InfraEnv` CR 
```bash
oc get infraenv -n <infra env namespace> <infra env name> -o jsonpath='{.status.isoDownloadURL}'
```
The output should look like this:
```
https://assisted-image-service-assisted-installer.apps.example-acm-hub.com/byapikey/eyJhbGciOiJFUzI1NiIsInC93XVCJ9.eyJpbmZyYV9lbnZfaWQcTA0Y38sWVjYi02MTA0LTQ4NDMtODasdkOGIxYTZkZGM5ZTUifQ.3ydTpHaXJmTasd7uDp2NvGUFRKin3Z9Qct3lvDky1N-5zj3KsRePhAM48aUccBqmucGt3g/4.16/x86_64/minimal.iso
```
2. Download the ISO file using the URL and boot your hosts with the ISO file

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -14,22 +14,25 @@ After creating your host inventory (infrastructure environment) you can discover
...
 
-Each booted host creates an `Agent` resource in the same namespace. You must approve each host before you can use it.
+Each booted host creates an `Agent` resource in the same namespace. Approve each host so that you can use it.
```
CrystalChun: To approve the hosts, do the following:

1. List all of your `Agents`
```bash
oc get agent -n <infra env namespace>
```
The output will appear like this
```bash
NAME                                   CLUSTER   APPROVED   ROLE          STAGE
24a92a6f-ea35-4d6f-9579-8f04c0d3591e             false      auto-assign   
```
2. Approve any `Agent` that is not yet approved
```bash
oc patch agent -n <infra env namespace> <agent name> -p '{"spec":{"approved":true}}' --type merge
```
3. List all of the `Agent` CRs again to ensure they're all approved
```bash
oc get agent -n <infra env namespace>
```
The output will appear like this:
```bash
NAME                                   CLUSTER   APPROVED   ROLE          STAGE
173e3a84-88e2-4fe1-967f-1a9242503bec             true       auto-assign    
```

clusters/cluster_lifecycle/cim_add_host.adoc
swopebe: TY!

briantward: > ```shell
> unsupported.agent-install.openshift.io/assisted-service-configmap
> ```

This looks nice, but do we want to reference an "unsupported" option?  Does this affect the download URL default? I found we can simply replace `minimal.iso` with `full.iso` in the download URL.

CrystalChun: That's true! In that case, I'll add a comment below in the CLI section to use either minimal or full iso

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -14,22 +14,45 @@ After creating your host inventory (infrastructure environment) you can discover
...
+oc get infraenv -n <infra env namespace> <infra env name> -o jsonpath='{.status.isoDownloadURL}'
+----
+
+See the following output:
+
++
+[source,bash]
+----
+https://assisted-image-service-assisted-installer.apps.example-acm-hub.com/byapikey/eyJhbGciOiJFUzI1NiIsInC93XVCJ9.eyJpbmZyYV9lbnZfaWQcTA0Y38sWVjYi02MTA0LTQ4NDMtODasdkOGIxYTZkZGM5ZTUifQ.3ydTpHaXJmTasd7uDp2NvGUFRKin3Z9Qct3lvDky1N-5zj3KsRePhAM48aUccBqmucGt3g/4.16/x86_64/minimal.iso
+----
```
CrystalChun, line 60: A note here:
By default, the ISO provided is a minimal ISO. The minimal ISO does not contain the RootFS and will download the RootFS at a later stage.

If a full ISO is desired, replace `minimal.iso` in the URL above with `full.iso`.

swopebe: @CrystalChun so  say part of that below in the next section? Can you advise?

CrystalChun: Sorry for the confusion, since you mentioned it's not ideal to mention the "unsupported" option ([comment](https://github.com/stolostron/rhacm-docs/pull/6891/files#r1747221931)), I figured we would not have the below section at all (completely remove section `[#add-host-full-iso]`) and only have this note here.

swopebe: I gotcha! I didn't see @briantward's comment.

swopebe: Please review, knowing that there are some formatting issues I am working through. @CrystalChun @briantward

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -14,22 +14,49 @@ After creating your host inventory (infrastructure environment) you can discover
...
 . Select your infrastructure environment from the list.
 
 . Click *Add hosts* and select *With Discovery ISO*.
 
-You now see a URL to download the ISO. Booted hosts appear in the host inventory table. Hosts might take a few minutes to appear. You must approve each host before you can use it. You can select hosts from the inventory table by clicking *Actions* and selecting *Approve*.
++
+You now see a URL to download the ISO file. Booted hosts appear in the host inventory table. Hosts might take a few minutes to appear. 
+
++
+*Note:* By default, the ISO that is provided is a _minimal_ ISO. The minimal ISO does not contain the root file system, `RootFS`. The `RootFS` is downloaded later. To display full ISO, replace `minimal.iso` in the URL with `full.iso`.
```
CrystalChun, line 35: Note looks great, minor suggested word change:
```suggestion
*Note:* By default, the ISO that is provided is a _minimal_ ISO. The minimal ISO does not contain the root file system, `RootFS`. The `RootFS` is downloaded later. To get the full ISO, replace `minimal.iso` in the URL with `full.iso`.
```
May we also have this note added below after this line too https://github.com/stolostron/rhacm-docs/pull/6891/files#diff-a31f1109b991f81d062625730bb1e5e4a613d186c9885cc4feddc5fea2a6aa30R55

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -14,22 +14,49 @@ After creating your host inventory (infrastructure environment) you can discover
...
++
+[source,bash]
+----
+https://assisted-image-service-assisted-installer.apps.example-acm-hub.com/byapikey/eyJhbGciOiJFUzI1NiIsInC93XVCJ9.eyJpbmZyYV9lbnZfaWQcTA0Y38sWVjYi02MTA0LTQ4NDMtODasdkOGIxYTZkZGM5ZTUifQ.3ydTpHaXJmTasd7uDp2NvGUFRKin3Z9Qct3lvDky1N-5zj3KsRePhAM48aUccBqmucGt3g/4.16/x86_64/minimal.iso
+----
+
+. Use the URL to download the ISO file and boot your hosts with the ISO file.
 
-Each booted host creates an `Agent` resource in the same namespace. You must approve each host before you can use it.
+. Approve each host.
```
CrystalChun: Maybe a procedure like below could be added too:

---
To approve the hosts, do the following:

1. List all of your `Agents`
```bash
oc get agent -n <infra env namespace>
```
The output will appear like this
```bash
NAME                                   CLUSTER   APPROVED   ROLE          STAGE
24a92a6f-ea35-4d6f-9579-8f04c0d3591e             false      auto-assign   
```
2. Approve any `Agent` that is not yet approved
```bash
oc patch agent -n <infra env namespace> <agent name> -p '{"spec":{"approved":true}}' --type merge
```
3. List all of the `Agent` CRs again to ensure they're all approved
```bash
oc get agent -n <infra env namespace>
```
The output will appear like this:
```bash
NAME                                   CLUSTER   APPROVED   ROLE          STAGE
173e3a84-88e2-4fe1-967f-1a9242503bec             true       auto-assign    
```

clusters/cluster_lifecycle/cim_add_host.adoc
swopebe: Done @CrystalChun , if all is well, please approve. If not, let me know. :-)

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -14,22 +20,93 @@ After creating your host inventory (infrastructure environment) you can discover
...
+----
+
++
+See the following output:
+
++
+[source,bash]
+----
+https://assisted-image-service-assisted-installer.apps.example-acm-hub.com/byapikey/eyJhbGciOiJFUzI1NiIsInC93XVCJ9.eyJpbmZyYV9lbnZfaWQcTA0Y38sWVjYi02MTA0LTQ4NDMtODasdkOGIxYTZkZGM5ZTUifQ.3ydTpHaXJmTasd7uDp2NvGUFRKin3Z9Qct3lvDky1N-5zj3KsRePhAM48aUccBqmucGt3g/4.16/x86_64/minimal.iso
+----
```
CrystalChun, line 60: Could we add this note here too? In case the user doesn't read the UI portion above 

```suggestion
----

*Note:* By default, the ISO that is provided is a _minimal_ ISO. The minimal ISO does not contain the root file system, `RootFS`. The `RootFS` is downloaded later. To display full ISO, replace `minimal.iso` in the URL with `full.iso`.
```

CrystalChun (commented):
Just one more comment, but otherwise looks good!

CrystalChun (approved):
/lgtm

Thanks Brandi!

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -14,22 +20,95 @@ After creating your host inventory (infrastructure environment) you can discover
...
+[source,bash]
+----
+oc get infraenv -n <infra env namespace> <infra env name> -o jsonpath='{.status.isoDownloadURL}'
+----
+
++
+See the following output:
+
++
+[source,bash]
```
oafischer: This is not a command, so we need to remove the `[source,bash]`

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -14,22 +20,95 @@ After creating your host inventory (infrastructure environment) you can discover
...
+----
+oc get agent -n <infra env namespace>
+----
+
++
+You get an output that is similar to the following output:
+
++
+[source,bash]
```
oafischer: Same here

clusters/cluster_lifecycle/cim_add_host.adoc
```diff
@@ -14,22 +20,95 @@ After creating your host inventory (infrastructure environment) you can discover
...
+----
+oc get agent -n <infra env namespace>
+----
+
++
+You get an output that is similar to the following output with a `true` value:
 
-Each booted host creates an `Agent` resource in the same namespace. You must approve each host before you can use it.
++
+[source,bash]
```
oafischer: One more

oafischer (approved):
Looks good, just need to remove the source highlighting from the outputs.

clusters/cluster_lifecycle/cim_add_host.adoc
swopebe: @oafischer I add these because it is part of the terminal output; let's look and see if others are doing it. Why would we not highlight it as part of the terminal?

swopebe: DMing about this quickly

oafischer: You are right, if they are terminal outputs then they should be highlighted. I jumped in thinking it was a non-terminal output.

PR #6905, opened 2024-09-12
[ACM-14264]: Backport proxy prereq to HCP docs

lahinson:
Related Jira: https://issues.redhat.com/browse/ACM-14264

After the wording is approved, it needs to also be backported to the following versions:
- 2.10
- 2.9

### DIFF
--- a/clusters/hosted_control_planes/kubevirt_intro.adoc
+++ b/clusters/hosted_control_planes/kubevirt_intro.adoc
@@ -57,6 +57,8 @@ oc patch storageclass ocs-storagecluster-ceph-rbd -p '{"metadata": {"annotations
 oc get managedclusters local-cluster
 ----
 
+- If you are using a proxy, the `NO_PROXY` environment variable in {ocp-virt-short} must have an IP address of at least 172.20.0.1.
+
 [#firewall-port-reqs-kubevirt]
 == Firewall and port requirements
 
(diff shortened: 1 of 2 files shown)


clusters/hosted_control_planes/kubevirt_intro.adoc
```diff
@@ -57,6 +57,8 @@ oc patch storageclass ocs-storagecluster-ceph-rbd -p '{"metadata": {"annotations
 oc get managedclusters local-cluster
 ----
 
+- If you are using a proxy, the `NO_PROXY` environment variable in {ocp-virt-short} must have an IP address of at least 172.20.0.1.
```
enxebre: This apply to all platforms.
Also this won't be necessary after https://github.com/openshift/hypershift/pull/4693 is shipped by mce
I think we'll also need to document how to configure proxy in general, e.g. run hcp create cluster --render, edit hc.spec.config.proxy...

enxebre: Also it doesn't really must, it's a config choice. I'd say something like "if you are using proxy and you want traffic from pods to the kube api server to not use the proxy you need to add our default kas advertised address 172.20.0.1 to the no_proxy list.

swopebe (approved):
looks great

PR #6916, opened 2024-09-16
https://issues.redhat.com/browse/ACM-13475 Keep hub active docs

oafischer:
Train 19, 2.12 only


business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,145 @@
...
+ name: schedule-acm-msa
+ namespace: open-cluster-management-backup
+spec:
+ veleroSchedule: 0 */1 * * *
+ veleroTtl: 120h
+ useManagedServiceAccount: true
+ paused: true
+----
++
+*Note:* If you do not pause the backup schedule on the primary hub cluster before continuing to the next step, you can also back up the managed cluster by adding the `import.open-cluster-management.io/disable-auto-import` annotation to the `ManagedCluster` resource.
```
birsanv: This is incorrect, I wanted to point out that it's not desired to have the resource backed up with the annotation, for the reason descried in the next 2 sentences.

Replace lines 40- 43 with :

*Note:* If you do not pause the backup schedule on the primary hub cluster before continuing to the next step, you may end up backing up the `ManagedCluster` resource with the `import.open-cluster-management.io/disable-auto-import` annotation set, if a backup is running while the annotation is set on the resource.

The result of backing up the `ManagedCluster` resource with the `import.open-cluster-management.io/disable-auto-import` annotation set is that the managed cluster will not be automatically imported when you restore this backup data on a new hub cluster.

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,145 @@
...
+
++
+If you set the annotation, the managed cluster is not automatically imported when you restore it on a new hub cluster. If you set the annotation, the hub cluster does not connect to the managed cluster when the hub cluster cannot access the managed cluster.
+
+. *Optional:* Tag the primary hub cluster resources with backup annotation by applying the following `Restore` resource:
```
birsanv: I don't like having the optional is front of the title ..

Can we have a shorter title:
Tag the primary hub cluster resources with the backup annotation

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,145 @@
...
+
++
+*Note:* Completing the restore operation on the primary hub cluster prepares the hub cluster for reuse as a primary or secondary hub cluster after the recovery process completes. You can skip this optional step if you do not want to use this hub cluster as a primary or secondary hub cluster after the restore operation completes.
```
birsanv: Replace with : 

Apply the following restore operation on the primary hub cluster in order to prepare the hub cluster for reuse as a primary or secondary hub cluster after the recovery process completes. You can skip this step if you do not want to use this hub cluster as a primary or secondary hub cluster after the restore operation completes.

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+.. Apply the file by running the following command:
+
++
+[source,bash]
+----
+oc -f apply restore-acm.yaml
+----
+
+.. Run the `Restore` resource to tag all hub cluster resources that you want to back up with the `velero.io/backup-name: backupName` label, where `backupName` is the name of the latest backup created by the primary hub cluster.
```
birsanv: This is not an actual step, remove it from here. The Restore is already running as a result of `oc -f apply restore-acm.yaml` 

Replace lines 77-80 with : 
Running the `Restore` resource will result in tagging all hub cluster resources that should be backed up with the `velero.io/backup-name: backupName` label, where `backupName` is the name of the latest backup created by the primary hub cluster. As a result, any restore operation run on the primary cluster using a Restore resource with the `cleanupBeforeRestore` option set to `CleanupRestored` will process these resources, and will delete any of these resources if they are not part of the latest backup.

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,145 @@
...
+* <<dr4hub-keep-hub-alive-restore-delete,Deleting resources after restoration>>
+
+[#dr4hub-keep-hub-alive-restore-prepare]
+== Preparing the primary hub cluster
+
+Before running the restore operation and keeping the hub cluster active, you must prepare the primary hub cluster.
+
+Pause the backup schedule on the primary hub cluster and disable the automatic import operation for the managed clusters by completing the following steps: 
+
+. If the backup hub cluster has a `BackupSchedule` resource in the `Enabled` state, pause the resource by setting the `paused` property to `true` on the `BackupSchedule`. See the following example:
```
birsanv: Let's keep this title brief and explain below the title what we are doing and why

Line 23 should be 
. Pause BackupSchedule

and below : 
If the backup hub cluster has a `BackupSchedule` resource in the `Enabled` state, pause the resource by setting the `paused` property to `true` on the `BackupSchedule`. See the following example:

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,145 @@
...
+----
+oc -f apply restore-acm.yaml
+----
+
+.. Run the `Restore` resource to tag all hub cluster resources that you want to back up with the `velero.io/backup-name: backupName` label, where `backupName` is the name of the latest backup created by the primary hub cluster.
+
++
+*Note:* Hub cluster resources that you tag show that they were created by a previous restored backup. Any new restore operation runs on the primary cluster by using the `cleanupBeforeRestore: CleanupRestored` option, processes hub cluster resources, and deletes the resources if they are not part of the latest backup.
+
+. Disable the automatic import for managed clusters you want to move by setting the `import.open-cluster-management.io/disable-auto-import: ''` annotation on the `ManagedCluster` global resource for all managed clusters that you want to move to the restoration hub cluster. See the following example:
```
birsanv: Replace with : 

. Disable the automatic import for managed clusters

set the `import.open-cluster-management.io/disable-auto-import: ''` annotation on the `ManagedCluster` global resource for all managed clusters that will be moved to the restore hub.
This annotation will prevent the backup hub trying to recover any managed cluster after the managed cluster is moved to the restore hub.

See the following example:

Remove line 92

birsanv (changes requested):

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: If it's an optional step we do have to indicate it like that but I can try to make the step shorter

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: Since it's already in the example we can probably make it even shorter than that, I'll update it

oafischer:
Thanks for the review @birsanv, made some updates if you want to take a look at it again

business_continuity/backup_restore/backup_active_restore.adoc
birsanv: verified fixed

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,141 @@
...
+.. Apply the file and run the restore operation by running the following command:
+
++
+[source,bash]
+----
+oc -f apply restore-acm.yaml
+----
+
++
+Running the `Restore` resource tags all hub cluster resources that are backed up with the `velero.io/backup-name: backupName` label, where `backupName` is the name of the latest backup created by the primary hub cluster. As a result, any restore operation that runs on the primary cluster uby sing a `Restore` resource with the `cleanupBeforeRestore` option set to `CleanupRestored` processes these resources, and deletes all resources if they are not part of the latest backup.
```
birsanv: Please make these change : 

Running the `Restore` resource tags all hub cluster resources that *should be* backed up with the `velero.io/backup-name: backupName` label, where `backupName` is the name of the latest backup created by the primary hub cluster. As a result, any restore operation that runs on the primary cluster *using*  a `Restore` resource with the `cleanupBeforeRestore` option set to `CleanupRestored` processes these resources, and deletes *any resource which is no longer* part of the latest backup.

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,141 @@
...
+
++
+[source,yaml]
+----
+annotations:
+   import.open-cluster-management.io/disable-auto-import: ''
+----
+
++
+*Note:* Set the annotation on the `ManagedCluster` global resource for all managed clusters that you want to move to the restoration hub cluster. This annotation prevents the backup hub cluster from recovering any managed cluster after the managed cluster moves to the restoration hub cluster.
```
birsanv: typo : restoration hub should be **restore** hub - 2 places

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,141 @@
...
+. Apply the file and run the restore operation by running the following command:
+
++
+[source,bash]
+----
+oc -f apply restore-acm.yaml
+----
+
+[#dr4hub-keep-hub-alive-restore-delete]
+== Deleting resources after restoration
```
birsanv: Can we call this : **Clean up managed cluster resources on backup hub**

(The purpose of this step is to clean up managed cluster info from the primary hub; it just happens that this involves deleting the managed cluster resources )

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,141 @@
...
+
+After the restoration and after the managed clusters successfully connect to the new hub cluster, you can clean up the managed cluster resources from the initial backup hub cluster. You might want to skip cleaning the resources if you want to restore the data to the backup after your recovery test completes.
```
birsanv: typo : restoration should be restore

birsanv (changes requested):
there are still a few changes that I am requesting Oliver

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: We can't say "should" but I assume the intention is "the ones you want to back up", right? I'll update it

oafischer:
@birsanv Made some more updates

business_continuity/backup_restore/backup_active_restore.adoc
birsanv: The user has no control over what is going to be backed up so 'the ones you want to back up' is not correct.

I meant to say : the ones that would be backed up if you would create a BackupSchedule and start backing up resources.

So I meant something like 

Running the Restore resource tags all hub cluster resources that would be backed up if you enabled the BackupSchedule. These resources will be tagged with the velero.io/backup-name: backupName label, where backupName is the name of the latest backup created by the primary hub cluster. As a result, any restore operation that runs on the primary cluster using a Restore resource with the cleanupBeforeRestore option set to CleanupRestored processes these resources, and deletes any resource which is no longer part of the latest backup.

business_continuity/backup_restore/backup_active_restore.adoc
birsanv: this looks good, thanks !

business_continuity/backup_restore/backup_active_restore.adoc
birsanv: looks good, thanks

business_continuity/backup_restore/backup_active_restore.adoc
birsanv: missed one typo here - should be restore

after the managed cluster moves to the **restoration** hub cluster.

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: I see, thanks for clarifying. I updated it, let me know if it looks good now. Thanks!

business_continuity/backup_restore/backup_active_restore.adoc
birsanv: This is still not exactly clear ..

Issues that I have with the current content : 

line 80 : 
Is the title so I expect is short and concise. 
See the following example should be in the next line .. 
There is no info what to do with that annotation, where to set it. You have to read the Note to get it

I suggest : 


. Disable the automatic import for managed clusters. 

Set the `import.open-cluster-management.io/disable-auto-import` annotation on the `ManagedCluster` global resource for all managed clusters that you want to move to the restore hub cluster. This annotation prevents the backup hub cluster from recovering any managed cluster after the managed cluster moves to the restoration hub cluster.

See the following example:

+
[source,yaml]
----
annotations:
   import.open-cluster-management.io/disable-auto-import: ''
----

+

business_continuity/backup_restore/backup_active_restore.adoc
birsanv: I would still take out the **Note**, just make it a normal content

And the title, take out *Optional:* , this step is not optional unless you decide to not use the cluster for a backup operation, which is a very remote option .. and we already tell that this is optional **can skip this optional step if you do not want to use this hub cluster** ..
If you add the Optional in the title it makes it look less important; my opinion

business_continuity/backup_restore/backup_active_restore.adoc
birsanv: looks good, thanks

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: OK I took out the optional and turned the paragraph into regular content

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: Updated

business_continuity/backup_restore/backup_active_restore.adoc
birsanv: looks good, thanks

business_continuity/backup_restore/backup_active_restore.adoc
birsanv: looks good

business_continuity/backup_restore/backup_active_restore.adoc
birsanv: looks good

birsanv (approved):
lgtm

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
+[#keep-hub-active-restore]
+= Keeping the primary hub cluster active during a restore process
+
+The restore operation typically happens after a disaster occurs and the primary hub cluster fails. As a result, the primary hub cluster is not available when the restore operation starts.
```
jc-berger: Optional: user focus/active voice: 

```suggestion
The restore operation typically happens after a disaster occurs and the primary hub cluster fails. As a result, you cannot access the primary hub cluster when the restore operation starts.
```

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+In some cases, you might want to keep the primary hub cluster active during and after a hub cluster restoration. For example, you might want to run a disaster recovery test and not want the primary hub cluster to stop during the test.
```
jc-berger: Optional: especially since my rewrite might not be better, lol. 

But consider leading with the "why the user would do this" then give the action:

```suggestion
If you want to run a disaster recovery test and not want the primary hub cluster to stop during the test,  keep the primary hub cluster active during and after a hub cluster restoration.
```

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+If you keep the primary hub cluster active while restoring the hub cluster data on a different hub cluster, the initial hub cluster can access the managed clusters. As a result, both hub clusters compete for the managed clusters. Any policy or application changes you make on the restored hub cluster might be undone if the initial hub cluster takes control of the managed clusters.
+
+Continue reading to learn how to keep the primary hub cluster active during and after a restore operation.
```
jc-berger, line 10: User focus/action voice

```suggestion
To keep the primary hub cluster active during and after a restore operation, complete the following sections: 
```

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
+* <<dr4hub-keep-hub-alive-restore-prepare,Preparing the primary hub cluster>>
+* <<dr4hub-keep-hub-alive-restore-run,Running the restore operation on the new hub cluster>>
+* <<dr4hub-keep-hub-alive-restore-delete,Deleting resources after restoration>>
+
+[#keep-hub-active-restore-prepare]
+== Preparing the primary hub cluster
+
+Before running the restore operation and keeping the hub cluster active, you must prepare the primary hub cluster.
+
+Pause the backup schedule on the primary hub cluster and disable the automatic import operation for the managed clusters by completing the following steps:
```
jc-berger: You give these actions as steps, so what do you think about making this section more concise? 

```suggestion
To prepare the primary hub cluster, complete the following steps: 
```

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
+ name: schedule-acm-msa
+ namespace: open-cluster-management-backup
+spec:
+ veleroSchedule: 0 */1 * * *
+ veleroTtl: 120h
+ useManagedServiceAccount: true
+ paused: true
+----
++
+*Note:* If you do not pause the backup schedule on the primary hub cluster before continuing to the next step, you might back up the `ManagedCluster` resource with the `import.open-cluster-management.io/disable-auto-import` annotation set if a backup is running while the annotation is set on the resource.
```
jc-berger: You have multiple notes, so maybe a bullet list? 

```suggestion
*Notes:*
-  If you do not pause the backup schedule on the primary hub cluster before continuing to the next step, you might back up the `ManagedCluster` resource with the `import.open-cluster-management.io/disable-auto-import` annotation set if a backup is running while the annotation is set on the resource.
```

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
+
++
+Backing up the `ManagedCluster` resource with the `import.open-cluster-management.io/disable-auto-import` annotation causes the managed cluster to not automatically import when you restore the backup data on a new hub cluster.
```
jc-berger:
```suggestion
- Backing up the `ManagedCluster` resource with the `import.open-cluster-management.io/disable-auto-import` annotation causes the managed cluster to not automatically import when you restore the backup data on a new hub cluster.

```

jc-berger: User focus, and for consistency, keep similar sentence structure to the previous Notes statement: 

```suggestion
If you back up the `ManagedCluster` resource with the `import.open-cluster-management.io/disable-auto-import` annotation,  the managed cluster cannot automatically import when you restore the backup data on a new hub cluster.
```

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+. Tag the primary hub cluster resources with the backup annotation. Complete the following steps:
+
++
+Apply the following restore operation on the primary hub cluster to prepare the hub cluster for reuse as a primary or secondary hub cluster after the recovery process completes. You can skip this step if you do not want to use this hub cluster as a primary or secondary hub cluster after the restore operation completes.
```
jc-berger: Since this action comes right after you say "complete the following steps" I think it needs to be a sub step just like `Create a file` and  `Add the following` You can add an optional note if you want. 

See both of my suggestions, and choose which one you think is best. 

```suggestion
. . Apply the following restore operation on the primary hub cluster to prepare the hub cluster for reuse as a primary or secondary hub cluster after the recovery process completes. Skip this step if you do not want to use this hub cluster as a primary or secondary hub cluster after the restore operation completes.
```

jc-berger:
```suggestion
. . *Optional:* If you do not want to use this hub cluster as a primary or secondary hub cluster after the restore operation completes, .Apply the following restore operation on the primary hub cluster to prepare the hub cluster for reuse as a primary or secondary hub cluster after the recovery process completes. 
```

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
+.. Apply the file and run the restore operation by running the following command:
+
++
+[source,bash]
+----
+oc -f apply restore-acm.yaml
+----
+
++
+Running the `Restore` resource tags all hub cluster resources that would be backed up if you enabled the `BackupSchedule`. The resources are tagged with the `velero.io/backup-name: backupName` label, where `backupName` is the name of the latest backup created by the primary hub cluster. As a result, any restore operation that runs on the primary cluster by using a `Restore` resource with the `cleanupBeforeRestore` option set to `CleanupRestored` processes these resources, and deletes all resources if they are not part of the latest backup.
```
jc-berger: Is this a note to provide users extra context? Or is it an optional step if they want to tag hub cluster resources? 

See both suggestions: 

```suggestion
*Note:* When you run the `Restore` resource, it tags all hub cluster resources that would be backed up if you enabled the `BackupSchedule`. The resources are tagged with the `velero.io/backup-name: backupName` label, where `backupName` is the name of the latest backup created by the primary hub cluster. As a result, any restore operation that runs on the primary cluster by using a `Restore` resource with the `cleanupBeforeRestore` option set to `CleanupRestored` processes these resources, and deletes all resources if they are not part of the latest backup.
```

jc-berger:
```suggestion
. . *Optional:* Run the `Restore` resource to tag all hub cluster resources that would be backed up if you enabled the `BackupSchedule`. The resources are tagged with the `velero.io/backup-name: backupName` label, where `backupName` is the name of the latest backup created by the primary hub cluster. As a result, any restore operation that runs on the primary cluster by using a `Restore` resource with the `cleanupBeforeRestore` option set to `CleanupRestored` processes these resources, and deletes all resources if they are not part of the latest backup.
```

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+. Disable the automatic import for managed clusters.
+
++
+Set the following annotation on the `ManagedCluster` global resource for all managed clusters that you want to move to the restore hub cluster. This annotation prevents the backup hub cluster from recovering any managed cluster after the managed cluster moves to the restore hub cluster:
```
jc-berger: Isn't this another step? See suggestion:

```suggestion
. Set the following annotation on the `ManagedCluster` global resource for all managed clusters that you want to move to the restore hub cluster. By setting the annotation, you prevent the backup hub cluster from recovering any managed cluster after the managed cluster moves to the restore hub cluster:
```

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
+[source,yaml]
+----
+annotations:
+   import.open-cluster-management.io/disable-auto-import: ''
+----
+
+[#keep-hub-active-restore-run]
+== Running the restore operation on the new hub cluster
+
+You can now run the restore operation on the new hub cluster and move your managed clusters. The managed clusters connect to the new hub cluster and do not move back to the initial hub cluster because the automatic import feature is now disabled on the backup hub cluster.
```
jc-berger: See following suggestion to help with user flow: 

```suggestion
After preparing the primary hub cluster,  run the restore operation on the new hub cluster and move your managed clusters. The managed clusters connect to the new hub cluster and do not move back to the initial hub cluster because the automatic import feature is now disabled on the backup hub cluster.
```

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+Run the restore operation on the new hub cluster by completing the following steps:
+
+. Open or create a file named `restore-acm.yaml`.
```
jc-berger: Optional:

We don't need to give the users various options. Consider choosing which verb, `Open` or `Create` that you think is best applicable for this step, and only use that verb.

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
++
+[source,bash]
+----
+oc -f apply restore-acm.yaml
+----
+
+[#keep-hub-active-restore-clean]
+== Cleaning up managed cluster resources on the backup hub cluster
+
+After the restore and after the managed clusters successfully connect to the new hub cluster, you can clean up the managed cluster resources from the initial backup hub cluster. You might want to skip cleaning the resources if you want to restore the data to the backup after your recovery test completes.
```
jc-berger:
```suggestion
After the restore and after the managed clusters successfully connect to the new hub cluster, clean up the managed cluster resources from the initial backup hub cluster. If you want to restore the data to the backup after your recovery test completes, skip cleaning the resources 
```

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+To clean up the managed clusters from the backup hub cluster, delete the `ManagedCluster` global resource from the backup hub cluster for each of the managed clusters that you moved to the new hub cluster by using the restore operation.
```
jc-berger: Optional:

If it's important for the user to complete these actions, maybe we should make it a procedure? What do you think of the following steps? 
```suggestion
To clean up the managed clusters from the backup hub cluster, complete the following steps:

. Make sure that the managed cluster status is `Unknown` on the primary hub before deleting the `ManagedCluster` global resource. If the status is not `Unknown`, your workloads are uninstalled from the managed cluster.
. Decide if you want to remove the `ManagedCluster` global resource. Doing so deletes the managed cluster namespace.
. Delete the `ManagedCluster` global resource from the backup hub cluster for each of the managed clusters that you moved to the new hub cluster by using the restore operation.
```

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
+
+*Important:*
```
jc-berger, line 142: If you don't like the procedure steps I suggested above, I would recommend moving the Important content before telling the user to delete the managed clusters. Something like the following: 

```suggestion
*Important:* 

*  Make sure that the managed cluster status is `Unknown` on the primary hub before deleting the `ManagedCluster` global resource. If the status is not `Unknown`, your workloads are uninstalled from the managed cluster.
* Removing the `ManagedCluster` global resource also deletes the managed cluster namespace.

If you decide to clean up the managed clusters from the backup hub cluster, delete the `ManagedCluster` global resource from the backup hub cluster for each of the managed clusters that you moved to the new hub cluster by using the restore operation.
```

business_continuity/backup_restore/backup_active_restore.adoc
swopebe: and _do_ not want

swopebe: extra space here

business_continuity/backup_restore/backup_active_restore.adoc
swopebe: complete the following sections:

they are not completing a _section_; they are completing a process. You could say _see the following sections_, up to you what change you want to make here. I also don't think line 10 is wrong, but like Jake's suggestion, minus the object change.

business_continuity/backup_restore/backup_active_restore.adoc
swopebe: extra space

business_continuity/backup_restore/backup_active_restore.adoc
swopebe: Back up and restore has these really long paragraphs throughout the doc, and I know this is because it comes to us in drafts. We really need to break these up into smaller chunks when we get them.

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,138 @@
...
+. Disable the automatic import for managed clusters.
+
++
+Set the following annotation on the `ManagedCluster` global resource for all managed clusters that you want to move to the restore hub cluster. This annotation prevents the backup hub cluster from recovering any managed cluster after the managed cluster moves to the restore hub cluster:
++
+[source,yaml]
+----
+annotations:
+   import.open-cluster-management.io/disable-auto-import: ''
+----
```
swopebe, line 95: formatting error here; this should be grayed out.

business_continuity/backup_restore/backup_active_restore.adoc
swopebe: 94 is doing the opposite of some of the other suggestions, in that we are returning to the gerund here. (extra space here, too)

After preparing becomes: _After you prepare._ 

The original 94 is also active. This seems to deviate from the previous paragraphs a bit.

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: I'll keep this one as is

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: Updated

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: I'll update this separately to follow @swopebe's suggestion

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: Updated separately since there are formatting errors in the suggestion

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: Updated

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: Not sure about this one, both apply. Open applies if the optional step was done from the previous section. Create applies if the user did not do the optional step. What's the best way to resolve this?

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: The procedure is fine, I went with that

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: Fixed

oafischer:
Thanks for the review @swopebe @jc-berger I've addressed all the comments with one question left, would appreciate one final review

business_continuity/backup_restore/backup_active_restore.adoc
swopebe: then I think you need to say that.

Open the file that you created. If you did not...then create. (But if they need a file created, then the optional step is not fully optional, no?)

swopebe: I would look at the optional section again and see if you can prevent the user from having to ask themselves if they need to create a file at this point, or can they just open it. (But then I have to ask why "open a file" is a single step. Usually we would say "Go to the file and add the following content" or "Add the following content to a file named ..."

oafischer: Good point on the last comment, I think we can just direct them to that file(name), if it's not there it should be pretty clear that the user needs to create it first.

swopebe: not sure I agree with that fully because we are assuming something here. Will DM.

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -0,0 +1,147 @@
...
++
+- If you back up the `ManagedCluster` resource with the `import.open-cluster-management.io/disable-auto-import` annotation, the managed cluster cannot automatically import when you restore the backup data on a new hub cluster.
+
+
+. Tag the primary hub cluster resources with the backup annotation. Complete the following steps:
+
+
+.. *Optional:* If you do not want to use this hub cluster as a primary or secondary hub cluster after the restore operation completes, apply the following restore operation on the primary hub cluster to prepare the hub cluster for reuse as a primary or secondary hub cluster after the recovery process completes.
+
+.. Create a file named `restore-acm.yaml`.
```
swopebe, line 49: @oafischer it seems to me that creating the file is not optional. (?) If they need it later. What seems to be optional is line 54. (?)

So we can move create up above the optional explanation, maybe.

So later the open or create question is resolved because the user has the file. So they can just "Go to x file and add the following..." -- would that work?

swopebe (approved):

business_continuity/backup_restore/backup_active_restore.adoc
oafischer: That's a really good solution, I'll update it. Thanks for the help!

oafischer:
All comments addressed, merging

PR #6937, opened 2024-09-17
Update create_clustersetbinding.adoc

bemarek:
As per definition of ManagedClusterSetBinding.spec.clusterSet:

>ClusterSet is the name of the ManagedClusterSet to bind. It must match the instance name of the ManagedClusterSetBinding.

### DIFF
--- a/clusters/cluster_lifecycle/create_clustersetbinding.adoc
+++ b/clusters/cluster_lifecycle/create_clustersetbinding.adoc
@@ -35,7 +35,7 @@ apiVersion: cluster.open-cluster-management.io/v1beta2
 kind: ManagedClusterSetBinding
 metadata:
   namespace: <namespace>
-  name: <cluster_name>
+  name: <cluster_set>
 spec:
   clusterSet: <cluster_set>
 ----


dockerymick:
@bemarek is there an issue related to this PR?

bemarek:
What do you mean by issue? I've suggested the change which, in my opinion is reasonable. The details are in my initial comment.

dockerymick:
@bemarek we have a process where a doc issue needs to be created before we make the change. There is a doc issue template. This helps us keep track of changes for other versions and is guidance on what issues to prioritize. Is this also applicable to earlier versions? I assume that this is also applies to earlier versions such as 2.11 and 2.10. If we did not make this update, would it break the user environment or is the a customer portal issue associated with this?

I appreciate you updating the file. I can create a doc issue now, but in the future please create a Doc issue

clusters/cluster_lifecycle/create_clustersetbinding.adoc
```diff
@@ -35,7 +35,7 @@ apiVersion: cluster.open-cluster-management.io/v1beta2
 kind: ManagedClusterSetBinding
 metadata:
   namespace: <namespace>
-  name: <cluster_name>
+  name: <cluster_set>
```
dockerymick, line 38: @bemarek If we update this to `<cluster_set_name>`, would it still be accurate to your suggestion?

dockerymick:
Here is the link to the doc issue: https://issues.redhat.com/browse/ACM-14706

PR #6943, opened 2024-09-18
https://issues.redhat.com/browse/ACM-13563 Fix OCP links

oafischer:
All links were manually searched and picked/replaced individually.


dockerymick (commented):
Based on our [support statement](https://github.com/stolostron/rhacm-docs/blob/2.12_stage/clusters/install_upgrade/install_intro.adoc), I believe these should be pointing to 4.14.

oafischer:
Updated to 4.14 @dockerymick

PR #6949, opened 2024-09-18
https://issues.redhat.com/browse/ACM-13088

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-13088


dhaiducek:
This looks good--thanks for the quick update, @jc-berger!

Similarly, could you also update this line in the `config_gk_operator.adoc`?

https://github.com/stolostron/rhacm-docs/blob/2b1041f2fda221c15df0c5aa292e8a5382d8999d/governance/gatekeeper_operator/config_gk_operator.adoc?plain=1#L4

Maybe something like this?

> Install the {gate} from the {ocp} catalog to install Gatekeeper on your cluster. See _Adding Operators to a cluster_ in the {ocp-short} documentation in the _Additional resources_ section. With {acm-short} you can use a policy to install the {gate} by using the governance framework.

drkho:
Sounds good to me. Thanks, all.

governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -1,7 +1,7 @@
 [#config-gk-operator]
 = Configuring the {gate} 
 
-To install Gatekeeper on your hub cluster, install the {gate} from the {ocp} catalog. See _Adding Operators to a cluster_ in the {ocp-short} documentation and go to the _Additional resources_ section. Configure the {acm-short} Gatekeeper policy by using the governance framework.
+Install the {gate} from the {ocp} catalog to install Gatekeeper on your cluster. See _Adding Operators to a cluster_ in the {ocp-short} documentation and go to the _Additional resources_ section. With {acm-short} you can use a policy to install the {gate} by using the governance framework.
```
swopebe: "install to install", do we mean this or do we mean to _Access the catalog to install_? 

Please remove the citation from the short description, as this is only a description and should not interrupt the user. This trail to the OCP doc and then to additional info is very odd to me at the beginning of the topic. Should this be appended to a step if it is something the user needs? Or a prereq?

Also, throughout the doc, I think we mean _OLM catalog._ (see the rest of the doc for how we write that, I don't think we call anything the OCP catalog)

dockerymick (approved):
This looks good to me! Thanks Jake!

governance/gatekeeper_operator/gk_operator_overview.adoc
```diff
@@ -1,7 +1,7 @@
 [#gk-operator-overview]
 = {gate} overview 
 
-Gatekeeper is a validating webhook with auditing capabilities. Install the {gate} on a {ocp} cluster from the {ocp} operator catalog. With {acm}, you can install Gatekeeper on your hub cluster by using the {gate} policy. After you install Gatekeeper, use it for the following benefits: 
+The {gate} installs Gatekeeper which is a validating webhook with auditing capabilities. Install the {gate} on a {ocp} cluster from the {ocp} operator catalog. With {acm}, you can install Gatekeeper on your hub cluster by using the {gate} policy. After you install Gatekeeper, use it for the following benefits:
```
swopebe: missing comma here:

Gatekeeper, which


and again, please look at the name of the catalog here and throughout.

governance/gatekeeper_operator/config_gk_operator.adoc
swopebe: For some reason we don't have a prereq section, which we discussed a few meetings ago. All the tasks generally need that and it may be where you put your citations or links.

swopebe (changes requested):

governance/gatekeeper_operator/gk_operator_overview.adoc
swopebe: @dhaiducek can you check into this? I think we mean the OLM Catalog?

swopebe (changes requested):

governance/gatekeeper_operator/gk_operator_overview.adoc
swopebe: See here where we point to OLM: https://github.com/stolostron/rhacm-docs/blob/64fecc1cc6932390eb7f411d4c5243b43e4c69c7/governance/ansible_grc.adoc#prerequisites

governance/gatekeeper_operator/config_gk_operator.adoc
dhaiducek: @swopebe You're correct, it's the OLM catalog.

The "install to install" is technically correct. The flow is that the operator is installed on the cluster through OLM resources, after which the creation of the custom Gatekeeper operator resource on the cluster tells the Gatekeeper operator to install Gatekeeper.

dhaiducek (approved):
LGTM! Thanks for the update!

governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -1,16 +1,18 @@
...
-To install Gatekeeper on your hub cluster, install the {gate} from the {ocp} catalog. See _Adding Operators to a cluster_ in the {ocp-short} documentation and go to the _Additional resources_ section. Configure the {acm-short} Gatekeeper policy by using the governance framework.
+Install the {gate} from the {olm} catalog to install Gatekeeper on your cluster. With {acm-short} you can use a policy to install the {gate} by using the governance framework. After you install the {gate}, configure the {gate} custom resource to install Gatekeeper.
 
-After you install the {gate}, you must configure the {gate} custom resource to install Gatekeeper.
+[#configure-gk-operator-prerequisites]
+== Prerequisites
 
-*Required access*: Cluster administrator
+- *Required access*: Cluster administrator. 
+- See _Adding Operators to a cluster_ and the _Additional resources_ section in the link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.15/html/operators/administrator-tasks#olm-adding-operators-to-a-cluster[{ocp-short} documentation].
```
swopebe, line 11: Is this a prereq to configuring? I am still a bit confused about why the user needs to go to this link. TY for using the correct version of the link and thinking ahead there!

swopebe: This looks really good! Making it clear if or why this is a prereq is my only remaining comment. (If it is not a prereq and just supplemental info, then we don't need the section. I just wanted to make sure we did not place this in the short description.)

- *Required access*: Cluster administrator. (Can be in the prereqs but it can also stand alone if you have no actual prereqs.)

swopebe (approved):

PR #6899, opened 2024-09-10
https://issues.redhat.com/browse/ACM-14160 Adding double quotes where needed and making some improvements

dockerymick:

PR #6953, opened 2024-09-18
https://issues.redhat.com/browse/ACM-14160 for 2.11

dockerymick:
Previous PR: https://github.com/stolostron/rhacm-docs/pull/6899/files


dockerymick:
There is an OCP link that the linter flagged. leaving that for now since Oliver is updating the links

PR #6969, opened 2024-09-24
https://issues.redhat.com/browse/ACM-14439: some note and prerequirment for globalhub

ldpliu:
https://issues.redhat.com/browse/ACM-14439

### DIFF
--- a/global_hub/global_hub_import_hub_hosted.adoc
+++ b/global_hub/global_hub_import_hub_hosted.adoc
@@ -6,7 +6,8 @@ To enable `local-cluster` on your managed hub cluster, you must import it in the
 [#global-hub-importing-prereq]
 == Prerequisite
 
-Enable `local-cluster` in the {global-hub} cluster.
+- Enable `local-cluster` in the {global-hub} cluster.
+- The {acm-short} version should be 2.12.X or later
 
 [#global-hub-importing-hosted-mode]
 == Importing a hub cluster in the hosted mode 
@@ -16,6 +17,7 @@ To import an existing hub cluster that has `local-cluster` enabled, you must use
 *Note:*
 
 - Make sure this `kubeconfig` or API token is always usable. If it expires, regenerate the `auto-import-secret` in the managed hub cluster namespace with the following command: `oc create secret generic auto-import-secret --from-file=kubeconfig=./managedClusterKubeconfig -n <Managedhub Namespace>`. 
+- In this mode, {global-hub} do not support backup and restore
 
 Complete the following steps: 
 


global_hub/global_hub_import_hub_hosted.adoc
```diff
@@ -7,6 +7,10 @@ Import a managed hub cluster in the hosted mode to import any managed hub cluste
 == Prerequisites
 
 - Enable `local-cluster` in the {global-hub} cluster.
+- The {acm-short} version should be 2.12.X or later
```
dockerymick, line 10: Install the latest version of {acm-short}.

ldpliu: It just need 2.12.x or later, if we use `latest version` here, we must change it in next release.

> Install the latest version of {acm-short}.

dockerymick: Thanks, so 2.13 would not support this info in the next release?

ldpliu: 2.13 will support this.
But in 2.13, if we use 'latest version', it means user must use 2.13, but we also support 2.12.

global_hub/global_hub_import_hub_hosted.adoc
```diff
@@ -22,3 +26,4 @@ With this annotation, {global-hub} does the following actions:
 
 - In the hosted mode, you only get support for import managed hub clusters that use `kubeconfig` or API tokens. {acm} uses this `kubeconfig` or API token to generate the `auto-import-secret`, which connects to the managed hub cluster.
 - Make sure this `kubeconfig` or API token is always usable. If it expires,  regenerate the `auto-import-secret` in the managed hub namespace, `oc create secret generic auto-import-secret --from-file=kubeconfig=./managedClusterKubeconfig -n <Managedhub Namespace>`
+- In this mode, {global-hub} do not support backup and restore
```
hchenxa: seems like when using the host mode, the external-kubeconfig will come from the creds user provided when import the cluster, and if the token is expired, it may have issues. @ldpliu

ldpliu: > seems like when using the host mode, the external-kubeconfig will come from the creds user provided when import the cluster, and if the token is expired, it may have issues. @ldpliu

Yes, so we add this note here.
https://github.com/stolostron/rhacm-docs/pull/6969/files/adc8cfa46bfb4ebc3d9090a1b46bb9102b3c47b9#diff-426c3333e3d6698409184255e75546191f82b8b362dc8f713fb9e77a00c289e1L24

hchenxa: What I mean here is how do we know if the token is expired in the external-kubeconfig and we should re-generate it?

jc-berger:
Closing this PR, because I made new PR, implementing the content here and the feedback within the discussion: https://github.com/stolostron/rhacm-docs/pull/7018

PR #7011, opened 2024-10-04
Add `containerArguments` docs to Gatekeeper

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-14700


governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -74,11 +80,20 @@ spec:
...
+
+<1> (3.14) Enable the `auditEventsInvolvedNamespace` parameter to manage the namespace audit event you want to create. When you enable this parameter, the Gatekeeper controller deployment runs with the following argument: `--audit-events-involved-namespace=true`.
+<2> (3.14) Enable the `admissionEventsInvolvedNamespace` parameter to manage the namespace admission event you want to create. When you enable this parameter, the Gatekeeper controller deployment runs with the following argument: `--admission-events-involved-namespace=true`.
+<3> (3.14) To manage your webhook operations, you can use the following values for the `operations` parameter, `"CREATE"`, `"UPDATE"`, `"CONNECT"`, and `"DELETE"`.
+<4> (3.17) Specify `containerArguments` by providing a list of argument names (omitting the leading dashes) and values to pass to the container. An omitted value is treated as `true`. Arguments provided are ignored if the argument is set previously by the operator or configurations from other fields. Furthermore, see this list of flags that is deny listed and aren't currently supported:
```
dockerymick: Why are there version numbers before the descriptions? In the what's new entry, we will specifically speak about the `containerArguments`. The callout number is necessary for just the first instance of the parameter. Remove `<4>` from line 56.
 
```suggestion
<1> Enable the `auditEventsInvolvedNamespace` parameter to manage the namespace audit event you want to create. When you enable this parameter, the Gatekeeper controller deployment runs with the following argument: `--audit-events-involved-namespace=true`.
<2> <4> Specify `containerArguments` by providing a list of argument names and values to pass to the container. Be sure to omit leading dashes. An omitted value is treated as `true`. Arguments provided are ignored if the argument is set previously by the operator or configurations from other fields. See the following list of flags that are deny-listed and not currently supported:
<3> Enable the `admissionEventsInvolvedNamespace` parameter to manage the namespace admission event you want to create. When you enable this parameter, the Gatekeeper controller deployment runs with the following argument: `--admission-events-involved-namespace=true`.
<4> To manage your webhook operations, you can use the following values for the `operations` parameter, `"CREATE"`, `"UPDATE"`, `"CONNECT"`, and `"DELETE"`.
```

I don't quite understand "(omitting the leading dashes)". Is there a smaller list of supported flags?

dockerymick (changes requested):

governance/gatekeeper_operator/config_gk_operator.adoc
dhaiducek: - The version numbers are there because these docs get cherry-picked back to all supported versions of ACM since Gatekeeper doesn't explicitly follow the same release schedule as ACM, so it's necessary to delineate changes that occurred in a particular version since this same doc will be mirrored/cherry-picked. (We'll consult with you all for the content, but we're handling the "What's New" ourselves directly in the description on the container page--it won't live in the ACM doc. See the link here: https://github.com/stolostron/rhacm-docs/blob/2.12_stage/governance/gatekeeper_operator/general_support.adoc#general-support)
- I'm fine removing the callout, but I'll note there is precedence in OCP to have a second callout: https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html-single/hosted_control_planes/index#hcp-bm-hosts_hcp-deploy-dc-bm
- In practice, container arguments are specified as `--argument=<value>` or, equivalently, `--argument <value>` (See [the Gatekeeper deployment](https://github.com/open-policy-agent/gatekeeper/blob/475c9b3c55dabe2bdfd328efa7fd98099b19b343/deploy/gatekeeper.yaml#L4936-L4942), for instance). For the `containerArguments` feature in the Gatekeeper operator, I opted to omit these leading dashes so that the argument name would be provided as `name: argument` rather than `name: --argument`.

dockerymick: @dhaiducek 

* Wouldn't the changes be mentioned in the Gatekeeper what's new topic though? 
* From my perspective, it would be best to be more specific of which `containerArgument`. For example, `audit.containerArguments` and `webhook.containerArguments`. If that isn't helpful then we need to at least order it. Line 33 should be have "<2>", line 48 "<3>", line 51 "<4>", and line 56 "<5>"

dhaiducek: > * Wouldn't the changes be mentioned in the Gatekeeper what's new topic though?

Yes, though the "what's new" topic won't live here in this repo, so the version numbers need to stay.

> * From my perspective, it would be best to be more specific of which `containerArgument`. For example, `audit.containerArguments` and `webhook.containerArguments`. If that isn't helpful then we need to at least order it. Line 33 should be have "<2>", line 48 "<3>", line 51 "<4>", and line 56 "<5>"

Both `containerArgument` fields are identical in code, so having the same reference makes the most sense to me. But the numbering would need to stay as it is if it's not removed--`asciidoc` won't be able to align the callout to the particular reference at the bottom (and would likely result in a build failure):
ref: https://github.com/openshift/openshift-docs/blob/3b72358b9255d7b257d96e49db63a8a6104a6304/modules/hcp-bm-hosts.adoc?plain=1#L29-L39
ref: https://discuss.asciidoctor.org/Way-to-indicate-a-callout-covers-multiple-lines-td6168.html

governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -74,11 +80,20 @@ spec:
...
+
+<1> (3.14) Enable the `auditEventsInvolvedNamespace` parameter to manage the namespace audit event you want to create. When you enable this parameter, the Gatekeeper controller deployment runs with the following argument: `--audit-events-involved-namespace=true`.
+<2> (3.14) Enable the `admissionEventsInvolvedNamespace` parameter to manage the namespace admission event you want to create. When you enable this parameter, the Gatekeeper controller deployment runs with the following argument: `--admission-events-involved-namespace=true`.
+<3> (3.14) To manage your webhook operations, you can use the following values for the `operations` parameter, `"CREATE"`, `"UPDATE"`, `"CONNECT"`, and `"DELETE"`.
+<4> (3.17) Specify `containerArguments` by providing a list of argument names and values to pass to the container. Be sure to omit leading dashes from the argument name. An omitted value is treated as `true`. Arguments provided are ignored if the argument is set previously by the operator or configurations from other fields. See the following list of flags that are deny-listed and aren't currently supported:
```
dockerymick: I still don't understand why it is important to delineate the versions here though? Are the parameters not supported in later versions or are they only supported for those versions?

release_notes/whats_new.adoc
```diff
@@ -88,9 +88,9 @@ See link:../console/console.adoc#home-page[Console overview].
...
 
-* You can now view the metadata of a root policy by using the `.PolicyMetadata` hub cluster template variable. See the link:../governance/template_support_intro.adoc#template-coomparison-table[Comparison of hub cluster and managed cluster templates].
+* You can now view the metadata of a root policy by using the `.PolicyMetadata` hub cluster template variable. See the link:../governance/template_support_intro.adoc#template-comparison-table[Comparison of hub cluster and managed cluster templates].
 
-* You can now use the `hubTemplateOptions.serviceAccountName` field to specify a service account to use for all hub cluster template lookups. See the link:../governance/template_support_intro.adoc#template-coomparison-table[Comparison of hub cluster and managed cluster templates].
+* You can now use the `hubTemplateOptions.serviceAccountName` field to specify a service account to use for all hub cluster template lookups. See the link:../governance/template_support_intro.adoc#template-comparison-table[Comparison of hub cluster and managed cluster templates].
```
dockerymick: great catch !

governance/gatekeeper_operator/config_gk_operator.adoc
dhaiducek: They don't exist in versions prior to the ones listed. Since this is a unified doc, it's a doc meant for users with Gatekeeper 3.11, 3.14, 3.15, or 3.17 installed.

dhaiducek: ...so these are the versions where the feature appeared.

dockerymick: Thanks for explaining. So would it be accurate if we say "For version 3.14 and later, enable the `auditEventsInvolvedNamespace` parameter..."

governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -74,11 +80,20 @@ spec:
...
+<3> For version 3.14 and later, to manage your webhook operations, you can use the following values for the `operations` parameter, `"CREATE"`, `"UPDATE"`, `"CONNECT"`, and `"DELETE"`.
+<4> For version 3.17 and later, specify `containerArguments` by providing a list of argument names and values to pass to the container. Be sure to omit leading dashes from the argument name. An omitted value is treated as `true`. Arguments provided are ignored if the argument is set previously by the operator or configurations from other fields. See the following list of flags that are deny-listed and aren't currently supported:
+- port
+- prometheus-port
+- health-addr
+- validating-webhook-configuration-name
+- mutating-webhook-configuration-name
+- disable-cert-rotation
+- client-cert-name
+- tls-min-version
```
dockerymick:
```suggestion
<4> For version 3.17 and later, specify `containerArguments` by providing a list of argument names and values to pass to the container. Be sure to omit leading dashes from the argument name. An omitted value is treated as `true`. Arguments that you provide are ignored if the argument is set previously by the operator or configurations from other fields. See the following list of flags that are deny-listed and are not currently supported:
- `port`
- `prometheus-port`
- `health-addr`
- `validating-webhook-configuration-name`
- `mutating-webhook-configuration-name`
- `disable-cert-rotation`
- `client-cert-name`
- `tls-min-version`
```

dockerymick (approved):

dockerymick:
@dhaiducek thanks for accepting my suggestions. I just recently asked for another doc peer review

governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -74,11 +80,20 @@ spec:
     some-annotation: "this is a test"
     other-annotation: "another test"
 ----
-*Note:* For versions 3.14 or later, you can implement the following features from this YAML:
-
-<1> Enable the `auditEventsInvolvedNamespace` parameter to manage the namespace audit event you want to create. When you enable this parameter, the Gatekeeper controller deployment runs with the following argument: `--audit-events-involved-namespace=true`.
-<2> Enable the `admissionEventsInvolvedNamespace` parameter to manage the namespace admission event you want to create. When you enable this parameter, the Gatekeeper controller deployment runs with the following argument: `--admission-events-involved-namespace=true`.
-<3> To manage your webhook operations, you can use the following values for the `operations` parameter, `"CREATE"`, `"UPDATE"`, `"CONNECT"`, and `"DELETE"`.
+*Note:* Starting in the versions specified, you can implement the following features from this YAML:
```
jc-berger: Is it the version the user specified? If so, consider rephrasing to something like the following: 

```suggestion
*Note:* For the version you specified, you can implement the following features from this YAML:
```

governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -74,11 +80,20 @@ spec:
...
+
+<1> For version 3.14 and later, enable the `auditEventsInvolvedNamespace` parameter to manage the namespace audit event you want to create. When you enable this parameter, the Gatekeeper controller deployment runs with the following argument: `--audit-events-involved-namespace=true`.
+<2> For version 3.14 and later, enable the `admissionEventsInvolvedNamespace` parameter to manage the namespace admission event you want to create. When you enable this parameter, the Gatekeeper controller deployment runs with the following argument: `--admission-events-involved-namespace=true`.
+<3> For version 3.14 and later, to manage your webhook operations, you can use the following values for the `operations` parameter, `"CREATE"`, `"UPDATE"`, `"CONNECT"`, and `"DELETE"`.
```
jc-berger: In your earlier callouts, you use the imperative voice/command. So I wouldn't use "you can..." here. 

```suggestion
<3> For version 3.14 and later, to manage your webhook operations, use the following values for the `operations` parameter, `"CREATE"`, `"UPDATE"`, `"CONNECT"`, and `"DELETE"`.
```

governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -74,11 +80,20 @@ spec:
...
+<4> For version 3.17 and later, specify `containerArguments` by providing a list of argument names and values to pass to the container. Be sure to omit leading dashes from the argument name. An omitted value is treated as `true`. Arguments that you provide are ignored if the argument is set previously by the operator or configurations from other fields. See the following list of flags that are deny-listed and are not currently supported:
```
jc-berger: Optional:  I actually like "be sure" because it's more of a reminder for the user, but for consistency, consider making it an action verb/imperative like all the other sentences here:

```suggestion
<4> For version 3.17 and later, specify `containerArguments` by providing a list of argument names and values to pass to the container. Omit the leading dashes from the argument name. An omitted value is treated as `true`. Arguments that you provide are ignored if the argument is set previously by the operator or configurations from other fields. See the following list of flags that are deny-listed and are not currently supported:
```

release_notes/whats_new.adoc
```diff
@@ -88,9 +88,9 @@ See link:../console/console.adoc#home-page[Console overview].
 
 * To add more clarity for compliance messages in your configuration policies, you can now customize compliance messages by using the `spec.customMessage` fields. See the link:../governance/config_policy_ctrl.adoc#configuration-policy-yaml-table[Kubernetes configuration policy controller].
 
-* You can now view the metadata of a root policy by using the `.PolicyMetadata` hub cluster template variable. See the link:../governance/template_support_intro.adoc#template-coomparison-table[Comparison of hub cluster and managed cluster templates].
+* You can now view the metadata of a root policy by using the `.PolicyMetadata` hub cluster template variable. See the link:../governance/template_support_intro.adoc#template-comparison-table[Comparison of hub cluster and managed cluster templates].
```
jc-berger: Again, optional, but since the existing bullet list here in the what's new doc uses the sentence structure, "To < do action >, you can < do this >" consider structuring your content in a similar way: 

```suggestion
* To view the metadata of a root policy, you can use the `.PolicyMetadata` hub cluster template variable. See the link:../governance/template_support_intro.adoc#template-comparison-table[Comparison of hub cluster and managed cluster templates].
```

jc-berger: Or to make a little shorter and more imperative:

```suggestion
* To view the metadata of a root policy, use the `.PolicyMetadata` hub cluster template variable. See the link:../governance/template_support_intro.adoc#template-comparison-table[Comparison of hub cluster and managed cluster templates].
```

release_notes/whats_new.adoc
jc-berger:
```suggestion
* To specify a service account to use for all hub cluster template lookups, You can now use the `hubTemplateOptions.serviceAccountName` field. See the link:../governance/template_support_intro.adoc#template-comparison-table[Comparison of hub cluster and managed cluster templates].
```

jc-berger (approved):
/lgtm

governance/gatekeeper_operator/config_gk_operator.adoc
dhaiducek: No, these configurations depend on the version of the operator that's installed whether they're available.

dockerymick: Can we actually remove the note and just start with the callouts? @dhaiducek

dhaiducek: Oops, sorry I missed this comment @dockerymick! Yes, I'll remove it!

dhaiducek:
Thanks for the update, @dockerymick! This LGTM!
(Build errors are the result of linking to 2.12 support matrices that aren't yet accessible.)

drkho:
💯 LGTM

dockerymick:
@dhaiducek thanks for initiating and collaborating on the PR, and thanks for the reviews and approvals @jc-berger and @drkho ! Merging this PR

PR #7014, opened 2024-10-07
https://issues.redhat.com/browse/ACM-13552 Deprecations/removals n-2 only

oafischer:


clusters/release_notes/deprecate_remove.adoc
```diff
@@ -17,19 +17,13 @@ Learn when parts of the product are deprecated or removed from {mce-short}. Cons
...
-|===
-| Product or category | Affected item | Version | Recommended action | More details and links
-
-| ManagedServiceAccount
-| The `v1alpha1` API is upgraded to `v1beta1` because `v1alpha1` is deprecated.
-| 2.4
-| Use `v1beta1`.
-| None
+//|===
+//| Product or category | Affected item | Version | Recommended action | More details and links
```
swopebe: I am only cautious of removing this from deprecations bc it was at some point supposed to be removed. We may want to ask Foundation team about this one before we take it away, unless there is already an issue somewhere to remove it?

oafischer: I'll ask about this

oafischer: @swopebe It is indeed still deprecated, so we'll leave this one in but keep the rest as is?

PR #7022, opened 2024-10-08
https://issues.redhat.com/browse/ACM-14386

dockerymick:


release_notes/whats_new.adoc
```diff
@@ -57,6 +57,8 @@ View tasks and support information at link:../clusters/cluster_mce_overview.adoc
 
 * To create and mount secrets to your `alertmanager` pods for access to arbitrary content, you can add the contents to your `MultiClusterObservability` resource. See link:../observability/observability_alerts.adoc#mount-secrets-alertmanager[Mounting secrets within the Alertmanager pods].
 
+* Grafana is updated to version 11.1.5 to make arbitrary and adhoc metric queries from the Grafana _Explore_ page. See link:../observability/design_grafana.adoc#using-grafana-dashboards[Using Grafana dashboards].
```
jacobbaungard: > to make arbitrary and adhoc metric queries from the Grafana _Explore_ page

I'm not sure we should include this. To clarify this was possible before, however in _some_ situations (i.e those using Chromium based browser) there was a bug preventing this from working correctly.

I would either rewrite this, to more specifically call out that it fixes that bug, or simply leave it out entirely, and just say the version was upgraded.

dockerymick: Thanks @jacobbaungard for your feedback. We typically do not call out fixes in What's new, but we do mention them in the Errata. With that being said, I agree with you on only mentioning the Grafana update.

dockerymick:
```suggestion
* Grafana is updated to version 11.1.5. See link:../observability/design_grafana.adoc#using-grafana-dashboards[Using Grafana dashboards].
```

jacobbaungard (approved):
/lgtm

dockerymick:
I'm not familiar with the message is received when trying to fix the conflict. Since this is a small change, I will close this PR and create a new one

PR #7032, opened 2024-10-09
Version updates 

swopebe:


clusters/cluster_lifecycle/release_image_disconn.adoc
```diff
@@ -3,7 +3,7 @@
 
 In some cases, you need to maintain a custom list of release images when the hub cluster has no Internet connection. You can create your own custom list of release images that are available when creating a cluster. Complete the following steps to manage your available release images while disconnected:
 
-. While you are on a connected system, navigate to the link:https://github.com/stolostron/acm-hive-openshift-releases/tree/backplane-2.6[acm-hive-openshift-releases GitHub repository] to access the cluster image sets that are available.
+. While you are on a connected system, navigate to the link:https://github.com/stolostron/acm-hive-openshift-releases/tree/backplane-2.7[acm-hive-openshift-releases GitHub repository] to access the cluster image sets that are available.
```
jc-berger: Optional: though not a big deal, IBM SG suggests saying "go to" instead of navigate. 

```suggestion
. When you are on a connected system, go to the link:https://github.com/stolostron/acm-hive-openshift-releases/tree/backplane-2.7[acm-hive-openshift-releases GitHub repository] to access the available cluster image sets.
```

troubleshooting/must_gather.adoc
```diff
@@ -28,7 +28,7 @@ See the following procedure to start using the `must-gather` command:
 . Log in to your cluster. Add the {acm} image that is used for gathering data and the directory. Run the following command, where you insert the image and the directory for the output:
```
jc-berger, line 28: Optional: so "run the following command" immediately precedes the command, consider rephrasing this last sentence: 

```suggestion
. Log in to your cluster. Add the {acm} image that is used for gathering data and the directory. To insert the image and the directory for the output, run the following command:
```

jc-berger (approved):
/lgtm

troubleshooting/must_gather.adoc
swopebe: This will stay for now--not part of this issue scope. It is absolutely **not** written well, I agree!

But the second way is not saying what we need to say, either. 

We need to tell the user to r**eplace variables.** The command is **To add the image**, not to `insert the image and the directory for the output`. Inserting the values is what the user does with the variables. Output here is confusing. I have another issue for mustgather and I will work on that within that scope.

Example of how to write "run a command" and "replace values" situation; this can also be two sentences.

`Run the following commands to collect all of the information, replacing <2.x> with the supported version for both <acm-must-gather>, for example 2.10, and <multicluster-engine/must-gather>, for example 2.5.`

I will go over this in the next meeting: How to tell the user to replace variables.

PR #7046, opened 2024-10-10
Fix cluster book for 2.11

dockerymick:
Screenshot 2024-10-10 at 3 20 13 PM

I noticed that the file existed within the folder but it was empty. Should this file be empty?


swopebe:
This can close. This is a new file that just needed to be added.

PR #7066, opened 2024-10-14
https://issues.redhat.com/browse/ACM-13093

jc-berger:
PR to change Third Party Titles, see Jira: https://issues.redhat.com/browse/ACM-13093


governance/integrate_policy.adoc
JustinKuli, line 1: I don't feel like this section is quite right. It feels like this "custom annotations" sentence is a leftover from something for a different section - users can use annotations for standards, categories, and controls, and that can help organize their policies - it doesn't have much to do with the policy generator.

Also, the Compliance Operator part is more of an example of what can be done with the generator; it's not something a user "must" do in order to utilize the generator.

If this chapter will only be about the policy generator (which is currently the only content it has), I think it would be more discover-able if it was titled something like "Integrate the policy generator" and jumped straight into the intro  information in https://github.com/stolostron/rhacm-docs/blob/2.12_stage/governance/policy_generator.adoc. What do you think?

governance/gatekeeper_operator/manage_gk.adoc
```diff
@@ -119,5 +119,3 @@ For more details, see the following resources:
 
 - For an explanation of the optional parameters that can be used for the {gate} policy, see link:https://github.com/open-policy-agent/gatekeeper/blob/master/charts/gatekeeper/README.md[Gatekeeper Helm Chart].
 
-- For a list of topics to integrate third-party policies with the product, see xref:../../governance/third_party_policy_intro.adoc#integrate-third-party-policy-controllers[Integrate third-party policy controllers].
```
jc-berger, line 122: @JustinKuli since we don't talk about third party policies in this section, I removed this reference. If you have an alternative doc we can link to here, please let me know. Thanks!

governance/integrate_policy.adoc
jc-berger: @JustinKuli thanks! We can make the "Policy Generator" content the new 'leveloffset=1' 

However, since we do have a 4.2 section (Generating a policy), I thought we could still keep a brief overview, prompting users to 2.1 (Policy Generator) to understand the Policy Generator then to 4.2 (Generating a policy) so they have an example of what they can do with it. 

Let me know what you think, thanks :)

JustinKuli: I'm not as familiar with the `leveloffset` stuff, but as you described it, I think it will be good. [integrate_policy_generator.adoc](https://github.com/stolostron/rhacm-docs/pull/7066/files#diff-dd590e1e917a3a650b52b0613a247ebf03a82a5f553ec5a76ebc40ecc25cac09) definitely looks better than what we had before, so thanks for that!

governance/gatekeeper_operator/manage_gk.adoc
JustinKuli: I think just removing it is fine. I don't have anything else to add right now.

JustinKuli (approved):
Looks good to me! Thanks for this

swopebe (approved):

swopebe:
looks great.

PR #7090, opened 2024-10-16
https://issues.redhat.com/browse/ACM-14538

dockerymick:
I added the step in the command line section, it seemed like it wouldn't be best to add it in the console section


clusters/release_notes/mce_whats_new.adoc
```diff
@@ -30,7 +30,9 @@ Learn more about new features for specific components.
 
 Learn about new features and enhancements for Cluster management with {mce-short}.
 
-You can now set a duration to choose when the `kubeconfig` bootstrap in the klusterlet manifest expires. To learn more, see xref:../cluster_lifecycle/import_agent.adoc#cluster-import-agent[Importing a cluster].
+* You can now set a duration to choose when the `kubeconfig` bootstrap in the klusterlet manifest expires. To learn more, see xref:../cluster_lifecycle/import_agent.adoc#cluster-import-agent[Importing a cluster].
+
+* Import information for {rosa} clusters can now be found and pre-filled in the console. Use the {rosa} command line interface to receive more import information. See link:../discovery/enable_discovery.adoc#import-discovered-rosa-cli[Importing from the command line interface].
```
Randy424: What does this link point to? I was thinking the rosa cli docs were here:https://docs.redhat.com/en/documentation/red_hat_openshift_service_on_aws/4/html/rosa_cli/rosa-get-started-cli#rosa-get-started-cli

dockerymick: This is for what's new and pointing to the Importing from the command line interface topic where I added the step. I can add the OpenShift Service link to the Additional resources section

clusters/discovery/enable_discovery.adoc
```diff
@@ -93,6 +93,14 @@ oc patch discoveredcluster <name> -n <namespace> --type='json' -p='[{"op": "repl
 oc get managedcluster <name>
 ----
 
+. To get a description of your cluster ID, use the {rosa} command line interface by running the following command:
+
++
+[source,bash]
+----
+rosa describe cluster --cluster=<cluster-name> | grep -o '^ID:.*
+----
```
dockerymick, line 104: @Randy424 can this command be run from the MCE cluster?

Randy424: I'm not sure if this answers the question completely but the command is not connected to the MCE cluster that will import the discovered cluster. It is just used to get information from a the target ROSA cluster on your Red Hat OpenShift Cluster Manager console. I usually run it from my local machine, no need to authenticate to the MCE cluster, all you need is your Cluster manager sign-in.

dockerymick: It does answer it. I'm not sure, but it seems like this shouldn't be added as a step, nor for what's new. Maybe we can add the link (https://docs.redhat.com/en/documentation/red_hat_openshift_service_on_aws/4/html/rosa_cli/rosa-get-started-cli#rosa-get-started-cli), to this topic since cluster ID is mentioned: https://github.com/stolostron/rhacm-docs/blob/2.12_stage/clusters/discovery/discovery_config_ui.adoc#view-discovered-clusters

dockerymick: During meeting with @Randy424 

To get a description of your {rosa} cluster ID,  run the following command from the {rosa} command line interface:

clusters/discovery/enable_discovery.adoc
```diff
@@ -93,9 +93,18 @@ oc patch discoveredcluster <name> -n <namespace> --type='json' -p='[{"op": "repl
 oc get managedcluster <name>
 ----
 
+. To get a description of your cluster ID, use the {rosa} command line interface by running the following command:
```
dockerymick:
```suggestion
. To get a description of your {rosa} cluster ID, run the following command from the {rosa} command line interface:
```

clusters/discovery/enable_discovery.adoc
swopebe: Why would this not be a step? If you add that link, I would add it to the prereqs.

clusters/cluster_lifecycle/import_agent.adoc
```diff
@@ -102,17 +110,22 @@ oc patch clustermanager cluster-manager --type=merge -p '{"spec":{"registrationC
...
 ----
 curl --cacert ca.crt -H "Authorization: Bearer $token" https://$agent_registration_host/agent-registration/crds/v1 | oc apply -f -
 ----
 
 . Run the following command to import the managed cluster to the hub cluster. Replace `<clusterName>` with the name of you cluster. Replace `<duration>` with a time value. For example, `4h`:
 +
 *Optional:* Replace `<klusterletconfigName>` with the name of your KlusterletConfig.
+
 +
+[source,bash]
```
swopebe, line 125: We are losing this when we move to a different language anyway. Much of the legacy doc is missing this in cluster and install...I am just leaving it.

clusters/release_notes/mce_whats_new.adoc
swopebe: From the What's new topic, we don't point to OCP for our ACM function. We would point to our documentation.

swopebe: But is this 2.12 feature? It is 2.11 and 2.12 in the issue, which suggests not.

swopebe:
I am confused on this thread and the link to the issue takes me to the PR. 

This seems to have gotten complicated. If it has 2.11 on it and not related to a developer story, there is no need to update what's new, but there should have been either a bug or task for us to follow from development to reduce that confusion @Randy424.

If this is just an enhancement for 2.11 refresh, that's fine. I don't see all the updates that the reporter asked for.

If the user needs ROSA CLI, that should be in prereqs. 

Let me know when it's ready for another pass, or if this is for 2.12 feature.

https://issues.redhat.com/browse/ACM-14538

clusters/release_notes/mce_whats_new.adoc
Randy424: @swopebe the feature itself isn't new, but we are looking for the right place to home information about how to get a particular input value used in our recently added cluster import (from 2.11). The 2.11 feature let's us import ROSA clusters through discovery and through the usual import procedure. One of the fields necessary for import, "cluster ID", isn't easily found in the OpenShift Cloud Manager console. We wanted to offer one way for ACM customers who have a ROSA cluster to get this cluster id value. Using the CLI is the only way to get this value manually, as our console and the openshift console do not display it.

I don't know if "what's new" is the right place for it, but I do think we need to offer this information to our users.

dockerymick: Yea, my bad I forgot to remove the entry. Randy and I met the other day and agreed that this was a refresh item

dockerymick:
```suggestion
```

swopebe: Ok, I see.

1. You can add it to the What's new for 2.11, but I don't think anyone is looking at that anymore.
2. Obviously an update to the topic itself is part of this issue--that looks great.
3. But if you want more, to say, help in a customer call or face-to-face...have you considered placing this content as a _limitation?_ @Randy424 

Example:

## Users might not know to enter the `cluster ID` in the OpenShift Cloud Manager console. (or something like this)

If you did not access the `cluster ID` in the OpenShift Cloud Manager console, you can still get a description of your {rosa} `cluster ID` from the terminal. You need to ROSA CLI (link). (something like this)

Run the following command from the {rosa} command line interface:

+
[source,bash]
----
rosa describe cluster --cluster=<cluster-name> | grep -o '^ID:.*
----

(And this is just me...if you guys are running in to this a lot, maybe open an issue for OCP and see if they can make it better in the UI, but that's me.)

If you go forth with a limitation, this can be done in a new issue UNLESS Mikela has time with the delayed GA date. I am happy to take it, Mikela, if you are bogged down. Technically I am the Discover focal.

swopebe (approved):
See comment about what we can do further. Otherwise approved.

clusters/release_notes/mce_whats_new.adoc
Randy424: Ok, I really like how this is presented, though I'm not sure what is meant by a "limitation." 
We can open a ticket against OCM-UI for the cluster ID issue, I think that's a great idea too!

swopebe: A limitation would be like a known issue. It's just another type of release note.
8:57
we don't typically link to it in the main doc, but if you like that line I wrote-just add it to the doc...we can do that. We can just embed a line that says if you did not get to it in the console, then here is where you do it.

We don't link to known issues or limitations in the main tasks because they change, interrupt the user, and if it is that important, it should just be part of the doc.

dockerymick: Thanks Brandi for your input here. So it sounds like this is good to merge, but there should also be an issue created for the OCM-UI team to add a limitation about the Cluster ID. Is that correct? Since this is a refresh item, we don't necessarily need a review from QE

Randy424: I don't know if we want to add a limitation in the OCM-UI docs for that. I was thinking more along the lines of an enhancement request. A limitation document on the OCM side would be a bit weird because OCM users don't use the ID. ACM user's will. So the limitation is relevant to the ACM user and to the ACM product. I don't know if that is besides the point, though, I guess I could see it the other way too, as technically OCM-UI needs to take action to expose the ID. Maybe the user might check OCM-UI docs for that information, and that's why we would include a limitation there as well? Because in the OCM product is where the ID field **should** be? So they own the limitation?

Randy424: The RFE is here btw: https://issues.redhat.com/browse/OCMUI-2585

dockerymick:
Reached out through Slack to confirm that this can be merged and will create a doc issue to add the limitation @Randy424 @swopebe . Merging this PR

PR #7091, opened 2024-10-17
https://issues.redhat.com/browse/ACM-14676

oafischer:
2.12, 2.11, and 2.10

### DIFF
--- a/networking/submariner/subm_configure.adoc
+++ b/networking/submariner/subm_configure.adoc
@@ -79,10 +79,10 @@ annotations:
   submariner.io/worker-sg-id: <worker_security_group_id> <4>
 ----
 +
-<1> Replace with your control plan security group ID. Typically, the name is similar to `<infra-id>-master-sg`.
+<1> Replace with your control plane security group ID. Typically, you can find the ID from the controle plane security group that has a name similar to `<infra-id>-master-sg>`.
 <2> Replace with a comma-separated list of public subnet IDs in your custom VPC.
-<3> Repalce with your custom VPC ID.
-<4> Replace with your worker security group ID. Typically, the name is similar to `<infra-id>-worker-sg`.
+<3> Replace with your custom VPC ID.
+<4> Replace with your worker security group ID. Typically, you can find the ID from the worker security group that has a name similar to `<infra-id>-worker-sg`.
 
 [#submariner-globalnet]
 == Globalnet
--- a/networking/submariner/subm_configure.adoc
+++ b/networking/submariner/subm_configure.adoc
@@ -73,10 +73,10 @@ oc edit submarinerconfig -n <managed-cluster-ns> submariner
 [source,yaml]
 ----
 annotations:
-  submariner.io/control-plane-sg-id: <your_group_ID> <1>
-  submariner.io/subnet-id-list: <your_list> <2>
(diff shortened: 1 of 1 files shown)


networking/submariner/subm_configure.adoc
```diff
@@ -52,6 +52,38 @@ View the following table to see the Submariner ports that you must enable:
...
+[source,yaml]
+----
+annotations:
+  submariner.io/control-plane-sg-id: <control_plane_group_id> <1>
+  submariner.io/subnet-id-list: <subnet_id_list> <2>
+  submariner.io/vpc-id: <custom_vpc_id> <3>
+  submariner.io/worker-sg-id: <worker_security_group_id> <4>
+----
++
+<1> Replace with your control plan security group ID. Typically, the name is similar to `<infra-id>-master-sg`.
```
aswinsuryan: This description I gave created confusion for the users. it should mean this  "Typically, the name is similar to `<infra-id>-master-sg` for the security group we need to take the ID of that". They thought they need to use name here , where as it is supposed to be ID.

networking/submariner/subm_configure.adoc
```diff
@@ -52,6 +52,38 @@ View the following table to see the Submariner ports that you must enable:
...
+<2> Replace with a comma-separated list of public subnet IDs in your custom VPC.
+<3> Repalce with your custom VPC ID.
+<4> Replace with your worker security group ID. Typically, the name is similar to `<infra-id>-worker-sg`.
```
aswinsuryan: Same comment as above here.

networking/submariner/subm_configure.adoc
oafischer: I see, how does the following sound then?

"Typically, you can find the ID as part of the following name: `<infra-id>-master-sg`"

aswinsuryan: > I see, how does the following sound then?
> 
> "Typically, you can find the ID as part of the following name: `<infra-id>-master-sg`"

 something like this "Typically, you can find the ID from the security group which has the following name `<infra-id>-master-sg`"

oafischer:
Updated, if you want to take another look @aswinsuryan. Thanks!

aswinsuryan (approved):

networking/submariner/subm_configure.adoc
```diff
@@ -52,6 +52,38 @@ View the following table to see the Submariner ports that you must enable:
...
+[source,yaml]
+----
+annotations:
+  submariner.io/control-plane-sg-id: <control_plane_group_id> <1>
+  submariner.io/subnet-id-list: <subnet_id_list> <2>
+  submariner.io/vpc-id: <custom_vpc_id> <3>
+  submariner.io/worker-sg-id: <worker_security_group_id> <4>
+----
++
+<1> Replace with your control plane security group ID. Typically, you can find the ID from the controle plane security group that has a name similar to `<infra-id>-master-sg>`.
```
dockerymick:
```suggestion
<1> Replace with your control plane security group ID. Typically, you can find the ID from the control plane security group that has a name similar to `<infra-id>-master-sg>`.
```

dockerymick (approved):
Left a suggestion for spelling. Otherwise, this looks good to me!

PR #7094, opened 2024-10-17
https://issues.redhat.com/browse/ACM-13207

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-13207


business_continuity/backup_restore/backup_schedule.adoc
```diff
@@ -91,7 +91,10 @@ For descriptions of the various `BackupSchedule` statuses, see the following tab
 [#avoid-backup-collision]
 == Avoiding backup collisions
 
-Backup collisions might occur if the hub cluster changes from being a passive hub cluster to becoming a primary hub cluster, or the other way around, and different managed clusters back up data at the same storage location.
+Backup collisions might occur in the following situations:
+
+* If the hub cluster changes status from a passive to primary hub cluster, or from a primary to passive hub cluster.
```
birsanv: This is not correct, it implies the collision happens If the hub cluster changes status. That is not enough.
If the hub cluster changes status AND other hub clusters back up data at the same storage location.

Can we keep the initial content ? Replace 94-97 with the proposed : 
 _Backup collisions might occur if the hub cluster changes status from passive to primary hub cluster, or from primary to passive, and different hub clusters back up data at the same storage location._

( note that line 99 is a continuation of this sentence; that will not make sense with these changes )

birsanv (changes requested):

birsanv (approved):
lgtm

dockerymick (approved):

PR #7117, opened 2024-10-19
https://issues.redhat.com/browse/ACM-14653--what's new for adv config

swopebe:


dislbenn (approved):
LGTM

release_notes/acm_whats_new.adoc
```diff
@@ -28,9 +29,10 @@ Some features and components are identified and released as link:https://access.
 
 Access the link:https://access.redhat.com/articles/7086905[{acm-short} Support matrix] to learn about hub cluster and managed cluster requirements and support for each component. For lifecycle information, see link:https://access.redhat.com/support/policy/updates/openshift[{ocp} Life Cycle policy].
 
+[#install-acm-whats-new]
+== Installation
 
-//[#installation]
-//== Installation no epics this release 2.11;be sure to add bullet up top if that changes
+* You can enable the `SiteConfig` component from the `MultiClusterHub` custom resource that is deployed on your cluster. By default, the `SiteConfig` component is disabled. Learn more at xref:../install/adv_config_install.adoc#advanced-config-hub[MultiClusterHub advanced configuration].
```
amolnar-gh: The intention of this paragraph is not entirely clear to me. Are we introducing SiteConfig? If yes, then why not link to the docs that tell about the operator and how to use it? Why only the link to the MultiClusterHub advanced configuration that only covers enabling the operator?

That's my only concern but maybe I'm missing something. Otherwise lgtm!

swopebe: @amolnar-rh we can link to it here, but this is more about the advanced config.

I think it makes sense to do it in line 44.

PR #7121, opened 2024-10-21
https://issues.redhat.com/browse/ACM-14360 Import resources doc

oafischer:
Train 19, 2.12 only


clusters/cluster_lifecycle/import_ocp.adoc
```diff
@@ -172,3 +173,38 @@ oc get infraenv -n managed-cluster some-other-infraenv -ojson | jq ".status.<url
...
+
+- `Agent`
+- `AgentClassification`
+- `AgentClusterInstall`
+- `BareMetalHost`
+- `ClusterDeployment`
+- `InfraEnv`
+- `NMStateConfig`
+
+You can move management by creating a copy of the original resources to load balance, scale down, and scale up your managed cluster in the new hub cluster.
```
carbonin: I'm not sure what "load balance" is referring to here.

We really don't intend to be able to manage a cluster from multiple hubs at once.

carbonin (commented):
I think we also should call out that the resources should be removed from the original hub and how to do that.

Specifically we should mention setting `preserveOnDelete` in the `ClusterDeployment`.

I'm hoping @danielerez has exact steps he followed for this, but he's out for a while still :worried:

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: I'll remove it for now for time reasons.

oafischer:
Thanks for the suggestion @carbonin. Our cut off date is nearing quickly (end of this week). Since this is a train item we need to push it before that. Are we OK with merging the content as is for now? We can definitely revisit the resource removal steps with Daniel, maybe after GA, in a separate issue.

carbonin:
>Are we OK with merging the content as is for now?

Yup, looks good for now

clusters/cluster_lifecycle/import_ocp.adoc
```diff
@@ -172,3 +173,38 @@ oc get infraenv -n managed-cluster some-other-infraenv -ojson | jq ".status.<url
 ----
 
 . *Optional:* If you want to use {acm-short} features, such as policies, on your {ocp-short} cluster, create a `ManagedCluster` resource. Make sure that the name of your `ManagedCluster` resource matches the name of your `ClusterDeplpoyment` resource. If you are missing the `ManagedCluster` resource, your cluster status is `detached` in the console.
+
+[#import-ocp-cluster-resources]
+== Importing cluster resources
+
+You can import an {ocp-short} managed cluster that was installed by the {ai} on one hub cluster to another hub cluster along with all cluster resources. You can import the following resources and continue to manage your cluster with them:
```
jc-berger: Is the user importing an ocp managed cluster with all the cluster resources? 

Or is the user importing an ocp cluster to a hub cluster and that hub cluster has the cluster resources? 

Although this rewrite might add word count, I think some clarification could help. What do yo think of the following rewrite? 

```suggestion
You can import an {ocp-short} managed cluster, along with its resources, that was installed by the {ai} on one hub cluster to another hub cluster. You can import the following resources and continue to manage your cluster with them:
```

clusters/cluster_lifecycle/import_ocp.adoc
```diff
@@ -172,3 +173,38 @@ oc get infraenv -n managed-cluster some-other-infraenv -ojson | jq ".status.<url
...
+
+- `Agent`
+- `AgentClassification`
+- `AgentClusterInstall`
+- `BareMetalHost`
+- `ClusterDeployment`
+- `InfraEnv`
+- `NMStateConfig`
+
+You can move management by creating a copy of the original resources and scale down or scale up your managed cluster in the new hub cluster.
```
jc-berger: I'm not sure what you mean by "move management" Earlier, we were talking about importing managed cluster resources. 

Do you mean _change_ the management of these hub clusters? If so, think about rewriting it to something like: 

```suggestion
You can control the management of your clusters by creating a copy of the original resources and either scale down or scale up your managed cluster in the new hub cluster.
```

clusters/cluster_lifecycle/import_ocp.adoc
```diff
@@ -172,3 +173,38 @@ oc get infraenv -n managed-cluster some-other-infraenv -ojson | jq ".status.<url
...
+
+*Important:* This feature is only available if the managed cluster you are importing was installed by the {ai}.
```
jc-berger: When you say "this feature" do you mean the "move management"? I didn't really read that as a feature. I understood more as an action the user could take. What do you think about rewriting this note to something like: 

```suggestion
*Important:* You can only scale your imported managed clusters if they were installed by the the {ai}.
```

clusters/cluster_lifecycle/import_ocp.adoc
```diff
@@ -172,3 +173,38 @@ oc get infraenv -n managed-cluster some-other-infraenv -ojson | jq ".status.<url
...
+
+To create a copy of the `Agent` resource on the target hub cluster, complete the following steps. You can use the following example to create copies of your other resources as well:
```
jc-berger: How does copying an Agent help the overall goal to import cluster resources? If I understand correctly, making a copy of Agent helps ensure you import all the cluster resources. If I'm right, what do you think of the following restructure? 

```suggestion
[#import-copy-resource]
=== Creating a copy of resources

To help ensure the cluster resources get imported, create a copy of the `Agent` resource on the target hub cluster. You can also create copies of your other resources you want imported. To create copies, complete the following steps and use the following example: 
```

oafischer: The Agent is a cluster resource. The topic is about importing the cluster resources, which are listed in the intro. I chose Agent as an example since it's the first resource alphabetically, but you could use any of the other resources as an example. Let me know if that makes sense, right now I'd leave this section as is.

I did wonder if maybe it would make sense to add the variables for each of the resources since I believe they aren't all intuitive. `Agent` is `agent`, but `MultiClusterHub` is `mch`. Would adding those make this section clearer?

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: The feature is being able to import the cluster resources from a different hub cluster and being able to continue using them.

Previously, you could only scale up but with this new feature, you can scale down as well, so it's important to keep that distinction. I agree that the placement of the note is maybe a bit confusing. Maybe moving it further up might help?

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: Moving management refers to cluster management being moved from one hub cluster to a different hub cluster. I can add that if it helps clarify.

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: I like that, I'd maybe go even further?

```suggestion
If your {ocp-short} managed cluster was installed by the {ai}, you can move the managed cluster and its resources from one hub cluster to another hub cluster. You can import the following resources and continue to manage your cluster with them:
```
Maybe this makes the note further down unnecessary as well? We wanted to emphasize that you shouldn't do this if the cluster was not installed by AI. I'm not sure if the relative clause here is enough to achieve that, even if it's at the beginning of the sentence. What do you think @jc-berger?

jc-berger:
```suggestion
If your {ocp-short} managed cluster was installed by the {ai}, you can move the managed cluster and its resources from one hub cluster to another hub cluster. 

*Important:* You can only import your {ocp-short} managed clusters if they were installed by the the {ai}.

You can import the following resources and continue to manage your cluster with them:
```

jc-berger: I love your rewrite! I also think it's smart to add the "Important" note earlier. 

I think the "important" can be removed. However, to your point about the clause "If your ocp managed cluster.." is enough to convey that users _should not do this_ unless the clusters have been installed by AI. I don't think it really communicates _don't do it_. 

In that case, the "important" might be necessary to communicate they can only import these clusters if they were installed by ai.

clusters/cluster_lifecycle/import_ocp.adoc
jc-berger: Thanks for the explanation. What do you think of this rewrite then? 

```suggestion
You can change the management of your clusters by moving it from one hub cluster to a different hub cluster. To do this, create a copy of the original resources and scale down or scale up your managed cluster in the new hub cluster.
```

clusters/cluster_lifecycle/import_ocp.adoc
jc-berger: Agreed, see my new suggestion above :)

clusters/cluster_lifecycle/import_ocp.adoc
jc-berger: I think you can keep to one example, like Agent. But sure, go ahead and add variables. Basically, I don't think you need to document more steps for copying MCH if that's what you mean. 

 But how it's currently written, I didn't see how copying an Agent helps with importing managed clusters. What do you think about this? 

```suggestion
For an example on importing your cluster resources, create a copy of the `Agent` resource on the target hub cluster by completing the following steps:  
```

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: Yep that's the dilemma, we shouldn't be negative in the doc as per the style guide, but I agree I don't think the relative clause alone is strong enough. I think your positioning for the important note works well. I'll use that suggestion. Thanks!

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: I think I see where the confusion is. The goal of this topic is mainly importing the cluster resources and letting users know that they can continue using those resources for managing. Importing a cluster is the topic that comes before this one and is a separate thing, but the problem is that the resource import with full functionality is only something you can do on imported clusters that were installed by AI. So we we need to mention the cluster import stuff on a general level somewhere.

My thinking was that this new topic comes right after the cluster import topic, so the context and purpose should be clear to the user, but I can see how it might be a little confusing out of context. I love your suggestion of starting by stating that it's an example, I think that makes the modularity of the steps a little clearer.

I must add though, technically you aren't importing the resource, you are creating a copy and I think we need to say that because that is what the steps in this procedure describe. Functionally though it is importing. I think the distinction is not super important to the end user, but it does need to be in there somewhere.

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: Hmm, the meaning here is changing a little too much. We need to convey the following:

- You can move management of your cluster from one hub to another hub
- The cluster is managed by you through the cluster resources, so moving the resources is essentially moving management from one hub to another hub
- By doing this, you can scale down and scale up the cluster on the new hub. Before this feature, you could not import all the existing resources, and you were not able to scale down the cluster
- I'm careful with saying import because technically you are just creating a new copy of the old resources on the new hub, although functionally it is essentially importing

Maybe what I'm trying to say is you can move the **resources for managing the cluster**? Might be a little clearer than saying moving management. What do you think @jc-berger?

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: If all that makes sense, how about this? @jc-berger 
```suggestion
Import your cluster resources by creating a copy of your resources on your target hub cluster. See the following example to learn how to create a copy of the `Agent` resource and apply similar steps to your other resources:
```

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: I don't think the suggested statement is accurate. Scaling up was possible before I believe? I'm also not sure how an AI installed cluster vs not AI installed cluster affects or limits features.

I know for sure that the whole importing and continuing to use existing resources feature is what is dependent on the an AI installed cluster, so maybe we should focus on that.

clusters/cluster_lifecycle/import_ocp.adoc
carbonin: Maybe something like this?

```suggestion
You can manage a cluster from a new hub by creating a copy of the original resources and applying them to the new hub. You will then be able to scale down or scale up your managed cluster from the new hub cluster.
```

clusters/cluster_lifecycle/import_ocp.adoc
carbonin: Right, so you can import and scale up an existing cluster regardless of how it was installed originally.

This new change is related to continuing to manage and view the hosts in the cluster as they were on the original hub cluster where they were installed (via assisted). The fact that the record of these hosts is preserved allows us to enable scale down (where previously we could only scale up).

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: I like that suggestion, thanks Nick. With the necessary style edits it would be this:

```suggestion
You can manage a cluster from a new hub cluster by creating a copy of the original resources and applying them to the new hub cluster. You can then scale down or scale up your managed cluster from the new hub cluster.
```

If that sounds good to everyone I'll go ahead and merge it

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: How about:

**Important:** You can only scale down imported {ocp-short} managed clusters if they were installed by the {ai}.

?

clusters/cluster_lifecycle/import_ocp.adoc
```diff
@@ -172,3 +173,42 @@ oc get infraenv -n managed-cluster some-other-infraenv -ojson | jq ".status.<url
 ----
 
 . *Optional:* If you want to use {acm-short} features, such as policies, on your {ocp-short} cluster, create a `ManagedCluster` resource. Make sure that the name of your `ManagedCluster` resource matches the name of your `ClusterDeplpoyment` resource. If you are missing the `ManagedCluster` resource, your cluster status is `detached` in the console.
+
+[#import-ocp-cluster-resources]
+== Importing cluster resources
+
+If your {ocp-short} managed cluster was installed by the {ai}, you can move the managed cluster and its resources from one hub cluster to another hub cluster.
```
jc-berger, line 180: Great intro! Awesome revisions :D

clusters/cluster_lifecycle/import_ocp.adoc
jc-berger: @oafischer yes, merge it, it's great! Thanks so much for taking the time and the follow up to make changes while maintaining technical accuracy!

clusters/cluster_lifecycle/import_ocp.adoc
jc-berger: @oafischer That's good by me!

clusters/cluster_lifecycle/import_ocp.adoc
jc-berger: @oafischer thanks so much for taking time to explain it to me! Yes, this revision of yours definitely improves the user flow and allows this section to build upon the previous section. 

Commit the change :)

jc-berger (approved):
/lgtm

oafischer:
Made the updates, thanks for indulging in my discussions @jc-berger!

@carbonin If you don't mind checking over the PR one more time to make sure we still have all the technical info we want to convey (and that everything is still accurate), that would be great. Thanks as well!

QE should also be taking a look now.

carbonin:
Looks good :+1: 

Not sure if you want me to actually `lgtm` though

oafischer:
Just letting me know that it looks OK is enough, thanks @carbonin. I'll wait to see if we can get QE approval on this as well, if not I'll merge it soon. Thanks again for all the reviews everyone.

clusters/cluster_lifecycle/import_ocp.adoc
```diff
@@ -172,3 +173,40 @@ oc get infraenv -n managed-cluster some-other-infraenv -ojson | jq ".status.<url
...
+
+- `Agent`
+- `AgentClassification`
+- `AgentClusterInstall`
+- `BareMetalHost`
+- `ClusterDeployment`
+- `InfraEnv`
+- `NMStateConfig`
+
+You can manage a cluster from a new hub cluster by creating a copy of the original resources and applying them to the new hub cluster. You can then scale down or scale up your managed cluster from the new hub cluster.
```
ddmitrie, line 192: It conflicts with line 182

clusters/release_notes/mce_whats_new.adoc
```diff
@@ -34,10 +34,13 @@ If you are using {mce-short} 2.6 and earlier, the hosted control planes document
 
 Learn about new features and enhancements for Cluster lifecycle with {mce-short}.
 
-* You can now set a duration to choose when the `kubeconfig` bootstrap in the klusterlet manifest expires. To learn more, see xref:../cluster_lifecycle/import_agent.adoc#cluster-import-agent[Importing a cluster].
+- You can now set a duration to choose when the `kubeconfig` bootstrap in the klusterlet manifest expires. To learn more, see xref:../cluster_lifecycle/import_agent.adoc#cluster-import-agent[Importing a cluster].
 
-* You can now connect to {ocm} with Service Account credentials. To learn more, see xref:../credentials/credential_ocm.adoc#creating-a-credential-for-openshift-cluster-manager[Creating a credential for Red Hat {ocm}].
+- You can now import all cluster resources and continue using them after moving a managed cluster that was installed by the {ai} from one hub cluster to another hub cluster. To learn more, see xref:../cluster_lifecycle/import_ocp.ado#import-ocp-cluster-resources[Importing cluster resources].
```
ddmitrie, line 39: There is no mentions how to remove managed cluster from the source hub. So it looks like spoke cluster will be managed by two hub clusters simultaneously. Not sure it is supported configuration.

oafischer: As mentioned in Slack, we will add this later when @danielerez returns.

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: I'm not sure I see the conflict. If the cluster was installed by AI it can be scaled down. If it wasn't, then it can't be scaled down. Is that not correct?

oafischer: Maybe it might be clearer if we merge both paragraphs. I can add the info from the important note to line 194, then just remove line 182. Would that help?

oafischer:
Thanks for the QE review @ddmitrie, I made a few changes based on your comments made here and on Slack that should address most of the concerns. Let me know if it's better now.

clusters/cluster_lifecycle/import_ocp.adoc
ddmitrie: It would, yes.

Do you mean that after moving the spoke to target hub cluster it can be scaled up as well as scaled down?

ddmitrie (approved):

PR #7127, opened 2024-10-21
https://issues.redhat.com/browse/ACM-14919

dockerymick:


release_notes/acm_whats_new.adoc
```diff
@@ -86,6 +86,10 @@ The following new summary cards are available:
 - Cluster health
 - Personalized view of your resources
 
+* You can now export data in a CSV file by using selecting the *Export* button. See link:../console/console_access.adoc#accessing-your-console[Accessing your console].
+
+* When you use the Search feature from the console, you now receive results that match the exact strings that you enter and range-based search parameters. See link:../observability/search_console.adoc#search-console-intro[Search in the console].
```
Randy424: This needs a small tweak. You will either get a fuzzy search result (if you use the ordinary search) or get an exact string match result if you use our advanced search feature, which is accessible by selecting the dropdown popper in the search input field.  See screenshots below for illustration.

Screenshot 2024-10-21 at 10 36 34 PM

Screenshot 2024-10-21 at 10 36 40 PM

oafischer (approved):
Looks good from doc perspective. Thanks!

release_notes/acm_whats_new.adoc
dockerymick: @Randy424 what is the name of the button for the drop-down arrow? Is it Advanced search? Why is it called "Fuzzy search" if the user is getting more with the additional filters? Should it be called "Advanced search"?

Randy424: We haven't given a name to the dropdown button. The name "Advanced search" was used mostly to describe the feature itself during development and it represents a wider array of search options, including our fuzzy search. "Fuzzy search" is just a duplicated field for the original search bar (which uses a fuzzy search method). The fields below it use the strict search. We included it to make it easier to use both search methods from within the popper.

Randy424: I think "Advanced search" works. 🤷🏾‍♂️

dockerymick: @Randy424 thanks for explaining! I googled "Fuzzy search" and it seems like saying approximate string match might be more clear. I'm not sure how the word "Fuzzy" is handled for translation

observability/search_console.adoc
```diff
@@ -151,7 +151,7 @@ See the following example where memory resources are applied to the indexer pod:
           memory: 1Gi 
 ----
 
-- Node placement for search pods:
+- Define the node placement for search pods.
```
dockerymick:
```suggestion
- You can define the node placement for search pods.
```

observability/search_console.adoc
```diff
@@ -151,7 +151,7 @@ See the following example where memory resources are applied to the indexer pod:
           memory: 1Gi 
 ----
 
-- Node placement for search pods:
+- You can define the node placement for search pods.
 +
 You can update the `Placement` of search pods by using the `nodeSelector` parameter, or the `tolerations` parameter. View the following example configuration:
```
dockerymick:
```suggestion
You can update the `Placement` resource of search pods by using the `nodeSelector` parameter, or the `tolerations` parameter. View the following example configuration:
```

Randy424 (approved):
/lgtm

PR #7159, opened 2024-10-23
2.9 refresh + Errata

oafischer:


swopebe:
Does anyone know why the previous Errata note is commented out in the conflict (prod branch). 
I see 2.9.5 did not get published. I am not sure why, as there was not issue or note to me about this. I may have missed it but something should have been communicated at some point. 

UPDATE:

We did discuss in the issue for 2.9.5

The commented out in prod was because the release did not go out, but then we never went back after.

I will remove the comments.

**Plan:**

Let's no longer close the Errata issues until after refresh and link is posted.
Let's avoid updating the prod branch because we would have to remember to go undo that.

swopebe (approved):

PR #7168, opened 2024-10-23
Update policy_generator.adoc updates

swopebe:
We need to initiate an update anyway, but reading this, I felt a little disoriented because we introduce many samples at once with a lot of stuff, but didn't say much about them. For screen readers, I like to point out things early in the intro.

Also, since we only have one call out, we can add it to the top.

### DIFF
--- a/governance/policy_generator.adoc
+++ b/governance/policy_generator.adoc
@@ -87,7 +87,7 @@ object-templates-raw: |
   {{- end }}
 ----
 
-- Then create a `PolicyGenerator` configuration file similar to the following YAML. For `path`, enter the path to your `manifest.yaml` file:
+- Next create a `PolicyGenerator` configuration file similar to the following YAML. For `path`, enter the path to your `manifest.yaml` file:
 
 +
 [source,yaml]
--- a/governance/policy_generator.adoc
+++ b/governance/policy_generator.adoc
@@ -102,7 +102,7 @@ policyDefaults:
 policies:
   - name: config-data
     manifests:
-      - path: manifest.yaml <1>
+      - path: manifest.yaml
 ----
 
 - The generated `Policy` resource, along with the generated `Placement` resource, and `PlacementBinding` resource, might resemble the following examples. *Note:* Specifications for resources are described in the _Policy Generator configuration reference table_. 
--- a/governance/policy_generator.adoc
+++ b/governance/policy_generator.adoc
@@ -36,7 +36,7 @@ generators:
 ----
 <1> The `policy-generator-config.yaml` file that is referenced in the previous example is a YAML file with the instructions of the policies to generate.
 
-- A simple `PolicyGenerator` configuration file might resemble the following example:
+- A simple `PolicyGenerator` configuration file might resemble the following example, with `policyDefaults` and `policies`:
 
 +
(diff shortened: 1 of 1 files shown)


governance/policy_generator.adoc
```diff
@@ -68,9 +68,7 @@ data:
   key2: value2
 ----
 
-- Use the `object-templates-raw` manifest to automatically generate a `ConfigurationPolicy` resource with the content you add. See the following examples:
-
-** For example, create a manifest file with the following syntax:
+- Use the `object-templates-raw` manifest to automatically generate a `ConfigurationPolicy` resource with the content you add. For example, create a manifest file with the following syntax:
```
swopebe: We only needed to say example once here

governance/policy_generator.adoc
```diff
@@ -106,9 +104,8 @@ policies:
     manifests:
       - path: manifest.yaml <1>
 ----
-<1> Path to your `manifest.yaml` file.
```
swopebe, line 109: This wasn't a complete sentence, but also, it can move up as the only callout.

governance/policy_generator.adoc
```diff
@@ -106,9 +104,8 @@ policies:
...
 
-- The generated `Policy` resource, along with the generated `Placement` and `PlacementBinding` might resemble the following example:
+- The generated `Policy` resource, along with the generated `Placement` resource, and `PlacementBinding` resource, might resemble the following examples. *Note:* Specifications for resources are described in the _Policy Generator configuration reference table_.
```
swopebe, line 108: Since we don't go into the specs here, and we have three large samples all together (we should separate them but I won't fuss about that now), I wanted to say more about how to find info about the values.

governance/policy_generator.adoc
```diff
@@ -89,7 +87,7 @@ object-templates-raw: |
   {{- end }}
 ----
 
-** Then create a `PolicyGenerator` configuration file similar to the following YAML:
+- Then create a `PolicyGenerator` configuration file similar to the following YAML. For `path`, enter the path to your `manifest.yaml` file:
```
dockerymick: This is intended to be a sub-bullet point of the "Use the object-templates-raw` manifest.

governance/policy_generator.adoc
```diff
@@ -106,9 +104,8 @@ policies:
     manifests:
       - path: manifest.yaml <1>
```
dockerymick:
```suggestion
      - path: manifest.yaml
```

governance/policy_generator.adoc
swopebe: I don't understand why we need a sub bullet, though.

swopebe: Actually, we should say.. Next...

However, if we are having them do something in chronological order, why are they not in a numbered list?

swopebe: took to slack

dockerymick (approved):
Thanks Brandi!

governance/policy_generator.adoc
swopebe: oops, missed this in refreshing the PR, will catch it in the next commit I make. TY

PR #7189, opened 2024-10-28
https://issues.redhat.com/browse/ACM-14686

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-14686


troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,103 @@
...
+The maximum transmission unit (MTU) can cause the end-to-end data plane test failure. Verify if the MTU causes the failure by running an end-to-end data plane test that uses a small packet size. To run this kind of test, complete the following step: 
+
+* In your Submariner workspace, run the following test: 
+
+[source,bash]
+----
+subctl verify --verbose --connectivity only --context <from_context> --tocontext <to_context> --image-override submariner-nettest=quay.io/submariner/nettest:devel --packet-size 200
+----
+
+If the test succeeds with this small packet size, you can prevent more test failures by using Submariner to customize the transmission control protocol (TCP) minimum security standards (MSS) `clamping` value set. Customize the TCP MSS by completing the following steps:
```
jc-berger: Dev and QA reviewers, reading the Jira, I understood this customization as a way to prevent further end to end test failures, please correct me if I am wrong.

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,103 @@
+[#trouble-submariner-test-failure]
+= Troubleshooting Submariner end-to-end test failures 
+
+After running Submariner end-to-end tests, you might get failures. Use the following sections to help you troubleshoot these end-to-end test failures.
```
jc-berger, line 4: Dev and QA reviewers, 

There were a lot of acronyms I was unfamiliar with. Please correct me if I was wrong about any of their full names and titles.

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,103 @@
...
+If the test succeeds with this small packet size, you can prevent more test failures by using Submariner to customize the transmission control protocol (TCP) minimum security standards (MSS) `clamping` value set. Customize the TCP MSS by completing the following steps: 
+
+. Delete the `Submariner-postrouting` chain in the `mangle` table by running the following command on all the `RouteAgent` pods: 
++
+[source,bash]
+----
+iptables -t mangle -F SUBMARINER-POSTROUTING
+----
+
+. Set the MSS `clamping` value set by annotating the `GW` node. For example, run the following command with a value of `1200`:
```
jc-berger: I couldn't determine what GW stands for. Does anyone know?

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,103 @@
...
+----
+oc delete pod -n submariner-operator -l app=submariner-routeagent 
+----
+
+[#symptom-submariner-dataplane-bare-metal-clusters]
+== Symptom: Submariner end-to-end test fails for bare metal clusters 
+
+The end-to-end data plane tests might fail for the bare metal cluster for the following reasons: 
+
+* The `inter-cluster` tunnel is healthy.
```
jc-berger: Dev and QA reviewers, reading the Jira, it seems to me that the test might fail in the bare metal cluster due to the following reasons. Is that right? IT seems odd to say it could fail because the tunnel is healthy. 

Anyway, please correct me if I'm wrong.

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,103 @@
+[#trouble-submariner-test-failure]
+= Troubleshooting Submariner end-to-end test failures 
+
+After running Submariner end-to-end tests, you might get failures. Use the following sections to help you troubleshoot these end-to-end test failures.  
+
+[#symptom-submariner-data-plane]
+== Symptom: Submariner end-to-end data plane test fails
+
+When the end-to-end data plane tests fails, the Submariner tests shows a successful connection even though the pods remain in the `listening` phase.
```
aswinsuryan: Something like this would be better "the Submariner tests show that connector pod can connect to listener pod, but later the pod get stuck in the listening phase"

troubleshooting/trouble_submariner_tests.adoc
aswinsuryan: it stands for gateway

troubleshooting/trouble_submariner_tests.adoc
aswinsuryan: Can be modified as "you can resolve the connection issues  by setting the transmission control protocol (TCP)  maximum segment size (MSS)" instead of "you can prevent more test failures by using Submariner to customize the transmission control protocol (TCP) minimum security standards (MSS) `clamping` value set."

aswinsuryan: MSS stands for maximum segment size

troubleshooting/trouble_submariner_tests.adoc
aswinsuryan: it should be like end to end to test fails even when the inter-cluster` tunnel is healthy and the if the VXLAN tunnels are used. It is like try the check sum disable step only when the above two pre-requisite are met.

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,103 @@
...
+
+[#symptom-submariner-data-plane]
+== Symptom: Submariner end-to-end data plane test fails
+
+When the end-to-end data plane tests fails, the Submariner tests shows a successful connection even though the pods remain in the `listening` phase. 
+
+[#resolving-submariner-data-plane]
+== Resolving the problem: Running a data plane test with a small packet size 
+
+The maximum transmission unit (MTU) can cause the end-to-end data plane test failure. Verify if the MTU causes the failure by running an end-to-end data plane test that uses a small packet size. To run this kind of test, complete the following step:
```
yboaron: Maybe we should elaborate on MTU issues, something like 
`Inter-cluster traffic over IPSec tunnel may fail because of MTU (Maximum Transmission Unit) issues, Verify if the MTU causes the failure by running .....`

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,103 @@
...
+* The virtual extensible local-area network (VXLAN) tunnels is used between clusters. 
+* The container network interface (CNI) is OpenShift software-defined networking (SDN). 
+
+[#resolving-submariner-dataplane-bare-metal-clusters]
+== Resolving the problem: Disabling the hardware offloading 
+
+A bug in the User Datagram Protocal (UDP) can be the root cause for these reasons contributing to the end-to-end data plane test failures for bare metal clusters. To troubleshoot this bug complete the following step:
+
+* In your UDP, disable the hardware offloading by applying the following YAML:  
++
```
oafischer: @jc-berger I'll let you address dev comments first but just a heads up that there is a formatting issue here as well

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,103 @@
...
+[#resolving-submariner-data-plane]
+== Resolving the problem: Running a data plane test with a small packet size 
+
+The maximum transmission unit (MTU) can cause the end-to-end data plane test failure. Verify if the MTU causes the failure by running an end-to-end data plane test that uses a small packet size. To run this kind of test, complete the following step: 
+
+* In your Submariner workspace, run the following test: 
+
+[source,bash]
+----
+subctl verify --verbose --connectivity only --context <from_context> --tocontext <to_context> --image-override submariner-nettest=quay.io/submariner/nettest:devel --packet-size 200
```
yboaron: should be 

 `--only connectivity` instaed of `--connectivity only`

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,103 @@
...
+
+If the test succeeds with this small packet size, you can prevent more test failures by using Submariner to customize the transmission control protocol (TCP) minimum security standards (MSS) `clamping` value set. Customize the TCP MSS by completing the following steps: 
+
+. Delete the `Submariner-postrouting` chain in the `mangle` table by running the following command on all the `RouteAgent` pods: 
++
+[source,bash]
+----
+iptables -t mangle -F SUBMARINER-POSTROUTING
+----
+
```
yboaron: This  section can be deleted

troubleshooting/trouble_submariner_tests.adoc
yboaron: This issue is relevant only if CNI is OpenShiftSDN **or** VxLAN is used for inter-cluster tunnels.

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,103 @@
...
+[#symptom-submariner-dataplane-bare-metal-clusters]
+== Symptom: Submariner end-to-end test fails for bare metal clusters 
+
+The end-to-end data plane tests might fail for the bare metal cluster for the following reasons: 
+
+* The `inter-cluster` tunnel is healthy. 
+* The virtual extensible local-area network (VXLAN) tunnels is used between clusters. 
+* The container network interface (CNI) is OpenShift software-defined networking (SDN).
```
yboaron: Should be

`The container network interface (CNI) is OpenShiftSDN`

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,103 @@
...
+The end-to-end data plane tests might fail for the bare metal cluster for the following reasons: 
+
+* The `inter-cluster` tunnel is healthy. 
+* The virtual extensible local-area network (VXLAN) tunnels is used between clusters. 
+* The container network interface (CNI) is OpenShift software-defined networking (SDN). 
+
+[#resolving-submariner-dataplane-bare-metal-clusters]
+== Resolving the problem: Disabling the hardware offloading 
+
+A bug in the User Datagram Protocal (UDP) can be the root cause for these reasons contributing to the end-to-end data plane test failures for bare metal clusters. To troubleshoot this bug complete the following step:
```
yboaron: `A bug in the User Datagram Protocal (UDP) checksum calculation by HW can be the root cause ...`

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,103 @@
...
+
+* In your UDP, disable the hardware offloading by applying the following YAML:
```
yboaron: We can remove 
`In your UDP,`

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,93 @@
...
++
+[source,bash]
+----
+oc delete pod -n submariner-operator -l app=submariner-routeagent 
+----
+
+[#symptom-submariner-data-plane-bare-metal-clusters]
+== Symptom: Submariner end-to-end test fails for bare-metal clusters 
+
+The end-to-end data plane tests might fail for the bare-metal cluster if the container network interface (CNI) is OpenShift software-defined networking (SDN), or if the virtual extensible local-area network (VXLAN) is used for the `inter-cluster` tunnels.
```
aswinsuryan:
```suggestion
The end-to-end data plane tests might fail for the bare-metal cluster if the container network interface (CNI) is OpenShiftSDN, or if the virtual extensible local-area network (VXLAN) is used for the `inter-cluster` tunnels.
```

aswinsuryan (approved):

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,93 @@
...
+* In your Submariner workspace, run the following test: 
+
+[source,bash]
+----
+subctl verify --verbose --only connectivity --context <from_context> --tocontext <to_context> --image-override submariner-nettest=quay.io/submariner/nettest:devel --packet-size 200
+----
+
+If the test succeeds with this small packet size, you can resolve the connection issues by setting the transmission control protocol (TCP) maximum segment size (MSS). Set the TCP MSS by completing the following steps: 
+
+. Set the TCP MSS `clamping` value set by annotating the gateway node. For example, run the following command with a value of `1200`:
```
yboaron:
```suggestion
. Set the TCP MSS `clamping` value by annotating the gateway node. For example, run the following command with a value of `1200`: 
```

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,93 @@
+[#trouble-submariner-test-failure]
+= Troubleshooting Submariner end-to-end test failures 
+
+After running Submariner end-to-end tests, you might get failures. Use the following sections to help you troubleshoot these end-to-end test failures.  
+
+[#symptom-submariner-data-plane]
+== Symptom: Submariner end-to-end data plane test fails
+
+When the end-to-end data plane tests fails, the Submariner tests show that the connector pod can connect to the listener pod, but later the connector pod gets stuck in the `listening` phase.
```
oafischer:
```suggestion
When the end-to-end data plane test fails, the Submariner tests show that the connector pod can connect to the listener pod, but later the connector pod gets stuck in the `listening` phase. 
```

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,93 @@
...
+
+After running Submariner end-to-end tests, you might get failures. Use the following sections to help you troubleshoot these end-to-end test failures.  
+
+[#symptom-submariner-data-plane]
+== Symptom: Submariner end-to-end data plane test fails
+
+When the end-to-end data plane tests fails, the Submariner tests show that the connector pod can connect to the listener pod, but later the connector pod gets stuck in the `listening` phase. 
+
+[#resolving-submariner-data-plane]
+== Resolving the problem: Running a data plane test with a small packet size
```
oafischer: Resolving should always have the same title as the Symptom that its trying to resolve.

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,93 @@
...
+== Symptom: Submariner end-to-end data plane test fails
+
+When the end-to-end data plane tests fails, the Submariner tests show that the connector pod can connect to the listener pod, but later the connector pod gets stuck in the `listening` phase. 
+
+[#resolving-submariner-data-plane]
+== Resolving the problem: Running a data plane test with a small packet size 
+
+The maximum transmission unit (MTU) can cause the end-to-end data plane test failure. For example, the MTU might cause the `inter-cluster` traffic over the Internet Protocol Security (IPsec) to fail. Verify if the MTU causes the failure by running an end-to-end data plane test that uses a small packet size. To run this kind of test, complete the following step: 
+
+* In your Submariner workspace, run the following test:
```
oafischer: I don't think we need all this for just a single step. We don't have this kind of formatting anywhere else in our doc that I could find, so maybe do something like:

`To run this type of test, run the following command in your Submariner workspace:`

Then remove the bullet point since we don't use single bullets.

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,93 @@
...
+
+. Set the TCP MSS `clamping` value set by annotating the gateway node. For example, run the following command with a value of `1200`: 
++
+[source,bash]
+----
+oc annotate node <node_name> submariner.io/tcp-clamp-mss=1200
+----
+
+. Restart all the `RouteAgent` pods by running the following commands: 
++
```
oafischer: There needs to be an empty line between steps and the +. We do this everywhere in our doc to avoid formatting issues.

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,93 @@
...
+oc delete pod -n submariner-operator -l app=submariner-routeagent 
+----
+
+[#symptom-submariner-data-plane-bare-metal-clusters]
+== Symptom: Submariner end-to-end test fails for bare-metal clusters 
+
+The end-to-end data plane tests might fail for the bare-metal cluster if the container network interface (CNI) is OpenShiftSDN, or if the virtual extensible local-area network (VXLAN) is used for the `inter-cluster` tunnels.
+
+[#resolving-submariner-dataplane-bare-metal-clusters]
+== Resolving the problem: Disabling the hardware offloading
```
oafischer: Resolving sections should have the same title as the Symptom.

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,93 @@
...
+
+. Restart all the `RouteAgent` pods by running the following commands: 
++
+[source,bash]
+----
+oc delete pod -n submariner-operator -l app=submariner-routeagent 
+----
+
+[#symptom-submariner-data-plane-bare-metal-clusters]
+== Symptom: Submariner end-to-end test fails for bare-metal clusters
```
oafischer, line 40: It seems we are packing multiple symptoms and resolutions into a file. I'm not opposed to this given that it's all around the same topic, but I'd be interested to hear some other views on this.

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,93 @@
...
+== Symptom: Submariner end-to-end test fails for bare-metal clusters 
+
+The end-to-end data plane tests might fail for the bare-metal cluster if the container network interface (CNI) is OpenShiftSDN, or if the virtual extensible local-area network (VXLAN) is used for the `inter-cluster` tunnels.
+
+[#resolving-submariner-dataplane-bare-metal-clusters]
+== Resolving the problem: Disabling the hardware offloading 
+
+A bug in the User Datagram Protocal (UDP) checksum calculation by the hardware can be the root cause for the end-to-end data plane test failures for bare-metal clusters. To troubleshoot this bug, complete the following step:
+
+* Disable the hardware offloading by applying the following YAML:
```
oafischer: There is still a formatting issue here. Also, same as before, we don't use single bullet points as per the style guide, so I would just merge the content here with the last sentence from the previous paragraph.

"To troubleshoot this bug, disable the hardware offloading by applying the following YAML content:"

prachiyadav (approved):

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,93 @@
+[#trouble-submariner-test-failure]
+= Troubleshooting Submariner end-to-end test failures 
+
+After running Submariner end-to-end tests, you might get failures. Use the following sections to help you troubleshoot these end-to-end test failures.  
+
+[#symptom-submariner-data-plane]
+== Symptom: Submariner end-to-end data plane test fails
+
+When the end-to-end data plane test fails, the Submariner tests show that the connector pod can connect to the listener pod, but later the connector pod gets stuck in the `listening` phase.
```
dockerymick: is "connector" and "listener" the name of the pods? If yes, we should monospace those words

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,93 @@
...
+
+. Set the TCP MSS `clamping` value by annotating the gateway node. For example, run the following command with a value of `1200`: 
+
++
+[source,bash]
+----
+oc annotate node <node_name> submariner.io/tcp-clamp-mss=1200
+----
+
+. Restart all the `RouteAgent` pods by running the following commands:
```
dockerymick:
```suggestion
. Restart all the `RouteAgent` pods by running the following command: 
```

troubleshooting/trouble_submariner_tests.adoc
```diff
@@ -0,0 +1,93 @@
...
+
+[#symptom-submariner-data-plane-bare-metal-clusters]
+== Symptom: Submariner end-to-end test fails for bare-metal clusters 
+
+The end-to-end data plane tests might fail for the bare-metal cluster if the container network interface (CNI) is OpenShiftSDN, or if the virtual extensible local-area network (VXLAN) is used for the `inter-cluster` tunnels.
+
+[#resolving-submariner-dataplane-bare-metal-clusters]
+== Resolving the problem: Submariner end-to-end test fails for bare-metal clusters 
+
+A bug in the User Datagram Protocal (UDP) checksum calculation by the hardware can be the root cause for the end-to-end data plane test failures for bare-metal clusters. To troubleshoot this bug, disable the hardware offloading by applying the following YAML:
```
dockerymick:
```suggestion
A bug in the User Datagram Protocal (UDP) checksum calculation by the hardware can be the root cause for the end-to-end data plane test failures for bare-metal clusters. To troubleshoot this bug, disable the hardware offloading by applying the following YAML file:  
```

dockerymick (changes requested):
Left a comment and couple of suggestions

dockerymick (approved):
this looks good to me! I made a commit to fix one of the links. Approving. Thanks Jake!

PR #7197, opened 2024-10-29
Moving a few entries since we have the Console section now

dockerymick:
Thanks for adding the console section Brandi. I figured it be best to move the entries out of Observability. In a refresh, I plan to move the search_console.adoc file to console per development request


release_notes/acm_whats_new.adoc
```diff
@@ -42,6 +42,29 @@ Learn about what is new in the {acm-short} integrated console.
...
+
++
+The following new summary cards are available: 
+
+- Cluster version
+- Nodes
+- Cluster recommendations
+- Cluster health
+- Personalized view of your resources
+
```
swopebe, line 61: I am changing this section bc we don't need the whole list in the what's new.

swopebe (approved):

PR #7200, opened 2024-10-30
https://issues.redhat.com/browse/ACM-14719 InfraEnv known issue

oafischer:
Also for 2.11, 2.10 and 2.9

### DIFF
--- a/clusters/release_notes/mce_known_issues.adoc
+++ b/clusters/release_notes/mce_known_issues.adoc
@@ -364,7 +364,7 @@ To avoid this error, upgrade your {ocp-short} to version 4.8.18 or later, or 4.9
 === Cannot use host inventory to boot with the discovery image and add hosts automatically
 //2.12:ACM-14719
 
-You cannot use a host inventory, or `InfraEnv` custom resource, for both booting with the discovery image and adding hosts automatically. If you used your old `InfraEnv` for `BareMetalHost` and want to boot the image yourself, you can work around the issue by creating a new `InfraEnv`.
+You cannot use a host inventory, or `InfraEnv` custom resource, for both booting with the discovery image and adding hosts automatically. If you used your previous `InfraEnv` for `BareMetalHost`, and want to boot the image yourself, you can work around the issue by creating a new `InfraEnv`.
 
 [#cluster-issues-msa]
 == Managed Service Account add-on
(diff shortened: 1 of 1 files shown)


CrystalChun (approved):
/lgtm

clusters/release_notes/mce_known_issues.adoc
```diff
@@ -360,6 +360,12 @@ If this problem occurs, it is typically on the following versions of {ocp-short}
 
 To avoid this error, upgrade your {ocp-short} to version 4.8.18 or later, or 4.9.7 or later.
 
+[#boot-discovey-auto-add-host]
+=== Cannot use host inventory for booting with the discovery image and adding hosts automatically
```
swopebe:
```suggestion
=== Cannot use host inventory to boot with the discovery image and add hosts automatically
```

clusters/release_notes/mce_known_issues.adoc
```diff
@@ -360,6 +360,12 @@ If this problem occurs, it is typically on the following versions of {ocp-short}
...
+//2.12:ACM-14719
+
+You cannot use a host inventory, or `InfraEnv` custom resource, for both booting with the discovery image and adding hosts automatically. If you used your old `InfraEnv` for `BareMetalHost` and want to boot the image yourself, you can work around the issue by creating a new `InfraEnv`.
```
swopebe: Let's not use old if we can avoid it (?); I have mentioned in meetings in the past and will mention again, better to use _previous_. I have always been told calling customers components and what not "old" can insult users who are not on the latest release or version on that component. It carries negative connotation. I know this isn't a legacy product where old means "you need to upgrade" but I think it is just good practice.

swopebe:
```suggestion
You cannot use a host inventory, or `InfraEnv` custom resource, to both boot with the discovery image and add hosts automatically. If you used your previous `InfraEnv` resource for the `BareMetalHost` resource, and you want to boot the image yourself, you can work around the issue by creating a new `InfraEnv` resource.
```

swopebe: Are there any steps for this or are we assuming they know how from previous procedures they used?

swopebe: comma here?

_image, and add_

swopebe (approved):
I won't hold it up contingent on the few improvements.

clusters/release_notes/mce_known_issues.adoc
oafischer: Steps for this are in the main AI docs. At this point users have already created the CR, so I think it's fair to assume they know to create another one. If necessary we can link to the main doc though.

oafischer: Changed it to previous

oafischer:
Made all the requested changes, merging

PR #7205, opened 2024-10-31
https://issues.redhat.com/browse/ACM-12550

oafischer:


clusters/release_notes/whats_new.adoc
```diff
@@ -42,6 +42,8 @@ Learn about new features and enhancements for Cluster lifecycle with {mce-short}
 
 - You can now use the {assist-install} to install a cluster in FIPS mode. See link:../../clusters/cluster_lifecycle/cim_enable.adoc#fips-install-cim[Installing a FIPS-enabled cluster by using the {assist-install}]. 
 
+- Managed clusters are now imported automatically if both an `AgentServiceConfig` and `ManagedCluster` custom resource are present.
```
paul-maidment: So my immediate take is that the documentation comment does not adequately describe what is actually happening.

We should be clear that we are only talking about "local-cluster"

The ManagedCluster CR that matches the following conditions will be imported as a local-cluster on a given hub

1: The ManagedCluster should have the label "local-cluster"
2: The ManagedCluster will have a label "ClusterID" this should match the cluster ID of the hub (it will in normal use)

Then the Name field of the ManagedCluster will be used as both the namespace and name of the imported local-cluster on the hub.

It's not accurate to say that any ManagedCluster will be imported as the local cluster as there may be numerous managed clusters in ACM that are not local clusters.

paul-maidment (changes requested):

clusters/release_notes/whats_new.adoc
oafischer: I'll change it to local cluster, adding all the details about the CR is too much for just a what's new entry though. Would you say it's enough to add that the `ManagedCluster` must meet certain conditions?

If we need more detail we need to find a place in the main doc to put it.

paul-maidment (approved):
/lgtm

paul-maidment (approved):
/approve
/lgtm

dockerymick (approved):
Looks good to me! Thanks Oliver!

PR #7208, opened 2024-11-01
https://issues.redhat.com/browse/ACM-14244

dockerymick:


troubleshooting/trouble_policy_addon_crash.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+[source,bash]
+----
+OOMKilled
+----
+
+[#resolving-grc-policy-framework-pod]
+== Resolving the problem: Troubleshooting _governance-policy-framework-addon_ pod on managed clusters
+
+To resolve this error, you must reduce the cached memory for the pod. Complete the following steps:
```
mprahl: It seems like this documentation is talking about the implementation details of the code change that reduced the cache size. We should not be mentioning code things like this in the documentation.

I'll let @gparvin relay what his intentions were with the Jira.

gparvin: To resolve the issue can we just link to the Known Issue?  That way we aren't managing the same content twice?  I think the known issue has the correct steps to resolve the problem, the user shouldn't try to reduce memory in other ways. Thanks!

dockerymick: Which known issue?

gparvin: It's this one https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.11/html/release_notes/release-notes#policy-status-repeated-updates. Thanks

dockerymick: @gparvin thanks for your patience here. Since there is a known issue with the solution, what do you think about us only adding the symptom as a third bullet point in the known issue?

dockerymick: @gparvin here is the link to the new PR, in case you agree with my suggestion: https://github.com/stolostron/rhacm-docs/pull/7250

gparvin: > @gparvin thanks for your patience here. Since there is a known issue with the solution, what do you think about us only adding the symptom as a third bullet point in the known issue?

When troubleshooting you would generally show the symptoms first and then the solution.  Could you re-organize this some so it flows more like a troubleshooting item? Thanks

dockerymick:
Closing this PR based on my comment/suggestion here: https://github.com/stolostron/rhacm-docs/pull/7208#discussion_r1847257876

Thanks Gus for your help and patience. We agreed to add more details to the existing known issue since the solutions provided, solve the issue: https://github.com/stolostron/rhacm-docs/pull/7250#issuecomment-2512986015

PR #7218, opened 2024-11-04
https://issues.redhat.com/browse/ACM-13550--review of the flow of wha…

swopebe:
…t's new and notification about where to find MCE release notes. Gonna merge this because the what's new entries were already reviewed. We are heading in to final build. All links work. The only major change is moving the header out of cluster known issues, which I had raised to Oliver last week.

### DIFF
--- a/clusters/release_notes/mce_known_issues.adoc
+++ b/clusters/release_notes/mce_known_issues.adoc
@@ -30,7 +30,6 @@ Cluster management known issues and limitations are part of the _Cluster lifecyc
 * <<cluster-issues-install,Installation>>
 * <<cluster-issues-mce,Cluster management>>
 * <<cluster-issues-cim,Central infrastructure management>>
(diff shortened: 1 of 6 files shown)


clusters/release_notes/mce_known_issues.adoc
```diff
@@ -30,7 +30,6 @@ Cluster management known issues and limitations are part of the _Cluster lifecyc
 * <<cluster-issues-install,Installation>>
 * <<cluster-issues-mce,Cluster management>>
 * <<cluster-issues-cim,Central infrastructure management>>
-* <<cluster-issues-msa,Managed Service Account add-on>>
```
swopebe, line 33: header not needed, placed in install bc it seems like adv config.

swopebe: could also go in cluster management

PR #7229, opened 2024-11-12
https://issues.redhat.com/browse/ACM-14396--editing

swopebe:


networking/submariner/subm_disconnected.adoc
```diff
@@ -6,11 +6,13 @@ Deploying Submariner on disconnected clusters can help with security concerns by
...
+After following the steps that are outlined in link:../../install/install_disconnected.adoc#install-on-disconnected-networks[Install in disconnected network environments], you must configure Submariner during the installation to support deployment on disconnected clusters. 
 
-. Mirror the images in the local registry. Make sure to mirror the `Submariner Operator bundle` image in the local registry before deploying Submariner on disconnected clusters.
+See the following procedure:
 
-. Choose the Submariner operator version that is compatible with {acm-short}. For instance, use `0.16.0` for {acm-short} version 2.9.
+. Mirror the images in the local registry. Make sure to mirror the `Submariner Operator bundle` image in the local registry before you deploy Submariner on disconnected clusters.
+
+. Choose the Submariner Operator version that is compatible with your {acm-short} version. For instance, use `0.16.0` for {acm-short} version 2.9.
```
swopebe, line 15: please review this addition.

syangsao: Looks good to me

networking/submariner/subm_disconnected.adoc
```diff
@@ -6,11 +6,13 @@ Deploying Submariner on disconnected clusters can help with security concerns by
 [#configuring-submariner-disconnected]
 == Configuring Submariner on disconnected clusters
 
-After following the steps outlined in link:../../install/install_disconnected.adoc#install-on-disconnected-networks[Install in disconnected network environments], you must configure Submariner during the installation to support deployment on disconnected clusters. See the following topics:
+After following the steps that are outlined in link:../../install/install_disconnected.adoc#install-on-disconnected-networks[Install in disconnected network environments], you must configure Submariner during the installation to support deployment on disconnected clusters.
```
jc-berger: Optional: 

"Completing" could be more straightforward than "following", and it seems like we could use the imperative voice n the second part:

```suggestion
After completing the steps in link:../../install/install_disconnected.adoc#install-on-disconnected-networks[Install in disconnected network environments], configure Submariner during the installation to support deployment on disconnected clusters. 
```

networking/submariner/subm_disconnected.adoc
```diff
@@ -6,11 +6,13 @@ Deploying Submariner on disconnected clusters can help with security concerns by
 [#configuring-submariner-disconnected]
 == Configuring Submariner on disconnected clusters
 
-After following the steps outlined in link:../../install/install_disconnected.adoc#install-on-disconnected-networks[Install in disconnected network environments], you must configure Submariner during the installation to support deployment on disconnected clusters. See the following topics:
+After following the steps that are outlined in link:../../install/install_disconnected.adoc#install-on-disconnected-networks[Install in disconnected network environments], you must configure Submariner during the installation to support deployment on disconnected clusters. 
 
-. Mirror the images in the local registry. Make sure to mirror the `Submariner Operator bundle` image in the local registry before deploying Submariner on disconnected clusters.
+See the following procedure:
```
jc-berger: See or complete? I think we want the users to complete these steps? 

Optional: "procedure" is correct and a RH term, but I think in our RHACM docs, we tend to say "steps" or "task" more than "procedure" 

```suggestion
Complete the following steps:
```

networking/submariner/subm_disconnected.adoc
```diff
@@ -6,11 +6,13 @@ Deploying Submariner on disconnected clusters can help with security concerns by
...
 
-. Choose the Submariner operator version that is compatible with {acm-short}. For instance, use `0.16.0` for {acm-short} version 2.9.
+. Mirror the images in the local registry. Make sure to mirror the `Submariner Operator bundle` image in the local registry before you deploy Submariner on disconnected clusters.
```
jc-berger: Optional: 

I like the first sentence as a general action for the user to complete. But since you're more specific in the second sentence, what do you think about removing the first sentence? 

```suggestion
. Mirror the `Submariner Operator bundle` image in the local registry before you deploy Submariner on disconnected clusters.
```

networking/submariner/subm_disconnected.adoc
jc-berger, line 15: Although not incorrect, IBM SG says to use "select" over "choose" for UI elements, and therefore, encourages the verb choice "select" in general. 

```suggestion
. Select the Submariner Operator version that is compatible with your {acm-short} version. For instance, use `0.16.0` for {acm-short} version 2.9.
```

jc-berger (approved):
/lgtm

networking/submariner/subm_disconnected.adoc
swopebe: Not sure if this needs to change, if it is a drop-down or something in the UI...I don't know what the user sees...so I think Choose is more general. Select indicates it could be a box, as well. Without knowledge of what the user sees in the process, I think choose is safe.

swopebe: I can ask for later versions liek 2.11 and beyond--for 2.9, let's just go with choose.

networking/submariner/subm_disconnected.adoc
swopebe: It can be either, I don't think we get this hardlined on how we say it. I like your change. Before there was nothing there.

PR #7240, opened 2024-11-15
https://issues.redhat.com/browse/ACM-14961

oafischer:
2.12 only

### DIFF
--- a/clusters/release_notes/mce_known_issues.adoc
+++ b/clusters/release_notes/mce_known_issues.adoc
@@ -340,7 +340,7 @@ If you define configurations in the `ManagedClusterAddon` to override some confi
 progressing... mca and work configs mismatch
 ----
 
-When you check the `ManagedClusterAddon` status, part of the configurations has an empty `spec` hash, even if the configurations exist. See the following example:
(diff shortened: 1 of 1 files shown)


clusters/release_notes/mce_known_issues.adoc
```diff
@@ -330,6 +330,55 @@ When you upgrade an {ocp-short} Dedicated cluster by using the `ClusterCurator`
...
+=== `ManagedClusterAddon` status becomes stuck
+
+If you define configurations in the `ManagedClusterAddon` to override some configurations in the `ClusterManagementAddon`, the `ManagedClusterAddon` might become stuck at the following status:
+
+[source,bash]
+----
+progressing... mca and work configs mismatch
+----
+
+When you check the `ManagedClusterAddon` status, part of the configurations have an empty `spec` hash, even if the configurations exist. See the following example:
```
jc-berger: if "have" is the verb for the subject, "part" then I believe it should be "has" and maybe adding "a" before part would help since it's singular. 

```suggestion
When you check the `ManagedClusterAddon` status, a part of the configurations has an empty `spec` hash, even if the configurations exist. See the following example:
```

jc-berger (approved):
/lgtm

PR #7247, opened 2024-11-18
https://issues.redhat.com/browse/ACM-13028

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-13028

### DIFF
--- a/release_notes/known_issues_continuity.adoc
+++ b/release_notes/known_issues_continuity.adoc
@@ -33,10 +33,10 @@ For more about deprecations and removals, see xref:../release_notes/deprecate_re
 Backup and restore known issues and limitations are listed here, along with workarounds if they are available.
 
 [#limited-support-ocp]
-=== Backup and restore has limited support with {ocp}
+=== Backup and restore in {acm-short} 2.10.4 has limited support with {ocp-short} 4.13
 //2.10:ACM-13028
 
-{acm} 2.10.4 uses OpenShift APIs for Data Protection (OADP) 1.4 which has limited support for {ocp-short} 4.13. To have full support for the backup and restore procedure, move to {ocp-short} 4.14 or later when you use {acm} 2.10.4.  
+{acm-short} 2.10.4 uses OpenShift APIs for Data Protection (OADP) 1.4, which has limited support for {ocp-short} 4.13. To use the supported backup and restore function, move to {ocp-short} 4.14 or later when you use {acm} 2.10.4.  
 
 [#cluster-proxy-addon-issue]
 === The _cluster-proxy-addon_ fails during hub cluster backup and restore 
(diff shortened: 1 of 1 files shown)


release_notes/known_issues_continuity.adoc
```diff
@@ -32,6 +32,12 @@ For more about deprecations and removals, see xref:../release_notes/deprecate_re
 
 Backup and restore known issues and limitations are listed here, along with workarounds if they are available.
 
+[#limited-support-ocp]
+=== Backup and restore has limited support with {ocp}
+//2.10:ACM-13028
+
+{acm} 2.10 uses OpenShift APIs for Data Protection (OADP) 1.4 which has limited support for {ocp-short} 4.13. To have full support for the backup and restore procedure, move to {ocp-short} 4.14 or later when you use {acm} 2.10.  
+
```
sahare, line 40: Based on the table here https://github.com/stolostron/cluster-backup-operator?tab=readme-ov-file#oadp-version-relationship , ACM 2.10.4 uses OADP 1.4. ACM 2.10 uses OADP 1.3 so I think this text needs to be updated according to the instructions in the Jira issue.

sahare (changes requested):

sahare (approved):
Looks good, Thanks!

release_notes/known_issues_continuity.adoc
swopebe: **Comma here:**

_1.4, which_

**Recommend changing:**

To use the supported backup and restore function, move to...

Use **{acm-short}** this deep into the file (for conciseness)

swopebe (changes requested):

release_notes/known_issues_continuity.adoc
swopebe: I would put the OCP and ACM version in the title and also use the shortened version of the conref. Since this is a known issue on a fix, it is worth mentioning early.

oafischer (approved):

release_notes/known_issues_continuity.adoc
swopebe: This has not changed in one spot:

Use {acm-short} this deep into the file (for conciseness)

Otherwise this looks good.

swopebe: Looks like it was merged without my review, so will bring that up on slack. This is a small change; I will just remind that we use the shortened version of conrefs in the files this long--though I think this is in the guides.

jc-berger: @swopebe I can make changes before closing the Jira. 

All the conrefs I've used here in this content are the short conrefs. Am I missing something?

swopebe: line 39: `{acm} 2.10.4.` and after we shorten the ones before.

PR #7279, opened 2024-11-25
https://issues.redhat.com/browse/ACM-15857 - namespace and clusterName must match in ClusterInstance

amolnar-gh:


achuzhoy (approved):

mce_acm_integration/siteconfig/siteconfig_install_clusters.adoc
```diff
@@ -182,26 +182,29 @@ spec:
...
       [...]
 ----
 <1> The `namespace` in the `ClusterInstance` custom resource must match the target namespace that you defined.
 <2> Reference the `name` of one or more extra manifests `ConfigMap` objects.
 <3> Reference the `name` of your pull secret.
-<4> Reference the `name` of the cluster-level templates under the `spec.templateRefs` field. The `namespace` must match the namespace where the Operator is installed.
-<5> Reference the `name` of the BMC secret.
-<6> Reference the `name` of the node-level templates under the `spec.nodes.templateRefs` field. The `namespace` must match the namespace where the Operator is installed.
+<4> The value of the `clusterName` field in the `ClusterInstance` custom resource must match the value of the `namespace` field.
+<5> Reference the `name` of the cluster-level templates under the `spec.templateRefs` field. The `namespace` must match the namespace where the Operator is installed.
```
sakhoury: The namespace requirement here is only if the user wants to apply the default cluster-level installation template.

mce_acm_integration/siteconfig/siteconfig_install_clusters.adoc
```diff
@@ -182,26 +182,29 @@ spec:
...
+<6> Reference the `name` of the BMC secret.
+<7> Reference the `name` of the node-level templates under the `spec.nodes.templateRefs` field. The `namespace` must match the namespace where the Operator is installed.
```
sakhoury: Same here, the namespace requirement here is only if the user wants to apply the default node-level installation template. For example, the user may reference their own templates which can exist in other namespaces.

jc-berger (approved):
/lgtm

PR #7288, opened 2024-11-26
https://issues.redhat.com/browse/ACM-11974--small edits to catch up

swopebe:


clusters/install_upgrade/install_disconnected.adoc
```diff
@@ -3,7 +3,7 @@
 
 You might need to install the {mce-short} on {ocp} clusters that are not connected to the Internet. The procedure to install on a disconnected engine requires some of the same steps as the connected installation.
 
-*Important:* You must install {mce-short} on a cluster that does not have Red Hat Advanced Cluster Management for Kubernetes earlier than 2.5 installed. The {mce-short} cannot co-exist with Red Hat Advanced Cluster Management for Kubernetes on versions earlier than 2.5 because they provide some of the same management components. It is recommended that you install {mce-short} on a cluster that has never previously installed Red Hat Advanced Cluster Management. If you are using Red Hat Advanced Cluster Management for Kubernetes at version 2.5.0 or later then {mce-short} is already installed on the cluster with it.
+*Important:* Install {mce-short} on a cluster that does not have {acm-short} installed because the two products provide some of the same management components. Install {mce-short} on a cluster that has never previously installed {acm-short}. If you are using {acm-short} version 2.5.0 or later, {mce-short} is already installed on your cluster.
```
amolnar-gh: A small suggestion depending on who's doing the installation. I see it was phrased similarly before the change but it doesn't sound right that the cluster installs RHACM, however, I could be missing context/knowledge :) 
```suggestion
*Important:* Install {mce-short} on a cluster that does not have {acm-short} installed because the two products provide some of the same management components. Install {mce-short} on a cluster that has never had {acm-short} previously installed. If you are using {acm-short} version 2.5.0 or later, {mce-short} is already installed on your cluster.
```

amolnar-gh (approved):
One small comment but feel free to ignore if it doesn't make sense.

PR #7294, opened 2024-11-28
ACM-16051: importing cluster resources

danielerez:
Following https://github.com/stolostron/rhacm-docs/pull/7121, added further information and steps for the flow of moving a managed cluster between ACM hubs (i.e. by importing cluster resources).

### DIFF
--- a/clusters/cluster_lifecycle/import_ocp.adoc
+++ b/clusters/cluster_lifecycle/import_ocp.adoc
@@ -179,7 +179,7 @@ oc get infraenv -n managed-cluster some-other-infraenv -ojson | jq ".status.<url
 
 If your {ocp-short} managed cluster was installed by the {ai}, you can move the managed cluster and its resources from one hub cluster to another hub cluster. 
 
-You can manage a cluster from a new hub cluster by creating a copy of the original resources and applying them to the new hub cluster. You can then scale down or scale up your managed cluster from the new hub cluster.
+You can manage a cluster from a new hub cluster by saving a copy of the original resources and applying them to the new hub cluster. You can then scale down or scale up your managed cluster from the new hub cluster.
 
 *Important:* You can only scale down imported {ocp-short} managed clusters if they were installed by the {ai}.
 
@@ -196,8 +196,8 @@ You can import the following resources and continue to manage your cluster with 
 - `Secret`
 ** Includes `admin-kubeconfig` and `bmc-secret`
 
-[#copy-apply-cluster-resources]
-=== Copying and applying managed cluster resources
+[#store-apply-cluster-resources]
+=== Saving and applying managed cluster resources
 
 Complete the following steps to import managed cluster resources and apply them to a new hub cluster: 
 
--- a/clusters/cluster_lifecycle/import_ocp.adoc
+++ b/clusters/cluster_lifecycle/import_ocp.adoc
@@ -184,7 +184,7 @@ You can manage a cluster from a new hub cluster by saving a copy of the original
 *Important:* You can only scale down imported {ocp-short} managed clusters if they were installed by the {ai}.
 
 [#import-ocp-cluster-resources-table]
-== Manage cluster resource table
+=== Manage cluster resource table
 
 You can import the following resources and continue to manage your cluster with them:
 
--- a/clusters/cluster_lifecycle/import_ocp.adoc
+++ b/clusters/cluster_lifecycle/import_ocp.adoc
@@ -231,13 +231,15 @@ You can import the following resources and continue to manage your cluster with 
 
 To save a copy of your managed cluster resources and apply them to a new hub cluster, complete the following steps:
 
-. Get your resources from your source hub cluster by running the following command. Repeat the command for every resource you want to import by replacing `<resource_name>` with the name of the resource. Replace other values where needed:
+. Get your resources from your source hub cluster by running the following command. Replace values where needed:
 
 +
 [source,bash]
 ----
 oc –kubeconfig <source_hub_kubeconfig> -n <managed_cluster_name> get <resource_name> <cluster_provisioning_namespace> -oyaml > <resource_name>.yaml
 ----
+
+.. Repeat the command for every resource you want to import by replacing `<resource_name>` with the name of the resource.
 
 . Remove the `ownerReferences` property from the following resources by running the following commands:
 +
@@ -255,7 +257,7 @@ yq --in-place -y 'del(.metadata.ownerReferences)' AgentClusterInstall.yaml
 yq --in-place -y 'del(.metadata.ownerReferences)' AdminKubeconfigSecret.yaml
 ----
 
-. Detach the managed cluster from the source hub cluster by running the following command to your `ManagedCluster` resource. Replace values where needed:
+. Detach the managed cluster from the source hub cluster by running the following command. Replace values where needed:
 
 +
 [source,bash]
@@ -265,7 +267,7 @@ oc –kubeconfig <target_hub_kubeconfig> delete ManagedCluster <cluster_name>
 
 . Create a namespace on the target hub cluster for the managed cluster. Use a similar name as the source hub cluster.
 
-. Apply your stored resources on the target hub cluster individually by running the following command. Repalce values where needed:
+. Apply your stored resources on the target hub cluster individually by running the following command. Replace values where needed:
 +
 *Note:* Replace `<resource_name>.yaml` with `.` if you want to apply all the resources as a group instead of individually.
 +
--- a/clusters/cluster_lifecycle/import_ocp.adoc
+++ b/clusters/cluster_lifecycle/import_ocp.adoc
@@ -179,22 +179,53 @@ oc get infraenv -n managed-cluster some-other-infraenv -ojson | jq ".status.<url
 
 If your {ocp-short} managed cluster was installed by the {ai}, you can move the managed cluster and its resources from one hub cluster to another hub cluster. 
 
-You can manage a cluster from a new hub cluster by saving a copy of the original resources and applying them to the new hub cluster. You can then scale down or scale up your managed cluster from the new hub cluster.
+You can manage a cluster from a new hub cluster by saving a copy of the original resources, applying them to the new hub cluster, and deleting the original resources. You can then scale down or scale up your managed cluster from the new hub cluster.
 
 *Important:* You can only scale down imported {ocp-short} managed clusters if they were installed by the {ai}.
 
+[#import-ocp-cluster-resources-table]
+== Manage cluster resource table
+
 You can import the following resources and continue to manage your cluster with them:
 
-- `Agent`
-- `AgentClassification`
-- `AgentClusterInstall`
-- `BareMetalHost`
-- `ClusterDeployment`
-- `InfraEnv`
-- `NMStateConfig`
-- `ManagedCluster`
-- `Secret`
-** Includes `admin-kubeconfig` and `bmc-secret`
+|===
+| Resource | Optional or required | Description
+
+| `Agent`
+| Required
+| 
+
+| `AgentClassification`
+| Optional
+| 
+
+| `AgentClusterInstall`
+| Required
+| 
+
+| `BareMetalHost`
+| Required
+| 
+
+| `ClusterDeployment`
+| Required
+| 
+
+| `InfraEnv`
+| Required
+| 
+
+| `NMStateConfig`
+| Optional
+| 
+
+| `ManagedCluster`
+| Required
+| 
+
+| `Secret`
+| Required
+| Includes `admin-kubeconfig` and `bmc-secret`
 
 [#save-apply-cluster-resources]
 === Saving and applying managed cluster resources
--- a/clusters/cluster_lifecycle/import_ocp.adoc
+++ b/clusters/cluster_lifecycle/import_ocp.adoc
@@ -226,6 +226,7 @@ You can import the following resources and continue to manage your cluster with 
 | `Secret`
 | Required
 | Includes `admin-kubeconfig` and `bmc-secret`
+|===
 
 [#save-apply-cluster-resources]
 === Saving and applying managed cluster resources
--- a/clusters/cluster_lifecycle/import_ocp.adoc
+++ b/clusters/cluster_lifecycle/import_ocp.adoc
@@ -188,27 +188,76 @@ You can import the following resources and continue to manage your cluster with 
 - `ClusterDeployment`
 - `InfraEnv`
 - `NMStateConfig`
+- `ManagedCluster`
+- `Secret` (admin-kubeconfig and bmc-secret)
 
 You can manage a cluster from a new hub cluster by creating a copy of the original resources and applying them to the new hub cluster. You can then scale down or scale up your managed cluster from the new hub cluster.
 
 *Important:* You can only scale down imported {ocp-short} managed clusters if they were installed by the {ai}.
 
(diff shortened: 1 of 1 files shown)


swopebe:
@danielerez thanks for the PR. We normally like to move doc ourselves because it impacts the build and we set up each book as similarly as possible to the larger collection. We welcome smaller PRs from developers, for sure.

Can you point us to the issue where this move was requested? We also want to understand what we are doing to the content and why, so that is why we are asking for the issue.

**Please see the playbook for the ACM Team (Doc section) here:** https://docs.google.com/document/d/1YTqpZRH54Bnn4WJ2nZmjaCoiRtqmrc2w6DdQxe_yLZ8/edit?tab=t.0#heading=h.9fvyr2rdriby

I think @oafischer (who may already have your issue; but if not, please open one?) needs to be aware of this PR as the writer for this area to ensure it is tracked in the issue, written within standards. We can leave Jake to finalize with the peer review. 

(I see a closed issue connected to a different PR, but we do need an active issue on such content changes.)

danielerez:
> @danielerez thanks for the PR. We normally like to move doc ourselves because it impacts the build and we set up each book as similarly as possible to the larger collection. We welcome smaller PRs from developers, for sure.
> 
> Can you point us to the issue where this move was requested? We also want to understand what we are doing to the content and why, so that is why we are asking for the issue.

So this change is required for expanding the details and steps of the flow.
I.e. addressing the content added prior feature completeness: https://github.com/stolostron/rhacm-docs/pull/7121/files#r1816294651
@oafischer Let me know if we need to create a new ticket specifically for that. Or, can we just reuse the original [issue](https://issues.redhat.com/browse/ACM-14360)? 

> 
> **Please see the playbook for the ACM Team (Doc section) here:** https://docs.google.com/document/d/1YTqpZRH54Bnn4WJ2nZmjaCoiRtqmrc2w6DdQxe_yLZ8/edit?tab=t.0#heading=h.9fvyr2rdriby
> 
> I think @oafischer (who may already have your issue; but if not, please open one?) needs to be aware of this PR as the writer for this area to ensure it is tracked in the issue, written within standards. We can leave Jake to finalize with the peer review.
> 
> (I see a closed issue connected to a different PR, but we do need an active issue on such content changes.)

swopebe:
To move content after the doc issue is closed, yes, we do need a new issue. If it was a small change, no biggie. Moving content impacts the build and the writer has more editing (usually) in these cases. This is especially true if it spans versions. Thanks so much.

danielerez:
> To move content after the doc issue is closed, yes, we do need a new issue. If it was a small change, no biggie. Moving content impacts the build and the writer has more editing (usually) in these cases. This is especially true if it spans versions. Thanks so much.

Sure, created a new ticket for this change: https://issues.redhat.com/browse/ACM-16051

clusters/cluster_lifecycle/import_ocp.adoc
```diff
@@ -188,27 +188,76 @@ You can import the following resources and continue to manage your cluster with
 - `ClusterDeployment`
 - `InfraEnv`
 - `NMStateConfig`
+- `ManagedCluster`
+- `Secret` (admin-kubeconfig and bmc-secret)
```
oafischer: Are these 3 different resources, or what do the brackets indicate in this case?

clusters/cluster_lifecycle/import_ocp.adoc
```diff
@@ -188,27 +188,76 @@ You can import the following resources and continue to manage your cluster with
...
 
 You can manage a cluster from a new hub cluster by creating a copy of the original resources and applying them to the new hub cluster. You can then scale down or scale up your managed cluster from the new hub cluster.
 
 *Important:* You can only scale down imported {ocp-short} managed clusters if they were installed by the {ai}.
 
-Import your cluster resources by creating a copy of your resources on your target hub cluster. See the following example to learn how to create a copy of the `Agent` resource and apply similar steps to your other resources:
+[#store-resources-source-hub]
+=== Store resources from the source hub
```
oafischer: What is the goal with the 4 new topics you added? Are these all the steps required to import the cluster resources and then clean up the remaining ones on the source hub? If so, these 4 topics need to be converted to steps and added to two topics, one for importing and one for deleting.

If you don't mind I can go ahead and rearrange it directly in the PR.

oafischer (changes requested):
Thanks for creating the PR and the new issue @danielerez. The content looks good, but we need to rearrange and edit a bunch of stuff to conform with style and company guidelines. We need overarching topics that describe the user goals.

In this case I can see 2 goals, one being importing the resources, and the second one being deleting/cleaning up. The topics you added would be considered steps to reach those goal, so we need to add them to those 2 topics I mentioned.

I'm happy to do this directly in the PR if you don't mind. Just let me know what you prefer.

clusters/cluster_lifecycle/import_ocp.adoc
danielerez: So these are just for implying the names of the secrets.
I.e. for a managed cluster in namespace 'spoke':
* Secret with name 'spoke-admin-kubeconfig'
* Secret with name 'spoke-bmc-secret' (optional - if using BMH)
Not sure if we need to include these full details, so added the brackets as a hint to store all required secrets.

clusters/cluster_lifecycle/import_ocp.adoc
danielerez: Sure, this is just a suggestion, feel free to rearrange it as you see fit.
Anyway, right, maybe it can be converted into an import step and a cleanup step (destroy cluster).
Though I suppose the import step should still include these inner-steps, just to make it clearer.
I.e.
1. import:
1.1. store
1.2. modify
1.3. detach
1.4. apply on target
2. cleanup (destroy)

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: We generally try avoid brackets as per our style guide, though if necessary we can keep them.

Are those 2 secrets the only ones you can import? Meaning there are more but you can only import those 2?

Or are we just listing the 2 that exist for clarity?

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: Sounds good, I'll go ahead and rearrange it as needed and make some edits while I'm at it. I'll let you know when I'm done so you can take another look afterwards.

danielerez:
> Thanks for creating the PR and the new issue @danielerez. The content looks good, but we need to rearrange and edit a bunch of stuff to conform with style and company guidelines. We need overarching topics that describe the user goals.
> 
> In this case I can see 2 goals, one being importing the resources, and the second one being deleting/cleaning up. The topics you added would be considered steps to reach those goal, so we need to add them to those 2 topics I mentioned.
> 
> I'm happy to do this directly in the PR if you don't mind. Just let me know what you prefer.

Sure, this is just a suggestion, feel free to rearrange it as you see fit.
Anyway, right, maybe it can be converted into an import step and a cleanup step (destroy cluster).
Though I suppose the import step should still include these inner-steps, just to make it clearer.
E.g.
1. Import step:
1.1. Store resources
1.2. Modify resources
1.3. Detach cluster from source hub
1.4. Apply resources on target hub
2. Cleanup
Destroy cluster on source hub

clusters/cluster_lifecycle/import_ocp.adoc
danielerez: Potentially there could be more secrets, depending on the exact deployment.
I've emphasized these secrets just for clarity indeed, especially as the `admin-kubeconfig` Secret is always mandatory.
Feel free to illustrate it in a more conventional way instead of the brackets:) Maybe a short note is more adhering to our style guide(?)

oafischer: A note might be better but I need to play around with it a bit and see what works best. I'll tackle this along with the rearranging.

clusters/cluster_lifecycle/import_ocp.adoc
```diff
@@ -188,12 +192,14 @@ You can import the following resources and continue to manage your cluster with
...
 - `NMStateConfig`
+- `ManagedCluster`
+- `Secret`
+** Includes `admin-kubeconfig` and `bmc-secret`
 
-You can manage a cluster from a new hub cluster by creating a copy of the original resources and applying them to the new hub cluster. You can then scale down or scale up your managed cluster from the new hub cluster.
-
-*Important:* You can only scale down imported {ocp-short} managed clusters if they were installed by the {ai}.
+[#copy-apply-cluster-resources]
+=== Copying and applying managed cluster resources
```
danielerez: nit: instead of 'Copying', maybe 'Storing' or 'Saving' is a clearer terminology?

clusters/cluster_lifecycle/import_ocp.adoc
```diff
@@ -188,12 +192,14 @@ You can import the following resources and continue to manage your cluster with
...
 
-Import your cluster resources by creating a copy of your resources on your target hub cluster. See the following example to learn how to create a copy of the `Agent` resource and apply similar steps to your other resources:
+Complete the following steps to import managed cluster resources and apply them to a new hub cluster: 
 
 . Get the `Agent` resource from your source hub cluster by running the following command. Replace values where needed:
```
danielerez: Can we emphasize here that it should be performed for all resources? I.e. just since, in first glance, it looks like only 'Agent' is required.

oafischer: Good call, I'll add that.

oafischer: Just to clarify, does every resource have to be copied or are there cases where a user might only need to or want to copy some resources?

What do you think about making the resource name in the command a variable as well?

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: I'm OK with storing or saving. We should build a connection to the previous paragraph where we explain this feature. We used copying in that paragraph, but I can change that to storing or saving as well if it's more accurate.

danielerez: Sure, makes sense.

clusters/cluster_lifecycle/import_ocp.adoc
danielerez: Yeah, good idea. Some of the resources are mandatory, a few are optional depending on the env (e.g. NMStateConfig, AgentClassification). So right, we can simply use a var for the name.

oafischer: If some are mandatory we need to state that. We usually use a table with a column that reads optional or required for that. I will convert the resource list to table and add that column. For the resources where it depends on the environment, can you provide a one or two sentence explanation of what it depends on? We can add that to a Description column in the table.

danielerez: Well, not sure if it's the right place to explain this thoroughly, as every resource has its responsibility.
But let's see:
- `Agent` - Mandatory
- `AgentClassification` - Optional (requires to classify Agents by providing a filter query)
- `AgentClusterInstall` - Mandatory
- `BareMetalHost` - Optional (required when using `baremetal` platform)
- `ClusterDeployment` - Mandatory
- `InfraEnv` - Mandatory
- `NMStateConfig` - Optional (network configuration to apply on the hosts)
- `ManagedCluster` - Mandatory
- `Secret` - `admin-kubeconfig` is Mandatory, `bmc-secret` is Optional (required when using BareMetalHosts)

oafischer: Thanks for the list, let me know if the table looks good to you.

danielerez:
Looks good to me. Thanks @oafischer !

oafischer:
I'll hand this over for peer review now and approve it once any last comments are addressed. I'll let you know when you can go ahead and merge.

danielerez:
> I'll hand this over for peer review now and approve it once any last comments are addressed. I'll let you know when you can go ahead and merge.

Sounds good. Many thanks again!

clusters/cluster_lifecycle/import_ocp.adoc
```diff
@@ -179,36 +179,106 @@ oc get infraenv -n managed-cluster some-other-infraenv -ojson | jq ".status.<url
...
+| Required
+| The `admin-kubeconfig` secret is required. The `bmc-secret` secret is only required if you are using `BareMetalHosts`.
+|===
+
+[#save-apply-cluster-resources]
+=== Saving and applying managed cluster resources
+
+To save a copy of your managed cluster resources and apply them to a new hub cluster, complete the following steps:
+
+. Get your resources from your source hub cluster by running the following command. Repeat the command for every resource you want to import by replacing `<resource_name>` with the name of the resource. Replace other values where needed:
```
jc-berger: These seem like several steps because there's multiple actions listed here. What do you think about this? 

```suggestion
. Get your resources from your source hub cluster by running the following command:
```

clusters/cluster_lifecycle/import_ocp.adoc
```diff
@@ -179,36 +179,106 @@ oc get infraenv -n managed-cluster some-other-infraenv -ojson | jq ".status.<url
...
 
-Import your cluster resources by creating a copy of your resources on your target hub cluster. See the following example to learn how to create a copy of the `Agent` resource and apply similar steps to your other resources:
++
+[source,bash]
+----
+oc –kubeconfig <source_hub_kubeconfig> -n <managed_cluster_name> get <resource_name> <cluster_provisioning_namespace> -oyaml > <resource_name>.yaml
+----
 
-. Get the `Agent` resource from your source hub cluster by running the following command. Replace values where needed:
+. Remove the `ownerReferences` property from the following resources by running the following commands:
```
jc-berger, line 244:
```suggestion
. Repeat the command for every resource you want to import by replacing `<resource_name>` with the name of the resource. Replace other values where needed.
. Remove the `ownerReferences` property from the following resources by running the following commands:
```

jc-berger: I see why you had "Get resources, repeat command, replace other values" all as one command, so feel free to keep it as one. See if you think this flow makes it a little easier to read. 

We could also make the "Repeat the. command for every resource..." as a substep, like this: 
```suggestion
.. Repeat the command for every resource you want to import by replacing `<resource_name>` with the name of the resource. Replace other values where needed.
. Remove the `ownerReferences` property from the following resources by running the following commands:
```

clusters/cluster_lifecycle/import_ocp.adoc
jc-berger: Reading the rest of the procedure, I see you often end a step with "replace other values where needed" which is good for consistency. 

So to determine if it should be multiple steps, I guess we need to decide if "Repeat the command for every resource you want to import by replacing `<resource_name>` with the name of the resource." is a separate action or part of the one action where users get their resources from the source hub cluster. 

It seems to me like a second action, so maybe it should be its own step, thoughts?

jc-berger (approved):
/lgtm

clusters/cluster_lifecycle/import_ocp.adoc
oafischer: I can see it both ways, but I like your sub-step solution. I think I'll go with that.

oafischer (approved):
All comments addressed, feel free to merge @danielerez.

danielerez:
> All comments addressed, feel free to merge @danielerez.

@oafischer Seems I'm missing the `approved` label for merge. Can you maybe approve?

oafischer:
I can just merge it as well if you are happy with everything @danielerez

PR #7296, opened 2024-12-02
https://issues.redhat.com/browse/ACM-15319 CIM regression KI

oafischer:
2.12 only

### DIFF
--- a/clusters/release_notes/mce_known_issues.adoc
+++ b/clusters/release_notes/mce_known_issues.adoc
@@ -458,4 +458,4 @@ See link:../../add-ons/configure_nodeselector_tolerations_addons.adoc#configure-
 === Nodes shut down after removing `BareMetalHost` resource
 //2.12:ACM-15319
 
-If you remove the `BareMetalHost` resource from a hub cluster, the nodes shut down. You can turn on the nodes again by manually powering them on.
+If you remove the `BareMetalHost` resource from a hub cluster, the nodes shut down. You can manually power on the nodes again.
--- a/clusters/release_notes/mce_known_issues.adoc
+++ b/clusters/release_notes/mce_known_issues.adoc
@@ -453,3 +453,9 @@ If the `managed-serviceaccount` add-on is required, you can work around the issu
 . Update the `managed-serviceaccount` `ManagedClusterAddon` in the local cluster namespace to use the `addonDeploymentConfig` custom resource you created.
 
 See link:../../add-ons/configure_nodeselector_tolerations_addons.adoc#configure-nodeselector-tolerations-addons[Configuring nodeSelectors and tolerations for klusterlet add-ons] to learn more about how to use the `addonDeploymentConfig` custom resource to configure `tolerations` and `nodeSelector` for add-ons.
+
+[#nodes-shut-down-bmh]
+=== Nodes shut down after removing `BareMetalHost` resource
+//2.12:ACM-15319
+
+If you remove the `BareMetalHost` resource from a managed cluster, the nodes shut down.
--- a/clusters/release_notes/mce_known_issues.adoc
+++ b/clusters/release_notes/mce_known_issues.adoc
@@ -458,4 +458,4 @@ See link:../../add-ons/configure_nodeselector_tolerations_addons.adoc#configure-
 === Nodes shut down after removing `BareMetalHost` resource
 //2.12:ACM-15319
 
-If you remove the `BareMetalHost` resource from a managed cluster, the nodes shut down.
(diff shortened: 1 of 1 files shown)


clusters/release_notes/mce_known_issues.adoc
```diff
@@ -453,3 +453,9 @@ If the `managed-serviceaccount` add-on is required, you can work around the issu
 . Update the `managed-serviceaccount` `ManagedClusterAddon` in the local cluster namespace to use the `addonDeploymentConfig` custom resource you created.
 
 See link:../../add-ons/configure_nodeselector_tolerations_addons.adoc#configure-nodeselector-tolerations-addons[Configuring nodeSelectors and tolerations for klusterlet add-ons] to learn more about how to use the `addonDeploymentConfig` custom resource to configure `tolerations` and `nodeSelector` for add-ons.
+
+[#nodes-shut-down-bmh]
+=== Nodes shut down after removing `BareMetalHost` resource
+//2.12:ACM-15319
+
+If you remove the `BareMetalHost` resource from a managed cluster, the nodes shut down.
```
carbonin: This should be the BMH in the hub cluster, right?

cc @trewest

trewest: That is correct. This issue occurs when you remove the relevant spoke BMH from the hub cluster.

oafischer: Thanks for catching that, updated!

carbonin (approved):

clusters/release_notes/mce_known_issues.adoc
swopebe: @carbonin @trewest @oafischer -- for a simple known issue like this, we still need to ask ourselves what the user needs to know and when so that they don't have to stop and think: What did I miss?

1. There is no information about how to get out of this. 
2. If there is no way out, we really need to say at some point in the doc:

_Do not remove the `BareMetalHost` resource._ 

If you already say that in the main documentation and they are still doing it, we need to make it clearer there and not here in the known issues.

3. Let's say only a few users would do this and get caught up, but that most are no impacted and that is why we choose to doc it here. I think we still need to say something more:

You must reinstall the resource to get your nodes to run....
You can manually restart the nodes by....

Something to tell them how to get out of it or what the next step is.

carbonin: This is a bug so ideally we would allow the user to remove that BMH and not have the node shut down so I don't think it makes sense for us to tell them not to remove the BMH in the main doc.

The user shouldn't need to reinstall anything to get the node back, just powering it back on will do the job, but we don't know anything about how they do power management outside of the BMO integration. But if you think it's valuable to say "power the node back on" then I think it's fine to put that here.

oafischer: Thanks for the detailed explanation @carbonin. I'll add the part about powering the node back on. Let me know if that's sufficient @swopebe.

swopebe: The documentation should also prevent a bug; prevent a customer call. We also don't want to make the user have to stop and think what they did wrong or how they start over. @oafischer let me know what you decide to add here.

oafischer: As discussed in meeting, I'll add the recovery step of powering the node back on, so that we don't have to remember to remove any notes in the main doc once this issue is fixed.

clusters/release_notes/mce_known_issues.adoc
```diff
@@ -453,3 +453,9 @@ If the `managed-serviceaccount` add-on is required, you can work around the issu
 . Update the `managed-serviceaccount` `ManagedClusterAddon` in the local cluster namespace to use the `addonDeploymentConfig` custom resource you created.
 
 See link:../../add-ons/configure_nodeselector_tolerations_addons.adoc#configure-nodeselector-tolerations-addons[Configuring nodeSelectors and tolerations for klusterlet add-ons] to learn more about how to use the `addonDeploymentConfig` custom resource to configure `tolerations` and `nodeSelector` for add-ons.
+
+[#nodes-shut-down-bmh]
+=== Nodes shut down after removing `BareMetalHost` resource
+//2.12:ACM-15319
+
+If you remove the `BareMetalHost` resource from a hub cluster, the nodes shut down. You can turn on the nodes again by manually powering them on.
```
swopebe:
```suggestion
If you remove the `BareMetalHost` resource from a hub cluster, the nodes shut down. You can manually power on the nodes again.
```

swopebe (approved):

PR #7310, opened 2024-12-05
https://issues.redhat.com/browse/ACM-16174

jc-berger:


rhaignere (approved):
lgtm, thanks!

clusters/release_notes/mce_errata.adoc
```diff
@@ -11,12 +11,16 @@ If no release notes are listed, the product does not have an Errata release at t
 
 * Delivers updates to one or more product container images.
 
+* Contains no issues for this release.
```
swopebe: Where did we start using this line? I may be missing some discussion or something. :-)

_Contains no issues for this release._

swopebe: We don't normally say that and it's not exactly accurate, because there are issues for the release. We only link to public issues. That is why we only say container updates and we link to any public issues. 

We can remove that line from last release, and if in the ACM release note, we can do the same. We don't want to say there are no issues bc there are always issues. If not linked, it's implied there are no public issues.

swopebe (changes requested):

PR #7321, opened 2024-12-09
https://issues.redhat.com/browse/ACM-16247--removing what's new and adding notes for when we update

swopebe:


jc-berger (approved):
/lgtm

PR #7323, opened 2024-12-09
https://issues.redhat.com/browse/ACM-16247--removing deprecations/removal notices that do not fall in to the N-2 for 2.13

swopebe:


dockerymick (approved):
Thanks Brandi!

PR #7359, opened 2025-01-08
https://issues.redhat.com/browse/ACM-15276 BMH role

oafischer:
Also for 2.12, 2.11 and 2.10.

### DIFF
--- a/clusters/assisted_installer/ai_auto_add_host.adoc
+++ b/clusters/assisted_installer/ai_auto_add_host.adoc
@@ -60,28 +60,32 @@ apiVersion: metal3.io/v1alpha1
 kind: BareMetalHost
 metadata:
   name: <bmh-name>
-  namespace: <your_infraenv_namespace> <1>
+  namespace: <your-infraenv-namespace> <1>
   annotations:
     inspect.metal3.io: disabled
+    bmac.agent-install.openshift.io/hostname: <hostname> <2>
+    bmac.agent-install.openshift.io/role: <role> <3>
   labels:
-    infraenvs.agent-install.openshift.io: <your-infraenv> <2>
+    infraenvs.agent-install.openshift.io: <your-infraenv> <4>
 spec:
   online: true
-  automatedCleaningMode: disabled <3>
-  bootMACAddress: <your-mac-address>  <4>
+  automatedCleaningMode: disabled <5>
+  bootMACAddress: <your-mac-address>  <6>
   bmc:
-    address: <machine-address> <5>
-    credentialsName: <bmc-secret-name> <6>
+    address: <machine-address> <7>
+    credentialsName: <bmc-secret-name> <8>
   rootDeviceHints:
-    deviceName: /dev/sda <7>
+    deviceName: /dev/sda <9>
 ----
 <1> The namespace must be the same as the namespace of your `InfraEnv`.
-<2> The name must match the name of your `InfrEnv` and exist in the same namespace.
-<3> If you do not set a value, the `metadata` value is automatically used.
-<4> Make sure the MAC address matches the MAC address of one of the interaces on your host.
-<5> Use the address of the BMC. See _Port access for the out-of-band management IP address_ for more information.
-<6> Make sure that the `credentialsName` value matches the name of the BMC secret you created.
-<7> *Optional:* Select the installation disk. See _The BareMetalHost spec_ for the available root device hints. After the host is booted with the Discovery Image and the corresponding `Agent` resource is created, the installation disk is set according to this hint.
+<2> Replace with the name of your host.
+<3> Possible values are `master` or `worker`.
+<4> The name must match the name of your `InfrEnv` and exist in the same namespace.
+<5> If you do not set a value, the `metadata` value is automatically used.
+<6> Make sure the MAC address matches the MAC address of one of your host interfaces.
+<7> Use the address of the BMC. See _Port access for the out-of-band management IP address_ for more information.
+<8> Make sure that the `credentialsName` value matches the name of the BMC secret you created.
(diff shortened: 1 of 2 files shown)


clusters/assisted_installer/ai_auto_add_host.adoc
```diff
@@ -60,28 +60,32 @@ apiVersion: metal3.io/v1alpha1
...
 ----
 <1> The namespace must be the same as the namespace of your `InfraEnv`.
-<2> The name must match the name of your `InfrEnv` and exist in the same namespace.
-<3> If you do not set a value, the `metadata` value is automatically used.
-<4> Make sure the MAC address matches the MAC address of one of the interaces on your host.
-<5> Use the address of the BMC. See _Port access for the out-of-band management IP address_ for more information.
-<6> Make sure that the `credentialsName` value matches the name of the BMC secret you created.
-<7> *Optional:* Select the installation disk. See _The BareMetalHost spec_ for the available root device hints. After the host is booted with the Discovery Image and the corresponding `Agent` resource is created, the installation disk is set according to this hint.
+<2> Replace with the name of your host.
+<3> Possible values are `master` or `worker`.
```
carbonin: These are optional

oafischer: Updated

carbonin (approved):

clusters/assisted_installer/ai_auto_add_host.adoc
```diff
@@ -60,28 +60,32 @@ apiVersion: metal3.io/v1alpha1
...
-<4> Make sure the MAC address matches the MAC address of one of the interaces on your host.
-<5> Use the address of the BMC. See _Port access for the out-of-band management IP address_ for more information.
-<6> Make sure that the `credentialsName` value matches the name of the BMC secret you created.
-<7> *Optional:* Select the installation disk. See _The BareMetalHost spec_ for the available root device hints. After the host is booted with the Discovery Image and the corresponding `Agent` resource is created, the installation disk is set according to this hint.
+<2> *Optional:* Replace with the name of your host.
+<3> *Optional:* Possible values are `master` or `worker`.
+<4> The name must match the name of your `InfrEnv` and exist in the same namespace.
+<5> If you do not set a value, the `metadata` value is automatically used.
+<6> Make sure the MAC address matches the MAC address of one of your host interfaces.
+<7> Use the address of the BMC. See _Port access for the out-of-band management IP address_ for more information.
```
swopebe: This looks really good. The only thing I would caution is that the user has to go search for this resource without any clues. 

See _section_ later in this topic? Or something.

swopebe (approved):
See comment for suggestion, but looks really good.

clusters/cluster_lifecycle/scale_node_ctrl_plane.adoc
```diff
@@ -22,6 +22,14 @@ oc patch agent <AGENT-NAME> -p '{"spec":{"role": "master"}}' --type=merge
...
 *Note:* CSRs are not automatically approved.
 
+. If you are using a `BareMetalHost`, add the following line to your `BareMetalHost` annotations:
+
++
+[source,yaml]
+----
+    bmac.agent-install.openshift.io/role: master
+----
+
```
yatanaka-1007, line 32: Thank you for taking care of the scaling parts!
We have to add the annotation when we create a BareMetalHost resource. We cannot add the annotation after creating it, because Agent resource is automatically approved and the node installation will be automatically completed. 
And, setting the agent to `master` is required only when we're using discovery ISO method, not BareMetalHost.
I wrote my suggestion below. What do you think about it?

```
. Complete the steps in xref:scale_node_ocp.adoc#add-nodes-cim-ocp[Adding worker nodes to {ocp-short} clusters] for your the new control plane node.
  If you are using a `BareMetalHost` to add a node, add the following line to your `BareMetalHost` annotations when creating the `BareMetalHost` resource:

+
[source,yaml]
----
    bmac.agent-install.openshift.io/role: master
----

. If you are using discovery ISO to add a node, set the agent to `master` before you approve the agent by running the following command:
+
[source,bash]
----
oc patch agent <AGENT-NAME> -p '{"spec":{"role": "master"}}' --type=merge
----
+
*Note:* CSRs are not automatically approved.
```

oafischer: Thanks for the suggestion @yatanaka-1007, I added it. For the last step, it seems identical to step 2 however. Is there a way we can merge them?

clusters/assisted_installer/ai_auto_add_host.adoc
oafischer: Clarified that this can be found in the additional resources section

clusters/cluster_lifecycle/scale_node_ctrl_plane.adoc
yatanaka-1007: Thank you for updating @oafischer 
The step2 is for Discovery ISO method.
The step3 is for BareMetalNode method.
Both are for the same purpose but not perfectly identical.

What do you think about merging them like this?

```
. If you are using discovery ISO to add a node, set the agent to `master` before you approve the agent by running the following command:
+
[source,bash]
----
oc patch agent <AGENT-NAME> -p '{"spec":{"role": "master"}}' --type=merge
----
+
*Note:* CSRs are not automatically approved.

If you are using a `BareMetalHost` to add a node, add the following line to your `BareMetalHost` annotations when creating the `BareMetalHost` resource:
+
[source,yaml]
----
    bmac.agent-install.openshift.io/role: master
----
```

oafischer: @yatanaka-1007 That works, just updated it. Thanks!

PR #7362, opened 2025-01-09
https://issues.redhat.com/browse/ACM-15868 Add External platform

oafischer:
2.13 only

### DIFF
--- a/clusters/cluster_lifecycle/create_cluster_on_prem.adoc
+++ b/clusters/cluster_lifecycle/create_cluster_on_prem.adoc
@@ -244,7 +244,7 @@ spec:
 ----
 
 <1> Add your platform name.
-<2> Leave empty to disable the Cloud Controller Manager during installation. If set to `External`, you must add custom manifests to enable the cloud controller manager.
+<2> Leave empty to disable the Cloud Controller Manager during installation. If set to `External`, you must add custom manifests to enable the Cloud Controller Manager.
 
 [#on-prem-creating-your-cluster-with-the-cli-nmstateconfig]
 === Optional: Create the NMStateConfig custom resource
--- a/clusters/cluster_lifecycle/create_cluster_on_prem.adoc
+++ b/clusters/cluster_lifecycle/create_cluster_on_prem.adoc
@@ -3,7 +3,7 @@
 
 You can use the console to create on-premises {ocp} clusters. You can create {sno} clusters, multi-node clusters, and compact three-node clusters. You can create these clusters on VMware vSphere, Red Hat OpenStack, Nutanix, External, or in a bare metal environment.
 
-There is no platform integration with the platform where you install the cluster, as the platform value is set to `platform=none`. A {sno} cluster has only a single node, which hosts the control plane services and the user workloads. This configuration can be helpful when you want to minimize the resource footprint of the cluster. 
+Since the platform value is set to `platform=none`, there is no platform integration with the platform where you install the cluster. A {sno} cluster has only a single node, which hosts the control plane services and the user workloads. This configuration can be helpful when you want to minimize the resource footprint of the cluster. 
 
 You can also provision multiple {sno} clusters on edge resources by using the zero touch provisioning feature, which is a feature that is available with {ocp}. For more information about zero touch provisioning, see _Clusters at the network far edge_ in the {ocp-short} documentation.
 
--- a/clusters/cluster_lifecycle/create_cluster_on_prem.adoc
+++ b/clusters/cluster_lifecycle/create_cluster_on_prem.adoc
@@ -189,7 +189,7 @@ spec:
 <1> Use the name of your `AgentClusterInstall` resource.
 <2> Use the pull secret that you downloaded in <<on-prem-creating-your-cluster-with-the-cli-pull-secret,Add the pull secret to the namespace>>. 
 
-[#on-prem-creating-your-cluster-with-the-cli-agentclusterinstall]
+[#create-agentclusterinstall-cr]
 === Create the AgentClusterInstall custom resource
 
 In the `AgentClusterInstall` custom resource, you can specify many of the requirements for the clusters. For example, you can specify the cluster network settings, platform, number of control planes, and worker nodes. 
--- a/clusters/cluster_lifecycle/create_cluster_on_prem.adoc
+++ b/clusters/cluster_lifecycle/create_cluster_on_prem.adoc
@@ -1,7 +1,7 @@
 [#creating-a-cluster-on-premises]
 = Creating a cluster in an on-premises environment
 
-You can use the console to create on-premises {ocp} clusters. The clusters can be {sno} clusters, multi-node clusters, and compact three-node clusters on VMware vSphere, Red Hat OpenStack, Nutanix, External, or in a bare metal environment.
+You can use the console to create on-premises {ocp} clusters. You can create {sno} clusters, multi-node clusters, and compact three-node clusters. You can create these clusters on VMware vSphere, Red Hat OpenStack, Nutanix, External, or in a bare metal environment.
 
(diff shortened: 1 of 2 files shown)


clusters/cluster_lifecycle/create_cluster_on_prem.adoc
```diff
@@ -225,7 +225,27 @@ spec:
...
+  namespace: demo-worker4
+spec:
+  platformType: External
+  external:
+    platformName: "myplatform" <1>
+    cloudControllerManager: External <2>
+----
+
+<1> Add your platform name.
+<2> Leave empty to disable the Cloud Controller Manager during installation.
```
adriengentil:
```suggestion
<2> Leave empty to disable the Cloud Controller Manager during installation.
```

what do you think about adding that when `cloudControllerManager` is set to `External`, we expect the user to add custom manifests that will enable the cloud controller manager.

oafischer: I added your suggestion, let me know if it looks good now. Thanks!

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
```diff
@@ -225,7 +225,26 @@ spec:
...
+  namespace: demo-worker4
+spec:
+  platformType: External
+  external:
+    platformName: "myplatform" <1>
+    cloudControllerManager: External <2>
+----
+
+<1> Add your platform name.
+<2> Leave empty to disable the Cloud Controller Manager during installation. If set to `External`, you must add custom manifests to enable the cloud controller manager.
```
adriengentil:
```suggestion
<2> Leave empty to disable the Cloud Controller Manager during installation. If set to `External`, you must add custom manifests to enable the Cloud Controller Manager.
```

nit

adriengentil (commented):
Looks good to me. @makentenza do you think it's enough, or should we add more details about what is the external platform?

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
```diff
@@ -1,9 +1,9 @@
 [#creating-a-cluster-on-premises]
 = Creating a cluster in an on-premises environment
 
-You can use the console to create on-premises {ocp} clusters. The clusters can be {sno} clusters, multi-node clusters, and compact three-node clusters on VMware vSphere, Red Hat OpenStack, Nutanix, or in a bare metal environment.
+You can use the console to create on-premises {ocp} clusters. The clusters can be {sno} clusters, multi-node clusters, and compact three-node clusters on VMware vSphere, Red Hat OpenStack, Nutanix, External, or in a bare metal environment.
```
jc-berger: What do you think about making it a bit more user focused and action oriented? Something like: 

```suggestion
You can use the console to create on-premises {ocp} clusters. You can create {sno} clusters, multi-node clusters, and compact three-node clusters. You can create these clusters on VMware vSphere, Red Hat OpenStack, Nutanix, External, or in a bare metal environment.
```

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
```diff
@@ -1,9 +1,9 @@
...
 
-There is no platform integration with the platform where you install the cluster, as the platform value is set to `platform=none`. A {sno} cluster contains only a single node, which hosts the control plane services and the user workloads. This configuration can be helpful when you want to minimize the resource footprint of the cluster. 
+There is no platform integration with the platform where you install the cluster, as the platform value is set to `platform=none`. A {sno} cluster has only a single node, which hosts the control plane services and the user workloads. This configuration can be helpful when you want to minimize the resource footprint of the cluster.
```
jc-berger: I think it might read a bit easier if we flip the order of this first sentence, what do you think? Something like: 

```suggestion
Since the platform value is set to `platform=none`, there is no platform integration with the platform where you install the cluster. A {sno} cluster has only a single node, which hosts the control plane services and the user workloads. This configuration can be helpful when you want to minimize the resource footprint of the cluster. 
```

jc-berger: What do you think about make the last sentence more authoritative by making it a statement of what users do rather than simply saying it's helpful? 

```suggestion
There is no platform integration with the platform where you install the cluster, as the platform value is set to `platform=none`. A {sno} cluster has only a single node, which hosts the control plane services and the user workloads. Use this configuration when you want to minimize the resource footprint of the cluster. 
```

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
jc-berger: Not sure if I'm understanding the second bullet here. Do we mean something like the following?: 

 ```suggestion
<2> Leave `empty` as it is so that is disables the Cloud Controller Manager during installation. If set to `External`, you must add custom manifests to enable the Cloud Controller Manager.
```

**Review by jc-berger — CHANGES_REQUESTED** [review]

File: clusters/cluster_lifecycle/create_cluster_on_prem.adoc (hunk 2 above)
Comment (oafischer) [reply clusters/cluster_lifecycle/create_cluster_on_prem.adoc]: For this bullet we can't use `empty` since it's not a valid option. We can just say leave it empty or maybe say unpopulated, though that word might be less accessible.

**Review by jc-berger — APPROVED** [review]
/lgtm

File: clusters/cluster_lifecycle/create_cluster_on_prem.adoc [hunk 5]
```diff
@@ -225,7 +225,26 @@ spec:
...
+  namespace: demo-worker4
+spec:
+  platformType: External
+  external:
+    platformName: "myplatform" <1>
+    cloudControllerManager: External <2>
+----
+
+<1> Add your platform name.
+<2> Leave empty to disable the Cloud Controller Manager during installation. If set to `External`, you must add custom manifests to enable the Cloud Controller Manager.
```
jc-berger:
```suggestion
<2> The field is automatically empty, `{ }`, so it disables the Cloud Controller Manager during installation. If you set it to `External`, you must add custom manifests to enable the Cloud Controller Manager.
```

jc-berger:
```suggestion
<2> If you set the field to empty, `{}`, it disables the Cloud Controller Manager during installation. If you set it to `External`, you must add custom manifests to enable the Cloud Controller Manager.
```

oafischer:
```suggestion
<2> Set the field to empty (`{ }`) to disable the Cloud Controller Manager during installation. If you set it to `External`, you must add custom manifests to enable the Cloud Controller Manager.
```

makentenza (commented):
Few comments from my side:

- Why are we keeping references for both `platform: none` and `platform: external`? Should we deprecate `platform: none` and just use `platform: external` for both integrated and non-integrated platforms?
- Is this going to support Oracle Cloud Infrastructure out of the box or do we need to any changes in the ACM product for that? If adding `platform: external` can support OCI clusters out of the box let's add a new section "Creating a cluster on Oracle Cloud Infrastructure" outside the "on-prem" section

adriengentil:
> Is this going to support Oracle Cloud Infrastructure out of the box or do we need to any changes in the ACM product for that? If adding `platform: external` can support OCI clusters out of the box let's add a new section "Creating a cluster on Oracle Cloud Infrastructure" outside the "on-prem" section

yes, this change enables Oracle Cloud integration when `platformName` is set to `oci`. So, let's document it as well (maybe as part of another PR?).

oafischer:
I recommend creating a new issue to document the Oracle Cloud integration.

Are we good with this PR for now?

oafischer:
Given approval so far and the fact that the PM comments address things that would be done in a separate PR anyway, I'll go ahead and merge this for now. For any additional work, feel free to open another issue @adriengentil @makentenza. Thanks!

PR #7374, opened 2025-01-13
https://issues.redhat.com/browse/ACM-16773--deprecation of API docume…

swopebe:
…ntation only (ACM APIs content added to announce


oafischer (approved):
Looks good, thanks!

PR #7377, opened 2025-01-15
2.12 refresh Jan 22

oafischer:


jc-berger (approved):
/lgtm

clusters/assisted_installer/ai_enable_cim.adoc
```diff
@@ -293,11 +293,14 @@ Your {cim} service is configured. You can verify that it is healthy by checking
 [#fips-install-cim]
 == Installing a FIPS-enabled cluster by using the {ai}
 
-When you install a {ocp-short} cluster version 4.15 and earlier that is in FIPS mode, you must specify that the installers run Red Hat Enterprise Linux (RHEL) version 8 in the `AgentServiceConfig` resource.
+When you install a {ocp-short} cluster that is version 4.15 and earlier and is in FIPS mode, you must specify that the installers run Red Hat Enterprise Linux (RHEL) version 8 in the `AgentServiceConfig` resource. When you install a {ocp-short} clusterthat is version 4.16 and later and is in FIPS mode, do not specify any RHEL version for the installers.
```
swopebe: Hmmm, I don't think I saw this update @oafischer but we should consider this for versioning; this is something we have to change each time. (For 2.13, for instance, we don't need anything at all about 4.15. For 2.12, we only presented 4.15 in the docs. 

With stuff like this, I think we should chat about how to handle it.

That said, the file format seems to be off here. Can we look at this before we refresh 2.9?

 I would have requested some changes on this, but for now, can we please look at this format before we refresh 2.9? Let's chat in the morning before we merge?

oafischer: We've looked at this section before, the version here doesn't change. What might change though is that we remove the entire section about 4.15 and earlier, if support for 4.15 is dropped in a future ACM version.

swopebe (approved):

PR #7439, opened 2025-01-30
https://issues.redhat.com/browse/ACM-16788 ISO type

oafischer:
2.13 only


clusters/assisted_installer/ai_create_cli.adoc
```diff
@@ -107,6 +107,7 @@ spec:
       value: audit=0
   additionalTrustBundle: <bundle>
   osImageVersion: <version>
+  ImageType: <minimal_or_full>
```
carbonin: The values are `minimal-iso` or `full-iso`, not sure if you want to reflect that here.

Maybe it would by more clear for this to just be `<type>`?

oafischer: Good point, I'll go with `<type>` since ideally we want the user to read the table where there is more info on it.

carbonin (approved):

clusters/assisted_installer/ai_add_host.adoc
```diff
@@ -28,11 +28,10 @@ Download the ISO file by completing the following steps:
 
 . Click *Add hosts* and select *With Discovery ISO*.
 
-+
-You now see a URL to download the ISO file. Booted hosts appear in the host inventory table. Hosts might take a few minutes to appear. 
+. Select *Minimal image file* or *Full image file* and generate the Discovery ISO.
```
jc-berger: Very minor and by no means necessary, but I think you mean _to_ generate? That is, one action (selecting files generates the ISO) rather than two action (selecting a file then generating the ISO)

```suggestion
. Select *Minimal image file* or *Full image file* to generate the Discovery ISO.
```

clusters/assisted_installer/ai_add_host.adoc
```diff
@@ -60,7 +59,7 @@ https://assisted-image-service-assisted-installer.apps.example-acm-hub.com/byapi
 ----
 
 +
-*Note:* By default, the ISO that is provided is a _minimal_ ISO. The minimal ISO does not contain the root file system, `RootFS`. The `RootFS` is downloaded later. To display full ISO, replace `minimal.iso` in the URL with `full.iso`.
+*Note:* By default, the minimal ISO is provided. You can specify the full ISO by setting `ImageType` to `full-iso` in your `InfraEnv` resource. See _Creating a host inventory by using the command line interface_ in the additional resources section to learn more.
```
jc-berger, line 64: Optional:

I defer to your judgement as the writer here, but since there's already an "Additional resources" section, I would put this reference to creating a host inventory under additional resources. 

In my opinion, referencing other docs in the middle of your doc just makes it likely the user will leave and not return to your awesome work here. That's fine and even helpful if it's important for them to complete another task in a separate doc, but that doesn't seem to be super high importance here. 

```suggestion
*Note:* By default, the minimal ISO is provided. You can specify the full ISO by setting `ImageType` to `full-iso` in your `InfraEnv` resource. 
```

clusters/assisted_installer/ai_add_host.adoc
```diff
@@ -112,4 +111,4 @@ NAME                                   CLUSTER   APPROVED   ROLE          STAGE
 [#additional-resources-add-host]
 == Additional resources
 
-- xref:../assisted_installer/ai_create_cli.adoc#create-host-inventory-cli[Creating a host inventory by using the command line interface]
+- To learn more about the `InfraEnv` resource, see xref:../assisted_installer/ai_create_cli.adoc#create-host-inventory-cli[Creating a host inventory by using the command line interface].
```
jc-berger:
```suggestion
- To learn more about the `InfraEnv` resource, see xref:../assisted_installer/ai_create_cli.adoc#create-host-inventory-cli[Creating a host inventory by using the command line interface].
- To learn more about specifying ISO sizes, see _Creating a host inventory by using the command line interface_ in the additional resources section.
```

jc-berger (approved):
/lgtm

clusters/assisted_installer/ai_add_host.adoc
oafischer: In this case, selecting either option doesn't generate the ISO, you have to press another button for that. It's very clear in the UI, but I can add another step to clarify that. I was trying to keep UI documentation to a minimum but maybe it would be better to spell it out here.

clusters/assisted_installer/ai_add_host.adoc
oafischer: I see what you mean, the whole reason we have removed inline links in certain spots was to avoid this situation. In this case you are right, it's not something the user has to visit, which is why the actual link is in additional resources, not here in the note. We do need the pointer here though, so that the user knows that there is additional info they can look up.

jc-berger (approved):
/lgtm

PR #7458, opened 2025-02-04
https://issues.redhat.com/browse/ACM-16886

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-16886

### DIFF
--- a/applications/subscription_report.adoc
+++ b/applications/subscription_report.adoc
@@ -88,7 +88,7 @@ The cluster-level status is located in `<namespace:<your-managed-cluster-1>` on 
   - `Failed`
   - `propagationFailed`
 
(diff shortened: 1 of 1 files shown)


tanfengshuang (commented):
Thanks for your fix, I have a question here: This bug also exists in version 2.5 ~ 2.12, will this fix apply to other versions(2.5~2.11)?

swopebe:
If the bug came in on 2.12, that should be the first place the update is made (maybe this was discussed already), then we normally ask if it is forwarded to 2.13--almost always it is.

We only update for things like this N-2. 

The product lifecycle applies here, in that 2.8 and earlier are officially EOL, so we would not update these versions.

swopebe (approved):

PR #7470, opened 2025-02-06
https://issues.redhat.com/browse/ACM-14532-draft of customer request …

swopebe:
…for ai that was not tracked on Kanban (Also for later versions) Will review as first draft and then send to developers for tech review. This is nearly done.

### DIFF
--- a/clusters/cluster_lifecycle/cim_auto_add_host.adoc
+++ b/clusters/cluster_lifecycle/cim_auto_add_host.adoc
@@ -1,7 +1,7 @@
 [#auto-add-host-host-inventory]
 = Automatically adding bare metal hosts to the host inventory
 
-After creating your host inventory (infrastructure environment) you can discover your hosts and add them to your inventory. You can automate booting the Discovery Image of your infrastructure environment by making the bare metal operator communicate with the Baseboard Management Controller (BMC) of each bare metal host by creating a `BareMetalHost` resource and associated BMC secret for each host. The automation is set by a label on the `BareMetalHost` that references your infrastructure environment.
+After creating your host inventory (infrastructure environment), you can discover your hosts and add them to your inventory. You can automate booting the Discovery Image of your infrastructure environment by making the bare metal operator communicate with the Baseboard Management Controller (BMC) of each bare metal host. Create a `BareMetalHost` resource and associated BMC secret for each host. The automation is set by a label on the `BareMetalHost` that references your infrastructure environment.
 
 The automation performs the following actions:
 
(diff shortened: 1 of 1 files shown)


clusters/cluster_lifecycle/cim_auto_add_host.adoc
```diff
@@ -1,7 +1,7 @@
 [#auto-add-host-host-inventory]
 = Automatically adding bare metal hosts to the host inventory
 
-After creating your host inventory (infrastructure environment) you can discover your hosts and add them to your inventory. You can automate booting the Discovery Image of your infrastructure environment by making the bare metal operator communicate with the Baseboard Management Controller (BMC) of each bare metal host by creating a `BareMetalHost` resource and associated BMC secret for each host. The automation is set by a label on the `BareMetalHost` that references your infrastructure environment.
+After creating your host inventory (infrastructure environment), you can discover your hosts and add them to your inventory. You can automate booting the Discovery Image of your infrastructure environment by making the bare metal operator communicate with the Baseboard Management Controller (BMC) of each bare metal host. Create a `BareMetalHost` resource and associated BMC secret for each host. The automation is set by a label on the `BareMetalHost` that references your infrastructure environment.
```
swopebe: After creating your host inventory (infrastructure environment), you can discover your hosts and add them to your inventory.

I always open a file to try to understand what I am updating and on this line, I wonder if we are making sense.

If you created a host inventory, didn't you already discover hosts and add them to the inventory (a collection of hosts).

I could be wrong but are we saying that you created an environment to start discovering and adding hosts? @carbonin

swopebe: Changed a really long run-on sentence.

carbonin: I think the confusion here is that "host inventory" is being used as a synonym for "infrastructure environment".

I don't know why, but I think that's what is happening.

The infrastructure environment is the actual CR that defines the image (discovery image) that needs to be booted on a host in order for it to be added.

While we're here, can we change "The automation is set" to "The automation is triggered"?

clusters/cluster_lifecycle/cim_auto_add_host.adoc
```diff
@@ -145,7 +145,11 @@ oc delete bmh <bmh-name>
 [#auto-remove-host-steps-no-bmh]
 === Removing managed cluster nodes without a bare metal host
 
-If you do not have a bare metal host on your hub cluster and want remove managed cluster nodes from a managed cluster, follow the link:https://docs.openshift.com/container-platform/4.14/nodes/nodes/nodes-nodes-working.html#deleting-nodes[Deleting nodes] instructions in the {ocp-short} documentation.
+If you do not have a bare metal host on your hub cluster and you want to remove managed cluster nodes from a managed cluster, you can unbind the `clusterDeploymentName` field from the `Agent` specification, or delete the `Agent` custom resource that corresponds with the node that you are removing.
```
carbonin:
```suggestion
If you do not have a bare metal host on your hub cluster and you want to remove managed cluster nodes from a managed cluster, you can unbind the agent by removing the `clusterDeploymentName` field from the `Agent` specification, or delete the `Agent` custom resource that corresponds with the node that you are removing.
```

carbonin (commented):
Generally some related info was added in https://github.com/stolostron/rhacm-docs/pull/7471

It's possible we're in a different doc here or a different branch (this seems to be 2.11 and the other is 2.12), but in either case maybe it's worth consolidating or a link between these two sections.

clusters/cluster_lifecycle/cim_auto_add_host.adoc
swopebe: The _why_ was lack of technical understanding when the content was built, but we have over time corrected that with Oliver.

**Host inventory** is the collection of host, right? And if so, they are placed in the _infrastructure environment_?

I think it's clear that we aren't sure here, so at least let's get this description changed. 

I won't say "triggered" in the doc and I can say _initiated_ or _started_.

swopebe:
@carbonin the 2.12+ doc is better because the incoming writer spent time understanding the component and that file did change. For 2.11, this is what we are adding and all we need to review. For 2.12, the change will look different (although the terms I address in the short description will change based on your feedback).

So for 2.11, is everything ok now that I have applied your feedback?

carbonin (approved):

jc-berger (approved):
/lgtm

PR #7481, opened 2025-02-07
https://issues.redhat.com/browse/ACM-16770

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-16770

### DIFF
--- a/business_continuity/backup_restore/backup_restore.adoc
+++ b/business_continuity/backup_restore/backup_restore.adoc
@@ -338,9 +338,11 @@ spec:
 [#advanced-restore-options]
 == Using advanced restore options
 
-View the following YAML examples to use the advanced options with the `velero.io.restore` `spec` option to filter out resources you want to restore.
+To filter out resources you want to restore, use the advanced restore option, `velero.io.restore` specification.  
 
-* Use the following sample with the `velero.io.restore` `spec` options to restore resources only from the `vb-managed-cls-2` managed cluster namespace and to exclude the global `MultiCluster` resources:
+To restore resources only from the `vb-managed-cls-2` managed cluster namespace and to exclude the global `MultiCluster` resources, complete the following steps:
+
+. Create a file with the following YAML sample: 
 
 +
 [source,yaml]
@@ -362,7 +364,7 @@ spec:
 ----
 +
 
-* Set the following `velero.io.restore spec` options when you create the {acm} restore resource:
+. Set the following `velero.io.restore` specification options when you create the {acm} restore resource:
 
 +
 [source,yaml]
@@ -377,7 +379,7 @@ spec:
 ----
 +
 
-* When you set any of the Velero restore filters, set the  `cleanupBeforeRestore` field  to `None` to prevent having to clean up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter.
+. Set the  `cleanupBeforeRestore` field  to `None` to prevent having to clean up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter. 
 
 [#viewing-restore-events]
 == Viewing restore events
--- a/business_continuity/backup_restore/backup_restore.adoc
+++ b/business_continuity/backup_restore/backup_restore.adoc
@@ -338,9 +338,9 @@ spec:
 [#advanced-restore-options]
 == Using advanced restore options
 
-View the following YAML examples to use the advanced options with the `velero.io.restore` spec option to filter out resources you want to restore.
+View the following YAML examples to use the advanced options with the `velero.io.restore` `spec` option to filter out resources you want to restore.
 
-* Use the following sample with the `velero.io.restore` spec options to restore resources only from the `vb-managed-cls-2` managed cluster namespace and to exclude the global `MultiCluster` resources:
+* Use the following sample with the `velero.io.restore` `spec` options to restore resources only from the `vb-managed-cls-2` managed cluster namespace and to exclude the global `MultiCluster` resources:
 
 +
 [source,yaml]
@@ -360,8 +360,9 @@ spec:
   includedNamespaces:
     - vb-managed-cls-2  
 ----
++
 
-* Set the following `velero.io.restore spec` options when you create the {acm}restore resource:
+* Set the following `velero.io.restore spec` options when you create the {acm} restore resource:
 
 +
 [source,yaml]
@@ -376,9 +377,7 @@ spec:
 ----
 +
 
-*Note:*
-
-** When you set any of these Velero restore filters, set the field, `cleanupBeforeRestore`, to `None`. Doing this, you prevent having to clean up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter.
+* When you set any of the Velero restore filters, set the  `cleanupBeforeRestore` field  to `None` to prevent having to clean up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter.
 
 [#viewing-restore-events]
 == Viewing restore events
--- a/business_continuity/backup_restore/backup_restore.adoc
+++ b/business_continuity/backup_restore/backup_restore.adoc
@@ -338,13 +338,11 @@ spec:
 [#advanced-restore-options]
 == Using advanced restore options
 
-To filter out resources you want to restore, use the advanced restore option, `velero.io.restore` specification.  
+There are advanced restore options defined by the `velero.io.restore` specification. You can use these advanced restore options to filter out resources you want to restore. 
 
-To restore resources only from the `vb-managed-cls-2` managed cluster namespace and to exclude the global `MultiCluster` resources, complete the following steps:
+For example, you can use the following YAML sample to restore resources only from the 
+`vb-managed-cls-2` managed cluster namespace and to exclude the global `MultiCluster` resources: 
 
-. Create a file with the following YAML sample: 
-
-+
 [source,yaml]
 ----
 apiVersion: cluster.open-cluster-management.io/v1beta1
@@ -362,11 +360,9 @@ spec:
   includedNamespaces:
     - vb-managed-cls-2  
 ----
-+
 
-. Set the following `velero.io.restore` specification options when you create the {acm} restore resource:
+If you want more options for the  `velero.io.restore` specification when you create the {acm} restore resource, see the following restore filers: 
 
-+
 [source,yaml]
 ----
 spec: 
@@ -377,9 +373,8 @@ spec:
   LabelSelector
   OrLabelSelector
 ----
-+
 
-. Set the  `cleanupBeforeRestore` field  to `None` to prevent having to clean up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter. 
+*Note:* For any of the Velero restore filters, set the  `cleanupBeforeRestore` field  to `None` to prevent having to clean up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter. 
 
 [#viewing-restore-events]
 == Viewing restore events
--- a/business_continuity/backup_restore/backup_restore.adoc
+++ b/business_continuity/backup_restore/backup_restore.adoc
@@ -361,7 +361,7 @@ spec:
     - vb-managed-cls-2  
 ----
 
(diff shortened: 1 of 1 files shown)


business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,49 @@ spec:
...
+metadata:
+  name: restore-filter-sample 
+  namespace: open-cluster-management-backup
+spec:
+  cleanupBeforeRestore: None 
+  veleroCredentialsBackupName: latest 
+  veleroResourcesBackupName: latest 
+  veleroManagedClustersBackupName: latest
+  excludedResources:
+    - ManagedCluster
```
birsanv, line 359: missing
  includedNamespaces:
    - vb-managed-cls-2
so, add this after line 359:

  includedNamespaces:
    - vb-managed-cls-2

birsanv (changes requested):

birsanv (approved):
lgtm

business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,51 @@ spec:
 
 ** When you set any of these Velero restore filters, set the `cleanupBeforeRestore` to `None` to avoid cleaning up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter. 
 
+[#advanced-restore-options]
+== Using advanced restore options
+
+View the following YAML examples to use the advanced options with the `velero.io.restore` spec option to filter out resources you want to restore.
```
swopebe: please change "spec" here and throughout.

business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,51 @@ spec:
...
+  veleroCredentialsBackupName: latest 
+  veleroResourcesBackupName: latest 
+  veleroManagedClustersBackupName: latest
+  excludedResources:
+    - ManagedCluster
+  includedNamespaces:
+    - vb-managed-cls-2  
+----
+
+* Set the following `velero.io.restore spec` options when you create the {acm}restore resource:
```
swopebe: Missing a space here.

business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,51 @@ spec:
...
+  excludedResources
+  excludedNamespaces
+  LabelSelector
+  OrLabelSelector
+----
++
+
+*Note:*
+
+** When you set any of these Velero restore filters, set the field, `cleanupBeforeRestore`, to `None`. Doing this, you prevent having to clean up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter.
```
swopebe:
```suggestion
** When you set any of these Velero restore filters, set the `cleanupBeforeRestore` field to `None` to prevent having to clean up hub cluster resources that are part of the restored backup, but are not being restored because of the applied filter.
```

swopebe: @jc-berger if this is a preventative step, which is what it seems like, should it not be integrated in to the doc instead of just a note? It was hard to see what you meant by "these," (self referential).

Suggested edits made above are to reduce some confusion I had when I read it due to too many commas. This example doesn't need the field name offset by commas if you just call it what it is; we can reduce punctuation and words like "By doing this..." I removed, but feel free to play with it.
 
We can merge the step and the note, can't we? The step says **set the fields,** but there is nothing to set in the list of specifications or the step. Set to what?

The details for the step are in the note. 
 
 I find that structure a bit confusing. I think they can be combined.

swopebe (changes requested):

business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,50 @@ spec:
...
+spec: 
+  includedResources
+  includedNamespaces
+  excludedResources
+  excludedNamespaces
+  LabelSelector
+  OrLabelSelector
+----
++
+
```
swopebe, line 376: I still think the note is part of what should be a step here (it's telling the user to set something, but it's a bullet?), I think that was the original feed back. 

The step is telling the user to set something. Then the user has to go to the note to see what to set. This can be merged in to a concise step. This may be legacy doc, but I am really confused by it's a bullet.

business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,50 @@ spec:
 
 ** When you set any of these Velero restore filters, set the `cleanupBeforeRestore` to `None` to avoid cleaning up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter.
```
swopebe, line 336: line 336 and line 380 are the same?

business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,50 @@ spec:
 
 ** When you set any of these Velero restore filters, set the `cleanupBeforeRestore` to `None` to avoid cleaning up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter. 
 
+[#advanced-restore-options]
+== Using advanced restore options
+
+View the following YAML examples to use the advanced options with the `velero.io.restore` `spec` option to filter out resources you want to restore.
+
+* Use the following sample with the `velero.io.restore` `spec` options to restore resources only from the `vb-managed-cls-2` managed cluster namespace and to exclude the global `MultiCluster` resources:
```
swopebe: The file is a procedure, indicated by the gerund "Using". Typically that looks like:

1. Create a file with the following YAML sample to restore resources....(with more details about the sample in the step) --just an example

2. Set the __ to ___. (That is in a note)

3. Apply the file with command. (if they need to)

I didn't mean code **spec**.; I meant use "specification." (There is a different way to write this if you want to use that. I believe its something like `spec.cleanupBeforeRestore: None`, but you will need to check in to that--we can look around. 

Best to just use the word **specification** and not code it. 

The way this is written, I am not sure what we are telling the user to do with the YAML. 
Is this a doc guiding the user through steps, or just letting them see some samples? It starts like a task, and probably should be, but needs a little tweaking.

Happy to help in a quick live meeting.

swopebe (changes requested):

business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,52 @@ spec:
 
 ** When you set any of these Velero restore filters, set the `cleanupBeforeRestore` to `None` to avoid cleaning up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter.
```
jc-berger, line 336: @birsanv Brandi brought up a good point. Looking at lines 294-336, it's practically identical to the content we have under the section "Using advanced restore options" I see that the YAML sample is slightly different, but otherwise, it's very similar content. 

Can we remove lines 294-336? Or, modify it more so it's more distinct from "Using advanced restore options"

business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,52 @@ spec:
...
 ** When you set any of these Velero restore filters, set the `cleanupBeforeRestore` to `None` to avoid cleaning up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter. 
 
+[#advanced-restore-options]
+== Using advanced restore options
+
+To filter out resources you want to restore, use the advanced restore option, `velero.io.restore` specification.  
+
+To restore resources only from the `vb-managed-cls-2` managed cluster namespace and to exclude the global `MultiCluster` resources, complete the following steps:
+
+. Create a file with the following YAML sample:
```
jc-berger: @birsanv we changed this content quite a bit here, turning it into a procedure. What do you think?

birsanv (approved):
lgtm

business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,47 @@ spec:
...
+  veleroCredentialsBackupName: latest 
+  veleroResourcesBackupName: latest 
+  veleroManagedClustersBackupName: latest
+  excludedResources:
+    - ManagedCluster
+  includedNamespaces:
+    - vb-managed-cls-2  
+----
+
+If you want more options for the  `velero.io.restore` specification when you create the {acm} restore resource, see the following restore filers:
```
swopebe:
```suggestion
If you want more options for the  `velero.io.restore` specification when you create the {acm} restore resource, see the following restore filters: 
```

business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,47 @@ spec:
 
 ** When you set any of these Velero restore filters, set the `cleanupBeforeRestore` to `None` to avoid cleaning up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter. 
 
+[#advanced-restore-options]
+== Using advanced restore options
+
+There are advanced restore options defined by the `velero.io.restore` specification. You can use these advanced restore options to filter out resources you want to restore.
```
swopebe:
```suggestion
There are advanced restore options that are defined by the `velero.io.restore` specification. You can use these advanced restore options to filter out resources that you want to restore. 
```

business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,47 @@ spec:
 
 ** When you set any of these Velero restore filters, set the `cleanupBeforeRestore` to `None` to avoid cleaning up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter. 
 
+[#advanced-restore-options]
+== Using advanced restore options
+
+There are advanced restore options defined by the `velero.io.restore` specification. You can use these advanced restore options to filter out resources you want to restore. 
+
+For example, you can use the following YAML sample to restore resources only from the 
+`vb-managed-cls-2` managed cluster namespace and to exclude the global `MultiCluster` resources:
```
swopebe: I don't see where it "exclude the global **`MultiCluster`** resources:" in the sample. Is this line 358?

business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,47 @@ spec:
 
 ** When you set any of these Velero restore filters, set the `cleanupBeforeRestore` to `None` to avoid cleaning up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter. 
 
+[#advanced-restore-options]
+== Using advanced restore options
+
+There are advanced restore options defined by the `velero.io.restore` specification. You can use these advanced restore options to filter out resources you want to restore. 
+
+For example, you can use the following YAML sample to restore resources only from the
```
swopebe, line 343:
```suggestion
For example, you can use the following `Restore` YAML sample to restore resources only from the 
```

swopebe: here I am referring to the `kind` of resource in the intro-- for retrievability and for accessibility. It is a `Restore` resource.

business_continuity/backup_restore/backup_restore.adoc
```diff
@@ -334,6 +335,47 @@ spec:
...
+spec: 
+  includedResources
+  includedNamespaces
+  excludedResources
+  excludedNamespaces
+  LabelSelector
+  OrLabelSelector
+----
+
+*Note:* For any of the Velero restore filters, set the  `cleanupBeforeRestore` field  to `None` to prevent having to clean up hub cluster resources that are part of the restored backup but are not being restored because of the applied filter.
```
swopebe:
```suggestion
*Note:* For any of the Velero restore filters, set the  `cleanupBeforeRestore` field  to `None` to prevent having to clean up hub cluster resources that are part of the restored backup, but are not being restored because of the applied filter. 
```

swopebe (approved):
Some suggestions for style standards and spelling, etc... One question in that I think what is in the YAML intro and what is in the actual YAML present possible inaccuracy. Please confirm that and then feel free to merge.

business_continuity/backup_restore/backup_restore.adoc
jc-berger: Yes, that is my understanding, but I actually see that we say "ManagedCluster" in line 359

@birsanv should we change it to `MultiCluster` in the yaml? Or simply say that this yaml sample excludes a `ManagedCluster`?

PR #7482, opened 2025-02-07
https://issues.redhat.com/browse/ACM-16770

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-16770


birsanv (approved):

swopebe (changes requested):
same feedback as 7481, I didn't realize they were the same.

swopebe:
This is on my list to review but I think that is because it needs the same changes from another version, so my review will be the same as that PR.

PR #7499, opened 2025-02-17
Update global hub postgresql for the 2.13

yanmxa:
https://issues.redhat.com/browse/ACM-16279

### DIFF
--- a/troubleshooting/trouble_global_hub_access_postgres.adoc
+++ b/troubleshooting/trouble_global_hub_access_postgres.adoc
@@ -46,7 +46,7 @@ EOF
 Run the following command to get your the credentials:
 +
 ----
-# Get the Host
+# Host
 oc get svc multicluster-global-hub-postgresql-lb -n multicluster-global-hub -ojsonpath='{.status.loadBalancer.ingress[0].hostname}'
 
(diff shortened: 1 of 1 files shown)


troubleshooting/trouble_global_hub_access_postgres.adoc
```diff
@@ -38,19 +38,22 @@ spec:
...
   type: LoadBalancer
 EOF
 ----
 +
 Run the following command to get your the credentials:
 +
 ----
-# Host
-oc get svc postgres-ha -ojsonpath='{.status.loadBalancer.ingress[0].hostname}'
+# Get the Host
```
jc-berger: What do you think about keeping it in the original title `# Host` 

I notice that's how we document it on line 65. 

For all the other titles, we just have nouns, like `Database` and `Passwords` so I don't think we'd want to change the naming convention here by telling users `Get the Host` 

What do you think? 

```suggestion
# Host
```

jc-berger (commented):
see my one comment and let me know what you think, thanks :)

ldpliu:
> see my one comment and let me know what you think, thanks :)

It's ok for me.

PR #7501, opened 2025-02-17
https://issues.redhat.com/browse/ACM-16419 Replace custom metrics doc

oafischer:
Also for 2.13, 2.12, and 2.11.

Needs some guidance on naming (capitalize Platform?). Please check YAML content, formatting may have broken in the process of copying.

### DIFF
--- a/observability/customize_observability.adoc
+++ b/observability/customize_observability.adoc
@@ -152,12 +152,12 @@ You can monitor User workload metrics by setting the configuration on the manage
 kind: ConfigMap
 apiVersion: v1
 metadata:
-name: observability-metrics-custom-allowlist
-namespace: <monitored_namespace> <1>
-data:
-uwl_metrics_list.yaml:
-names:
-  <sample_metrics> <2>
+  name: observability-metrics-custom-allowlist
+  namespace: <monitored_namespace> <1>
+  data:
+  uwl_metrics_list.yaml:
+  names:
+    <sample_metrics> <2>
 ----
 
 <1> Add the namespace where you want to capture the metric.
--- a/observability/customize_observability.adoc
+++ b/observability/customize_observability.adoc
@@ -77,87 +77,93 @@ data:
 [#adding-custom-metrics]
 == Adding custom metrics
 
-To start collecting metrics from all of your managed clusters, add metrics to the `metrics_list.yaml` file. Complete the following steps:
+To monitor metrics from a remote cluster by using {acm}, check if the metric is exported as a Platform or a User workload metric. Use one of the following four methods to find the metric type:
 
-. Before you add a custom metric, verify that `mco observability` is enabled with the following command: 
+* Find the metric type in the documentation of the solution you want to monitor.
 
-+
-[source,bash]
-----
-oc get mco observability -o yaml
-----
+* Find the metric type by contacting the support for your product.
 
-. Check for the following message in the `status.conditions.message` section reads:
+* Find the metric type in the console by selecting your cluster from the dropdown and navigating to *Observe* > *Metrics*.
+** In the Prometheus column, Platform metrics are listed as the metric provider.
+** In the Prometheus column, User workload metrics are listed as `openshift-user-workload-monitoring`.
 
-+
-[source,bash]
-----
-Observability components are deployed and running
-----
+* Find the metric type by checking the annotation that the `ServiceMonitor` for the observed resource uses.
+** Platform metrics use `operator.prometheus.io/controller-id: openshift-platform-monitoring/prometheus-operator`.
+** User workload metrics use `operator.prometheus.io/controller-id: openshift-user-workload-monitoring/prometheus-operator`.
 
-. Create the `observability-metrics-custom-allowlist` config map in the `open-cluster-management-observability` namespace with the following command:
+If the metric is a Platform, continue at <<adding-platform-metrics,Adding Platform metrics>>. If the metric is a User workload, continue at <<adding-user-workload-metrics,Adding user workload metrics>>.
 
-+
-[source,bash]
-----
-oc apply -n open-cluster-management-observability -f observability-metrics-custom-allowlist.yaml
-----
+[#adding-platform-metrics]
+=== Adding Platform metrics
 
-. Add the name of the custom metric to the `metrics_list.yaml` parameter. Your YAML for the config map might resemble the following content:
+You can monitor Platform metrics creating a `ConfigMap` on the hub cluster in the `open-cluster-management-observability` namespace called `observability-metrics-custom-allowlist`. See the following `ConfigMap` example:
 
-+
 [source,yaml]
 ----
 kind: ConfigMap
 apiVersion: v1
 metadata:
-  name: observability-metrics-custom-allowlist
+name: observability-metrics-custom-allowlist
+namespace: open-cluster-management-observability
 data:
-  metrics_list.yaml: |
-    names: <1>
-      - node_memory_MemTotal_bytes
+metrics_list.yaml: |
+names: <1>
+    node_memory_MemTotal_bytes
+    rules: <2>
+    record: apiserver_request_duration_seconds:histogram_quantile_90
+    expr: histogram_quantile(0.90,sum(rate(apiserver_request_duration_seconds_bucket {job=\"apiserver\", verb!=\"WATCH\"}[5m])) by (verb,le))
+----
+
+<1> *Optional:* Add the name of the custom metrics you want to collect from your managed cluster.
+<2> *Optional:* Add one value for the `expr` and `record` parameter pair to define the query expression.
+
+The metrics are collected as the name that you define in your managed cluster `record` parameter. The metric values you get are the results after you run the query expression. You can use one section or both sections, which applies to every cluster with monitoring enabled.
+
+If you want to use the configuration for a single cluster, use the following configuration on your managed cluster in the `open-cluster-management-addon-monitoring` namespace:
+[source,yaml]
+----
+kind: ConfigMap
+apiVersion: v1
+metadata:
+name: observability-metrics-custom-allowlist
+namespace: open-cluster-management-addon-observability
+data:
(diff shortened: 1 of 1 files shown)


observability/customize_observability.adoc
```diff
@@ -77,87 +77,94 @@ data:
...
+    expr: histogram_quantile(0.90,sum(rate(apiserver_request_duration_seconds_bucket {job=\"apiserver\", verb!=\"WATCH\"}[5m])) by (verb,le))
 ----
 
-. Add the name of the custom metric to the `metrics_list.yaml` parameter. Your YAML for the config map might resemble the following content:
+<1> *Optional:* Add the names of the custom metrics you want to collect from your managed cluster.
+<2> *Optional:* Add one value for the `expr` and `record` parameter pair to define the query expression.
+
+The name for the metrics you collect is the same name that you define in your managed cluster `record` parameter. After you run the query expression, you get the metric value results. You can use one section or both sections, which applies to every cluster with monitoring enabled.
+
+If you want to use the configuration for a single cluster, use the following example and on your managed cluster in the `open-cluster-management-addon-monitoring` namespace:
```
oafischer: Please check this, the paragraph didn't make much sense to me in the original draft.

observability/customize_observability.adoc
```diff
@@ -77,87 +77,94 @@ data:
...
 
-+
-[source,bash]
-----
-oc get mco observability -o yaml
-----
+* Find the metric type by contacting the support for your product.
 
-. Check for the following message in the `status.conditions.message` section reads:
+* Find the metric type in the console by selecting your cluster from the dropdown and navigating to *Observe* > *Metrics*.
```
jacobbaungard: I am not fully understanding exactly these steps. I looked in a console with user workload metrics enabled, but with no targets. I saw no metrics annotated as "Platform" anywhere.

observability/customize_observability.adoc
```diff
@@ -77,87 +77,94 @@ data:
...
+** In the Prometheus column, Platform metrics are listed as the metric provider.
+** In the Prometheus column, User workload metrics are listed as `openshift-user-workload-monitoring`.
 
-+
-[source,bash]
-----
-Observability components are deployed and running
-----
+* Find the metric type by checking the annotation that the `ServiceMonitor` for the observed resource uses.
```
jacobbaungard, line 86: You can also find ServiceMonitors in the webconsole by going to observe->targets. On this page there is a filter on the top right corner where one can choose the "Source" Platform or User. However this doesn't give an actual list of metrics, just the service monitor/target information

observability/customize_observability.adoc
```diff
@@ -77,87 +77,94 @@ data:
...
+apiVersion: v1
+metadata:
+name: observability-metrics-custom-allowlist
+namespace: open-cluster-management-observability
+data:
+metrics_list.yaml: |
+names: <1>
+  - node_memory_MemTotal_bytes
+    rules: <2>
+  - record: apiserver_request_duration_seconds:histogram_quantile_90
```
jacobbaungard: Did we have this example from before? I believe `rules` are alerting rules and would normally need something like `>0` (i.e there needs to be a condition that can evalutate to true/false).

The example here is a recording rule, and I'm pretty sure this needs to go under `recording_rules` and not just `rules`. Adding alerting rules here is probably not the best ideas in most cases.

Will need to double check this, but pretty sure it needs to be like this.

observability/customize_observability.adoc
jacobbaungard: The example above which creates a configmap on the Hub, will collect those metrics for all managed clusters.

This section, is about collecting custom metrics from just a single managed-cluster.

```suggestion
<2> *Optional:* Add one value for the `expr` and `record` parameter pair to define the query expression.

The name for the metrics you collect is the same name that you define in your managed cluster `record` parameter. After you run the query expression, you get the metric value results. You can use one section or both sections, which applies to every cluster with monitoring enabled.

If you only want to collect custom metrics from a single managed cluster, use the following example and apply the config map on your managed cluster in the `open-cluster-management-addon-observability` namespace:
```

observability/customize_observability.adoc
```diff
@@ -77,87 +77,94 @@ data:
...
+uwl_metrics_list.yaml: <2>
+names:
+  <sample_metrics> <3>
 ----
-+
-<1> Enter the key for the config map data.
-<2> Enter the value of the config map data in YAML format. The `names` section includes the list of metric names, which you want to collect from the `test` namespace. After you create the config map, the observability collector collects and pushes the metrics from the target namespace to the hub cluster.
+
+<1> Add the namespace where you want to capture the metric.
+<2> Add the key for the config map data.
```
jacobbaungard: This key is not user configurable and always needs to be `uwl_metrics_list.yaml`, so I think we could just remove the additional annotation (i.e the <2> there).

observability/customize_observability.adoc
```diff
@@ -77,87 +77,94 @@ data:
...
+<3> Add the value of the config map data in YAML format. Add list of metric names that you want to collect from the `test` namespace to the `names` section. After you create the config map, the observability collector collects and pushes the metrics from the target namespace to the hub cluster.
+
+The previous example monitors the User workload metric `sample_metrics` from the namespace `monitored_namespace`. If you create the configuration in the `open-cluster-management-addon-monitoring` namespace instead, the metrics are collected from all the namespaces of the managed cluster.
```
jacobbaungard:
```suggestion
The previous example monitors the User workload metric `sample_metrics` from the namespace `monitored_namespace`. If you create the configuration in the `open-cluster-management-addon-observability` namespace instead, the metrics are collected from all the namespaces of the managed cluster.
```

observability/customize_observability.adoc
oafischer: Updated

observability/customize_observability.adoc
oafischer: Updated

observability/customize_observability.adoc
oafischer: @gerikis

observability/customize_observability.adoc
oafischer: Any update here @jacobbaungard? I can add `recording_rules` if it's missing and is required.

observability/customize_observability.adoc
oafischer: Is this something you think we should add? @jacobbaungard @gerikis? I would probably add it as a note if we want to add it.

jacobbaungard: I think that might work well as an additional note as suggested, but not essential.

observability/customize_observability.adoc
jacobbaungard: Confirmed, please use `recording_rules` instead of `rules`.

jacobbaungard:
Looks good except for the `rules` -> `recording_rules` and then the unresolved comment at L86. Maybe the section (L86-L88) can be removed completely if we don't get clarification.

observability/customize_observability.adoc
oafischer: Added, thanks!

observability/customize_observability.adoc
oafischer: Updated

oafischer:
Updated with the latest comments if you want to take one last look @jacobbaungard @gerikis. Thanks!

observability/customize_observability.adoc
```diff
@@ -77,87 +77,93 @@ data:
...
+kind: ConfigMap
+apiVersion: v1
+metadata:
+name: observability-metrics-custom-allowlist
+namespace: open-cluster-management-observability
+data:
+metrics_list.yaml: |
+names: <1>
+  - node_memory_MemTotal_bytes
+    recording_ules: <2>
```
jacobbaungard:
```suggestion
    recording_rules: <2>
```

observability/customize_observability.adoc
```diff
@@ -77,87 +77,93 @@ data:
...
+name: observability-metrics-custom-allowlist
+namespace: open-cluster-management-addon-observability
 data:
-  metrics_list.yaml: |
-    names: <1>
-      - node_memory_MemTotal_bytes
+metrics_list.yaml: |
+names: <1>
+  - node_memory_MemTotal_bytes
     rules: <2>
```
jacobbaungard:
```suggestion
    recording_rules: <2>
```

oafischer:
All comments from @jacobbaungard addressed and approval given in scrum

observability/customize_observability.adoc
```diff
@@ -77,87 +77,93 @@ data:
...
 
-+
 [source,yaml]
 ----
 kind: ConfigMap
 apiVersion: v1
 metadata:
-  name: observability-metrics-custom-allowlist
-  namespace: test
+name: observability-metrics-custom-allowlist
```
swopebe: This format looks way off, there is usually the indention under metadata. See the repo for how this is done, even with name.

Example:

```
[source,yaml]
----
apiVersion: cluster.open-cluster-management.io/v1beta1
kind: ClusterCurator
metadata:
  name: my-cluster
  namespace: clusters
  labels:
    open-cluster-management: curator
```

observability/customize_observability.adoc
```diff
@@ -77,87 +77,93 @@ data:
...
+=== Adding Platform metrics
+
+You can monitor Platform metrics by creating a `ConfigMap` on the hub cluster in the `open-cluster-management-observability` namespace. Use `observability-metrics-custom-allowlist` as the name. See the following `ConfigMap` example that you can use to monitor Platform metrics:
+
+[source,yaml]
 ----
-oc apply -n open-cluster-management-observability -f observability-metrics-custom-allowlist.yaml
+kind: ConfigMap
+apiVersion: v1
+metadata:
```
swopebe, line 105: Format?

oafischer:
Formatting fixed

swopebe (approved):

PR #7506, opened 2025-02-17
[2.11][ACM-17851] Updated advanced configuration deployed components for Installer

dislbenn:
As we continue to onboard new components into ACM and MCE, we need to ensure that we are keeping the component list up-to-date.

Related issue: https://issues.redhat.com/browse/ACM-17851

### DIFF
--- a/clusters/install_upgrade/adv_config_install.adoc
+++ b/clusters/install_upgrade/adv_config_install.adoc
@@ -22,7 +22,7 @@ Add one or more of the following attributes to the `MultiClusterEngine` custom r
 | hypershift-local-hosting | Enables local hosting capabilities for within the local cluster environment | True
 | image-based-install-operator-preview | Provides site configuration to {sno} clusters to complete installation | False
 | local-cluster | Enables the import and self-management of the local hub cluster where the {mce-short} is deployed | True
-| managedserviceacccount | Syncronizes service accounts to managed clusters and collects tokens as secret resources back to the hub cluster | True
+| managedserviceacccount | Synchronizes service accounts to managed clusters then collects tokens as secret resources to give back to the hub cluster | True
 | server-foundation | Provides foundational services for server-side operations within the multicluster environment | True
 |===
 
--- a/clusters/install_upgrade/adv_config_install.adoc
+++ b/clusters/install_upgrade/adv_config_install.adoc
@@ -161,7 +161,7 @@ spec:
     node-role.kubernetes.io/infra: ""
 ----
 
-To define a set of node selectors for the {acm-short} hub cluster, see link:../../install/adv_config_install.adoc#node-selector[nodeSelector] in the {acm-short} documentation.
+To define a set of node selectors for the {acm-short} hub cluster, see link:../../install/adv_config_install.adoc#node-selector[nodeSelector] in the product documentation.
 
 [#tolerations]
 == tolerations
--- a/clusters/install_upgrade/adv_config_install.adoc
+++ b/clusters/install_upgrade/adv_config_install.adoc
@@ -178,7 +178,7 @@ spec:
 
 The previous infra-node toleration is set on pods by default without specifying any tolerations in the configuration. Customizing tolerations in the configuration will replace this default behavior.
 
-To define a list of tolerations for the {acm} hub cluster,, see link:../../install/adv_config_install.adoc#tolerations[tolerations] in the {acm-short} documentation.
+To define a list of tolerations for the {acm-short} hub cluster, see link:../../install/adv_config_install.adoc#tolerations[tolerations] in the product documentation.
 
 [#serviceaccount-addon]
 == ManagedServiceAccount add-on
--- a/clusters/install_upgrade/adv_config_install.adoc
+++ b/clusters/install_upgrade/adv_config_install.adoc
@@ -20,9 +20,8 @@ Add one or more of the following attributes to the `MultiClusterEngine` custom r
 | hive | Provisions and performs initial configuration of {ocp-short} clusters | True
 | hypershift | Hosts {ocp-short} control planes at scale with cost and time efficiency, and cross-cloud portability | True
 | hypershift-local-hosting | Enables local hosting capabilities for within the local cluster environment | True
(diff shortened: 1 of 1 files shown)


swopebe:
@dislbenn I updated this for edge manager, so you don't need to send that PR. I don't think I got issues for these before. IBIO we have no doc for but we are taking the doc from OCP next release.

dislbenn:
@swopebrandi1 For this PR, we were missing the IBIO component within the advanced configuration section for MCE. This component was a tech preview and was delivered in MCE 2.6; however, the docs were never updated to reflect that.

clusters/install_upgrade/adv_config_install.adoc
```diff
@@ -20,8 +20,9 @@ Add one or more of the following attributes to the `MultiClusterEngine` custom r
 | hive | Provisions and performs initial configuration of {ocp-short} clusters | True
 | hypershift | Hosts {ocp-short} control planes at scale with cost and time efficiency, and cross-cloud portability | True
 | hypershift-local-hosting | Enables local hosting capabilities for within the local cluster environment | True
+| image-based-install-operator-preview | Provides site configuration to {sno} clusters to complete installation | False
 | local-cluster | Enables the import and self-management of the local hub cluster where the {mce-short} is deployed | True
-| managedserviceacccount | Syncronizes service accounts to managed clusters and collects tokens as secret resources back to the hub cluster | False
+| managedserviceacccount | Syncronizes service accounts to managed clusters and collects tokens as secret resources back to the hub cluster | True
```
jc-berger: Does it collect tokens then give these tokens back to the hub cluster? If so, perhaps a small edit would make that process clearer. What do you think of this edit? 

```suggestion
| managedserviceacccount | Synchronizes service accounts to managed clusters then collects tokens as secret resources to give back to the hub cluster | True
```

jc-berger: I used "then" thinking that the "collect" action needed to be done after "synchronize" if they don't have that causal relationship, just use "and" like you did :) 

```suggestion
| managedserviceacccount | Synchronizes service accounts to managed clusters and collects tokens as secret resources to give back to the hub cluster | True
```

clusters/install_upgrade/adv_config_install.adoc
```diff
@@ -160,6 +161,8 @@ spec:
     node-role.kubernetes.io/infra: ""
 ----
 
+To define a set of node selectors for the {acm-short} hub cluster, see link:../../install/adv_config_install.adoc#node-selector[nodeSelector] in the {acm-short} documentation.
```
jc-berger: I don't think we need {acm-short} documentation since we're giving users a specific link to click, so not sure if we need to provide this type of direction. 

```suggestion
To define a set of node selectors for the {acm-short} hub cluster, see link:../../install/adv_config_install.adoc#node-selector[nodeSelector] in the product documentation.
```

clusters/install_upgrade/adv_config_install.adoc
```diff
@@ -176,7 +179,9 @@ spec:
 
 The previous infra-node toleration is set on pods by default without specifying any tolerations in the configuration. Customizing tolerations in the configuration will replace this default behavior.
 
-[#serviceaccount-addon-intro]
+To define a list of tolerations for the {acm} hub cluster,, see link:../../install/adv_config_install.adoc#tolerations[tolerations] in the {acm-short} documentation.
```
jc-berger: Same as above and totally optional! 

```suggestion
To define a list of tolerations for the {acm-short} hub cluster, see link:../../install/adv_config_install.adoc#tolerations[tolerations] in the product documentation.
```

jc-berger (approved):
/lgtm

clusters/install_upgrade/adv_config_install.adoc
swopebe: I disagree that we don't need "in the documentation" in this case,  but we can remove the product name--agree that is redundant. Because the link name is simply [node-selector] and can be **anything** if read out loud to a non-sighted user, I will use that to indicate it is in the product doc. I am always thinking of the non-sighted user and trying to orient them better.

dislbenn:
For this PR, we can remove the IBIO component entry since it was dev-preview in ACM 2.11.

clusters/install_upgrade/adv_config_install.adoc
```diff
@@ -21,7 +21,7 @@ Add one or more of the following attributes to the `MultiClusterEngine` custom r
 | hypershift | Hosts {ocp-short} control planes at scale with cost and time efficiency, and cross-cloud portability | True
 | hypershift-local-hosting | Enables local hosting capabilities for within the local cluster environment | True
 | local-cluster | Enables the import and self-management of the local hub cluster where the {mce-short} is deployed | True
-| managedserviceacccount | Syncronizes service accounts to managed clusters and collects tokens as secret resources back to the hub cluster | False
+| managedserviceacccount | Synchronizes service accounts to managed clusters, and collects tokens as secret resources back to the hub cluster | True
 | server-foundation | Provides foundational services for server-side operations within the multicluster environment | True
```
swopebe, line 25: keeping this as it came from the developer "to give" may not be accurate. Added comma and removed "then" because that may also not be accurate chronologically.

swopebe: @dislbenn can you help here with this phrase:

tokens as secret resources **back** to the hub cluster

to send back to the hub cluster?

swopebe: removed IBIO bc developer said it was dev preview and we would not doc that.

dislbenn: @swopebe I think we can add "to give" back to the statement. This is what is defined within the `ManagedServiceAccount` repo:

> Managed Service Account" is an OCM addon developed over [addon-framework](https://github.com/open-cluster-management-io/addon-framework) for synchronizing [ServiceAccount](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/) to the managed clusters and collecting the tokens from these local service accounts as secret resources back to the hub cluster.

swopebe:
Merging, though some discussion with dev after peer review, the changes were small.

PR #7508, opened 2025-02-17
https://issues.redhat.com/browse/ACM-17809

jc-berger:
See Jira:   https://issues.redhat.com/browse/ACM-17809

### DIFF
--- a/business_continuity/volsync/volsync_replicate.adoc
+++ b/business_continuity/volsync/volsync_replicate.adoc
@@ -10,18 +10,21 @@ Before installing VolSync on your clusters, you must have the following requirem
 
 * A configured {ocp} environment on a supported {acm-short} hub cluster
 
+* The storage driver that you use for your source persistent volume must be CSI-compatible and able to support snapshots. 
+
+If you want to use the `rsync-tls` mover to replicate data between clusters, you must have the following requirements: 
+
 * At least two managed clusters for the same {acm-short} hub cluster
 
 * Network connectivity between the clusters that you are configuring with VolSync. If the clusters are not on the same network, you can configure the link:../../networking/submariner/subm_intro.adoc#submariner[Submariner multicluster networking and service discovery] and use the `ClusterIP` value for `ServiceType` to network the clusters, or use a load balancer with the `LoadBalancer` value for `ServiceType`.
 
-* The storage driver that you use for your source persistent volume must be CSI-compatible and able to support snapshots. 
 
 [#volsync-install-clusters]
 == Installing VolSync on the managed clusters
 
 To enable VolSync to replicate the persistent volume claim on one cluster to the persistent volume claim of another cluster, you must install VolSync on both the source and the target managed clusters.
 
-VolSync does not create its own namespace, so it is in the same namespace as other {ocp-short} all-namespace operators. Any changes that you make to the operator settings for VolSync also affects the other operators in the same namespace, such as if you change to manual approval for channel updates. 
+You can install VolSync as a deployment on the `volsync-system` namespace on the managed cluster. Doing this provides support for non-OpenShift managed clusters. 
 
 You can use either of two methods to install VolSync on two clusters in your environment. You can either add a label to each of the managed clusters in the hub cluster, or you can manually create and apply a `ManagedClusterAddOn`, as they are described in the following sections:
 
@@ -45,13 +48,21 @@ The VolSync service pod is installed on the managed cluster.
 
 . Add the same label the other managed cluster. 
 
-. Run the following command on each managed cluster to confirm that the VolSync operator is installed:
+. Run the following command on the hub to confirm that the VolSync operator is installed:
 +
 ----
-oc get csv -n openshift-operators
+oc -n <managed-cluster-name> get ManagedClusterAddOn volsync
 ----
 +
-There is an operator listed for VolSync when it is installed.
+
+. Verify that the command output shows that the `managedclusteraddon` has a `Ready` status. 
+
+. Run the following command on the managed cluster to confirm that the VolSync operator is installed:
++
+----
+kubectl -n volsync-system get deployment volsync
+----
++
 
 * Complete the following steps from the command-line interface:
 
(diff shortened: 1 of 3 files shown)


business_continuity/volsync/volsync_replicate.adoc
```diff
@@ -10,18 +10,21 @@ Before installing VolSync on your clusters, you must have the following requirem
...
 
-* The storage driver that you use for your source persistent volume must be CSI-compatible and able to support snapshots. 
 
 [#volsync-install-clusters]
 == Installing VolSync on the managed clusters
 
 To enable VolSync to replicate the persistent volume claim on one cluster to the persistent volume claim of another cluster, you must install VolSync on both the source and the target managed clusters.
 
-VolSync does not create its own namespace, so it is in the same namespace as other {ocp-short} all-namespace operators. Any changes that you make to the operator settings for VolSync also affects the other operators in the same namespace, such as if you change to manual approval for channel updates. 
+You can install VolSync as a deployment on the `volsync-system` namespace on the managed cluster. Doing this provides support for non-OpenShift managed clusters.
```
tesshuflower: Is it possible to say:  "when you install volsync from ACM on the managed cluster, it will be installed into the `volsync-system` namespace on the managed cluster"

I don't want to say "you can install volsync as a deployment ...." etc because there is no option to set any other namespace....

This also doesn't make it support non-openshift clusters.

I'd prefer to say something along the lines of:

"Previously VolSync was installed as an OLM operator with an associated CSV, but now it will be installed as a deployment in the volsync-system namespace. Note that this means non-OpenShift managed clusters are now supported."

^ I think you're probably trying to change the wording which is fine - the main point here is I'm pointing out to users who already have this deployed, that the OLM operator won't be there anymore.  And also that volsync itself will appear as a deployment in the volsync-system namepace.

release_notes/acm_whats_new.adoc
```diff
@@ -79,7 +79,9 @@ See link:../governance/grc_intro.adoc#governance[Governance] to learn more about
 [#dr4hub-whats-new]
 == Backup and restore
 
-To learn about disaster recovery solutions for your hub cluster, see link:../business_continuity/backup_restore/backup_install.adoc#backup-intro[Backup and restore].
+* You can now install VolSync as a deployment on the `volsync-system` namespace on the managed cluster. In earlier versions, you needed to install VolSync as an {olm} operator. By installing VolSync as a deployment on the `volsync-system` namespace, you have support for non-OpenShift managed clusters.
```
tesshuflower: I think again I would prefer to reword a bit.

I'm trying to get across that the main changes in ACM 2.13 and beyond with volsync are:
- You can now deploy volsync from ACM onto ACM managed clusters even when those managed clusters are not openshift.
- Previously VolSync was deployed as an OLM operator on openshift clusters.  Now it will instead appear as a deployment in the `volsync-system` namespace. 

I'd prefer not to say you "can install it as a deployment", as the user is not actually creating a deployment and don't want to be confusing - they are deploying VolSync from ACM as a managed cluster addon (as has always been done - so we don't need to spell this part out in the whats new) - but the end result is no longer an OLM operator with CSV that they see on the managed cluster.  The Volsync operator is still installed, it' s just running in the `volsync-system` namespace now.

tesshuflower (changes requested):
I've left some comments, @jc-berger please let me know if they don't make sense or if you want to discuss further

tesshuflower (approved):
/approve
/lgtm


Thanks Jake!

business_continuity/volsync/volsync_replicate.adoc
```diff
@@ -10,18 +10,23 @@ Before installing VolSync on your clusters, you must have the following requirem
 
 * A configured {ocp} environment on a supported {acm-short} hub cluster
 
+* The storage driver that you use for your source persistent volume must be CSI-compatible and able to support snapshots. 
+
+If you want to use the `rsync-tls` mover to replicate data between clusters, you must have the following requirements: 
+
 * At least two managed clusters for the same {acm-short} hub cluster
 
 * Network connectivity between the clusters that you are configuring with VolSync. If the clusters are not on the same network, you can configure the link:../../networking/submariner/subm_intro.adoc#submariner[Submariner multicluster networking and service discovery] and use the `ClusterIP` value for `ServiceType` to network the clusters, or use a load balancer with the `LoadBalancer` value for `ServiceType`.
```
oafischer: Optional: How do you feel about indenting this to keep the prereqs streamlined?

```suggestion
* If you want to use the `rsync-tls` mover to replicate data between clusters, you must have the following requirements: 

** At least two managed clusters for the same {acm-short} hub cluster

** Network connectivity between the clusters that you are configuring with VolSync. If the clusters are not on the same network, you can configure the link:../../networking/submariner/subm_intro.adoc#submariner[Submariner multicluster networking and service discovery] and use the `ClusterIP` value for `ServiceType` to network the clusters, or use a load balancer with the `LoadBalancer` value for `ServiceType`.
```

business_continuity/volsync/volsync_replicate.adoc
```diff
@@ -45,13 +50,21 @@ The VolSync service pod is installed on the managed cluster.
 
 . Add the same label the other managed cluster. 
 
-. Run the following command on each managed cluster to confirm that the VolSync operator is installed:
+. Run the following command on the hub to confirm that the VolSync operator is installed:
++
+----
+oc -n <managed-cluster-name> get ManagedClusterAddOn volsync
+----
```
oafischer:
```suggestion
+
[source,bash]
----
oc -n <managed-cluster-name> get ManagedClusterAddOn volsync
----
```

release_notes/acm_whats_new.adoc
```diff
@@ -79,7 +79,9 @@ See link:../governance/grc_intro.adoc#governance[Governance] to learn more about
 [#dr4hub-whats-new]
 == Backup and restore
 
-To learn about disaster recovery solutions for your hub cluster, see link:../business_continuity/backup_restore/backup_install.adoc#backup-intro[Backup and restore].
+* In earlier versions, you had to install VolSync as an {olm} operator on {ocp-short} clusters. Now, when you install VolSync from {acm-short} on the managed cluster, it gets installed into the `volsync-system` namespace on that managed cluster. Now you can deploy VolSync from {acm-short} onto managed clusters, even if those clusters are not {ocp-short} clusters.
```
oafischer:
```suggestion
* When you install VolSync from {acm-short} on your managed cluster, VolSync is now installed in the `volsync-system` namespace on your managed cluster.

* You can now deploy VolSync from {acm-short} onto non-{ocp-short} managed clusters.
```

Should this be 2 separate points? I would also recommend linking to the VolSync docs for at least one of these bullet points.

PR #7512, opened 2025-02-17
https://issues.redhat.com/browse/ACM-17507--adv config edge mngr

swopebe:


dislbenn (approved):
LGTM

release_notes/acm_whats_new.adoc
```diff
@@ -34,6 +34,9 @@ Access the link:https://access.redhat.com/articles/7086905[{acm-short} Support m
 [#install-acm-whats-new]
```
swopebe, line 34: This is "What's new" from the installer perspective and won't go into much about edge manager. That will be in a new entry.

amolnar-gh (approved):

PR #7515, opened 2025-02-19
https://issues.redhat.com/browse/ACM-13112 Fix wrong location

oafischer:
This content was moved from Install to Cluster.

This is old content, not new content! But there are some improvements regarding formatting and Style. Please review them and check for technical accuracy.

Also for 2.9, 2.10, 2.11, 2.12.


gparvin (commented):
My main feedback here is this content uses ACM features from within the MCE documentation so generally I would think that should be in the mce_acm_integration book.  Maybe check with Brandi to validate where this is being organized.

clusters/install_upgrade/upgrade_cluster_disconnected_policies.adoc
```diff
@@ -0,0 +1,557 @@
+[#upgrade-disconnect-cluster-policy]
+= Upgrading disconnected clusters by using policies
+
+If you have the {acm} hub cluster, which uses the `MultiClusterHub` operator to manage, upgrade, and install hub cluster components, you can use {update-service} with {acm-short} policies to upgrade multiple clusters in a disconnected environment.
```
swopebe, line 4: this placement is fine for me and according to Josh, since we make it clear that you need ACM and it is an "upgrade" policy offered in the product for cluster management.

swopebe (approved):

clusters/install_upgrade/upgrade_cluster_disconnected_policies.adoc
swopebe: Approved by @jnpacker in a live meeting.

swopebe: Can we also go back to 2.11 with this change?

oafischer: Thanks for confirming with Josh and approving, I'll go ahead and merge.

This was labeled as 2.9 to 2.13, so I will backport to 2.9 (unless we want to change that decision).

PR #7521, opened 2025-02-19
ACM-18041: Adjust `ConfigurationPolicy` `customMessage` doc (GRC)

dhaiducek:
- Removes irrelevant fields from the deprecated API doc (though I would alternatively be okay with stripping out the entire file and leaving the intro if that's preferable to the removals)
- Updates the custom message doc to clarify its use

ref: https://issues.redhat.com/browse/ACM-18041

### DIFF
--- a/governance/config_policy_ctrl.adoc
+++ b/governance/config_policy_ctrl.adoc
@@ -96,7 +96,7 @@ spec:
 
 | `spec.customMessage`
 | Optional
-| Use this section to configure the compliance message sent by the configuration policy based on the current compliance. Each message configuration is a string that can contain Go templates. The context variables `.DefaultMessage` and `.Policy` are available for use in the templates. You can access the default message using the `.DefaultMessage` parameter. The `.Policy` context variable contains the current policy object, including its status. For example, you can access the state of each related object by specifying `.Policy.status.relatedObjects[*].object`. If you set an `evaluationInterval` other than `watch`, only the kind, name, and namespace of the related objects are available.
+| Configure the compliance message sent by the configuration policy based on the current compliance. Each message configuration is a string that can contain Go templates. The `.DefaultMessage` and `.Policy` context variables are available for use in the templates. You can access the default message by using the `.DefaultMessage` parameter. The `.Policy` context variable contains the current policy object, including its status. For example, you can access the state of each related object by specifying the `.Policy.status.relatedObjects[*].object` field. If you set a value for the `evaluationInterval` field other than `watch`, only the kind, name, and namespace of the related objects are available.
 
 | `spec.customMessage.compliant`
 | Optional
--- a/governance/config_policy_ctrl.adoc
+++ b/governance/config_policy_ctrl.adoc
@@ -12,7 +12,7 @@ If you have existing Kubernetes manifests that you want to put in a policy, the 
 [#configuration-policy-yaml]
 == Configuration policy YAML structure
 
-To find the description of a field on your managed cluster, run the command `oc explain --api-version=policy.open-cluster-management.io/v1 ConfigurationPolicy.<field-path>`, and replace `<field-path>` with the path to the field that you need.
+You can find the description of a field on your managed cluster by running the `oc explain --api-version=policy.open-cluster-management.io/v1 ConfigurationPolicy.<field-path>` command. Replace `<field-path>` with the path to the field that you need.
 
 [source,yaml]
 ----
--- a/governance/config_policy_ctrl.adoc
+++ b/governance/config_policy_ctrl.adoc
@@ -9,8 +9,10 @@ When the `remediationAction` for the configuration policy controller is set to `
 
 If you have existing Kubernetes manifests that you want to put in a policy, the Policy Generator is a useful tool to accomplish this.
 
-[#configuration-policy-sample]
-== Configuration policy sample
+[#configuration-policy-yaml]
+== Configuration policy YAML structure
+
+To find the description of a field on your managed cluster, run the command `oc explain --api-version=policy.open-cluster-management.io/v1 ConfigurationPolicy.<field-path>`, and replace `<field-path>` with the path to the field that you need.
 
 [source,yaml]
 ----
@@ -30,35 +32,40 @@ spec:
     noncompliant: {}
   severity: low
   evaluationInterval:
-    compliant:
-    noncompliant:
+    compliant: ""
+    noncompliant: ""
+  object-templates-raw: ""
   object-templates: <2>
-  - complianceType: musthave
-    objectDefinition:
-      apiVersion: v1
-      kind: Pod
-      metadata:
-        name: pod
-      spec:
-        containers:
-        - image: pod-image
-          name: pod-name
-          ports:
-          - containerPort: 80
-  - complianceType: musthave
-    objectDefinition:
-    apiVersion: v1
-    kind: ConfigMap
-    metadata:
-      name: myconfig
-      namespace: default
-      data:
-      testData: hello
-    spec:
-...
+   - complianceType: musthave
+     metadataComplianceType:
+     recordDiff: ""
+     recreateOption: ""
+     objectSelector:
+       matchLabels: {}
+       matchExpressions: []
+     objectDefinition:
+       apiVersion: v1
+       kind: Pod
+       metadata:
+         name: pod
+       spec:
+         containers:
+          - image: pod-image
+            name: pod-name
+            ports:
+             - containerPort: 80
+   - complianceType: mustonlyhave
+     objectDefinition:
+       apiVersion: v1
+       kind: ConfigMap
+       metadata:
+         name: myconfig
+         namespace: default
+         data:
+           testData: hello
 ----
 <1> Configuration policies that specify an object without a name can only be set to `inform`. When the `remediationAction` for the configuration policy is set to `enforce`, the controller applies the specified configuration to the target managed cluster.
-<2> A Kubernetes object is defined in the `object-templates` array in the configuration policy, where fields of the configuration policy controller is compared with objects on the managed cluster. You can also use templated values within configuration policies. For more information, see _Template processing_.
+<2> A Kubernetes object is defined in the `object-templates` array in the configuration policy, where fields of the configuration policy controller is compared with objects on the managed cluster. You can also use templated values within configuration policies. For more advanced use cases, specify a string in `object-templates-raw` to create the `object-templates` that you want. For more information, see _Template processing_.
 
 [#configuration-policy-yaml-table]
 == Configuration policy YAML table
@@ -89,22 +96,15 @@ spec:
 
 | `spec.customMessage`
 | Optional
-a| Based on the current compliance, use this section to configure the compliance messages sent by the configuration policy to use one of the specified Go templates. You can use the default message from the `.DefaultMessage` parameter and `.Policy` object variable of the current state of the policy from the configuration policy controller. See the state of each related object in the following parameter section of the configuration policy controller:
-
-[source,yaml]
-----
-.Policy.status.relatedObjects[*].object
-----  
-
-If you set an `evaluationInterval`, only identifiable information is available.
+| Use this section to configure the compliance message sent by the configuration policy based on the current compliance. Each message configuration is a string that can contain Go templates. The context variables `.DefaultMessage` and `.Policy` are available for use in the templates. You can access the default message using the `.DefaultMessage` parameter. The `.Policy` context variable contains the current policy object, including its status. For example, you can access the state of each related object by specifying `.Policy.status.relatedObjects[*].object`. If you set an `evaluationInterval` other than `watch`, only the kind, name, and namespace of the related objects are available.
 
 | `spec.customMessage.compliant`
 | Optional
-| Use this field to configure custom messages for configuration policies that are compliant. UTF-8 encoded characters, including emoji and foreign characters are supported values.
+| Use this field to configure custom messages for configuration policies that are compliant. Go templates and UTF-8 encoded characters, including emoji and foreign characters, are supported values.
 
 | `spec.customMessage.noncompliant`
 | Optional
-| Use this field to configure custom messages for configuration policies that are non-compliant. UTF-8 encoded characters, along with emoji and foreign characters are supported values.
+| Use this field to configure custom messages for configuration policies that are non-compliant. Go templates and UTF-8 encoded characters, along with emoji and foreign characters, are supported values.
 
 | `spec.severity`
 | Required
--- a/governance/config_policy_ctrl.adoc
+++ b/governance/config_policy_ctrl.adoc
@@ -100,7 +100,7 @@ spec:
 
 | `spec.customMessage.compliant`
 | Optional
-| Use this field to configure custom messages for configuration policies that are compliant. Go templates and UTF-8 encoded characters, including emoji and foreign characters, are supported values.
+| Configure custom messages for configuration policies that are compliant. Go templates and UTF-8 encoded characters, including emoji and foreign characters, are supported values.
 
 | `spec.customMessage.noncompliant`
 | Optional
--- a/governance/config_policy_ctrl.adoc
+++ b/governance/config_policy_ctrl.adoc
@@ -104,7 +104,7 @@ spec:
 
 | `spec.customMessage.noncompliant`
 | Optional
-| Use this field to configure custom messages for configuration policies that are non-compliant. Go templates and UTF-8 encoded characters, along with emoji and foreign characters, are supported values.
+| Configure custom messages for configuration policies that are non-compliant. Go templates and UTF-8 encoded characters, including emoji and foreign characters, are supported values.
 
 | `spec.severity`
 | Required
--- a/apis/policy.json.adoc
+++ b/apis/policy.json.adoc
@@ -85,14 +85,9 @@ __required__|Parameters describing the policy to be created.|<<_rhacm-docs_apis_
   "kind": "Policy",
   "metadata": {
     "name": "test-policy-swagger",
-    "description": "Example body for Policy API Swagger docs"
   },
   "spec": {
     "remediationAction": "enforce",
-    "customMessage": {
-      "compliant": [],
-      "noncompliant": []
-    },
     "namespaces": {
       "include": [
         "default"
@@ -104,16 +99,8 @@ __required__|Parameters describing the policy to be created.|<<_rhacm-docs_apis_
     "policy-templates": {
     "kind": "ConfigurationPolicy",
     "apiVersion": "policy.open-cluster-management.io/v1",
-    "complianceType": "musthave",
-    "metadataComplianceType": "musthave",
     "metadata": {
-      "namespace": null,
       "name": "test-role"
-    },
-    "selector": {
-      "matchLabels": {
-        "cloud": "IBM"
-      }
     },
     "spec" : {
       "object-templates": {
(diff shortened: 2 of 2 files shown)


governance/adv_template_process.adoc
```diff
@@ -18,10 +19,40 @@ Hub cluster templates are resolved to the data in the referenced resources durin
 
 If you need to manually initiate an update, use the special annotation, `policy.open-cluster-management.io/trigger-update`, to indicate changes for the data referenced by the templates. Any change to the special annotation value automatically initiates template processing. Additionally, the latest contents of the referenced resource are read and updated in the policy definition that is propagated for processing on managed clusters. A way to use this annotation is to increment the value by one each time.
 
+[#bypass-template-processing]
+== Bypass template processing
+
+You might create a policy that contains a policy in `policy-templates` that is not intended to be processed by {acm-short}. By default, {acm-short} processes all templates.
```
jc-berger: do users purposefully creates these policies? or unintentionally? see the following suggestions. Do you think either of these are an improvement? 

```suggestion
If you create a policy that contains a policy in `policy-templates` that is not intended to be processed by {acm-short}, then by default, {acm-short} processes all templates.
```

jc-berger: or: 

```suggestion
You might unintentionally create a policy that contains a policy in `policy-templates` that is not intended to be processed by {acm-short}. By default, {acm-short} processes all templates.
```

governance/adv_template_process.adoc
```diff
@@ -18,10 +19,40 @@ Hub cluster templates are resolved to the data in the referenced resources durin
...
+
+To bypass template processing, you can change `{{ template content }}` to `+{{ `{{ template content }}` }}+` to have the template return a raw string to be processed by a subsequent templating engine.
```
jc-berger: If this is the change we want users to make, I think we can eliminate the "you can" 

```suggestion
To bypass template processing, change `{{ template content }}` to `+{{ `{{ template content }}` }}+` to have the template return a raw string to be processed by a subsequent templating engine.
```

governance/adv_template_process.adoc
```diff
@@ -18,10 +19,40 @@ Hub cluster templates are resolved to the data in the referenced resources durin
...
+
+Alternatively, for configuration policies you can add the following annotation in the relevant `ConfigurationPolicy` contained in your `Policy`: `policy.open-cluster-management.io/disable-templates: "true"`. When this annotation is included in a `ConfigurationPolicy`, the previous workaround is not necessary and template processing is bypassed.
```
jc-berger: is this an alternative step? Or is this the recommended action for users for configuration policies? what do you think of this rewrite? 

```suggestion
For configuration policies,  add the following annotation in the relevant `ConfigurationPolicy` contained in your `Policy`: `policy.open-cluster-management.io/disable-templates: "true"`. 

When you include this annotation in `ConfigurationPolicy`, the template processing gets bypassed.
```

governance/adv_template_process.adoc
```diff
@@ -18,10 +19,40 @@ Hub cluster templates are resolved to the data in the referenced resources durin
...
+
+[#object-selector]
+== Configuration policy object selector
```
jc-berger: wouldn't it be `objectSelector` like it is in the body paragraph? 

```suggestion
== Configuring a policy `objectSelector` 
```

governance/adv_template_process.adoc
```diff
@@ -18,10 +19,40 @@ Hub cluster templates are resolved to the data in the referenced resources durin
...
+
+`objectSelector` is available in `ConfigurationPolicy` in order to iterate over multiple objects using a label selector using the kind specified in the `objectDefinition`. `objectSelector` is similar to the `namespaceSelector`, except that it does not have name filtering built in. To further filter by name, the `skipObject` function is available. Calling `{{ skipObject }}` at any time inside of a Go template will signal to skip that particular object for the policy. View the following example that selects objects that have the label `foo: bar` but skips objects that have a name with the suffix `-prod`:
```
jc-berger: what do you think about hte following revisions to make it a bit more user focused? 

```suggestion
To iterate multiple objects using a label selector, use the `objectSelector` that is available in the  `ConfigurationPolicy` and specified in the `objectDefinition`. The `objectSelector` is similar to the `namespaceSelector`, except that it does not have name filtering built in. To  filter by name, use the `skipObject` function. If you reference `{{ skipObject }}` inside the Go template, it signals to skip that particular object for the policy. 

For an example, view the following YAML sample that selects objects that have the label `foo: bar` but skips objects that have a name with the suffix `-prod`:
```

governance/adv_template_process.adoc
```diff
@@ -18,10 +19,40 @@ Hub cluster templates are resolved to the data in the referenced resources durin
...
+  data:
+    key: '{{ if (hasSuffix "-prod" .ObjectName) }}{{ skipObject }}{{ end }}{{ hasNodesWithExactRoles "infra" }}'
+----
+
 [#raw-object-template-processing]
-== Object template processing
+== Configuration policy raw object template processing
 
-Set object templates with a YAML string representation. The `object-template-raw` parameter is an optional parameter that supports advanced templating use-cases, such as if-else and the `range` function. The following example is defined to add the `species-category: mammal` label  to any ConfigMap in the `default` namespace that has a `name` key equal to `Sea Otter`:
+The `object-template-raw` parameter is an optional parameter that supports advanced templating use-cases, such as if-else and the `range` function. `object-templates-raw` accepts a string containing Go templates that, once parsed, must result in an `object-templates` array. The following example is defined to add the `species-category: mammal` label  to any ConfigMap in the `default` namespace that has a `name` key equal to `Sea Otter`:
```
jc-berger: Some of the terminology I was unfamiliar with and didn't think was necessary. Please let me know if I omitted anything that was needed. 

I made some changes to bring the user in more. Let me know what you think: 

```suggestion
The `object-template-raw` parameter is an optional parameter that supports advanced templating use cases, such as, the `range` function. The `object-templates-raw` accepts a string containing Go templates, and when you use these templates, they must result in an `object-templates` array. 

For example, see the following YAML sample that adds the `species-category: mammal` label  to any ConfigMap in the `default` namespace that has a `name` key equal to `Sea Otter`:
```

governance/config_policy_ctrl.adoc
```diff
@@ -9,8 +9,10 @@ When the `remediationAction` for the configuration policy controller is set to `
 
 If you have existing Kubernetes manifests that you want to put in a policy, the Policy Generator is a useful tool to accomplish this.
 
-[#configuration-policy-sample]
-== Configuration policy sample
+[#configuration-policy-yaml]
+== Configuration policy YAML structure
+
+*NOTE:* You can find this information by issuing an `oc explain --api-version=policy.open-cluster-management.io/v1 ConfigurationPolicy.<field-path>` command, replacing `<field-path>` with the path to the field you're looking for.
```
jc-berger: I think this content would work better as an intro statement rather than a Note, what do you think? Also, I'm not sure how this content relates to the title, "Configuration policy YAML structure" 

This content is almost more like a procedure and action steps we want the user to complete. So, maybe we can make these steps and add some context about how finding a filed helps configure a policy YAML structure? 

```suggestion
To find a field, issue an `oc explain --api-version=policy.open-cluster-management.io/v1 ConfigurationPolicy.<field-path>` command and replace `<field-path>` with the path to the field need.

For an example, see the following YAML sample: 
```

governance/adv_template_process.adoc
```diff
@@ -1,5 +1,5 @@
 [#adv-template-processing]
-= Advanced template processing in configuration policies
+= Advanced template processing in policies
```
jc-berger: We want our titles to having gerunds

```suggestion
= Using advanced template processing in policies
```

governance/adv_template_process.adoc
```diff
@@ -18,10 +19,40 @@ Hub cluster templates are resolved to the data in the referenced resources durin
 
 If you need to manually initiate an update, use the special annotation, `policy.open-cluster-management.io/trigger-update`, to indicate changes for the data referenced by the templates. Any change to the special annotation value automatically initiates template processing. Additionally, the latest contents of the referenced resource are read and updated in the policy definition that is propagated for processing on managed clusters. A way to use this annotation is to increment the value by one each time.
 
+[#bypass-template-processing]
+== Bypass template processing
```
jc-berger:
```suggestion
== Bypassing template processing
```

governance/adv_template_process.adoc
```diff
@@ -18,10 +19,40 @@ Hub cluster templates are resolved to the data in the referenced resources durin
...
+objectDefinition:
+  apiVersion: v1
+  kind: ConfigMap
+  data:
+    key: '{{ if (hasSuffix "-prod" .ObjectName) }}{{ skipObject }}{{ end }}{{ hasNodesWithExactRoles "infra" }}'
+----
+
 [#raw-object-template-processing]
-== Object template processing
+== Configuration policy raw object template processing
```
jc-berger:
```suggestion
== Configuring policy raw object template processing
```

governance/config_policy_ctrl.adoc
```diff
@@ -9,8 +9,10 @@ When the `remediationAction` for the configuration policy controller is set to `
 
 If you have existing Kubernetes manifests that you want to put in a policy, the Policy Generator is a useful tool to accomplish this.
 
-[#configuration-policy-sample]
-== Configuration policy sample
+[#configuration-policy-yaml]
+== Configuration policy YAML structure
```
jc-berger, line 13:
```suggestion
== Configuring a policy YAML structure
```

governance/config_policy_ctrl.adoc
```diff
@@ -30,35 +32,40 @@ spec:
...
-<2> A Kubernetes object is defined in the `object-templates` array in the configuration policy, where fields of the configuration policy controller is compared with objects on the managed cluster. You can also use templated values within configuration policies. For more information, see _Template processing_.
+<1> When the `remediationAction` for the configuration policy is set to `enforce`, the controller applies the specified configuration to the target managed cluster. However, configuration policies that specify an object without a name can only be set to `inform`, unless the `objectSelector` is also configured.
+<2> A Kubernetes object is defined in the `object-templates` array in the configuration policy, where fields of the configuration policy controller is compared with objects on the managed cluster. You can also use templated values within configuration policies or, for more advanced use cases, alternatively specify a string in `object-templates-raw` to dynamically generate the desired `object-templates`. For more information, see _Template processing_.
```
jc-berger: Minor tweaks to match more the RH SG

```suggestion
<2> A Kubernetes object is defined in the `object-templates` array in the configuration policy, where fields of the configuration policy controller is compared with objects on the managed cluster. You can also use templated values within configuration policies. For more advanced use cases,  specify a string in `object-templates-raw` to create `object-templates` that you want. For more information, see _Template processing_.
```

governance/config_policy_ctrl.adoc
```diff
@@ -30,35 +32,40 @@ spec:
...
 
 [#configuration-policy-yaml-table]
 == Configuration policy YAML table
```
jc-berger, line 71:
```suggestion
== Configuring a policy YAML table
```

governance/template_support_intro.adoc
```diff
@@ -38,11 +38,11 @@ See the following table for a comparison of hub cluster and managed cluster temp
...
-| For `ConfigurationPolicy` resources, an `.ObjectNamespace` variable that resolves to the name of the selected namespace from the `spec.namespaceSelector` at runtime.
+| Context variables `.ManagedClusterName`, `.ManagedClusterLabels`, and `.PolicyMetadata` are available. `.ManagedClusterName` resolves to the name of the target cluster where the policy is propagated. `.ManagedClusterLabels` resolves to a map of keys and values of the labels on the managed cluster where the policy is propagated. `.PolicyMetadata` resolves to a map with the `name`, `namespace`, `labels`, and `annotations` keys with values from the root policy.
+| For `ConfigurationPolicy` resources, context variables `.ObjectName` and `.ObjectNamespace` are available inside `objectDefinition`. `.ObjectNamespace` resolves to the name of the selected namespace from the `spec.namespaceSelector` (or `metadata.namespace` if specified), assuming the object is not cluster-scoped. `.ObjectName` resolves to the name of the selected object from the `objectSelector` (or `metadata.name` if specified).
```
jc-berger:
```suggestion
| For `ConfigurationPolicy` resources, context variables `.ObjectName` and `.ObjectNamespace` are available inside `objectDefinition`. `.ObjectNamespace` resolves to the name of the selected namespace from the `spec.namespaceSelector` or `metadata.namespace` if specified, but only if  the object is not cluster-scoped. `.ObjectName` resolves to the name of the selected object from the `objectSelector` or `metadata.name` if specified.
```

jc-berger (changes requested):

governance/adv_template_process.adoc
dhaiducek: Creating a policy with Go templates that aren't intended to be processed by the ACM policy controllers would be intentional, but wouldn't happen all the time. In that case maybe it'd read better by flipping the order and combining the paragraphs?

>By default, {acm-short} processes all Go templates. You might create a policy that contains a policy in `policy-templates` that is not intended to be processed by {acm-short}. To bypass template processing, change `{{ template content }}` to `+{{ `{{ template content }}` }}+` to have the template return a raw string to be processed by a subsequent templating engine.

governance/adv_template_process.adoc
dhaiducek: It is indeed an alternative to the previous solution. With the previous solution, the Go templates can be individually overridden (in case there are other templates there that _should_ be processed by ACM). With this, the template resolution is shut off entirely. That said, maybe something like this for the previous paragraph:

>To bypass template processing for a particular template, change `{{ template content }}` to `+{{ `{{ template content }}` }}+`

And then this for this paragraph?

>Alternatively to bypass template resolution entirely for a configuration policy, add the following annotation in the relevant `ConfigurationPolicy` contained in your `Policy`: `policy.open-cluster-management.io/disable-templates: "true"`.

governance/adv_template_process.adoc
dhaiducek: Overall the changes make sense, though I think the "if-else" is helpful to note. Would this be clearer?

>such as the `if` conditional function and the `range` loop function

governance/config_policy_ctrl.adoc
dhaiducek: Sure, I'll make it an intro statement!

This is our API reference rather than a set of steps (though in the future we'd like to have sets of steps about composing policies, but that's for another time). Using `oc explain` will return similar information as that in this table.

governance/adv_template_process.adoc
dhaiducek: This one should stay since it's our policy kind. Maybe?

>Processing raw templates in configuration policies

Or?

>Processing raw templates in `ConfigurationPolicy`

governance/config_policy_ctrl.adoc
dhaiducek: Similar to above. This one is our YAML reference for `ConfigurationPolicy`. Not sure how to state that with a gerund 🤔

governance/config_policy_ctrl.adoc
dhaiducek: Similar to above. This one is our YAML reference for ConfigurationPolicy. Not sure how to state that with a gerund 🤔

governance/config_policy_ctrl.adoc
```diff
@@ -9,8 +9,10 @@ When the `remediationAction` for the configuration policy controller is set to `
 
 If you have existing Kubernetes manifests that you want to put in a policy, the Policy Generator is a useful tool to accomplish this.
 
-[#configuration-policy-sample]
-== Configuration policy sample
+[#configuration-policy-yaml]
+== Configuration policy YAML structure
+
+To find the description of a field on your managed cluster, issue an `oc explain --api-version=policy.open-cluster-management.io/v1 ConfigurationPolicy.<field-path>` command and replace `<field-path>` with the path to the field needed.
```
jc-berger: I'm not opposed to using the verb "issue" here, but we typically say "run a command" is there a  specific reason why we'd want to say "issue a command" here? 

 ```suggestion
To find the description of a field on your managed cluster, run an `oc explain --api-version=policy.open-cluster-management.io/v1 ConfigurationPolicy.<field-path>` command and replace `<field-path>` with the path to the field needed.
```

Comment (jc-berger) [inline governance/config_policy_ctrl.adoc]:
```suggestion
To find the description of a field on your managed cluster, issue an `oc explain --api-version=policy.open-cluster-management.io/v1 ConfigurationPolicy.<field-path>` command and replace `<field-path>` with the path to the field that you need. 
```

dhaiducek: Nah, just the wording I put in there--I'll update it. 😄

governance/config_policy_ctrl.adoc
```diff
@@ -9,8 +9,10 @@ When the `remediationAction` for the configuration policy controller is set to `
 
 If you have existing Kubernetes manifests that you want to put in a policy, the Policy Generator is a useful tool to accomplish this.
 
-[#configuration-policy-sample]
-== Configuration policy sample
+[#configuration-policy-yaml]
+== Configuration policy YAML structure
+
+To find the description of a field on your managed cluster, run the command `oc explain --api-version=policy.open-cluster-management.io/v1 ConfigurationPolicy.<field-path>`, and replace `<field-path>` with the path to the field that you need.
```
amolnar-gh: This is a step but I see that the title was just changed to indicate a reference instead of a procedure so I'm a bit torn on this sentence. Normally, I'd change it to something like this: 

```suggestion
. Find the description of a field on your managed cluster by running the following command:

+
[source,bash]
---- 
oc explain --api-version=policy.open-cluster-management.io/v1 ConfigurationPolicy.<field-path> <1>
----
<1> Replace `<field-path>` with the path to the field that you need.
```

However, this would completely imply that this is a procedure. You could also say:
```suggestion
You can find the description of a field on your managed cluster by running the `oc explain --api-version=policy.open-cluster-management.io/v1 ConfigurationPolicy.<field-path>` command. Replace `<field-path>` with the path to the field that you need.
```

governance/config_policy_ctrl.adoc
```diff
@@ -89,22 +96,15 @@ spec:
...
-
-[source,yaml]
-----
-.Policy.status.relatedObjects[*].object
-----  
-
-If you set an `evaluationInterval`, only identifiable information is available.
+| Use this section to configure the compliance message sent by the configuration policy based on the current compliance. Each message configuration is a string that can contain Go templates. The context variables `.DefaultMessage` and `.Policy` are available for use in the templates. You can access the default message using the `.DefaultMessage` parameter. The `.Policy` context variable contains the current policy object, including its status. For example, you can access the state of each related object by specifying `.Policy.status.relatedObjects[*].object`. If you set an `evaluationInterval` other than `watch`, only the kind, name, and namespace of the related objects are available.
```
amolnar-gh:
```suggestion
| Configure the compliance message sent by the configuration policy based on the current compliance. Each message configuration is a string that can contain Go templates. The `.DefaultMessage` and `.Policy` context variables are available for use in the templates. You can access the default message by using the `.DefaultMessage` parameter. The `.Policy` context variable contains the current policy object, including its status. For example, you can access the state of each related object by specifying the `.Policy.status.relatedObjects[*].object` field. If you set a value for the `evaluationInterval` field other than `watch`, only the kind, name, and namespace of the related objects are available.
```

governance/config_policy_ctrl.adoc
```diff
@@ -89,22 +96,15 @@ spec:
...
 
 | `spec.customMessage.compliant`
 | Optional
-| Use this field to configure custom messages for configuration policies that are compliant. UTF-8 encoded characters, including emoji and foreign characters are supported values.
+| Use this field to configure custom messages for configuration policies that are compliant. Go templates and UTF-8 encoded characters, including emoji and foreign characters, are supported values.
```
amolnar-gh:
```suggestion
| Configure custom messages for configuration policies that are compliant. Go templates and UTF-8 encoded characters, including emoji and foreign characters, are supported values.
```

governance/config_policy_ctrl.adoc
```diff
@@ -89,22 +96,15 @@ spec:
...
 
 | `spec.customMessage.noncompliant`
 | Optional
-| Use this field to configure custom messages for configuration policies that are non-compliant. UTF-8 encoded characters, along with emoji and foreign characters are supported values.
+| Use this field to configure custom messages for configuration policies that are non-compliant. Go templates and UTF-8 encoded characters, along with emoji and foreign characters, are supported values.
```
amolnar-gh:
```suggestion
| Configure custom messages for configuration policies that are non-compliant. Go templates and UTF-8 encoded characters, including emoji and foreign characters, are supported values.
```

amolnar-gh (approved):
Again, overall it looks great, just small nits! The biggest question is the Config policy YAML but you can ignore my suggestion, I just wanted to call attention to it.

PR #7528, opened 2025-02-20
ACM-18034: Doc `resourceRequirements` add-on configuration

dhaiducek:
This is an add-on configuration that goes into the add-on doc. However, for 2.13 it only applies to GRC. In 2.14, the remainder of the add-ons will implement it in [ACM-17874](https://issues.redhat.com/browse/ACM-17874).

ref: https://issues.redhat.com/browse/ACM-18034

### DIFF
--- a/add-ons/configure_addons.adoc
+++ b/add-ons/configure_addons.adoc
@@ -86,12 +86,12 @@ oc apply -f addondeploymentconfig
 . Use the configuration that you created as the global default configuration for your add-on by running the following command:
 +
 ----
-oc patch clustermanagementaddons <addon-name> --type='json' -p='[{"op":"add", "path":"/spec/supportedConfigs", "value":[{"group":"addon.open-cluster-management.io","resource":"addondeploymentconfigs", "defaultConfig":{"name":"deploy-config","namespace":"open-cluster-management-hub"}}]}]'
+oc patch clustermanagementaddons <addon-name> --type='json' -p='[{"op":"add", "path":"/spec/supportedConfigs", "value":[{"group":"addon.open-cluster-management.io","resource":"addondeploymentconfigs", "defaultConfig":{"name":"<config-name>","namespace":"<config-namespace>"}}]}]'
 ----
 +
-* Replace `addon-name` with your add-on name.
-* Replace `config-name` with the name of the `AddonDeploymentConfig` that you just created.
-* Replace `config-namespace` with the namespace of the `AddonDeploymentConfig` that you just created.
+* Replace `<addon-name>` with your add-on name.
+* Replace `<config-name>` with the name of the `AddonDeploymentConfig` that you just created.
+* Replace `<config-namespace>` with the namespace of the `AddonDeploymentConfig` that you just created.
 
 The `nodeSelector` and `tolerations` that you specified are applied to all of your add-on on each of the managed clusters.
 
--- a/add-ons/configure_addons.adoc
+++ b/add-ons/configure_addons.adoc
@@ -1,7 +1,7 @@
 [#configure-addons]
 = Configuring klusterlet add-ons
 
-In {acm-short}, you can configure the following klusterlet add-ons:
+In {acm-short}, you can configure the following klusterlet add-ons to improve the performance and functionality of your managed clusters:
 
 * application-manager
 * cert-policy-controller
@@ -17,17 +17,21 @@ In {acm-short}, you can configure the following klusterlet add-ons:
 * work-manager
 
 // This note only applies to ACM 2.13 and should be removed in ACM 2.14.
-*NOTE:* Configuring `resourceRequirements` is only available for the following add-ons:
+*Note:* Configuring `resourceRequirements` is only available for the following add-ons:
 
 * cert-policy-controller
 * config-policy-controller
 * governance-policy-framework
 
-Complete the following steps:
+[#applying-specifications-addons]
+== Applying specifications to klusterlet add-ons
+
+When you configure the klusterlet add-ons, you can apply specifications to all the add-ons on each of your managed clusters, such as the `nodeSelector` and `tolerations`. To configure the klusterlet add-on, complete the following steps:
 
 .  Use the `AddonDeploymentConfig` API to create an add-on configuration in any namespace on the hub cluster.
 
-. Create a file named `addondeploymentconfig.yaml` that is based on the following template:
+. Create a file named `addondeploymentconfig.yaml` with the following template:
+
 +
 [source,yaml]
 ----
@@ -49,15 +53,16 @@ spec:
         memory: 150Mi
 ----
 +
-<1> Replace `config-name` with the name of the `AddonDeploymentConfig` that you just created.
-<2> Replace `config-namespace` with the namespace of the `AddonDeploymentConfig` that you just created.
+<1> Replace `config-name` with the name of the `AddonDeploymentConfig` that you created.
+<2> Replace `config-namespace` with the namespace of the `AddonDeploymentConfig` that you created.
 <3> Replace `node-selector` with your node selector.
 <4> Replace `tolerations` with your tolerations.
 // The note here only applies to ACM 2.13 and should be removed in ACM 2.14.
-<5> (*NOTE:* Configuring `resourceRequirements` is only available for policy add-ons.) List resource requirements here to override the `resources` of the add-on workload containers. If an add-on container matches more than one of the items in the list, the last matching configuration is applied.
-<6> Replace `<workload-kind>` with the kind of workload, like "deployment". Replace `<workload-name>` with the name of the workload. Replace `<container-name>` with the name of the container. For any of these values, a `*` wildcard can be used to apply the configuration to all objects managed by the add-on. For example, `*:*:*` would apply to every container of every workload kind in any add-on to which this configuration is attached.
+<5> *Note:* Configuring `resourceRequirements` is only available for policy add-ons. List resource requirements here to override the `resources` of the add-on workload containers. If an add-on container matches more than one of the items in the list, the last matching configuration is applied.
+<6> Replace `<workload-kind>` with the kind of workload, for example: `deployment`. Replace `<workload-name>` with the name of the workload. Replace `<container-name>` with the name of the container. 
+.. For any of these values, you can use `\*` attribute to apply the configuration to all objects managed by the add-on. For example, if you used the  `\*:*:*` attribute , it would apply the configuration  to every container of every workload kind in any add-on the configuration is attached to.
 +
-A completed `AddOnDeployment` file might resemble the following example: 
+A completed `AddOnDeployment` resembles the following example: 
 +
 [source,yaml]
 ----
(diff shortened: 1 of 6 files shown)


swopebe:
This is connected to feature work: https://issues.redhat.com/browse/ACM-17563

add-ons/configure_addons.adoc
```diff
@@ -0,0 +1,128 @@
+[#configure-addons]
+= Configuring klusterlet add-ons
+
+In {acm-short}, you can configure the following klusterlet add-ons to improve the performance and functionality of your managed clusters:
```
jc-berger: Please let me know what you think about adding some context to the "benefit" or "why" users would configure kluterlet add-ons.

add-ons/configure_addons.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+// This note only applies to ACM 2.13 and should be removed in ACM 2.14.
+*Note:* Configuring `resourceRequirements` is only available for the following add-ons:
+
+* cert-policy-controller
+* config-policy-controller
+* governance-policy-framework
+
+[#applying-specifications-addons]
+== Applying specifications to klusterlet add-ons
```
jc-berger: since we had two different procedures here, I re-structured the content a bit. Please let me know what you think, and if I am correct that the goal of this procedure/task is to instruct users to apply specification to their klusterlet add-ons

add-ons/configure_addons.adoc
```diff
@@ -0,0 +1,128 @@
...
+----
++
+<1> Replace `config-name` with the name of the `AddonDeploymentConfig` that you created.
+<2> Replace `config-namespace` with the namespace of the `AddonDeploymentConfig` that you created.
+<3> Replace `node-selector` with your node selector.
+<4> Replace `tolerations` with your tolerations.
+// The note here only applies to ACM 2.13 and should be removed in ACM 2.14.
+<5> *Note:* Configuring `resourceRequirements` is only available for policy add-ons. List resource requirements here to override the `resources` of the add-on workload containers. If an add-on container matches more than one of the items in the list, the last matching configuration is applied.
+<6> Replace `<workload-kind>` with the kind of workload, for example: `deployment`. Replace `<workload-name>` with the name of the workload. Replace `<container-name>` with the name of the container. 
+.. For any of these values, you can use `\*` attribute to apply the configuration to all objects managed by the add-on. For example, if you used the  `\*:*:*` attribute , it would apply the configuration  to every container of every workload kind in any add-on the configuration is attached to.
```
jc-berger: So, from reading various troubleshooting, I saw that \ before an asterisks in ticks, ( `*`) would prevent the rendering error. It worked the first time, but not the second time. I'll continue to look around for ways to render `*:*:*` but I had to take a break and move on, loll

add-ons/configure_addons.adoc
```diff
@@ -0,0 +1,128 @@
+[#configure-addons]
```
swopebe: I would like to change the filename here because "addons" means a lot right now within ACM. Can we indicate this is klusterlet addons? This will help bc in the future I will be rearranging this area.

add-ons/configure_addons.adoc
dhaiducek: I ran into this. I think the alternative syntax is a backtick and then a plus sign:
```
`+*:*:*+`
```
ref: https://docs.asciidoctor.org/asciidoc/latest/text/literal-monospace/

Alternatively, I think you can escape every single asterisk like:
```
`\*:\*:\*`
```
I'll note though that ASCIIDoc is not Markdown, so it won't necessarily render the same in GitHub as in staging/prod.

add-ons/configure_addons.adoc
dhaiducek: So something like `configure_klusterlet_addons.adoc`, then?

add-ons/configure_addons.adoc
dhaiducek: Yes, I think "performance and functionality" makes sense!

add-ons/configure_addons.adoc
dhaiducek: The reflow is good! But perhaps there could be three topics here to make referring directly to the `AddOnDeploymentConfig` more straightforward?
- Setting up the `AddOnDeploymentConfig` to configure klusterlet add-ons
- Configuring a klusterlet add-on for all managed clusters:
  - Configure the `ClusterManagementAddOn` with the `AddOnDeploymentConfig` (which applies to the add-on for all managed clusters attached to the hub cluster)
- Configuring a klusterlet add-on for a single managed cluster:
  - Configure the `ManagedClusterAddOn` with the `AddOnDeploymentConfig` (which applies to the add-on for the particular managed cluster attached to that namespace on the hub cluster)

add-ons/configure_klusterlet_addons.adoc
```diff
@@ -0,0 +1,134 @@
+[#configure-klusterlet_addons]
```
dhaiducek: Looks like this tag should have dashes throughout?

```suggestion
[#configure-klusterlet-addons]
```

add-ons/configure_klusterlet_addons.adoc
```diff
@@ -0,0 +1,134 @@
...
+*Note:* Configuring `resourceRequirements` is only available for the following add-ons:
+
+* cert-policy-controller
+* config-policy-controller
+* governance-policy-framework
+
+[#setting-addondeploymentconfig-klusterlet-addons]
+== Setting up the _AddOnDeploymentConfig_ to configure klusterlet add-ons
+
+When you configure the klusterlet add-ons, you can apply specifications to all the add-ons on each of your managed clusters, such as the `nodeSelector` and `tolerations`. To configure the klusterlet add-on, complete the following steps:
```
dhaiducek: I think this should be "any of the add-ons"? (Since the add-ons are configured one at a time)
```suggestion
When you configure the klusterlet add-ons, you can apply specifications to any of the add-ons on each of your managed clusters, such as the `nodeSelector` and `tolerations`. To configure the klusterlet add-on, complete the following steps:
```

dhaiducek: (I've updated this in a subsequent commit, in case it should be reverted.)

dhaiducek:
FYI: I've added a dev commit adjusting some spacing, adding some YAML syntax, and minor wording updates.

I verified the `oc` commands just now, but I think it'd be best have someone from server foundation give it a quick look at this content also since it's technically their content, so getting a quick approval from them would be good to close the loop.

swopebe:
I will review this one and edit as needed for peer review.

add-ons/configure_klusterlet_addons.adoc
```diff
@@ -0,0 +1,148 @@
...
+* hypershift-addon
+* managed-serviceaccount 
+* observability-controller
+* search-collector
+* submariner
+* volsync
+* work-manager
+
+// This note only applies to ACM 2.13 and should be removed in ACM 2.14.
+*Note:* Configuring `resourceRequirements` is only available for the following add-ons:
```
swopebe: Please change this a bit, the comment in this case is fine to keep, though we try to keep the source free of comments.

It's not a note, since it is rather important to the release. I would use _important_

Something like:

For {acm-short} version 2.13, you can only configure.... (Make it active) 

(Since I have cleaned out a lot of unnecessary versioning in the doc to simplify our lives and the doc, I will know to review this next release because it will come up in my versioning work I do each release.)

add-ons/configure_klusterlet_addons.adoc
```diff
@@ -0,0 +1,148 @@
...
+* cert-policy-controller
+* config-policy-controller
+* governance-policy-framework
+
+[#setting-addondeploymentconfig-klusterlet-addons]
+== Setting up the _AddOnDeploymentConfig_ to configure klusterlet add-ons
+
+When you configure the klusterlet add-ons, you can apply specifications to any of the add-ons on each of your managed clusters, such as the `nodeSelector` and `tolerations`. To configure the klusterlet add-on, complete the following steps:
+
+.  Use the `AddonDeploymentConfig` API to create an add-on configuration in any namespace on the hub cluster.
```
swopebe, line 30: extra space here

add-ons/configure_klusterlet_addons.adoc
```diff
@@ -0,0 +1,148 @@
...
+
+. Create a file named `addondeploymentconfig.yaml` with the following template:
```
swopebe, line 31: We usually also have a command to give them to create the file in a two-step process.

add-ons/configure_klusterlet_addons.adoc
```diff
@@ -0,0 +1,148 @@
...
+        memory: 75Mi
+      limits:
+        memory: 150Mi
+----
++
+<1> Replace `config-name` with the name of the `AddonDeploymentConfig` that you created.
+<2> Replace `config-namespace` with the namespace of the `AddonDeploymentConfig` that you created.
+<3> Replace `<node-selector>` with your node selector.
+<4> Replace `<tolerations>` with your tolerations.
+// The note here only applies to ACM 2.13 and should be removed in ACM 2.14.
```
swopebe, line 58: Same thing here, version this and you can remove the comment. You can also remove _Note_ -- in fact, I will create an issue now for next release for us to remove this after checking in with the team.

add-ons/configure_klusterlet_addons.adoc
```diff
@@ -0,0 +1,148 @@
...
+<5> *Note:* Configuring `resourceRequirements` is only available for policy add-ons. List resource requirements here to override the `resources` of the add-on workload containers. If an add-on container matches more than one of the items in the list, the last matching configuration is applied.
+<6> Replace `<workload-kind>` with the kind of workload, for example: `deployment`. Replace `<workload-name>` with the name of the workload. Replace `<container-name>` with the name of the container. 
+.. For any of these values, you can use `\*` attribute to apply the configuration to all objects managed by the add-on. For example, if you used the `*:*:*` attribute , it would apply the configuration to every container of every workload kind in any add-on the configuration is attached to.
++
+A completed `AddOnDeployment` resembles the following example: 
++
```
swopebe, line 65: Needs extra line before (+)

add-ons/configure_klusterlet_addons.adoc
```diff
@@ -0,0 +1,148 @@
...
+        key: node-dedicated
+        value: acm-addon
+        operator: Equal
+----
+
+
+[#configuring-klusterlet-addons-all-managedclusters]
+== Configuring a klusterlet add-on for all managed clusters
+
+After you set up the `AddOnDeploymentConfig`, you can configure it with the `ClusterManagementAddOn` which then applies this add-on configuration to all your managed clusters attached to the hub cluster. To configure a klusterlet add-on for all managed clusters, complete the following steps:
```
swopebe: "managed clusters attached"

Anytime we see this, we need to add "that are" 

managed clusters **that are** attached...

See the [standard.](https://www.ibm.com/docs/en/ibm-style?topic=grammar-pronouns#relative-pronouns)

add-ons/configure_klusterlet_addons.adoc
```diff
@@ -0,0 +1,148 @@
...
+
+. Apply the `AddOnDeploymentConfig` file to your klusterlet add-on by running the following command:
```
swopebe, line 89: Need the extra line here.

add-ons/configure_klusterlet_addons.adoc
```diff
@@ -0,0 +1,148 @@
...
++
+[source,bash]
+----
+oc apply -f addondeploymentconfig.yaml
+----
+
+. Use the configuration that you created as the global default configuration for your add-on by running the following command:
++
```
swopebe, line 98: same

swopebe: We need to tell the user that they are patching something here @dhaiducek -- some more detail; from the line before, we just tell them to use the configuration and run a command.

add-ons/configure_klusterlet_addons.adoc
```diff
@@ -0,0 +1,148 @@
...
+
+[#configuring-klusterlet-addons-single-managedclusters]
+== Configuring a klusterlet add-on for a single managed cluster
+
+You can also override the global default `AddonDeploymentConfig` configuration for your add-on on a certain managed cluster. By overriding, you can configure a klusterlet add-on for a single managed cluster because the add-on configuration only applies to a the particular managed cluster attached to that namespace of the hub cluster. To override configurations, complete the following steps:
+
+. Use the `AddonDeploymentConfig` API to create another configuration to specify the `nodeSelector` and `tolerations` on the hub cluster.
+
+. Connect the new configuration that you created to your `ManagedClusterAddon` add-on on the managed cluster by running the following command:
++
```
swopebe, line 124: Need extra space and a little more detail about the "patch"?

swopebe (changes requested):
this looks really good, but there are a few comments, changes required, and suggestions. I will create the issue so that we can remove the comments.

add-ons/configure_klusterlet_addons.adoc
swopebe: Issue for next release is here: https://issues.redhat.com/browse/ACM-18367

swopebe (approved):

jc-berger (approved):
/lgtm

PR #7490, opened 2025-02-13
https://issues.redhat.com/browse/ACM-17505 -- Manage devices

amolnar-gh:
Issue: https://issues.redhat.com/browse/ACM-17505

To preview, select the file in [this folder](https://drive.google.com/drive/folders/1G984Za5pWqW_EW6qkaPFUoeeJTY22PlJ) with the same name as the branch name from this PR.

PR #7531, opened 2025-02-21
https://issues.redhat.com/browse/ACM-17505--Add Manage and Configure OS to Edge Manager

amolnar-gh:
Issue: https://issues.redhat.com/browse/ACM-17505
Preview: https://drive.google.com/drive/folders/1G984Za5pWqW_EW6qkaPFUoeeJTY22PlJ (click on file named after the branch)


edge_manager/edge_mgr_manage_os_devices.adoc
```diff
@@ -0,0 +1,136 @@
+[#manage-os-config]
+= Operating system configuration
+
+You can include an operating system-level host configuration in the image to provide maximum consistency and repeatability.
```
amolnar-gh: When saying "repeatability", does it refer to deploying/updating with the same image on multiple devices?

edge_manager/edge_mgr_update_os_devices.adoc
```diff
@@ -0,0 +1,62 @@
...
+[...]
+----
+
+. Apply the updated `Device` resource by running the following command:
+
++
+[source,bash]
+----
+flightctl apply -f some_device_name.yaml
+----
```
amolnar-gh: Can we add verification or monitoring?

edge_manager/edge_mgr_manage_config_git_repo.adoc
```diff
@@ -0,0 +1,64 @@
...
++
+See the following example output:
++
+[source,bash]
+----
+NAME           TYPE  REPOSITORY URL                                 ACCESSIBLE
+site-settings  git   https://github.com/<your_org>/<your_repo>.git  True
+----
+
+. To apply the `example-site` configuration to a device, update the device specification as follows:
```
avishayt: I would clarify three things:
1. Explicitly explain that this configuration will take the files and directories in the directory "example-site" of the git repository, at the branch or tag "production", and put those files and directories at "/" on the device.
2. "/" is not writeable in bootc systems - we expect the user to create a directory structure in git to describe where to place the files.
3. RHEM does not monitor the git repository branch/tag for changes.  Pushing updates to git will not cause RHEM to update the configuration.  RHEM will only update devices when the Device specification changes.  Therefore we recommend using version tags for targetRevision.  Git hashes can also be used.

edge_manager/edge_mgr_manage_os_devices.adoc
avishayt: Yes - ensuring all devices have that same configuration

edge_manager/edge_mgr_manage_os_devices.adoc
```diff
@@ -0,0 +1,136 @@
...
+[%header,cols="1,3"]
+|===
+|Parameter|Description
+|`Repository`|The name of a `Repository` resource defined in the {rhem}.
+
+|`TargetRevision`|The branch, tag, or commit of the repository to checkout.
+
+|`Path`|The subdirectory of the repository that contains the configuration.
+
+|`MountPath`|Optional. The directory in the file system of the device to which the configuration is written. By default, the value is the file system root `/`.
```
avishayt: @fzdarsky Are we removing this parameter?  If so, we should do it now

avishayt (commented):
Overall looks great, just a few comments

edge_manager/edge_mgr_manage_config_git_repo.adoc
amolnar-gh: I have two questions:

> "/" is not writeable in bootc systems - we expect the user to create a directory structure in git to describe where to place the files.

1) I can highlight it but in this case, it'd be clearer to provide a valid example. Could we use `/etc/example-site` instead? If yes, this could nicely tie in with the image building best practices ([site-specific config is recommended in `/etc`](https://github.com/flightctl/flightctl/blob/main/docs/user/building-images.md#prefer-configuration-in-usr-over-etc)).

> RHEM does not monitor the git repository branch/tag for changes. Pushing updates to git will not cause RHEM to update the configuration. RHEM will only update devices when the Device specification changes. Therefore we recommend using version tags for targetRevision. Git hashes can also be used.

2) How do you sync your device spec from a Git repo? Use git for version control and then you manually update the device spec? Btw, the phrasing [here](https://github.com/flightctl/flightctl/blob/main/docs/user/managing-devices.md#getting-configuration-from-a-git-repository) implies that it automatically syncs the config once you set up a repo with RHEM.

edge_manager/edge_mgr_manage_devices.adoc
```diff
@@ -0,0 +1,25 @@
+[#manage-devices]
```
fzdarsky: Duplicate with edge_manager/edge_mgr_manage_devices_intro.adoc?
Anyway, see my comments in https://github.com/stolostron/rhacm-docs/pull/7490, please.

edge_manager/edge_mgr_manage_os_devices.adoc
```diff
@@ -0,0 +1,136 @@
...
+You can include an operating system-level host configuration in the image to provide maximum consistency and repeatability.
+To update a configuration, you create a new operating system image and update devices with the new image.
+
+However, updating devices with a new image can be impractical in the following cases:
+
+* The configuration is missing in the image.
+* The configuration needs to be specific to a device.
+* The configuration needs to be updateable at runtime without updating the operating system image and rebooting.
+
+For these cases, you can declare a set of configuration files that can be present on the file system of the device.
```
fzdarsky:
```suggestion
For these cases, you can declare a set of configuration files that shall be present on the file system of the device.
```
You're declaring a desired state, not a possibility.

edge_manager/edge_mgr_manage_os_devices.adoc
```diff
@@ -0,0 +1,136 @@
...
+The {rhem} agent applies updates to the configuration files while ensuring that either all files are successfully updated in the file system, or rolled back to their pre-update state.
+If the user updates both an operating system and configuration set of a device at the same time, the {rhem} agent updates the operating system first, then applies the specified set of configuration files.
+
+You can also specify a list of configurations sets that the {rhem} agent applies in sequence.
```
fzdarsky:
```suggestion
You can also specify a list of configuration sets that the {rhem} agent applies in sequence.
```

edge_manager/edge_mgr_manage_os_devices.adoc
```diff
@@ -0,0 +1,136 @@
...
+In case of a conflict, the last applied configuration set is valid.
+
+*Important:* After the {rhem} agent updates the configuration on the disk, the configuration must be activated.
+Running applications need to reload the new configuration into memory for the configuration to become effective.
```
fzdarsky: Is it clear that "activation" = "reloading the new configuration into memory, such that it becomes effective"? Otherwise, maybe a colon would help?

```suggestion
*Important:* After the {rhem} agent updates the configuration on the disk, the configuration must be activated: Running applications need to reload the new configuration into memory for the configuration to become effective.
```

edge_manager/edge_mgr_manage_os_devices.adoc
```diff
@@ -0,0 +1,136 @@
...
+* <<config-git,Configuration from a Git repository>>
+* <<config-k8s-secret,Secrets from a Kubernetes cluster>>
+* <<config-http,Configuration from an HTTP server>>
+* <<config-inline,Configuration inline in the device specification>>
+
+[#config-git]
+=== Configuration from a Git repository
+
+You can store device configuration in a Git repository such as GitHub or GitLab.
+The {rhem} synchronizes the configuration from the repository to the file system of the device by adding a Git Config Provider.
```
fzdarsky: "by adding a Git Config Provider" seems to refer to "The {rhem} synchronizes" here, while it should refer to the "You can". Maybe:

```suggestion
You can store device configuration in a Git repository such as GitHub or GitLab. You can then add a Git Config Provider, so that the {rhem} synchronizes the configuration from the repository to the file system of the device.
```

edge_manager/edge_mgr_manage_os_devices.adoc
```diff
@@ -0,0 +1,136 @@
...
+The HTTP Config Provider takes the following parameters:
+
+[%header,cols="1,3"]
+|===
+|Parameter|Description
+|`Repository`|The name of a `Repository` resource defined in the {rhem}.
+
+|`Suffix`|The suffix to append to the base URL defined in the `Repository` resource. The suffix can include path and query parameters, for example `/path/to/endpoint?query=param`.
+
+|`FilePath`|The path to the file on the device file system that stores the returned value of the HTTP Server.
```
fzdarsky:
```suggestion
|`FilePath`|The absolute path to the file in the file system of the device to write the response of the HTTP server to.
```

fzdarsky: (For consistency with the equivalent fields in other providers.)

edge_manager/edge_mgr_manage_os_devices.adoc
```diff
@@ -0,0 +1,136 @@
+[#manage-os-config]
+= Operating system configuration
+
+You can include an operating system-level host configuration in the image to provide maximum consistency and repeatability.
+To update a configuration, you create a new operating system image and update devices with the new image.
```
fzdarsky:
```suggestion
To update this configuration, you create a new operating system image and update devices with the new image.
```
(to make clear the statement only refers to configuration included in the image.)

edge_manager/edge_mgr_manage_devices.adoc
amolnar-gh: A rebase should resolve this :)

edge_manager/edge_mgr_manage_os_devices.adoc
amolnar-gh: Or we can just remove "the configuration must be activated" from the sentence and merge the two so it's clear what needs to be done:
> After the {rhem} agent updates the configuration on the disk, the running applications need to reload the new configuration into memory for the configuration to become effective.

edge_manager/edge_mgr_manage_os_devices.adoc
fzdarsky: I'm somewhat on the fence, TBH. I acknowledge there are use cases where this param is useful and so wouldn't remove it from the API. Still, we should find a solution that we don't confuse users, especially UI users that aren't technical or technical but not familiar with K8s.

I'm wondering whether the solution is to better explain this param and the relationship to the `Path` param instead. I'm adding what _might_ work as suggestion.

edge_manager/edge_mgr_manage_os_devices.adoc
```diff
@@ -0,0 +1,135 @@
...
+[%header,cols="1,3"]
+|===
+|Parameter|Description
+|`Repository`|The name of a `Repository` resource defined in the {rhem}.
+
+|`TargetRevision`|The branch, tag, or commit of the repository to checkout.
+
+|`Path`|The subdirectory of the repository that contains the configuration. *Important:* The root directory (`/`) is not writeable in `bootc` systems.
+
+|`MountPath`|Optional. The directory in the file system of the device to which the configuration is written. By default, the value is the file system root `/`.
```
fzdarsky:
```suggestion
|`Path`|The absolute path to the directory in the repository from which files and subdirectories are synchronised to the device's file system. This directory corresponds to the file system root `/` on the device, unless the `MountPath` parameter is specified (see below).

|`MountPath`|Optional. The absolute path to the directory in the file system of the device to write the content of the repository to. By default, the value is the file system root `/`.
```

edge_manager/edge_mgr_manage_os_devices.adoc
```diff
@@ -0,0 +1,135 @@
...
+|===
+
+The `Repository` resource defines the Git repository, the protocol and access credentials that the {rhem} must use.
+The repository needs to be set up only once.
+After setting up, the repository can be used to configure individual devices or device fleets.
+
+[#config-k8s-secret]
+=== Secrets from a Kubernetes cluster
+
+The {rhem} can query the Kubernetes cluster that it is running on for a Kubernetes secret.
```
fzdarsky:
```suggestion
The {rhem} can query the Kubernetes cluster _that it is running on_ for a Kubernetes secret.
```

We might want to emphasise that - currently - _only_ the Kubernetes cluster that RHEM is running on (= the ACM Hub Cluster) can be used, not any external one.

edge_manager/edge_mgr_manage_os_devices.adoc
amolnar-gh: I changed it to:
> For these cases, you can declare a set of configuration files that *is* present on the file system of the device.

edge_manager/edge_mgr_manage_os_devices.adoc
```diff
@@ -0,0 +1,135 @@
...
+
+You can specify configuration inline in a device specification.
+When using the inline device specification, the {rhem} does not need to connect to external systems to fetch the configuration.
+
+The Inline Config Provider takes a list of file specifications, where each file specification takes the following parameters:
+
+[%header,cols="1,3"]
+|===
+|Parameter|Description
+|`Path`|The absolute path to the file on the device.
```
fzdarsky:
```suggestion
|`Path`|The absolute path to the file in the file system of the device to write the content to.
```

edge_manager/edge_mgr_manage_devices_intro.adoc
```diff
@@ -17,10 +17,7 @@ For more information, read the following sections:
 * xref:../edge_manager/edge_mgr_view_devices#view-device[Viewing device inventory and device details]
 * xref:../edge_manager/edge_mgr_label_devices#labels[Labels and label selectors]
 * xref:../edge_manager/edge_mgr_update_labels_on_devices#update-labels-cli[Updating labels using the CLI]
-////
-TODO Add links to remaining sections
-* xref:../edge_manager/edge_mgmt_manage_update_os#update-os[Updating the operating system]
-* xref:../edge_manager/edge_mgr_manage_os.adoc#manage-os-config[Managing the operating system configuration]
-////
+* xref:../edge_manager/edge_mgr_update_os_devices#update-os[Updating the operating system]
+* xref:../edge_manager/edge_mgr_manage_os_devices.adoc#manage-os-config[Managing the operating system configuration]
```
fzdarsky: Should the file be renamed to "...manage_os_configuration.adoc"?

edge_manager/edge_mgr_manage_os_devices.adoc
amolnar-gh: We can't italicize text. However, what about:
> The {rhem} can query only the Kubernetes cluster that the {rhem} is running on for a Kubernetes secret.

Samudelacruz77 (approved):
LGTM

Samudelacruz77 (approved):
LGTM

edge_manager/main.adoc
```diff
@@ -10,3 +10,6 @@ include::edge_mgr_enroll_devices.adoc[leveloffset=+3]
 include::edge_mgr_view_devices.adoc[leveloffset=+3]
 include::edge_mgr_label_devices.adoc[leveloffset=+3]
 include::edge_mgr_update_labels_on_devices.adoc[leveloffset=+3]
+include::edge_mgr_manage_os_config.adoc[leveloffset=+3]
+include::edge_mgr_manage_config_git_repo.adoc[leveloffset=+4]
+include::edge_mgr_update_os_devices.adoc[leveloffset=+3]
```
fzdarsky: I feel updating the OS logically belongs before managing OS configuration. Also because we give a full procedure for getting the my_device.yaml, editing it, and applying it again, which we no longer describe in later sections.

fzdarsky (approved):

edge_manager/main.adoc
amolnar-gh: Updated

edge_manager/edge_mgr_manage_config_git_repo.adoc
```diff
@@ -0,0 +1,66 @@
+[#device-config-git-cli]
+= Managing the device configuration from a Git repository on the CLI
+
```
swopebe, line 3: we can ask this later, but are there any prereqs and what is the access required? @avishayt

edge_manager/edge_mgr_manage_config_git_repo.adoc
```diff
@@ -0,0 +1,66 @@
...
+. Verify that the resource has been correctly created and is accessible by {rhem} by running the following command:
+
++
+[source,bash]
+----
+flightctl get repository/site-settings
+----
++
+See the following example output:
```
swopebe, line 37: just need the extra line here and this file looks good.

edge_manager/edge_mgr_manage_os_config.adoc
```diff
@@ -0,0 +1,136 @@
...
+
+[#config-git]
+=== Configuration from a Git repository
+
+You can store device configuration in a Git repository such as GitHub or GitLab.
+You can then add a Git Config Provider so that the {rhem} synchronizes the configuration from the repository to the file system of the device.
+
+The Git Config Provider takes the following parameters:
+
+[%header,cols="1,3"]
```
swopebe: we don't use these and out tables build fine, do we need them?

edge_manager/main.adoc
```diff
@@ -10,3 +10,6 @@ include::edge_mgr_enroll_devices.adoc[leveloffset=+3]
 include::edge_mgr_view_devices.adoc[leveloffset=+3]
 include::edge_mgr_label_devices.adoc[leveloffset=+3]
 include::edge_mgr_update_labels_on_devices.adoc[leveloffset=+3]
+include::edge_mgr_update_os_devices.adoc[leveloffset=+3]
+include::edge_mgr_manage_os_config.adoc[leveloffset=+3]
+include::edge_mgr_manage_config_git_repo.adoc[leveloffset=+4]
```
swopebe, line 15: we don't normally go in 4, I can look around. We had a reason for that and we said stick with 3--don't remember the why.

swopebe: NM, I see we go level 4 in the cluster doc. :-)

swopebe (approved):
See the few comments. We don't use these `[%header,cols="1,3"]` and I am not sure we need them, I think this would be the only file with them. But we came here with tables that were converted, so that is likely why.

I won't hold this up, but see the few comments. Looks so good. I won't get to the others till later.

edge_manager/edge_mgr_manage_config_git_repo.adoc
avishayt: No prereqs or special access needed

avishayt (approved):

edge_manager/edge_mgr_manage_os_config.adoc
amolnar-gh: I can delete them, we don't need them if the tables build fine

PR #7541, opened 2025-02-25
https://issues.redhat.com/browse/ACM-17505--Add Managing fleets sections

amolnar-gh:


edge_manager/edge_mgr_fleets.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+You can have both individually managed and fleet-managed devices at the same time.
+//How does having both work? How are the devices different from a management point of view?
+
+*Important:* A device cannot be member of more than one fleet at the same time.
+
+When you update the device template for the fleet, the {rhem} service rolls out the new configuration to all of the devices in the fleet.
+The {rhem} agent updates the device as soon as the agent detects the new target configuration.
+The update process is the same when you add an individual device to the fleet or when a device changes fleets.
+
```
amolnar-gh, line 30: Do we need to go into detail about the controller? From upstream:
> When a user updates a fleet's device template to a new version, a component called "fleet controller" starts the process of rolling out this version to all of the fleet's devices over time. When the controller has selected a device for update, it copies the new device template to the device's specification. The next time the device's agent checks in, it learns of the new specification and begins the update process.

edge_manager/edge_mgr_fleets.adoc
```diff
@@ -0,0 +1,117 @@
...
+* Minimizes the risk of configuration mistakes and configuration drift.
+* Automatically applies the target configuration when you add devices to the fleet or replace devices in the fleet.
+
+The fleet specification consists of the following features:
+
+Label selector:: Determines which devices are part of the fleet.
+Device template:: Defines the configuration the devices in the fleet.
+Policies:: Govern how devices are managed, for example how changes to the device template are rolled out to the devices.
+
+You can have both individually managed and fleet-managed devices at the same time.
```
amolnar-gh, line 22: How does having both work? How are the devices different from a management point of view?

fzdarsky: A fleet's label selector determines the labels a device must have in order to be considered part of the fleet.

If a device is "selected into the fleet" in this way, the fleet controller takes ownership of this device and manages it on behalf of the user. The fleet controller basically takes the device (specification) template, "fills in the blanks" specific to a device, and the result is the device specification for that device. That device specification looks exactly like in the "Managing (individual) Devices) section.

If a device isn't selected by any fleet (i.e. no fleet controller assumes ownership of the device), it is "user-managed" or "unmanaged". In this case, the user has to update the device specification manually (or through external automation).

edge_manager/edge_mgr_fleets.adoc
fzdarsky: I don't think the term fleet controller as such or how it works in detail are important (though mentioning this term may help users familiar with that concept from Kubernetes understand what it does and those not familiar would at least understand this thing "takes control").

However, you raised the important questions around "how are indiviually-managed and fleet-managed devices different?" And so I think it's important we mention that "fleet-managed" implies there is something (the fleet controller) that manages on behalf of the user and it works by taking a template and enforcing that on all devices.

edge_manager/edge_mgr_fleets.adoc
```diff
@@ -0,0 +1,111 @@
...
+
+//The same would apply to the other specification items described in link:managing-devices.md[Managing Devices].
+
+However, sometimes not all of the devices in the fleet need to have the exact same specification.
+The {rhem} allows templates to contain placeholders that are populated based on the device name or label values.
+
+The syntax of the placeholders matches that of https://pkg.go.dev/text/template[Go templates].
+However, you can only use simple text and actions.
+The use of conditionals or loops in the placeholders is not supported.
+You can reference anything from the metadata of a device, such as `{{ .metadata.labels.key }}` or `{{ .metadata.name }}`.
```
fzdarsky, line 82: Need to verify that "You can reference anything from the metadata" or just the labels.key and name.

edge_manager/edge_mgr_fleets.adoc
```diff
@@ -0,0 +1,111 @@
...
+* The `replace` function replaces all occurrences of a substring with another string. For example, `{{ replace \"old\" \"new\" .metadata.labels.key }}`.
+* The `getOrDefault` function returns a default value if accessing a missing label. For example, `{{ getOrDefault .metadata.labels \"key\" \"default\" }}`.
+
+You can combine the functions in pipelines, for example `{{ getOrDefault .metadata.labels \"key\" \"default\" | upper | replace \" \" \"-\" }}`.
+
+*Note:* Ensure using proper Go template syntax. For example, `{{ .metadata.labels.target-revision }}` is not valid because of the hyphen. Instead, you must refer to the field as `{{ index .metadata.labels \"target-revision\" }}`.
+
+Here are some examples of what you can do with placeholders in device templates:
+
+You can use the placeholders in device templates in the following ways:
```
fzdarsky: Redundant (remove the second line)?

edge_manager/edge_mgr_fleets.adoc
fzdarsky: @avishayt confirmed it's really _all_ fields. 👍

edge_manager/edge_mgr_fleets.adoc
```diff
@@ -0,0 +1,111 @@
...
+You can also use the following functions in your placeholders:
+
+* The `upper` function changes the value to upper case. For example, `{{ upper .metadata.name }}`.
+* The `lower` function changes the value to lower case. For example, `{{ lower .metadata.labels.key }}`.
+* The `replace` function replaces all occurrences of a substring with another string. For example, `{{ replace \"old\" \"new\" .metadata.labels.key }}`.
+* The `getOrDefault` function returns a default value if accessing a missing label. For example, `{{ getOrDefault .metadata.labels \"key\" \"default\" }}`.
+
+You can combine the functions in pipelines, for example `{{ getOrDefault .metadata.labels \"key\" \"default\" | upper | replace \" \" \"-\" }}`.
+
+*Note:* Ensure using proper Go template syntax. For example, `{{ .metadata.labels.target-revision }}` is not valid because of the hyphen. Instead, you must refer to the field as `{{ index .metadata.labels \"target-revision\" }}`.
```
fzdarsky:
```suggestion
* The `replace` function replaces all occurrences of a substring with another string. For example, `{{ replace "old" "new" .metadata.labels.key }}`.
* The `getOrDefault` function returns a default value if accessing a missing label. For example, `{{ getOrDefault .metadata.labels "key" "default" }}`.

You can combine the functions in pipelines, for example `{{ getOrDefault .metadata.labels "key" "default" | upper | replace " " "-" }}`.

*Note:* Ensure using proper Go template syntax. For example, `{{ .metadata.labels.target-revision }}` is not valid because of the hyphen. Instead, you must refer to the field as `{{ index .metadata.labels "target-revision" }}`.
```

fzdarsky: (I've opened a PR to fix the upstream docs accordingly.)

edge_manager/edge_mgr_fleets.adoc
amolnar-gh: I rephrased this section a bit, trying to make it clear. PTAL: https://github.com/stolostron/rhacm-docs/pull/7541/commits/cfc854dd3002da3080a51db43b1a8412c37d7c21#diff-942fa244bd647701d6e734b379717dc252262c4c02a93e45b1fa84aeb4de5500

edge_manager/edge_mgr_fleets.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+The following fields in device templates support placeholders:
+
+|===
+|Field |Placeholders supported in
+|Operating System Image |repository name, image name, image tag
+|Git Config Provider |target revision, path
+|HTTP Config Provider |URL suffix, path
+|Inline Config Provider |content, path
+|===
```
swopebe, line 110: Looks good. My commits: 544a95d311f56f943b5797bdab81868d8814e61f

edge_manager/edge_mgr_select_device_fleets.adoc
```diff
@@ -0,0 +1,53 @@
+[#select-devices]
+= Selecting devices into a fleet using the CLI
```
swopebe: by using, otherwise it can read that the _fleet is using the CLI_

Standard:

Use with caution using noun
To avoid ambiguity, replace this gerund with either by using or that uses.

To refer to something using one thing or another is often ambiguous. For example, the sentence Back up the server using ABC might mean Back up the server by using ABC or Back up the server that uses ABC. This advice also applies to gerunds other than using. For more information, see the style and grammar tips in the [Global audiences](https://www.ibm.com/docs/en/ibm_style/global-audiences.html) topic in IBM Style

swopebe: Will need to change in the title in any URL

edge_manager/edge_mgr_select_device_fleets.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+Define label selector to add devices into a fleet.
+Complete the following tasks:
+
+. Optional: Verify that the label selector returns the devices that you want to add to the fleet.
```
swopebe: If this is optional, and every step after this is based on this, it sounds like the entire process is optional. If that is the case, let's just remove "optional" here and in the description say:

You have the option to define label selector to add devices into a fleet.

swopebe: I was waiting for a not optional step since one is optional.

swopebe: see other changes in commit: dcb6dc4f3d43bd46e5b86f7053aa9bfba42594e1

swopebe (approved):

PR #7561, opened 2025-02-28
Implement Frank's comments for Edge Manager/Managing devices

amolnar-gh:


fzdarsky (approved):

edge_manager/edge_mgr_manage_devices_intro.adoc
```diff
@@ -1,17 +1,20 @@
...
+The device lifecycle also includes device management, such as organizing, monitoring, and updating your devices with the {rhem}.
+
 You can manage your devices individually or in a fleet.
-Managing your devices individually is advantageous in the following scenarios:
+The {rhem} allows you to manage a whole fleet of devices as a single object instead of managing many devices individually.
+You only need to specify the desired configuration once and then, the {rhem} applies the configuration to all devices in the fleet.
 
-* If a few devices have different configuration.
-* If you use external automation for updating the device.
+However, understanding individual device management is the foundation for managing devices in a fleet.
```
swopebe:
```suggestion
Understanding individual device management is the foundation for managing devices in a fleet.
```

edge_manager/edge_mgr_manage_devices_intro.adoc
```diff
@@ -23,4 +26,6 @@ TODO Add links to remaining sections
 * xref:../edge_manager/edge_mgr_manage_os.adoc#manage-os-config[Managing the operating system configuration]
 ////
 
+//To learn more about managing your devices in a fleet, see xref:../[]
+
 *Required access:* Cluster administrator
```
swopebe: We usually place required access at the top of the topics, so I will move that.

edge_manager/edge_mgr_manage_devices_intro.adoc
swopebe: We are encouraged, if not directed, to avoid "needless words" per IBM Style Manual, and I think we don't need `However,` in this case. I do think this line and the line after is important and you guys may want to consider making it part of the short description in line 4, but where it is located now is not "wrong."

swopebe (approved):

edge_manager/edge_mgr_manage_devices_intro.adoc
amolnar-gh: I'd leave that sentence there. This flow feels logical to me:
1) Short desc of device lifecycle and what's included.
2) Introduce concept of fleet vs individual device management. (+ Link to fleet management at the end as an advanced topic once we have the files)
3) Focus on when you'd want individual devices and introduce the sections for their management, which you need to understand before you start managing a fleet.

PR #7565, opened 2025-03-02
acm-18080 Fix links in 2.12

cadawson:


swopebe (approved):

swopebe (approved):
Merging to test

clusters/assisted_installer/ai_intro.adoc
```diff
@@ -9,7 +9,7 @@ The {ai} is an install method for {ocp-short} that uses agents to run pre-instal
 
 The {infra} is a {mce-short} component that manages and installs the workloads that run the {ai} service.
 
-You can use the console to create a host inventory, which is a pool of bare metal or virtual machines that you can use to create on-premises {ocp-short} clusters. These clusters can be standalone, with dedicated machines for the control plane, or link:../../clusters/hosted_control_planes/hosted_intro.adoc#hosted-control-planes-intro[hosted control planes], where the control plane runs as pods on a hub cluster.
+You can use the console to create a host inventory, which is a pool of bare metal or virtual machines that you can use to create on-premises {ocp-short} clusters. These clusters can be standalone, with dedicated machines for the control plane, or link:../clusters/hosted_control_planes/hosted_intro.adoc#hosted-control-planes-intro[hosted control planes], where the control plane runs as pods on a hub cluster.
```
swopebe, line 13: @cadawson hcp doc went to OCP and this should have been removed. (?) Not sure why it wasn't but we had an issue to remove all those and point to OCP.

clusters/install_upgrade/adv_config_install.adoc
```diff
@@ -179,7 +179,7 @@ spec:
 
 The previous infra-node toleration is set on pods by default without specifying any tolerations in the configuration. Customizing tolerations in the configuration will replace this default behavior.
 
-To define a list of tolerations for the {acm} hub cluster,, see link:../../install/adv_config_install.adoc#tolerations[tolerations] in the product documentation.
+To define a list of tolerations for the {acm} hub cluster, see link:../install/adv_config_install.adoc#tolerations[tolerations] in the product documentation.
```
swopebe, line 182: I don't even understand why this is in the adv config install for mce, but this was added by the developer in ACM 2.6, it seems--this book was not as defined. Anyway, these links seem to work, however they are point to ACM install topic, which is weird. (Something to address another time because now that MCE can standalone, they won't always have ACM.)

swopebe: So this doesn't make sense.

clusters/install_upgrade/install_intro.adoc
```diff
@@ -22,6 +22,6 @@ See the following documentation:
 * xref:../about/mce_networking.adoc#mce-network-configuration[Network configuration]
 * xref:../upgrade_cluster_disconnected_policies.adoc#upgrade-disconnect-cluster-policy[Upgrading disconnected clusters using policies]
 * xref:./adv_config_install.adoc#advanced-config-engine[MultiClusterEngine advanced configuration]
-* link:../../mce_acm_integration/mce_acm_intro.adoc#mce-acm-integration[{mce-short} with {acm-short} integration]
+* link:../mce_acm_integration/mce_acm_intro.adoc#mce-acm-integration[{mce-short} with {acm-short} integration]
```
swopebe, line 25: This link now gives a 404, so this is def. not the solution here.

PR #7578, opened 2025-03-04
https://issues.redhat.com/browse/ACM-18039

jc-berger:
Jira: https://issues.redhat.com/browse/ACM-18039


release_notes/acm_whats_new.adoc
```diff
@@ -71,6 +71,10 @@ See link:../observability/observe_environments_intro.adoc#observing-environments
 [#governance-whats-new]
 == Governance
 
+* `ValidatingAdmissionPolicyBindings` policy now appears in the `Discovered` policies tab in the Governance console. Additionally, you can also see other policy types that are in your managed clusters. Use the `ValidatingAdmissionPolicyBindings` policy to see details of the resources on your manged cluster, including resources that you can use to configure the behaviors of policies.
```
jc-berger: Am I right that this is a policy? Correct me if I'm wrong. 

Also, my understanding is that users use `ValidatingAdmissionPolicyBindings` to see details of the resources. Is that correct?

JustinKuli: The way it's phrased here makes it seem to me like we have a single policy that does something with ValidatingAdmissionPolicyBindings (like how we have a single policy to install the Compliance Operator, for example), but that isn't accurate.

`ValidatingAdmissionPolicy` is somewhat similar to a ConstraintTemplate in Gatekeeper - it defines parts of what will be a policy, but a VAP doesn't do anything on the cluster until a ValidatingAdmissionPolicyBinding is created. The VAPB provides some additional configuration.

So our table shows these ValidatingAdmissionPolicyBindings, because they are the resources that actually indicate an active policy on the cluster.

I would just say
```suggestion
* `ValidatingAdmissionPolicyBindings` now appear in the `Discovered` policies tab in the Governance console. Additionally, you can also see other policy types that are in your managed clusters. The `ValidatingAdmissionPolicyBindings` details include other resources on the cluster that you can use to configure its behavior.
```

oafischer (approved):

PR #7587, opened 2025-03-05
Add Field selectors to Edge Manager

amolnar-gh:


edge_manager/main.adoc
```diff
@@ -10,3 +10,4 @@ include::edge_mgr_enroll_devices.adoc[leveloffset=+3]
 include::edge_mgr_view_devices.adoc[leveloffset=+3]
 include::edge_mgr_label_devices.adoc[leveloffset=+3]
 include::edge_mgr_update_labels_on_devices.adoc[leveloffset=+3]
+include::edge_mgr_field_selector.adoc[leveloffset=+3]
```
fzdarsky: This makes "field selectors" appear on the same level as "enrolling devices" and "viewing devices", or? I feel it belongs "below" viewing devices, as it is a filtering mechanism when viewing devices, just like "label selectors" are.  Label selectors are more important, though, as labels are the primary means for users to organize their devices (or any other resource, really).

Off topic for this PR, but I'd love to find a way to make clear that organizing resources by label and viewing/filtering (by label or field) isn't limted to Devices, but also applies to Fleets and other resources.

amolnar-gh: See my suggestions [here](https://github.com/stolostron/rhacm-docs/pull/7587/commits/7a3a05a01736dc28a79c895dcd0ebbd94f4e7851). You're right, they're on the same level, I moved them under view devices.

I also highlighted that the different filtering methods are applicable to all resources, not just devices.

edge_manager/edge_mgr_field_selector.adoc
```diff
@@ -0,0 +1,250 @@
...
+|Resource Sync |`spec.repository`
+|===
+
+[#field-discovery]
+== Field discovery
+
+Some {rhem} resources might expose additional supported fields.
+You can discover the supported fields by using the `flightctl` command with the `--field-selector` option. 
+If you try to use an unsupported field, the error message lists the available supported fields.
+See the following example:
```
jc-berger: Since you have two examples, make it plural: 

```suggestion
See the following examples:
```

edge_manager/edge_mgr_field_selector.adoc
```diff
@@ -0,0 +1,250 @@
...
+`In`: Matches if the array overlaps with the specified values.
+
+`NotIn`: Matches if the array does not overlap with the specified values. `Exists`: Matches if the field is present. 
+
+`DoesNotExist`: Matches if the field is not present.
+
+|Array element
+|===
+
+*Note:* Using `Array[Index]` treats the element as the type defined for the array elements. For example string, timestamp, number, or boolean.
```
jc-berger: Monospace here? 

```suggestion
*Note:* Using `Array[Index]` treats the element as the type defined for the array elements. For example: `string`, `timestamp`, `number`, or `boolean`.
```

jc-berger (approved):
/lgtm

edge_manager/edge_mgr_field_selector.adoc
```diff
@@ -0,0 +1,250 @@
...
+`In`: Matches if the array overlaps with the specified values.
+
+`NotIn`: Matches if the array does not overlap with the specified values. `Exists`: Matches if the field is present. 
+
+`DoesNotExist`: Matches if the field is not present.
+
+|Array element
+|===
+
+*Note:* Using `Array[Index]` treats the element as the type defined for the array elements. For example: `string`, `timestamp`, `number`, or `boolean`.
```
asafbss: Please remove this. There is no field that supports it, and we are about to remove this functionality.

edge_manager/edge_mgr_field_selector.adoc
```diff
@@ -0,0 +1,250 @@
...
+flightctl get devices --field-selector 'metadata.owner=Fleet/pos-fleet, status.updated.status in (Unknown, OutOfDate)' -l 'region=us'
+----
+
+[#supported-operators]
+== Supported operators
+
+[%header,cols="2,1,4"]
+|===
+|Operator |Symbol |Description
+|Exists |`exists` |Checks if a field exists.
```
asafbss: The symbol is a mistake, and we need to fix this in our documentation as well.

To check if a field exists, we simply query the field name without any symbol or value, following the Kubernetes approach.

Example:
`--field-selector 'metadata.owner'`
This filters results to show only those that have an owner field.

asafbss: @avishayt  - CC

amolnar-gh: This is a bit tricky because it's just using the field you're filtering for, technically there's no symbol. I would add the example to the description for clarification. There are 3 options for the Symbol column:

* ```|Exists |None |Checks if a field exists. For example, the `--field-selector 'metadata.owner'` field selector returns resources that have the `metadata.owner` field.```

* ```|Exists |Not applicable |Checks if a field exists. For example, the `--field-selector 'metadata.owner'` field selector returns resources that have the `metadata.owner` field.```

* ```|Exists |`--field-selector <field>` |Checks if a field exists. For example, the `--field-selector 'metadata.owner'` field selector returns resources that have the `metadata.owner` field.```

I'd vote for the last option but it could be confusing that the `--field-selector` flag is included. As a fourth option, we could have only `<field>` and mak
e it clear with the example how to use it. WDYT?

asafbss: I vote for `<field>.`
@fzdarsky @avishayt please share your thoughts 🙏

avishayt: Yes, option 3 (<field>) makes the most sense to me

PR #7592, opened 2025-03-06
https://issues.redhat.com/browse/ACM-17496--Add Introduction to Edge Manager

amolnar-gh:


edge_manager/edge_mgr_arch.adoc
```diff
@@ -1,4 +1,4 @@
-[#edge-mgr-arch]
+[#overview]
```
swopebe:
```suggestion
[#rhem-overview]
```

edge_manager/edge_mgr_intro.adoc
```diff
@@ -1,4 +1,33 @@
 [#edge-mgr-intro]
-= Edge management
+= {rhem} overview
```
swopebe: hmmm, that is the overview up there in the first file, according to the anchor. and this anchor says "intro" -- normally we only have one intro file that contains the links to all the other doc, like site config. Why is the above file not the architecture anymore?

edge_manager/edge_mgr_intro.adoc
```diff
@@ -1,4 +1,33 @@
...
+** Secure decommissioning with complete device lifecycle management.
+This includes appropriate deregistration and credential revocation.
+
+* Scalable management:
+** Declarative configuration, enabling you to specify required states for operating systems, configuration files, and applications
+** Fleet-level administration, where you can apply policies and configurations to groups of devices rather than managing individually
+** Real-time status reporting that gives you visibility into device health, connectivity, and configuration compliance
+
+By using the {rhem} on {acm}, you can manage non-Kubernetes workloads and your operating system configuration on a Red Hat Enterprise Linux machine consistently with how you manage it on {ocp}.
+
```
swopebe, line 8: This goes back to the discussion we have about the intro topics being too long and pushing down the links, customers not being able to see the links without scrolling. Ideally the intro is just a couple of paragraphs. The architecture has the info about the technology. You can also do an "About" file if this content is better there and not in architecture.

edge_manager/edge_mgr_arch.adoc
swopebe: ok, NM, have some questions about this change. See comments below. :-)

edge_manager/edge_mgr_intro.adoc
amolnar-gh: It should be in the architecture, overview will stay there, sorry about that! However, do we want to keep the title as "Edge Management" or should it be "Red Hat Edge Manager" (`{rhem}`)?

edge_manager/edge_mgr_intro.adoc
```diff
@@ -1,4 +1,33 @@
 [#edge-mgr-intro]
-= Edge management
+= {rhem}
 
-Short description here
\ No newline at end of file
+The {rhem} provides streamlined management of edge devices and applications through a declarative approach. 
+By defining the required state of your edge devices, which includes your operating system versions, host configurations, and application deployments, the {rhem} automatically implements and maintains these configurations across your entire device fleet.
+
```
swopebe, line 6: We get customer feedback that the links in topics like this are buried in the portal. To prevent that, the intro files must have short description (2-3 lines, not much more) and then the links to the collection for that component. I have asked for this to be in an "about" file. (rhem_about.adoc)

swopebe (approved):

edge_manager/edge_mgr_arch.adoc
```diff
@@ -1,7 +1,33 @@
...
 You can manage individual devices or an entire fleet by using the {rhem}.
+The {rhem} provides the following benefits:
+
+* Built for edge environments: 
+** Manage distributed edge devices from a centralized platform
+** Enhanced security with hardware root-of-trust
+** Autonomous agent operation during network interruptions
+** Simplified management for locations with limited or no on-site IT personnel
+
+* Simple and secure life cycle management:
```
swopebe: I think lifecycle is one word, -- that is what we use.

edge_manager/edge_mgr_arch.adoc
```diff
@@ -1,7 +1,33 @@
-[#edge-mgr-arch]
-= {rhem} architecture
+[#about]
```
swopebe: This needs to remain the _architecture_  in the title or we need to rename the file to _about_, which I think is not going to be as favored by the engineers but I could be wrong. We can ask if we have time but I think keeping it architecture with features listed does make sense. 

We mentioned just having a new section here in the architecture. Let's try a different route from "About" and change to "Features" -- that may make more sense and keep the tone you are looking for in this file?

**RHEM Architecture** (title)
short description

Learn more from the following sections:

* <<architecture,{rhem} architecture>>
* <<rhem-features, rhem features>>
* <<agent-service,{rhem} agent and service>>
* <<api-server,{rhem} API server>>

 -- RHEM function (?) (or features)

swopebe (changes requested):

edge_manager/edge_mgr_arch.adoc
swopebe: Or we can stick with About, but either adding it as a section or changing the file name to keep what you have in this PR. I just think they will expect an architecture file.

swopebe: We can also just have an about file, like we mentioned before, but I think it can fit here, too and get read more (people don't read our about files) lol

amolnar-gh: I rearranged this a bit as I don't want to separate the architecture from the agent and service section. Like this:
RHEM Architecture (title)
short description

Learn more from the following sections:

<<rhem-features, rhem features>>
<<architecture,{rhem} architecture>>
<<agent-service,{rhem} agent and service>>
<<api-server,{rhem} API server>>


I also suggest to rename it to rhem benefits instead of features because we have a list for rhem features, such as agent, service, API server, device, device fleet, etc. PTAL at the new draft and let me know what you think!

oafischer (approved):
Looks like Brandi's comments were addressed, generally looks good to me but I would still wait for her final approval.

edge_manager/edge_mgr_arch.adoc
```diff
@@ -1,7 +1,38 @@
-[#edge-mgr-arch]
+[#architecture]
```
swopebe:
```suggestion
[#rhem-architecture]
```

edge_manager/main.adoc
```diff
@@ -1,6 +1,6 @@
 include::modules/common-attributes.adoc[]
 
-= Edge manager
+= Edge manager (Technology Preview)
```
swopebe: We cannot manipulate this file title because it is brought it as-is by Pantheon. I am pretty sure a while ago we tested this and it failed. We can put it on the intro and other level two titles of files.

swopebe (approved):
There many be some clean up to do but we should merge to work on any build error.

PR #7594, opened 2025-03-07
https://issues.redhat.com/browse/ACM-16247 moving KI from Observability to Console

ragevou:
For 2.13


swopebe (approved):
This is fine, I talked to Jorge in their scrum and we are fine to keep this here. With the constant change to the Search code and where it fits in the product, the doc has moved around a few times.

PR #7599, opened 2025-03-10
https://issues.redhat.com/browse/ACM-16305 Update health checks doc

oafischer:
Also for 2.12

Draft that still needs some editing

### DIFF
--- a/gitops/gitops_manage_policy_def.adoc
+++ b/gitops/gitops_manage_policy_def.adoc
@@ -224,7 +224,7 @@ wget https://raw.githubusercontent.com/open-cluster-management-io/policy-collect
 
 . Apply the `argocd-policy-healthchecks.yaml` policy by navigating to *Governance* > *Create policy* in the console and pasting the content into the YAML editor.
 +
-*Note:* You can add Placement information to determine which clusters have the policy active in the YAML editor.
+*Note:* You can add Placement information in the YAML editor to determine which clusters have the policy active.
 
 . Verify that the health checks work as expected by viewing the _Summary_ tab of the `ArgoCD` resource. View the health details from the Argo CD console.
 
--- a/gitops/gitops_manage_policy_def.adoc
+++ b/gitops/gitops_manage_policy_def.adoc
@@ -206,7 +206,7 @@ roleRef:
 [#config-gitops-healthcheck]
 == Configuring policy health checks in {gitops-short}
 
-You can use {gitops-short} with the `ArgoCD` resource to define a custom logic that sets the health status of a resource based on the resource state. For example, you can define custom health checks that only reports a policy as healthy if your policy is compliant.
+You can use {gitops-short} with the `ArgoCD` resource to define a custom logic that sets the health status of a resource based on the resource state. For example, you can define custom health checks that only report a policy as healthy if your policy is compliant.
 
 You must add health checks for resources to the `resourceHealthChecks` field. 
 
@@ -214,7 +214,7 @@ You must add health checks for resources to the `resourceHealthChecks` field.
 
 To define health checks for your resource kinds complete the following steps:
 
-. To add a health check to your `CertificatePolicy`, `ConfigurationPolicy`, `OperatorPolicy`, and `Policy` resources, download the `argocd-policy-healthchecks.yaml` by running the following command:
+. Add a health check to your `CertificatePolicy`, `ConfigurationPolicy`, `OperatorPolicy`, and `Policy` resources by downloading the `argocd-policy-healthchecks.yaml`. Run the following command:
 
 +
 [source,bash]
@@ -222,15 +222,9 @@ To define health checks for your resource kinds complete the following steps:
 wget https://raw.githubusercontent.com/open-cluster-management-io/policy-collection/main/stable/CM-Configuration-Management/argocd-policy-healthchecks.yaml
 ----
 
-. To apply the `argocd-policy-healthchecks.yaml` policy, run the following command:
-
+. Apply the `argocd-policy-healthchecks.yaml` policy by navigating to *Governance* > *Create policy* in the console and pasting the content into the YAML editor.
 +
-[source,bash]
-----
-oc apply -f ./argocd-policy-healthchecks.yaml
-----
-
-. To apply the `argocd-policy-healthchecks.yaml` policy, navigate to *Governance* > *Create policy*, and paste it into the YAML editor. There, you can add Placement information to determine which clusters have the policy active.
+*Note:* You can add Placement information to determine which clusters have the policy active in the YAML editor.
 
 . Verify that the health checks work as expected by viewing the _Summary_ tab of the `ArgoCD` resource. View the health details from the Argo CD console.
 
--- a/gitops/gitops_manage_policy_def.adoc
+++ b/gitops/gitops_manage_policy_def.adoc
@@ -222,7 +222,7 @@ To define health checks for your resource kinds complete the following steps:
 wget https://raw.githubusercontent.com/open-cluster-management-io/policy-collection/main/stable/CM-Configuration-Management/argocd-policy-healthchecks.yaml
 ----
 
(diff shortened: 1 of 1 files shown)


gitops/gitops_manage_policy_def.adoc
```diff
@@ -266,6 +230,8 @@ wget https://raw.githubusercontent.com/open-cluster-management-io/policy-collect
 oc apply -f ./argocd-policy-healthchecks.yaml
 ----
 
+. Add your Placement information in the console by navigating to *Governance* > *Create policy*.
```
JustinKuli: I was thinking using the *Create policy* wizard would replace the `oc apply` just above this. Like:
```suggestion
. To apply the `argocd-policy-healthchecks.yaml` policy, navigate to *Governance* > *Create policy*, and paste it into the YAML editor. There, you can add Placement information to determine which clusters have the policy active.
```

gitops/gitops_manage_policy_def.adoc
```diff
@@ -206,50 +206,14 @@ roleRef:
 [#config-gitops-healthcheck]
 == Configuring policy health checks in {gitops-short}
 
-With the `ArgoCD` resoure, use {gitops-short} for you to define custom logic that determines the current health of specific resource based on the resource state. Define custom health checks to report the policy as healthy only when your policy is compliant. When you add a health check for a resource, you must add it as a `group` in the `resourceHealthChecks` field. 
+You can use {gitops-short} with the `ArgoCD` resource to define a custom logic that sets the health status of a resource based on the resource state. For example, you can define custom health checks that only reports a policy as healthy if your policy is compliant.
+
+You must add health checks for resources as a `group` in the `resourceHealthChecks` field.
```
JustinKuli:
```suggestion
You must add health checks for resources to the `resourceHealthChecks` field. 
```

"as a group" sounds either like you need to add all of them together somewhere, or is referring to the top key in the object that is added... But neither of those is really accurate.

Background: when adding a healthcheck to the ArgoCD object, it's similar to defining environment variables in a Pod. In a Pod, it looks like:
```yaml
env:
  - name: FOO
    value: hello
  - name: BAR
    value: world
```
But for health checks here, the `name` part is split into a `group` and `kind`; the `value` is the `check`. So it looks like:
```yaml
resourceHealthChecks:
  - group: policy.open-cluster-management.io
    kind: ConfigurationPolicy
    check: "..."
  - group: policy.open-cluster-management.io
    kind: OperatorPolicy
    check: "..."
```

oafischer:
Thanks for the review @JustinKuli, I implemented your suggestions with a few edits. Let me know if it looks good now.

JustinKuli (approved):

gitops/gitops_manage_policy_def.adoc
```diff
@@ -206,65 +206,25 @@ roleRef:
 [#config-gitops-healthcheck]
 == Configuring policy health checks in {gitops-short}
 
-With the `ArgoCD` resoure, use {gitops-short} for you to define custom logic that determines the current health of specific resource based on the resource state. Define custom health checks to report the policy as healthy only when your policy is compliant. When you add a health check for a resource, you must add it as a `group` in the `resourceHealthChecks` field. 
+You can use {gitops-short} with the `ArgoCD` resource to define a custom logic that sets the health status of a resource based on the resource state. For example, you can define custom health checks that only report a policy as healthy if your policy is compliant.
```
jc-berger: Minor, optional changes to make the intro more concise. 

```suggestion
Use {gitops-short} with the `ArgoCD` resource to define a custom logic that sets the health status of a resource based on the resource state. For example, you can define custom health checks that only report a policy as healthy if your policy is compliant.
```

gitops/gitops_manage_policy_def.adoc
```diff
@@ -206,65 +206,25 @@ roleRef:
...
+
+You must add health checks for resources to the `resourceHealthChecks` field.
```
jc-berger: Should this be step 1 in the procedure? 

Or a prereq? 

If you do make this a prereq, I would put the *Important* note _before_ it so users have the important notice at the forefront of their mind before beginning any sort of action or task.

gitops/gitops_manage_policy_def.adoc
```diff
@@ -206,65 +206,25 @@ roleRef:
...
 
 +
 [source,bash]
 ----
 wget https://raw.githubusercontent.com/open-cluster-management-io/policy-collection/main/stable/CM-Configuration-Management/argocd-policy-healthchecks.yaml
 ----
 
-. To apply the `argocd-policy-healthchecks.yaml` policy, run the following command:
-
+. Apply the `argocd-policy-healthchecks.yaml` policy by navigating to *Governance* > *Create policy* in the console and pasting the content into the YAML editor.
```
jc-berger: Not a big deal, but I think IBM SG doesn't want us to use the word "navigate" which I never totally agreed with lol 

```suggestion
. Apply the `argocd-policy-healthchecks.yaml` policy by going to *Governance* > *Create policy* in the console and pasting the content into the YAML editor.
```

gitops/gitops_manage_policy_def.adoc
```diff
@@ -206,65 +206,25 @@ roleRef:
...
 +
-[source,bash]
-----
-oc apply -f ./argocd-policy-healthchecks.yaml
-----
+*Note:* You can add Placement information in the YAML editor to determine which clusters have the policy active.
```
jc-berger, line 225: Should `Placement` be monospaced? 

```suggestion
*Note:* You can add `Placement` information in the YAML editor to determine which clusters have the policy active.
```

jc-berger (approved):
/lgtm

gitops/gitops_manage_policy_def.adoc
oafischer: @JustinKuli What do you think here? Any preferences?

gitops/gitops_manage_policy_def.adoc
oafischer: Well guess I learned something new, we've been using navigating throughout all of the ACM docs ever since I started lol. But I'll happily switch it out, thanks for pointing it out.

gitops/gitops_manage_policy_def.adoc
oafischer: I think we are referring to the resource in general in this case, but @JustinKuli can confirm.

JustinKuli: I was thinking "Placement information" would mean both `Placement` resources and `PlacementBinding` resources - the wizard makes both at once and makes it easy for the user.

gitops/gitops_manage_policy_def.adoc
JustinKuli: I think we might be getting tripped up on the "You must add ..." (if we're talking about a different part, let me know)

The point of the statement is to clarify *where* in the `ArgoCD` resource the healthchecks go, because it is potentially a very large object with many fields.

gitops/gitops_manage_policy_def.adoc
oafischer: Thanks for clarifying, I'll just leave it as is then since that is how we use "Placement information" in the Placement docs as well.

gitops/gitops_manage_policy_def.adoc
oafischer: How about: `You can only add health checks for resources to the resourceHealthChecks field. `

Might be a little clearer.

JustinKuli: To me, that implies that the user can't make any other changes to the ArgoCD object. Which probably isn't true.

What if the statement was added to the end of the first sentence? Like: "You can use {gitops-short} with the `ArgoCD` resource to define a custom logic that sets the health status of a resource based on the resource state by adding to the `resourceHealthChecks` field. For example, you can define custom health checks that only report a policy as healthy if your policy is compliant."

oafischer: I like that, updated it!

JustinKuli (approved):

PR #7603, opened 2025-03-11
https://issues.redhat.com/browse/ACM-16281 CIM audit

oafischer:


clusters/release_notes/mce_known_issues.adoc
```diff
@@ -259,29 +251,7 @@ If the {ai} agent starts slowly and you deploy a managed cluster, the managed cl
...
-  ALLOW_CONVERGED_FLOW: "false" <1> 
-----
-<1> When you set the parameter value to "false", you also disable any features enabled by the Ironic Python Agent. 
-
-. Apply the ConfigMap by running the following command:
-+
-----
-oc annotate --overwrite AgentServiceConfig agent unsupported.agent-install.openshift.io/assisted-service-configmap=my-assisted-service-config
-----
+To decide what port numbers you want to enable or disable without disabling the converged flow, see link:../../clusters/about/mce_networking.adoc#mce-network-configuration[Network configuration].
```
carbonin: I think we should still removing anything referencing disabling the converged flow as a valid option.
If we want to keep this section at all I'd say just make it more about referencing the networking doc if the cluster is pending or agents are not registering rather than ever mentioning that converged flow can be disabled.

We could also just not mention "converged flow" at all. This is really just how we do provisioning with BMO now.

oafischer: I'll remove if it's better not to mention it at all

oafischer: Are you OK with just removing this entire known issue?

carbonin (approved):

swopebe (approved):

PR #7615, opened 2025-03-12
Governance link goes to Splash page, changing link to xref (internal…

swopebe:
… link) to test.


swopebe:
if you go here: https://content-stage.docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.13/html-single/multicluster_engine_operator_with_red_hat_advanced_cluster_management/index#autoimport-disc-prereq

this is the mce-acm-integration folder with another folder in it. The link goes to the Splash page. Testing this to see if it is the fix. Only one at a time for now.

PR #7621, opened 2025-03-14
https://issues.redhat.com/browse/ACM-18680 Subm KI

oafischer:
2.13 only

### DIFF
--- a/release_notes/known_issues_network.adoc
+++ b/release_notes/known_issues_network.adoc
@@ -32,7 +32,7 @@ For more about deprecations and removals, see xref:../release_notes/acm_deprecat
 
 See the following known issues and limitations that might occur while using networking features.
 
-[#submariner-uninstall-before-move]
+[#source-ip-ocp-ovnk]
 === Source IP not retained for applications on {ocp-short} 4.18 with OVN-Kubernetes
 //2.13:ACM-18680
 
--- a/release_notes/known_issues_network.adoc
+++ b/release_notes/known_issues_network.adoc
@@ -36,7 +36,7 @@ See the following known issues and limitations that might occur while using netw
 === Source IP not retained for applications on {ocp-short} 4.18 with OVN-Kubernetes
 //2.13:ACM-18680
 
-If you are using {ocp-short} 4.18 with OVN-Kubernetes, the source IP is not retained when packets reach the destination pod. As a result, applications that rely on the source IP, such as `NetworkPolicy`, might not work correctly.
+If you are using {ocp-short} 4.18 with OVN-Kubernetes for Submariner, the source IP is not retained when packets reach the destination pod. As a result, applications that rely on the source IP, such as `NetworkPolicy`, might not work correctly.
 
 [#cluster-management-addon-needed]
 === Without _ClusterManagementAddon_ submariner add-on fails
--- a/release_notes/known_issues_network.adoc
+++ b/release_notes/known_issues_network.adoc
@@ -36,7 +36,7 @@ See the following known issues and limitations that might occur while using netw
 === Source IP not retained for applications on {ocp-short} 4.18 with OVN-Kubernetes
 //2.13:ACM-18680
 
-If you are using {ocp-short} 4.18 with OVN-Kubernetes, the source IP is not retained when packets reach the destination pod. As a result, applications that rely on the source IP, such as NetworkPolicy, might not work correctly.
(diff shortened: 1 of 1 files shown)


yboaron (approved):

release_notes/known_issues_network.adoc
```diff
@@ -32,6 +32,12 @@ For more about deprecations and removals, see xref:../release_notes/acm_deprecat
 
 See the following known issues and limitations that might occur while using networking features.
 
+[#source-ip-ocp-ovnk]
+=== Source IP not retained for applications on {ocp-short} 4.18 with OVN-Kubernetes
+//2.13:ACM-18680
+
+If you are using {ocp-short} 4.18 with OVN-Kubernetes, the source IP is not retained when packets reach the destination pod. As a result, applications that rely on the source IP, such as NetworkPolicy, might not work correctly.
+
```
swopebe, line 40: I have a tiny concern on this one. This is (from what I am reading) an _OCP_ known issue. 

If that is the case, remember that this should not be contained in our doc. We write in the known issues doc that we do not maintain OCP known issues and we tell the teams this, as well.

If this is in fact an _ACM_ known issue, we need to write in the known issue from the ACM side, meaning:

As a result, {acm-short} applications...

@yboaron can you provide some insight on how to make this more ACM-centric?

swopebe: Did we check on official way to write _OVN-Kubernetes_? I am not sure, that is why I ask.

release_notes/known_issues_network.adoc
```diff
@@ -32,6 +32,12 @@ For more about deprecations and removals, see xref:../release_notes/acm_deprecat
 
 See the following known issues and limitations that might occur while using networking features.
 
+[#source-ip-ocp-ovnk]
+=== Source IP not retained for applications on {ocp-short} 4.18 with OVN-Kubernetes
+//2.13:ACM-18680
+
+If you are using {ocp-short} 4.18 with OVN-Kubernetes, the source IP is not retained when packets reach the destination pod. As a result, applications that rely on the source IP, such as NetworkPolicy, might not work correctly.
```
swopebe:
```suggestion
If you are using {ocp-short} 4.18 with OVN-Kubernetes, the source IP is not retained when packets reach the destination pod. As a result, applications that rely on the source IP, such as `NetworkPolicy`, might not work correctly.
```

swopebe (changes requested):

release_notes/known_issues_network.adoc
oafischer: We don't have guidance but Red Hat uses (OVN)-Kubernetes to describe other entries. I went with what the project itself uses, which is OVN-Kubernetes. I'll wait and see what @yboaron will say about the other concern.

swopebe: Ok, I will approve, however, if this is an OCP-specific issue, we should not document those--they should and we point to their known issues in the intro.

swopebe (approved):

release_notes/known_issues_network.adoc
yboaron: Will try to summarize: 

1. OCP with OVNK always SNAT (changed the source IP address)
2. Applications running on top of Submariner may be sensitive to Source IP change
3. Until OCP 4.18 Submariner overrides this behavior and prevents source IP change 
4. In OCP 4.18, OVNK changed its internal implementation, and currently, Submariner cannot prevent SNAT. 

So, it is more like a Submariner issue

release_notes/known_issues_network.adoc
```diff
@@ -32,6 +32,12 @@ For more about deprecations and removals, see xref:../release_notes/acm_deprecat
 
 See the following known issues and limitations that might occur while using networking features.
 
+[#source-ip-ocp-ovnk]
+=== Source IP not retained for applications on {ocp-short} 4.18 with OVN-Kubernetes
+//2.13:ACM-18680
+
+If you are using {ocp-short} 4.18 with OVN-Kubernetes, the source IP is not retained when packets reach the destination pod. As a result, applications that rely on the source IP, such as `NetworkPolicy`, might not work correctly.
```
swopebe:
```suggestion
If you are using {ocp-short} 4.18 with OVN-Kubernetes for Submariner, the source IP is not retained when packets reach the destination pod. As a result, applications that rely on the source IP, such as `NetworkPolicy`, might not work correctly.
```

swopebe (approved):

PR #7622, opened 2025-03-14
Update trouble_global_hub_access_postgres.adoc discovered while looki…

swopebe:
…ng for examples of good t-shooting that this file is not following standards/format


swopebe:
closing to save comments for a tech debt issue.

PR #7639, opened 2025-03-17
https://issues.redhat.com/browse/ACM-18932--anchors, links, titles (out of sync and such)

swopebe:
… mess

### DIFF
--- a/edge_manager/edge_mgr_arch.adoc
+++ b/edge_manager/edge_mgr_arch.adoc
@@ -1,22 +1,13 @@
 [#rhem-architecture]
 = {rhem} architecture
 
-You can manage individual devices or an entire fleet by using the {rhem}.
+You can manage individual devices or an entire fleet by using the {rhem}. The {rhem} uses an agent-based architecture that allows for a scalable and robust device management, even with limited network conditions.
 
-Learn more from the following sections:
-
-* <<architecture,{rhem} architecture>>
-* <<agent-service,{rhem} agent and service>>
-* <<api-server,{rhem} API server>>
-
-[#rhem-architecture-con]
-== {rhem} architecture
-
-The {rhem} uses an agent-based architecture that allows for a scalable and robust device management, even with limited network conditions.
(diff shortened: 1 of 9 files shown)


edge_manager/edge_mgr_arch.adoc
```diff
@@ -1,22 +1,13 @@
 [#rhem-architecture]
 = {rhem} architecture
 
-You can manage individual devices or an entire fleet by using the {rhem}.
+You can manage individual devices or an entire fleet by using the {rhem}. The {rhem} uses an agent-based architecture that allows for a scalable and robust device management, even with limited network conditions.
```
swopebe, line 5: all I did was move this around because we had an anchor for architecture but that is the file and title, we don't need that extra header repeating the file name.

edge_manager/edge_mgr_images_special_considerations.adoc
```diff
@@ -1,14 +1,15 @@
-[#edge-images-special-considerations]
+[build-images-consider]
 = Special considerations for building images
+//let's make life easier on ourselves and reduce this file name and anchor...I will do filename when the build is up -- brandi 3/17
```
swopebe, line 3: will address this part later.

swopebe:
Merging bc I need to.

edge_manager/edge_mgr_arch.adoc
```diff
@@ -29,6 +20,11 @@ The {rhem} architecture has the following main features:
 * Device
 * Device fleet
 
+Learn more details from the following sections:
+
+* <<agent-service,{rhem} agent and service>>
+* <<api-server,{rhem} API server>>
+
 [#agent-service]
 === {rhem} agent and service
```
cadawson, line 29: Looks like this needs to be and ==, rather than an ===. Also check the API server section later. If you change this one, it will be a subsection of this one. If that's not what you want (which I don't think it is), then it should be ==, as well.

cadawson (commented):
Just a couple of heading-level questions. Looks good. Thanks!

edge_manager/edge_mgr_arch.adoc
swopebe: Yes! TY! That would explain the new error lol

PR #7673, opened 2025-03-28
Merge when ready: notes for peer review with writer/meeting

swopebe:


swopebe:
proposing slight restructure certificates, tagging, titles, headers, so we can clearly know what is hub, managed cluster, managed certificate.

Proposing that we don't need cert intro, add it to the larger intro.

jc-berger (approved):
/lgtm

PR #7686, opened 2025-04-03
https://issues.redhat.com/browse/ACM-19151--upgrade, Obs.

swopebe:


release_notes/known_issues_install.adoc
```diff
@@ -28,6 +28,61 @@ For your {ocp} cluster, see link:https://docs.redhat.com/documentation/en-us/ope
...
+
+For installations without the Observability component enabled, use the following workaround to resolve the error:
+
+*Important:* If the Observability component is installed on the cluster, the command causes the the Observability stack to be deleted. No known workaround exists for an upgrade with Observability enabled.
+
++
+[source,bash]
+----
+oc delete crd multiclusterobservabilities.observability.open-cluster-management.io
+----
```
swopebe, line 66: I am not sure what the actual workaround is. We give them a command to seemingly delete obs. and we tell them to use a work around if Obs is not enabled...but I see no workaround. (?)

Is the workaround just to delete obs?

And if so, then without it the user can upgrade with no error?

jacobbaungard: @swopebe , this command is a valid workaround for customers that don't use Observability.

To explain the general problem a bit more, this object, `multiclusterobservabilities.observability.open-cluster-management.io`, of the type CRD, is not able to be "re-created"/replaced due to a change in OCP4.18. This re-creation happens during upgrades of ACM and upgrade process therefore gets stuck with this error.

The object definition can however be created correctly from scratch, so when we delete the object in the command supplied there, the ACM upgrade process should quite soon after start up again and will successfully be able to create the object from scratch, progress and finish the update. This is not an object that customers change, so deleting it and creating it again is not a problem if we look at this object specifically.

The problem for Observability customers is that deleting this object has a side effect in that it deletes all Observability resources if any are created. That is big problem for customers that has Observability enabled, as it is not easy to just re-create the Observability stack without dataloss etc.

----

The bug will be fixed in ACM 2.13.2/ACM2.12.3, so that the object definition we ship will be able to be re-created correctly.

We've also received confirmation from QE, that if you end up in the stuck state, a valid workaround, is to update to a version with the fix included. This will allow the upgrade to process without any bad sideeffect. I wonder if it's worth to document this as well (even though those versions are not yet out) ?

Until ACM 2.13.2/ACM2.12.3 is out we should recommend that customers do not perform any ACM upgrades on OCP 4.18.

swopebe: Ok, yeah, not sure I am understanding that. Especially with the draft in the issue, but it seems **this is not an issue with observability? but a general upgrade, without Obs enabled?** I have asked on Slack--not sure if my question is not clear. But I have not gotten a clear answer.

This is the draft content:
```

[#local-cluster-upgrade-unknown]
== Upgrading {acm-short} with {ocp-short} 4.18 error installing CRDs
//2.13:19151

If you upgrade {acm-short} with with {ocp-short} 4.18, your upgrade fails because of an error installing the CRDs.
```

Why are we not just telling the user that they cannot upgrade with Observability enabled? Or if we just say don't upgrade on 4.18, that is simpler, but that is not what I got from the draft.

swopebe:
I did the best I can do but between the draft and what is in the comment above, I think it's even more confusing. I don't even know what we are trying to tell the user the workaround is--I cannot go through all the Slack threads. Sending it to install team for review, though.

1. The draft says "CRDs" but the rest of this explanation is about Observability.
2. The workaround is just a command. It seems it needs more but that is all I have in the draft. I need the draft to say what the work around is.

release_notes/known_issues_install.adoc
```diff
@@ -28,6 +28,62 @@ For your {ocp} cluster, see link:https://docs.redhat.com/documentation/en-us/ope
...
+//I am confused about why we are saying the command is a workaround--if that is true, we should tell them what this is doing, I think?
+
+Run the following command to check the progress of the {acm-short} upgrade after you enable the workaround.
```
swopebe: What work around?

release_notes/known_issues_install.adoc
```diff
@@ -28,6 +28,62 @@ For your {ocp} cluster, see link:https://docs.redhat.com/documentation/en-us/ope
 For more about deprecations and removals, see xref:../release_notes/acm_deprecate_remove.adoc#deprecations-removals-acm[Deprecations and removals for {acm-short}].
 
 [#local-cluster-upgrade-unknown]
+== Upgrading {acm-short} with {ocp-short} 4.18 error installing CRDs
```
swopebe: CRDs implies multiple. Yet, the draft talks about one.

release_notes/known_issues_install.adoc
```diff
@@ -28,6 +28,62 @@ For your {ocp} cluster, see link:https://docs.redhat.com/documentation/en-us/ope
...
+//2.13:19151
+
+If you upgrade {acm-short} with with {ocp-short} 4.18, your upgrade fails with an error from installing the CRDs.
+
```
swopebe, line 35: all crds?

release_notes/known_issues_install.adoc
```diff
@@ -28,6 +28,62 @@ For your {ocp} cluster, see link:https://docs.redhat.com/documentation/en-us/ope
...
+----
+message: 'Error installing CRDs: CustomResourceDefinition.apiextensions.k8s.io
+"multiclusterobservabilities.observability.open-cluster-management.io" is 
+invalid: spec.conversion.webhookClientConfig.caBundle: Invalid value: []byte{0x5c,0x6e}: unable to load root certificates: unable to parse bytes as PEM block'
+ reason: FailedDeployingComponent
+ status: "False"
+ type: Progressing
+desiredVersion: 2.13.0
+phase: Installing
+----
```
swopebe, line 58: for just obs? but we says CRDs?

release_notes/known_issues_install.adoc
```diff
@@ -28,6 +28,62 @@ For your {ocp} cluster, see link:https://docs.redhat.com/documentation/en-us/ope
...
+
+For installations without the Observability component enabled, use the following workaround to resolve the error:
+//really confused here--need to connect with engineer
+
+*Important:* If the Observability component is installed on the cluster, the command causes the the Observability stack to be deleted. No known workaround exists for an upgrade with Observability enabled.
```
swopebe: I changed this.

release_notes/known_issues_install.adoc
dislbenn: The only CRD that we've detected that was impacted is `multiclusterobservabilities.observability.open-cluster-management.io`; therefore, it is only one CRD.

release_notes/known_issues_install.adoc
dislbenn: Only the Observability CRD: `multiclusterobservabilities.observability.open-cluster-management.io`.

release_notes/known_issues_install.adoc
dislbenn: Correct, this is only for Observability.

release_notes/known_issues_install.adoc
swopebe: changed to a CRD, to still keep the title short.

release_notes/known_issues_install.adoc
swopebe: changed!

release_notes/known_issues_install.adoc
swopebe: I have more info now.

swopebe: 1. One thing that was confusing me that I understand now, 2.13.2 is the fix for this, but we cannot tell them that. So we have to say no current fix with Obs. enabled. We cannot preannounce in the doc.

2. The upgrade issue is with one resource.

3. I am still trying to understand and decide if we want to talk about a fresh install.

4. do not upgrade is not a general recommendation, but one for users with Obs. that they don't want to lose.

(I keep seeing _tell them not to upgrade,_ in general, but we are only concerned about those with Obs.) Correct?

release_notes/known_issues_install.adoc
swopebe: Now added info for existing vs new install

jc-berger (approved):
/lgtm

dislbenn (approved):
LGTM

release_notes/known_issues_install.adoc
```diff
@@ -28,6 +28,77 @@ For your {ocp} cluster, see link:https://docs.redhat.com/documentation/en-us/ope
...
+[source,bash]
+----
+oc delete crd multiclusterobservabilities.observability.open-cluster-management.io
+----
++
+*Important:* 
++
+- If the Observability component is `enabled` on the cluster, the previous command deletes the entire Observability stack. 
++
+- No known workaround exists for an upgrade with Observability enabled. You can unistall {acm-short} completely and reinstall without any CRDs without this failure.
```
jacobbaungard, line 72:
```suggestion
- No known workaround exists for an upgrade with Observability enabled. You can uninstall {acm-short} completely and reinstall without any CRDs without this failure.
```

Typo, fyi @swopebe

PR #7687, opened 2025-04-03
https://issues.redhat.com/browse/ACM-17885 Clarify external metrics ROSA

oafischer:
Also for 2.12, 2.11 and 2.10

### DIFF
--- a/observability/use_observability.adoc
+++ b/observability/use_observability.adoc
@@ -64,7 +64,7 @@ curl --cacert ./ca.crt -H "Authorization: Bearer {TOKEN}" https://{PROXY_ROUTE_U
 +
 *Note:* The `QUERY_EXPRESSION` is the standard Prometheus query expression. For example, query the metrics `cluster_infrastructure_provider` by replacing the URL in the previous command with the following URL: `https://{PROXY_ROUTE_URL}/api/v1/query?query=cluster_infrastructure_provider`. For more details, see link:https://prometheus.io/docs/prometheus/latest/querying/basics/[Querying Prometheus].
 
-. If you want to configure custom certificates for the `rbac-query-proxy` route, see link:../secure_cluster/cert_manage.adoc#replacing-certificates-for-rbac-query-proxy-route[Replacing certificates for rbac-query-proxy route].
+. If you want to configure custom certificates for the `rbac-query-proxy` route, see link:../secure_cluster/cert_manage.adoc#replacing-certificates-for-rbac-query-proxy-route[Replacing certificates for _rbac-query-proxy_ route].
 
 [#exporting-metrics-to-external-endpoints]
 === Exporting metrics to external endpoints
--- a/observability/use_observability.adoc
+++ b/observability/use_observability.adoc
@@ -38,7 +38,7 @@ MY_TOKEN=$(oc whoami --show-token)
 oc -n openshift-ingress get secret router-certs-default -o jsonpath="{.data.tls\.crt}" | base64 -d > ca.crt
 ----
 +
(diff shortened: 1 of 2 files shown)


secure_cluster/cert_manage.adoc
```diff
@@ -61,6 +61,40 @@ oc -n open-cluster-management-observability create secret tls alertmanager-byo-c
 oc -n open-cluster-management-observability create secret tls alertmanager-byo-cert --cert ./ingress.crt --key ./ingress.key
 ----
 
+[#replace-cert-rbac]
+== Replacing certificates for _alertmanager_ route
```
subbarao-meduri: The title seems to be wrong? I now see two titles "Replacing certificates for alertmanager route" in the branch.

secure_cluster/cert_manage.adoc
```diff
@@ -61,6 +61,40 @@ oc -n open-cluster-management-observability create secret tls alertmanager-byo-c
...
+
+You can replace certificates for the `rbac-query-proxy` route. See xref:./secure_cluster/cert_byo.adoc#openssl-commands-ca[Generating CA certificates by using OpenSSL commands] to create certificates.
```
subbarao-meduri: @oafischer is the hyperlink supposed to work? It was not working when I access the branch here:
https://github.com/stolostron/rhacm-docs/blob/of-17885/secure_cluster/cert_manage.adoc

subbarao-meduri (approved):
/lgtm

observability/use_observability.adoc
```diff
@@ -37,13 +37,21 @@ MY_TOKEN=$(oc whoami --show-token)
 ----
 oc -n openshift-ingress get secret router-certs-default -o jsonpath="{.data.tls\.crt}" | base64 -d > ca.crt
 ----
-
-. To access an endpoint of the default CA certificate from your hub cluster through mutual TLS, which verifies the identities of both parties in a network connection, create the `proxy-byo-ca` TLS secret with the following command:
-
++
+*Note:* The `router-certs-default` secret does not exist if your hub cluster is running on OpenShift Service on AWS. Instead, use the CA certificate that `spec.defaultCertificate.name` points to in the default ingress controller object and store the content of the `tls.crt` key in a local file. Complete the following steps:
```
jc-berger: The sentence is a little long. I don't think we need the "and" to help with comprehension. What do you think of this? 

```suggestion
*Note:* The `router-certs-default` secret does not exist if your hub cluster is running on OpenShift Service on AWS. Instead, use the CA certificate that `spec.defaultCertificate.name` points to in the default ingress controller object. Store the content of the `tls.crt` key in a local file. Complete the following steps:
```

secure_cluster/cert_manage.adoc
```diff
@@ -61,6 +61,40 @@ oc -n open-cluster-management-observability create secret tls alertmanager-byo-c
 oc -n open-cluster-management-observability create secret tls alertmanager-byo-cert --cert ./ingress.crt --key ./ingress.key
 ----
 
+[#replace-cert-rbac]
+== Replacing certificates for _rbac-query-proxy_ route
+
+You can replace certificates for the `rbac-query-proxy` route. See xref:../secure_cluster/cert_byo.adoc#openssl-commands-ca[Generating CA certificates by using OpenSSL commands] to create certificates.
```
jc-berger, line 67: Should we have a link this early in a procedure/task? Seems to me that it's likely a user would go to this link, Generating CA certs, then not complete this awesome procedure you have. 

What do you think about taking this xref and either putting it in a prereq (only if users need to generate a CA cert) or under Additional resources for users to continue to explore and learn?

jc-berger (approved):
/lgtm

secure_cluster/cert_manage.adoc
oafischer: I'll leave it as is for now because this is old doc that we are bringing back 1:1 that was removed for some reason. I'll trust that Mikela added the inline link because it was necessary.

PR #7688, opened 2025-04-04
https://issues.redhat.com/browse/ACM-16008 Add default metrics to doc

oafischer:
Also for 2.12 and 2.11

### DIFF
--- a/observability/obs_config.adoc
+++ b/observability/obs_config.adoc
@@ -87,6 +87,9 @@ View the following table of metric types that are supported by the framework:
 | Stable. See the _Search components_ section in the _Searching in the console_ documentation.
 |===
 
+[#obs-default-metrics]
+== Default metrics
+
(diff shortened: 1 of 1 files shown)


oafischer:
Approved by Christian in DM

philipgough (approved):
/lgtm

observability/obs_config.adoc
```diff
@@ -87,6 +87,18 @@ View the following table of metric types that are supported by the framework:
 | Stable. See the _Search components_ section in the _Searching in the console_ documentation.
 |===
 
+[#obs-default-metrics]
+== Default metrics
```
swopebe, line 91: One of those items that doesn't need a header but we have to get in where we fit in for now. I can see how things got this far in this area. It will be much easier when we fix it up a bit. TY!

swopebe (approved):

PR #7710, opened 2025-04-09
ACM-19800: DO NOT MERGE: modularization demo (emily)

emmurphy1:
These files illustrate changes to parts of the installation document that make files modularized in place.

Epic: https://issues.redhat.com/browse/ACM-19800#

### DIFF
--- a/install/adv_config_install.adoc
+++ b/install/adv_config_install.adoc
@@ -1,9 +1,7 @@
 [#advanced-config-hub]
-= _MultiClusterHub_ advanced configuration 
+= MultiClusterHub advanced configuration 
 
-{acm} is installed by using the `MultiClusterHub` operator, which deploys all of the required components. Some of the listed components are enabled by default. If a component is _disabled_, that resource is not deployed to the cluster until it is enabled. 
-
-The operator works to deploy the following components:
+{acm} is installed by using an operator that deploys all of the required components. Some of the listed components are enabled by default. If a component is _disabled_, that resource is not deployed to the cluster until it is enabled. The operator works to deploy the following components:
 
 .Table list of the deployed components
 |===
@@ -153,7 +151,7 @@ spec:
 
 The previous infra-node toleration is set on pods by default without specifying any tolerations in the configuration. Customizing tolerations in the configuration replaces this default.
 
-To define a list of tolerations for the {mce-short} hub cluster, see link:../clusters/install_upgrade/adv_config_install.adoc#tolerations[tolerations] in the {mce-short} documentation.
+To define a list of tolerations for the {mce-short} hub cluster, see link:../clusters/install_upgrade/adv_config_install.adoc#tolerations[tolreations] in the {mce-short} documentation.
 
 [#disable-hub-self-management]
 == disableHubSelfManagement
--- a/install/config_infra_nodes_acm.adoc
+++ b/install/config_infra_nodes_acm.adoc
@@ -1,3 +1,4 @@
+:_mod-docs-content-type: CONCEPT
 [#config-infra-node-acm]
 = Configuring infrastructure nodes for {acm-short}
 
@@ -5,13 +6,14 @@ Configure your {ocp-short} cluster to contain infrastructure nodes to run approv
 
 After adding infrastructure nodes to your {ocp-short} cluster, follow the xref:../install/install_connected.adoc#installing-from-the-cli[Installing from the {ocp-short} CLI] instructions and add configurations to the {olm} subscription and `MultiClusterHub` custom resource.
 
+:_mod-docs-content-type: PROCEDURE
 [#config-infra-nodes-ocp]
 == Configuring infrastructure nodes to the {ocp-short} cluster
 
 Follow the procedures that are described in link:https://docs.redhat.com/documentation/en-us/openshift_container_platform/4.16/html/machine_management/creating-infrastructure-machinesets[Creating infrastructure machine sets] in the {ocp-short} documentation. Infrastructure nodes are configured with a Kubernetes `taints` and `labels` to keep non-management workloads from running on them.
 
+.Procedure
 . To be compatible with the infrastructure node enablement provided by {acm-short}, ensure your infrastructure nodes have the following `taints` and `labels` applied:
-
 +
 [source,yaml]
 ----
@@ -25,7 +27,6 @@ spec:
 ----
 
 . Add the following additional configuration before applying the {olm} Subscription:
-
 +
 [source,yaml]
 ----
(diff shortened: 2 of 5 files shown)


install/adv_config_install.adoc
```diff
@@ -1,9 +1,7 @@
 [#advanced-config-hub]
-= _MultiClusterHub_ advanced configuration
```
swopebe, line 2: Line 2: 

This highlighting we have for translation, that word can stump a machine or translation focal--offsetting it tells them it is not an English word. (I know we don't handle it here, but I have lead translation and that is why we offset resources like that.) But always it's been a pain to italicize vs code. 

Can we change this to match the highlighting below:

`MultiClusterHub`

install/install_connected.adoc
```diff
@@ -28,14 +30,14 @@ For information on how to configure advanced configurations, see options in the
 
 [#connect-prerequisites]
 == Prerequisites
-
+//Brandi, I tried to build without the ID and the build failed because the ID is used as an XREF. I need to check with Jaromir about the extra ID in a module or assembly.
```
swopebe, line 33: @emmurphy1 I actually do not like us pointing to sections within the files for xrefs, especially prereqs. If this is the case, I would rather we re-examine that xref, which can likely just point to the file itself to move forward with your recommendation here.

install/install_connected.adoc
```diff
@@ -51,13 +53,15 @@ Before you install {acm-short}, see the following requirements:
...
 ** To import, you must use the `stable-2.0` channel of the klusterlet operator for {product-version}.
 
+:_mod-docs-content-type: PROCEDURE
 [#confirm-ocp-installation]
 == Confirm your {ocp-short} installation
 
 Verify that a {acm-short} hub cluster is not already installed on your {ocp-short} cluster. You cannot have more than one hub cluster. 
 
-You can only have one single {acm-short} hub cluster installation on each {ocp-short} cluster. Continue with the following steps if there is no {acm-short} hub cluster installed:
+You can only have one single {acm-short} hub cluster installation on each {ocp-short} cluster. Continue with the following steps if there is no {acm-short} hub cluster installed.
```
swopebe, line 63: colon bc of _following_

install/install_connected.adoc
```diff
@@ -78,18 +82,21 @@ console   https   reencrypt/Redirect     None
 
 . Open the URL in your browser and check the result. If the console URL displays `console-openshift-console.router.default.svc.cluster.local`, set the value for `openshift_master_default_subdomain` when you install {ocp-short}. See the following example of a URL: `https://console-openshift-console.apps.new-coral.purple-chesterfield.com`.
 
+.Next steps
```
swopebe, line 85: These two options that the user needs to complete installation. The line before it is to say they can choose between the CLI install or the console install, they have to choose one to complete this procedure. Next steps doesn't apply to this but I can see where it was read like that.

install/install_connected.adoc
```diff
@@ -138,13 +145,20 @@ metadata:
...
 
+:_mod-docs-content-type: PROCEDURE
 [#installing-from-the-cli]
 == Installing from the {ocp-short} CLI
 
-Install the operator and the objects. Complete the following steps:
+Install the operator and the objects. Complete the following steps.
+
+.Prequisites
+* one
```
swopebe, line 158: these are addressed above in the intro, as this is all part of installing the MCH. So I think removing headers here is probably best so that is clear and we don't lose those prereqs.

install/install_connected.adoc
swopebe: But we have a few "deep links" that may come up--some writers add them. We should not be pointing to prereqs only though. We can either rewrite them, or point to the introductions, which will contain them.

install/config_infra_nodes_acm.adoc
```diff
@@ -1,17 +1,19 @@
+:_mod-docs-content-type: CONCEPT
```
swopebe, line 1: I think what is confusing me here:

This is a task (or what CCS calls procedure) Every file has introductions between headers. That is what is line 2-8 (intro information for the task)

install/config_infra_nodes_acm.adoc
```diff
@@ -1,17 +1,19 @@
...
 [#config-infra-node-acm]
 = Configuring infrastructure nodes for {acm-short}
 
 Configure your {ocp-short} cluster to contain infrastructure nodes to run approved {acm-short} management components. Running components on infrastructure nodes avoids allocating {ocp-short} subscription quota for the nodes that are running {acm-short} management components.
 
 After adding infrastructure nodes to your {ocp-short} cluster, follow the xref:../install/install_connected.adoc#installing-from-the-cli[Installing from the {ocp-short} CLI] instructions and add configurations to the {olm} subscription and `MultiClusterHub` custom resource.
 
+:_mod-docs-content-type: PROCEDURE
 [#config-infra-nodes-ocp]
```
swopebe, line 10: **Configuring infrastructure nodes to the {ocp-short} cluster** is the procedure and reason for the file. It seems because part of it is in header 2, that is not clear internally? So can we just remove that header, then. Otherwise we have a CONCEPT that is only a short description and intro content for the procedure.

install/adv_config_install.adoc
emmurphy1: AFAIK, that would be fine.

install/install_connected.adoc
emmurphy1: OK. Either way works. I checked with Jaromir and he says it's fine to include additional IDs in a module or assembly as long as you don't include two ID for the same section. You have leave the Prerequisites section as it is. If we add the ASSEMBLY content type above the assembly ID Jaromir's converter will consider everything that follows to be part of an assembly/topic map. Just a reminder that his converter tool is expecting individual assembly or module files so won't work with the single file as is.

install/install_connected.adoc
emmurphy1: Sure. In this case you might have to do a bit of a rewrite. Here is the guidance from the template:
Only the following block headings can be reliably mapped to DITA: Prerequisites, Procedure, Verification, Troubleshooting, Troubleshooting steps, Next steps, Next step, Additional resources. They must appear in this order and, with the exception of Additional resources, are only allowed in a procedure module. You can also use block headings in figure, table, and example titles.

install/install_connected.adoc
emmurphy1: Yep. I was just showing you how .Prerequisites looks in the procedure module.

install/config_infra_nodes_acm.adoc
emmurphy1: Yes, that's fine. I actually did that locally and pushed but I don't see the changes in GitHub. Maybe because the PR is closed? But you have the idea. There is always more than one way to skin a cat and as long a your modules/assemblies contain all required template elements and don't include elements that aren't supported by DITA, you will be fine.

PR #7719, opened 2025-04-10
https://issues.redhat.com/browse/ACM-19432

jc-berger:
See Jira:  https://issues.redhat.com/browse/ACM-19432 

See Jira with resolved bug where I got this write up: https://issues.redhat.com/browse/ACM-19146#

### DIFF
--- a/clusters/release_notes/mce_errata.adoc
+++ b/clusters/release_notes/mce_errata.adoc
@@ -11,4 +11,4 @@ If no release notes are listed, the product does not have an Errata release at t
 
 * Delivers updates to one or more product container images.
 
-* Removes the Red Hat Enterprise Linux CoreOS (RHCOS) image from the default setting when you deploy the `infrastructure-operator`. (link:https://issues.redhat.com/browse/ACM-19146[ACM-19146])
+* Removes a problematic `rhcos` image from the default setting when you deploy the `infrastructure-operator`. (link:https://issues.redhat.com/browse/ACM-19146[ACM-19146])
--- a/clusters/release_notes/mce_errata.adoc
+++ b/clusters/release_notes/mce_errata.adoc
@@ -11,4 +11,4 @@ If no release notes are listed, the product does not have an Errata release at t
 
 * Delivers updates to one or more product container images.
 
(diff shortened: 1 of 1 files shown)


clusters/release_notes/mce_errata.adoc
```diff
@@ -11,4 +11,4 @@ If no release notes are listed, the product does not have an Errata release at t
 
 * Delivers updates to one or more product container images.
 
-* Removes the {ocp} image from the default setting when you deploy the `infrastructure-operator`. (link:https://issues.redhat.com/browse/ACM-19146[ACM-19146])
\ No newline at end of file
+* Removes the Red Hat Enterprise Linux CoreOS (RHCOS) image from the default setting when you deploy the `infrastructure-operator`. (link:https://issues.redhat.com/browse/ACM-19146[ACM-19146])
```
swopebe: I dropped a draft line in the bug yesterday and Matthew said it was OK. I tried to get that info to you but it may have gotten lost in the thread. 

We don't want to say it just removes the image. That is confusing because you need an image that is in production--this one is not. It removes a problematic version (dev stage) image.

From the JIRA: `The problematic image is not present in your mirror. The image names seem to be mismatching.`

To understand the problem when it is not spelled out for us (so we we can write it accurately), sometimes we can look at the description of the bug. (Here the engineer assumed our knowledge, and that happens sometimes.)

I just applied the following investigative writer actions:

Read the description, saw the "problematic" word describing the image. Googled the image, found the doc, lightly read the doc, gave a better description in the bug after I saw what it was and what it was not. Matthew agreed to it in the comments. 

It was the dev-stage image for `rhcos` mirrored on OCP (Matthew tested on 4.18, as he said in the comments, which is how I know we are incorrect to say OCP 4.19).

I am proposing this in the bug:

See suggested changes.

swopebe:
```suggestion
* Removes a problematic `rhcos` image from the default setting when you deploy the `infrastructure-operator`. (link:https://issues.redhat.com/browse/ACM-19146[ACM-19146])
```

PR #7722, opened 2025-04-10
https://issues.redhat.com/browse/ACM-19431

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-19432


swopebe:
@jc-berger:

This appears to have been merged without a peer review. Please get an approval from a peer on all of these, it's best to come to me bc I am in charge of release notes but if I am not able to, another writer can peer review.

For this one, we are using "to be" a lot here, I think we can reduce this: to be unable to be reapplied (reword it to reduce the use to "to be" -- this is a bit confusing).
This relates to the upgrade issue that I documented and I would like to get @jacobbaungard to review this wording, as he is reviewing my release note, as well.
Please check "could" in the modal guidance in the manual. We don't normally use that but I think viewing the standard is worth the effort/knowledge for us. I don't think it is approved but I could be wrong.

swopebe:
Also that line with to be is passive, which can also be changed to active voice (as required by IBM Style, though not always possible), please reopen and let me know when you are ready for a peer review. Thank you.

PR #7723, opened 2025-04-10
{qe/doc validation needed-internal bug} cleanup_reinstall.adoc 

cameronmwall:
Installer noticed that there were additional resources leaking and so we have added them to the cleanup scripts


oafischer:
For 2.13 and 2.14

PR #7726, opened 2025-04-14
https://issues.redhat.com/browse/ACM-19667

jc-berger:
See Previous PR that dealt with some of these changes: https://github.com/stolostron/rhacm-docs/pull/7673/files

See Jira: https://issues.redhat.com/browse/ACM-19667


secure_cluster/certificates.adoc
```diff
@@ -91,13 +90,7 @@ View the following table for a summarized list of the component pods that contai
...
 |===
 
-[#managed-cluster-certificates]
-=== Managed cluster certificates
-//This should not be it's own header, but it is very confusing to me anyway because we now talk about managed cluster certs, but where are these? Are they above? That is titled managed certs. 
-
-You can use certificates to authenticate managed clusters with the hub cluster. Therefore, it is important to be aware of troubleshooting scenarios associated with these certificates.
-
-The managed cluster certificates are refreshed automatically.
+Use these {acm-short} managed certificates to authenticate managed clusters within your hub cluster. These managed cluster certificates get refreshed automatically.
```
jc-berger: The note of troubleshooting seemed kind of random to me. I think we're simply telling users what they can do with these ACM managed certificates. 

@gparvin what do you think?

gparvin: We have a troubleshooting topic around customizing the hub API server certificate.
- https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.13/html/troubleshooting/troubleshooting#troubleshooting-imported-clusters-offline-after-certificate-change

The idea for certificates and managed clusters is:
1. They are managed automatically
2. If you customize the hub API server certificate, managed cluster will update their certificates automatically
3. Optional: To reconnect a managed cluster that is not reconnecting automatically, see the troubleshooting topic -- optional meaning we could skip this item totally

gparvin (commented):
I think this looks really good.  I replied to your comment just to try to clarify some of those details.  Thanks

secure_cluster/certificates.adoc
jc-berger: @gparvin thanks! I added some of the context you provided here. Since already have the "Additional resources" section, I thought the troubleshooting link would work well there, since it's optional as you say. 

Please let me know what you think!

gparvin (approved):

secure_cluster/securing_cluster_intro.adoc
```diff
@@ -8,8 +8,8 @@ Use the role-based access control and authentication to identify the user associ
 
 *Required access:* Cluster administrator
 
+* xref:../secure_cluster/certificates.adoc#certificates[Certificates]
```
swopebe: Can we move these to right above the first certificates?

 RBAC is probably first, then these, then BYO.

swopebe (approved):

PR #7745, opened 2025-04-23
https://issues.redhat.com/browse/ACM-19884 2.13 CIM feature

oafischer:


clusters/cluster_lifecycle/create_cluster_on_prem.adoc
```diff
@@ -216,14 +216,15 @@ spec:
...
+    controlPlaneAgents: 1 <4>
+  sshPublicKey: ssh-rsa <your-public-key-here> <5>
 ----
 
-<1> Specify the platform type of the environment where the cluster is created. Valid values are: `BareMetal`, `None`, `VSphere`, `Nutanix`, or `External`.  
+<1> Specify the platform type of the environment where the cluster is created. Possible values are: `BareMetal`, `None`, `VSphere`, `Nutanix`, or `External`.  
 <2> Use the same name that you used for your `ClusterDeployment` resource.
 <3> Use the `ClusterImageSet` that you generated in <<on-prem-creating-your-cluster-with-the-cli-cluster-image-set,Generate a ClusterImageSet>>.
-<4> You can specify your SSH public key, which enables you to access the host after it is installed.
+<4> Possible values are `1` and `3`. If you are using {ocp-short} version 4.18 or newer, your platform is `BareMetal` or `None`, and your CPU architecture is `x86_64`, you can also use `4` and `5`.
```
gamli75: The 'None' platform is for SNO, so we should remove it.
@danmanor, right?

danmanor: @gamli75 `None` is more for `user-managed-networking` which is required for `SNO`. This feature should work for both `Baremetal` (cmn) and `None` (umn) platforms

danmanor: @oafischer I think it would be great if we could add something that explains it a bit more, for example this is how it is shown in the UI - 
```
Control plane nodes manage workloads, maintain cluster state, and ensure stability. Using more than three nodes boosts fault tolerance and availability, reducing downtime during failures.
```

danmanor: @oafischer I would also change `using` to `installing` probably. WDYT ?

oafischer: I can change it to installing @danmanor. I can add a couple more details but keep in mind that this is the callout section for a resource sample, so we need to keep it on the shorter side to avoid bloating the content. I'll play around with the wording and see what I can do.

oafischer: So to confirm, we can keep the `None` platform?

oafischer: The user would would already have OCP installed at this point, correct? If that is the case, we'd need to stick with `using`.

oafischer: @danmanor Can you confirm?

I added the additional details you requested.

danmanor: @oafischer 

> The user would would already have OCP installed at this point, correct? If that is the case, we'd need to stick with using.

The MCE operator is running on an installed OCP cluster, but this code is about creating a spoke cluster.

danmanor (approved):

danmanor:
@oafischer looks good to me

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
```diff
@@ -216,14 +216,15 @@ spec:
...
+    controlPlaneAgents: 1 <4>
+  sshPublicKey: ssh-rsa <your-public-key-here> <5>
 ----
 
-<1> Specify the platform type of the environment where the cluster is created. Valid values are: `BareMetal`, `None`, `VSphere`, `Nutanix`, or `External`.  
+<1> Specify the platform type of the environment where the cluster is created. Possible values are: `BareMetal`, `None`, `VSphere`, `Nutanix`, or `External`.  
 <2> Use the same name that you used for your `ClusterDeployment` resource.
 <3> Use the `ClusterImageSet` that you generated in <<on-prem-creating-your-cluster-with-the-cli-cluster-image-set,Generate a ClusterImageSet>>.
-<4> You can specify your SSH public key, which enables you to access the host after it is installed.
+<4> Possible values are `1` and `3`. If you are using {ocp-short} version 4.18 or newer, your platform is `BareMetal` or `None`, and your CPU architecture is `x86_64`, you can also use `4` and `5`. Control plane nodes manage workloads, maintain cluster state, and ensure stability. You can increase fault tolerance and availability, and reduce downtime during failures, if you use more than three nodes.
```
amolnar-gh: What's the difference between using `4` or `5`? Why would I choose one over the other? Is it dependent on what I choose for platform type? For example, 4 - Baremetal, 5 - None? Also, do you need to explicitly define your CPU architecture in a resource or it's autodetected? If you need to set it, where do you do that? 

```suggestion
<4> Possible values are `1` and `3`. If you are using {ocp-short} version 4.18 or newer, the `platformType` field is set to `BareMetal` or `None`, and your CPU architecture is `x86_64`, you can also use `4` and `5`. Control plane nodes manage workloads, maintain cluster state, and ensure stability. You can increase fault tolerance and availability, and reduce downtime during failures, if you use more than three nodes.
```

clusters/release_notes/mce_whats_new.adoc
```diff
@@ -46,7 +46,9 @@ Learn about new features and enhancements for Cluster lifecycle with {mce-short}
 
 * You can now specify if you want a minimal ISO or full ISO when generating the Discovery ISO. To learn more, see xref:../assisted_installer/ai_add_host.adoc#add-host-steps-console[Adding hosts by using the console].
 
-* You can now use the `External` platform in {ai}, which allows you to use integration components from cloud providers. To learn more, see xref:../clusters/cluster_lifecycle/create_cluster_on_prem.adoc#create-agentclusterinstall-cr[Create the AgentClusterInstall custom resource]
+* You can now use the `External` platform in {ai}, which allows you to use integration components from cloud providers. To learn more, see xref:../clusters/cluster_lifecycle/create_cluster_on_prem.adoc#create-agentclusterinstall-cr[Create the AgentClusterInstall custom resource].
```
amolnar-gh: I would add this to tie it to the relevant field in the AgentClusterIntall resource. If you explicitly refer to the field, that's even better but might be a bit overkill. I'll leave it up to you 
```suggestion
* You can now use the `External` platform type in {ai}, which allows you to use integration components from cloud providers. To learn more, see xref:../clusters/cluster_lifecycle/create_cluster_on_prem.adoc#create-agentclusterinstall-cr[Create the AgentClusterInstall custom resource].
```

clusters/release_notes/mce_whats_new.adoc
```diff
@@ -46,7 +46,9 @@ Learn about new features and enhancements for Cluster lifecycle with {mce-short}
...
+
+* You can now install clusters with a highly available {ocp-short} control plane. To learn more, see xref:../clusters/cluster_lifecycle/create_cluster_on_prem.adoc#create-agentclusterinstall-cr[Create the AgentClusterInstall custom resource].
```
amolnar-gh, line 51: It's the exact same link as the previous one. Does this section cover both topics mentioned here?

amolnar-gh (commented):
Great work! I have a few questions that need clarification but maybe I'm just not familiar with the area :)

clusters/release_notes/mce_whats_new.adoc
oafischer: The previous one is already published, it's for a different feature. Not sure why GitHub is showing it as new

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
oafischer: The difference between `4` and `5` is that 5 uses 5 control plane nodes instead of 4. The reason for choosing more vs. less is to increase availability/decrease downtime. 4 and 5 is only available on BareMetal and None on x86_64.

CPU architecture does not need to be defined.

Let me know if that's clear and if you recommend any more changes. Thanks!

clusters/release_notes/mce_whats_new.adoc
amolnar-gh: That's fine, sometimes GH does the same for me. However, my problem is that they point to the exact same section. Is that intentional? Copypasting just the xref paths so it's easier to compare:
> xref:../clusters/cluster_lifecycle/create_cluster_on_prem.adoc#create-agentclusterinstall-cr
> xref:../clusters/cluster_lifecycle/create_cluster_on_prem.adoc#create-agentclusterinstall-cr

clusters/cluster_lifecycle/create_cluster_on_prem.adoc
amolnar-gh: That's much clearer thank you! I'm happy to approve

amolnar-gh (approved):

clusters/release_notes/mce_whats_new.adoc
oafischer: Yeah the doc for both features is in the same section

PR #7767, opened 2025-04-29
https://issues.redhat.com/browse/ACM-17815

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-17815

### DIFF
--- a/business_continuity/backup_restore/backup_arch.adoc
+++ b/business_continuity/backup_restore/backup_arch.adoc
@@ -1,7 +1,9 @@
 [#backup-restore-architecture]
 = Backup and restore operator architecture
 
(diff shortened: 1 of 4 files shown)


birsanv (approved):
lgtm

business_continuity/backup_restore/backup_arch.adoc
```diff
@@ -133,7 +139,8 @@ Aside from the activation data resources that are identified by using the `clust
 
 [#extend-backup-data]
 == Extending backup data
+//this seems to be missplaced since this is an architecture file and this is more about how to do something?
```
swopebe, line 142: This looks very good, but I made a couple small changes. 

I took a look at the file and it seems this part right here is task information, but placed in a concept. We should either move this to a task or rewrite it to just explain the label, removing the gerund from the title. 

That would look something like: 

== `backup` label function

swopebe (changes requested):

swopebe (approved):

PR #7768, opened 2025-04-29
https://issues.redhat.com/browse/ACM-20170 ISO type note

oafischer:
Also for 2.13

### DIFF
--- a/clusters/assisted_installer/ai_create_cli.adoc
+++ b/clusters/assisted_installer/ai_create_cli.adoc
@@ -182,7 +182,7 @@ See the following field descriptions in the `InfraEnv` table:
 
 | `ImageType`
 | Optional
-| Selects the ISO file type to download. Possible options are `minimal-iso` or `full-iso`. If you do not specify an ISO type, the minimal ISO is used by default. The minimal ISO does not contain the root file system, `RootFS`. If you select a minimal ISO, the `RootFS` is downloaded later.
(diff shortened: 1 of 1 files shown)


clusters/assisted_installer/ai_create_cli.adoc
```diff
@@ -182,7 +182,7 @@ See the following field descriptions in the `InfraEnv` table:
 
 | `ImageType`
 | Optional
-| Selects the ISO file type to download. Possible options are `minimal-iso` or `full-iso`. If you do not specify an ISO type, the minimal ISO is used by default. The minimal ISO does not contain the root file system, `RootFS`. If you select a minimal ISO, the `RootFS` is downloaded later.
+| Selects the ISO file type to download. Possible options are `minimal-iso` or `full-iso`. If you do not specify an ISO type, the minimal ISO is used by default. The minimal ISO does not contain the root file system, `RootFS`. If you select a minimal ISO, the `RootFS` is downloaded later. If you select a full ISO, you must also set `ISO_IMAGE_TYPE` to `full-iso` in the `AgentServiceConfig` resource for the `ImageType` parameter to take effect.
```
carbonin, line 185: None of this is accurate or true.

I'll check the issue to see where this is coming from.

carbonin (changes requested):

carbonin:
I think we can just close this until we get some more information about the underlying issue.

AgentServiceConfig doesn't have any field for ISO type so I'm not sure what is being suggested here or how a user would do what is written.

clusters/assisted_installer/ai_create_cli.adoc
oafischer: Thanks for confirming!

oafischer:
Thanks @carbonin and thanks for commenting in the issue as well. I can close the PR and the issue. You can see in the ticket that it came from support (which can sometimes give us inaccurate information), but that is exactly why we always ask engineers to review to catch stuff like this.

PR #7790, opened 2025-05-06
https://issues.redhat.com/browse/ACM-19943 spaces in yaml

ragevou:
for 2.12 and 2.13

### DIFF
--- a/business_continuity/backup_restore/tag_resources.adoc
+++ b/business_continuity/backup_restore/tag_resources.adoc
@@ -44,10 +44,10 @@ spec:
 [source,yaml]
 ----
 kind: Restore
-metadata:
+metadata: 
 name: restore-acm
 namespace: open-cluster-management-backup
-spec:
+spec: 
 cleanupBeforeRestore: None
 veleroManagedClustersBackupName: latest
 veleroCredentialsBackupName: latest
--- a/business_continuity/backup_restore/tag_resources.adoc
+++ b/business_continuity/backup_restore/tag_resources.adoc
@@ -25,13 +25,13 @@ To tag user created resources on the restore hub cluster with the `velero.io/bac
 apiVersion: cluster.open-cluster-management.io/v1beta1
 kind: BackupSchedule
 metadata:
- name: schedule-acm-msa
- namespace: open-cluster-management-backup
+  name: schedule-acm-msa
+  namespace: open-cluster-management-backup
 spec:
- veleroSchedule: 0 */1 * * *
- veleroTtl: 120h
- useManagedServiceAccount: true
- paused: false
+  veleroSchedule: 0 */1 * * *
+  veleroTtl: 120h
+  useManagedServiceAccount: true
+  paused: false
 ----
 
 . Verify that the `BackupSchedule` created the hub cluster backup resources and the resource status is `Completed`.
--- a/business_continuity/backup_restore/tag_resources.adoc
+++ b/business_continuity/backup_restore/tag_resources.adoc
@@ -45,19 +45,19 @@ spec:
 ----
 kind: Restore
 metadata: 
-name: restore-acm
-namespace: open-cluster-management-backup
+ name: restore-acm
+ namespace: open-cluster-management-backup
 spec:
-  cleanupBeforeRestore: None
-  veleroManagedClustersBackupName: latest
-  veleroCredentialsBackupName: latest
-  veleroResourcesBackupName: latest
-  excludedResources: <1>
-    - ManagedCluster
-  excludedNamespaces: <2>
-    - managed-cls-1
-    - managed-cls-2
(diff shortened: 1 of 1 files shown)


business_continuity/backup_restore/tag_resources.adoc
```diff
@@ -44,10 +44,10 @@ spec:
...
 kind: Restore
-metadata:
+metadata: 
 name: restore-acm
 namespace: open-cluster-management-backup
-spec:
+spec: 
 cleanupBeforeRestore: None
 veleroManagedClustersBackupName: latest
 veleroCredentialsBackupName: latest
```
swopebe: Adding an example of how spec is written in YAML from the other docs. The bug was for spacing that was incorrectly added to the doc. Maybe we are coming back to fix this and just haven't had time, but I think we still have to get the correct format:

```
[source,yaml]
----
spec:
  backupLocations:
    - velero:
        config:
          kmsKeyId: 502b409c-4da1-419f-a16e-eif453b3i49f
          profile: default
          region: us-east-1
----
```

swopebe: @birsanv can you confirm what the format is for the `Restore` resource kind?

swopebe: I took a second and did a simple search in the code and found what appears to be that same YAML snippet:

https://github.com/stolostron/rhacm-docs/blob/7105ee4d6fa416f01281d3ca91a230bdbb8c877e/business_continuity/backup_restore/backup_active_restore.adoc#L62

tesshuflower: > I took a second and did a simple search in the [code](https://github.com/stolostron/cluster-backup-operator) and found what appears to be that same YAML snippet as another doc:
> 
> https://github.com/stolostron/rhacm-docs/blob/7105ee4d6fa416f01281d3ca91a230bdbb8c877e/business_continuity/backup_restore/backup_active_restore.adoc#L62

@swopebe Yes this link looks better.

It looks like the formatting got messed up in this particular example and everything got pushed to the left.

it should look like the example you linked above, something like this:

```yaml
[source,yaml]
----
kind: Restore
metadata: 
 name: restore-acm
 namespace: open-cluster-management-backup
spec: 
 cleanupBeforeRestore: None
 veleroManagedClustersBackupName: latest
 veleroCredentialsBackupName: latest
 veleroResourcesBackupName: latest
 excludedResources: <1>
 - ManagedCluster
 excludedNamespaces: <2>
 - managed-cls-1
 - managed-cls-2
 - managed-cls-n
```


Note that I've used an indentation of 1 space to match the other yaml in this same page - typically I would use 2 spaces - but do we use 1 space everywhere in our docs?

swopebe: We should be matching what the source is. I will have to see...does it make a difference to the user? Either way this one was way off and should have been fixed during writing and review process. @ragevou if you are not able to get to this, please flip the issue to me since all that info has been added. Thank you.

swopebe: we need to follow what you guys have in your source...I will have to check other YAML, but from the looks of it, I agree with two. I am not sure about the link above--if that was also not added correctly.

swopebe: Screenshot 2025-05-08 at 7 22 51 PM
Screenshot 2025-05-08 at 7 23 28 PM
Screenshot 2025-05-08 at 7 23 53 PM

swopebe: Other examples from the doc ^^^

swopebe: https://github.com/stolostron/cluster-backup-operator has two spaces, it seem. I think two @tesshuflower

tesshuflower: I believe yaml will accept 1 space as well, but generally we use something like 2 or 4 spaces.  I think it's probably best just to be consistent within the docs whenever there's yaml?

Is there some yaml linter that could be used in the docs to ensure that all the yamls are formatted correctly?  Maybe it's difficult as they're not separate files, but embedded within the doc file.

sahare:
@ragevou Thanks for the changes. Based on the conversation above, I believe we should use two-space indentation in the YAML samples instead of one.

business_continuity/backup_restore/tag_resources.adoc
swopebe: There probably is in VS Code, however, we are losing our editor bc we have to move to DITA (Adobe) so we will have to see. I more try to get the writers to recognize YAML and prevent bugs such as this...we all have experience with YAML, not sure what happened here.

tesshuflower (approved):
/lgtm

PR #7809, opened 2025-05-07
https://issues.redhat.com/browse/ACM-20406--Add IBBF section to SiteConfig

amolnar-gh:
Related to https://github.com/stolostron/rhacm-docs/pull/7808

PR #7808, opened 2025-05-07
https://issues.redhat.com/browse/ACM-20406--Add cluster reinstallation concept for SiteConfig operator

amolnar-gh:

PR #7810, opened 2025-05-07
https://issues.redhat.com/browse/ACM-20406--Add enabling and labeling procedures for reinstallation with SiteConfig

amolnar-gh:
Related to https://github.com/stolostron/rhacm-docs/pull/7808 and https://github.com/stolostron/rhacm-docs/pull/7809


mce_acm_integration/siteconfig/siteconfig_reinstallation_enable.adoc
```diff
@@ -0,0 +1,133 @@
...
++
+[source,bash]
+----
+oc get configmap siteconfig-operator-configuration -n $NAMESPACE -o yaml  
+----
+
+[#labeling-resources-for-preservation]
+== Optional. Labeling resources for preservation
+
+Set the desired preservation mode in your `ClusterInstance` custom resource and if applicable, label your resources accordingly.
```
amolnar-gh: @sakhoury I wanted to clarify this section because the labeling can be a bit confusing at first (for me at least :) 
I clarified the following:
* You need to set the preservationMode in the ClusterInstance CR and to label the resources accordingly.
* The example we use for the command here is for the `All` preservation mode. I want to avoid confusion about why it doesn't work with ClusterIdentity.

However, while doing so I had a few questions:
1) When do we set the preservationMode? Does it matter if we set it before or after labeling? 
2) If you don't set the preservationMode, does it default to `None`? I don't remember if we mention it anywhere in the docs.
3) If we don't mention it, do we want to add a step here on how to do it?

mce_acm_integration/siteconfig/siteconfig_reinstallation_enable.adoc
```diff
@@ -0,0 +1,132 @@
...
+. Apply the changes by using one of the following methods:
+
+GitOps approach:: Commit and push the updated `ClusterInstance` resource to your Git repository.
+Manual application:: Apply the changes directly on the hub cluster:
+
++
+[source,bash]
+----
+oc apply -f clusterinstance-example.yaml
+----
```
amolnar-gh: @swopebe  Can we use this format throughout the docs? You can use SiteConfig with or without GitOps, so depending on that, the user might want to apply the changes via GitOps or manually to the hub cluster. However, currently we only instruct to apply the changes on the hub cluster, which is causing confusion.

mce_acm_integration/siteconfig/siteconfig_reinstallation_enable.adoc
```diff
@@ -0,0 +1,132 @@
...
+. Verify the update by running the following command:
+
++
+[source,bash]
+----
+oc get configmap siteconfig-operator-configuration -n $NAMESPACE -o yaml  
+----
+
+[#labeling-resources-for-preservation]
+== Optional. Labeling resources for preservation
```
amolnar-gh: The heading level is not correct for this section as per the reinstallation workflow: https://github.com/stolostron/rhacm-docs/pull/7808/files#diff-023ce400ad77728808806c824aa3f200538cefa8498d9c525c95c55f75c958c5R18

However, it needs to be done before initiating the reinstallation. I could separate this file into 3 smaller ones but not sure if that's the way to go. @swopebe thoughts?

mce_acm_integration/siteconfig/siteconfig_reinstallation_enable.adoc
swopebe: . **Optional:** If you are using {ocp} GitOps to initiate a cluster reinstallation, commit and push the updated `ClusterInstance` resource to your Git repository. 

. To manually apply your changes, run the following command:

mce_acm_integration/siteconfig/siteconfig_reinstallation_enable.adoc
```diff
@@ -0,0 +1,132 @@
+[#enabling-cluster-reinstallation-in-the-sco]
+= Enabling cluster reinstallation in the {sco}
```
swopebe, line 2: ok, this title is good, but I now I am wondering abut the entire reinstallation set up.

I think it is better to do an outline in the issue, rather than having to piece together the PRs.

mce_acm_integration/siteconfig/siteconfig_reinstallation_enable.adoc
```diff
@@ -0,0 +1,131 @@
...
+
+You can set the desired preservation mode in your `ClusterInstance` custom resource and if applicable, label your resources accordingly.
+//For more information, see xref:../../mce_acm_integration/siteconfig/siteconfig_cluster_reinstallation.adoc#resources-labeling-for-preservation[Resource labeling for preservation]
+
+* Label a resource for the `All` preservation mode by running the following command:
+
++
+[source,bash]
+----
+oc label configmap <your_configmap> "siteconfig.open-cluster-management.io/preserve="
```
sakhoury:
```suggestion
oc label configmap <your_configmap> "siteconfig.open-cluster-management.io/preserve=all"
```

mce_acm_integration/siteconfig/siteconfig_reinstallation_enable.adoc
sakhoury: 1. Set the preservationMode on the ClusterInstance when you want to trigger the reinstallation. This can be done when setting the `reinstall.generation` value. The labelling is to be done before applying the ClusterInstance with the preservationMode set.
2. The default is `None`.
3. Not sure I follow the question.

amolnar-gh: > If we don't mention it, do we want to add a step here on how to do it?

It's been a while since I wrote this comment but what I think I meant is to add a step to update the preservationMode field with the desired value. For example, in this procedure we're using the `All` preservation mode as an example but we don't have the step in the procedure to set it

mce_acm_integration/siteconfig/siteconfig_reinstallation_enable.adoc
```diff
@@ -0,0 +1,131 @@
+[#enabling-cluster-reinstallation]
+= Enabling cluster reinstallation in the {sco}
+
+You must explicitly enable cluster reinstallation in the `siteconfig-operator-configuration` `ConfigMap` resource, which you can complete well before initiating the process. By default, reinstallation is disabled.
+
+*Required access:* Cluster administrator
+
+Complete the following steps:
+
+. Create the `NAMESPACE` environment variable for the `siteconfig-operator-configuration` `ConfigMap` resource. The namespace must match the namespace where the {sco} is installed. Run the following command:
```
josclark42: I think this line is confusing. I think it can easily be simplified to:

. Set the `NAMESPACE` environment variable to match the namespace where the {sco} is installed. Run the following command:

sakhoury (commented):
/lgtm

PR #7811, opened 2025-05-07
https://issues.redhat.com/browse/ACM-20406--Add monitoring and verifying reinstallation for SiteConfig

amolnar-gh:
Related to https://github.com/stolostron/rhacm-docs/pull/7808, https://github.com/stolostron/rhacm-docs/pull/7809, and https://github.com/stolostron/rhacm-docs/pull/7810


mce_acm_integration/siteconfig/siteconfig_verify_reinstallation.adoc
```diff
@@ -0,0 +1,5 @@
+
```
amolnar-gh: I'm not happy with this file being so short but if I merge it with the monitoring file, the heading must be level 2, which is not correct as per the cluster reinstallation flow: https://github.com/stolostron/rhacm-docs/pull/7808/files#diff-0d1a43993cf2a881401b132d2ba9fbe30ccbdc000c26a0082fce857636cb7b87R15-R24

I could merge all the procedures into one file but that becomes unmanageable quickly. I'll come back to this.

swopebe: Can we not just have a _Monitoring_ file that covers all of it? Verifying would be a step.

mce_acm_integration/siteconfig/siteconfig_monitor_reinstallation.adoc
```diff
@@ -0,0 +1,151 @@
+[#optional-monitoring-the-cluster-reinstallation]
+= Optional. Monitoring the cluster reinstallation
```
swopebe: I think this should just be reinstallation, with monitoring and verifying sub sections.

swopebe: We don't normally have "optional" at the highest title, but we can be sure to emphasize that it is optional in the short description.

swopebe (approved):
I am approving, knowing we need to go back and look at the build. Also knowing we need a tech review on this before merge.

sakhoury (commented):
/lgtm
/approve

PR #7825, opened 2025-05-12
https://issues.redhat.com/browse/ACM-19142 adding content from image …

swopebe:
…volume size--editing out header threes


clusters/assisted_installer/ai_enable_cim.adoc
```diff
@@ -1,7 +1,8 @@
 [#enable-cim]
 = Enabling the {cim} service
 
-The {cim} service is provided with the {mce-short} and deploys {ocp-short} clusters. The {cim} service is deployed automatically when you enable the MultiClusterHub Operator on the hub cluster, but you have to enable the service manually.
+The {cim} service is provided with the {mce-short} and deploys {ocp-short} clusters. The {cim} service is deployed automatically when you enable the `MultiClusterHub` Operator on the hub cluster, but you have to enable the service manually.
+//This doc is in the MCE doc, which is supposed to be stand-alone for an OCP user. We can discuss MCH here, but we need to say: The {cim} service is deployed automatically if you installed {acm-short} and enabled the `MultiClusterHub` Operator on the hub cluster, but you have to enable the service manually in this case. (Do we tell them how?)
```
swopebe, line 5: Please review my recommended change to this intro section.

clusters/assisted_installer/ai_enable_cim.adoc
```diff
@@ -239,7 +241,7 @@ spec:
 +
 <2> Include the optional `unauthenticated_registry` parameter if you are using a mirror registry that does not require authentication. Entries on this list are not validated or required to have an entry in the pull secret. 
 +
```
swopebe, line 243: See this change for review

clusters/assisted_installer/ai_enable_cim.adoc
```diff
@@ -239,7 +241,7 @@ spec:
...
-<3> Replace `img_volume_size` with the size of the volume for the `imageStorage` field, for example `10Gi` per operating system image. The minimum value is `10Gi`, but the recommended value is at least `50Gi`. This value specifies how much storage is allocated for the images of the clusters. You need to allow 1 GB of image storage for each instance of Red Hat Enterprise Linux CoreOS that is running. You might need to use a higher value if there are many clusters and instances of Red Hat Enterprise Linux CoreOS.
+<3> Replace `img_volume_size` with the size of the volume for the `imageStorage` field, for example `10Gi` for each operating system image. The minimum value is `10Gi`, but the value for best practice is at least `50Gi`, especially if you are opting for the default set of images instead of defining a limited set using `osImages`. This value specifies how much storage is allocated for the images of the clusters. You need to allow at least 1 GB of image storage for each instance of Red Hat Enterprise Linux CoreOS that is running. You might need to use a higher value if there are many clusters and instances of Red Hat Enterprise Linux CoreOS.
```
swopebe: Review line 244

clusters/assisted_installer/ai_enable_cim.adoc
```diff
@@ -286,13 +288,16 @@ spec:
...
+<3> Replace `img_volume_size` with the size of the volume for the `imageStorage` field, for example `10Gi` for each operating system image. The minimum value is `10Gi`, but the recommended value is at least `50Gi`, especially in case you are opting for the default set of images instead of defining a limited set using `osImages`. This value specifies how much storage is allocated for the images of the clusters. You need to allow at least 1 GB of image storage for each instance of Red Hat Enterprise Linux CoreOS that is running. You might need to use a higher value if there are many clusters and instances of Red Hat Enterprise Linux CoreOS.
+
+See _Enabling {cim} in disconnected environments_ for details about how to define custom image set in `osImages`.
```
swopebe, line 294: Review line 291-294

clusters/assisted_installer/ai_enable_cim.adoc
```diff
@@ -286,13 +288,16 @@ spec:
...
 
 Your {cim} service is configured. You can verify that it is healthy by checking the `assisted-service` and `assisted-image-service` deployments and ensuring that their pods are ready and running. 
 
 [#fips-install-cim]
 == Installing a FIPS-enabled cluster by using the {ai}
```
swopebe, line 298: We can talk about this @oafischer but for transfer to DITA, we should consider breaking out this section. If you are want, we can create an issue and pursue it another time.

oafischer: I can add this task to some of the other CIM work I'm doing

clusters/assisted_installer/ai_enable_cim.adoc
oafischer: Looks good to me

clusters/assisted_installer/ai_enable_cim.adoc
oafischer: Looks good

clusters/assisted_installer/ai_enable_cim.adoc
oafischer: I think this works. We tell them how to do it in the 4 topics that follow the prereqs.

oafischer (approved):

clusters/assisted_installer/ai_enable_cim.adoc
swopebe: If we mention MCH in MCE, we need to mention that ACM automatically installs it.

oafischer: Yeah the suggestion that you commented looks good to me, you can add that.

clusters/assisted_installer/ai_enable_cim.adoc
carbonin: > for example `10Gi` for each operating system image

I don't think 10Gi per image is reasonable. I'd say something more like 2 per configured image.

> especially if you are opting for the default set of images instead of defining a limited set using `osImages`

@zszabo-rh how many images do we have in the default set now? Does 50Gi still make sense if we assume we need 2Gi per image?

> You need to allow at least 1 GB of image storage for each instance of Red Hat Enterprise Linux CoreOS that is running

I know this isn't a new change, but this doesn't make sense to me. This value has nothing to do with what is "running" it only changes based on the number of images available to use for install (i.e. the default set or whatever is configured in `osImages`).

> You might need to use a higher value if there are many clusters and instances of Red Hat Enterprise Linux CoreOS.

Same here. For example you could provision 1000 clusters all from the same image and only need a 2Gi volume here.

clusters/assisted_installer/ai_enable_cim.adoc
```diff
@@ -286,13 +288,16 @@ spec:
...
+<2> Replace `fs_volume_size` with the size of the volume for the `filesystemStorage` field, for example `200M` for each cluster and `2-3Gi` for each supported {ocp-short} version. The minimum value that is required is `1Gi`, but the recommended value is at least `100Gi`. This value specifies how much storage is allocated for storing logs, manifests, and `kubeconfig` files for the clusters. You might need to use a higher value if there are many clusters. 
+<3> Replace `img_volume_size` with the size of the volume for the `imageStorage` field, for example `10Gi` for each operating system image. The minimum value is `10Gi`, but the recommended value is at least `50Gi`, especially in case you are opting for the default set of images instead of defining a limited set using `osImages`. This value specifies how much storage is allocated for the images of the clusters. You need to allow at least 1 GB of image storage for each instance of Red Hat Enterprise Linux CoreOS that is running. You might need to use a higher value if there are many clusters and instances of Red Hat Enterprise Linux CoreOS.
```
carbonin: Same comment as above

clusters/assisted_installer/ai_enable_cim.adoc
swopebe: @zszabo-rh please see Nick's comments.

zszabo-rh: Yes I agree, this text needs much more refining than I originally planned.

The [default image set on master branch](https://github.com/openshift/assisted-service/blob/master/data/default_os_images.json) contains 39 items each requiring around 1.2GiB (full+minimal ISO together), which sums at ~47GiB storage space. 

For ACM branches we try to keep this number lower, e.g. [on 2.12 we have 16 images](https://github.com/openshift/assisted-service/blob/release-ocm-2.12/data/default_os_images.json) only, requiring ~20GiB tops. But currently we don't have any automation in place to ensure this optimization for each newly opened ACM branch, so I think it's safer to leave the 50GiB recommendation for default.

clusters/assisted_installer/ai_enable_cim.adoc
swopebe: Thanks @oafischer -- it looks like this will need some more work based on developer discussion above, so I will keep working on this. Just let me know if you end up breaking this out before I merge my branch. TY.

clusters/assisted_installer/ai_enable_cim.adoc
swopebe: What exactly needs to change from this PR?

carbonin:
```suggestion
<3> Replace `img_volume_size` with the size of the volume for the `imageStorage` field. This value specifies how much storage is allocated to store base Red Hat Enterprise Linux CoreOS images. The minimum value is `10Gi`, but the value for best practice is at least `50Gi`, especially if you are opting for the default set of images instead of defining a limited set using `osImages`. You need to allow at least `2Gi` of image storage for each instance of Red Hat Enterprise Linux CoreOS that is configured in `osImages`.
```

Maybe something like this?

The issue with the existing text is that this number has nothing to do with the number of clusters or anything that is "running".

clusters/assisted_installer/ai_enable_cim.adoc
carbonin: Is there any reason this shouldn't be the same as the previous one?
```suggestion
<3> Replace `img_volume_size` with the size of the volume for the `imageStorage` field. This value specifies how much storage is allocated to store base Red Hat Enterprise Linux CoreOS images. The minimum value is `10Gi`, but the value for best practice is at least `50Gi`, especially if you are opting for the default set of images instead of defining a limited set using `osImages`. You need to allow at least `2Gi` of image storage for each instance of Red Hat Enterprise Linux CoreOS that is configured in `osImages`.
```

swopebe: No, I just missed it.

carbonin (approved):

PR #7838, opened 2025-05-19
https://issues.redhat.com/browse/ACM-20885 -- remove upgrade logic in last release (see comment in Jira about removal--brandi)

ldpliu:

PR #7862, opened 2025-05-21
https://issues.redhat.com/browse/ACM-20885

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-20885 

See draft PR: https://github.com/stolostron/rhacm-docs/pull/7838/files


release_notes/acm_deprecate_remove.adoc
```diff
@@ -11,6 +11,11 @@ Learn when parts of the product are deprecated or removed from {acm}. Consider t
 
 *Important:* Cluster lifecycle components and features are within the {mce-short}, which is a software operator that enhances cluster fleet management. Release notes for {mce-short}-specific features are found in at link:../clusters/release_notes/mce_release_notes.adoc#mce-release-notes[Release notes for Cluster lifecycle with {mce-short}].
 
+[#global-hub-postgres-database-deprecation]
```
ldpliu: The reason is not right. we do not remove built-in postgres database.
We just remove the doc about `upgrade postgres in Globalhub`, this doc is only valid when upgrade globalhub 1.3 -> 1.4. In this release(Globalhub 1.5), globalhub will handle the database upgrade automatically, the is no need to do manually steps which include in [upgrade doc](https://github.com/stolostron/rhacm-docs/pull/7862/files#diff-43e8c6aac553e495a9380dd026f5e9d75636aff347a29d4b4d1d7edea2ddba9a). So we need to remove the doc.

release_notes/acm_deprecate_remove.adoc
```diff
@@ -11,6 +11,11 @@ Learn when parts of the product are deprecated or removed from {acm}. Consider t
 
 *Important:* Cluster lifecycle components and features are within the {mce-short}, which is a software operator that enhances cluster fleet management. Release notes for {mce-short}-specific features are found in at link:../clusters/release_notes/mce_release_notes.adoc#mce-release-notes[Release notes for Cluster lifecycle with {mce-short}].
 
+[#global-hub-postgres-deprecation]
+== {global-hub} deprecation of the Postgres documentation
```
ldpliu: It should be postgres `upgrade` document.

ldpliu:
It's OK for me, @yanmxa @yahliu Please help to review it

release_notes/acm_deprecate_remove.adoc
```diff
@@ -11,6 +11,11 @@ Learn when parts of the product are deprecated or removed from {acm}. Consider t
 
 *Important:* Cluster lifecycle components and features are within the {mce-short}, which is a software operator that enhances cluster fleet management. Release notes for {mce-short}-specific features are found in at link:../clusters/release_notes/mce_release_notes.adoc#mce-release-notes[Release notes for Cluster lifecycle with {mce-short}].
 
+[#global-hub-postgres-deprecation]
+== {global-hub} deprecation of the Postgres upgrade documentation
+
+In version 1.5 and later, {global-hub} automatically handles database upgrades. You no longer need to manually upgrade your data and databases with the Postgres database, so the Postgres upgrade documentation is removed.
```
dockerymick, line 17: This is usually in a table and since this is being deprecated, the global_hub_upgrade_postgres.adoc file should not be removed. The "Deprecated" label should be added in the title of that file:

`Recovering data from the upgraded built-in Postgres (Deprecated)`

I'm not sure what is the recommended action would be, however I expect to see the following deprecation table and entry:

```
|===

| {global-hub} 
| Postgres upgrade documentation
| {acm-short} 2.14
| Not sure
| None

|===
```
Is there a reason why you aren't using a table? 

Also the following statement seems like a what's new entry: 
"In version 1.5 and later, {global-hub} automatically handles database upgrades. You no longer need to manually upgrade your data and databases with the Postgres database, so the Postgres upgrade documentation is removed."

I suggest that you consider adding this to What's new: 

CHANGE: In version 1.5 and later, {global-hub} automatically handles database upgrades. You no longer need to manually upgrade your data and databases with the Postgres database, so the Postgres upgrade documentation is removed.

TO 

Automatic database upgrades are supported for {global-hub} version 1.5 and later. 


Let's discuss if you have any questions, concerns, or have another suggestion. Does the user need to do anything to verify that the upgrade is working?

dockerymick (changes requested):
Please see the comment that I made. I referred to the 2.13 Deprecation and removals doc as a guide: https://github.com/stolostron/rhacm-docs/blob/2.13_stage/release_notes/acm_deprecate_remove.adoc

release_notes/acm_deprecate_remove.adoc
ldpliu: I think it's fine to use the table, and user do not need anything else to verify the upgrade is working.

yanmxa:
looks good to me.

release_notes/acm_deprecate_remove.adoc
swopebe: The team is aware of the deprecation process and if not, it is also documented. Deprecations and removals are and have always been part of release notes. https://stolostron.github.io/rhacm-docs/deprecate_remove

Please follow this process for announcing deprecations and removals.

swopebe: We also don't need to deprecate previous versions of an operator, just functions or features if they are still supported, FYI. 

Please don't use "You no longer need" in the documentation when talking about a new version or a fix. In what's new we don't go in to what was. We go in to what is. In bug fixes we can talk about previous errors that were fixed, but not what's new. 

Let's go over this in the next meeting again.

jc-berger:
Closing. See new and revised Doc PR where postress is deprecated, not removed: https://github.com/stolostron/rhacm-docs/pull/7913

PR #7929, opened 2025-06-10
https://issues.redhat.com/browse/ACM-20629--link with wrong title, br…

swopebe:
…oken as well, too many headers instead of steps, reformatting to get rid of header 3 and 4, use proper gerund in title and in links


swopebe:
Because this is removing headers, fixing a link name, and fixing a broken link during restructure, I will merge and build with no approval.

PR #7943, opened 2025-06-11
[2.14_stage] Omit `403` errors for link checker

dhaiducek:
Currently the `https://developers.redhat.com/` site is throwing a `403 Unauthorized` error even though the site exists. Adding an exception (since we could run into this more if additional bot web scraping protections are put in place).


dockerymick (approved):
Thanks Dale!

PR #7961, opened 2025-06-14
https://issues.redhat.com/browse/ACM-20455--announce aws marketplace

swopebe:


release_notes/acm_whats_new.adoc
```diff
@@ -35,7 +35,9 @@ Some features and components are identified and released as link:https://access.
 [#install-acm-whats-new]
 == Installation
 
-You can now change the name of your local cluster. By default, the hub cluster manages itself as the `local-cluster`. When the `disableHubSelfManagement` field is set to `true`, which disables the local cluster feature, you can change the name. Use 34 or fewer characters for the `<your-local-cluster-name>` value. The `local-cluster` resource and the namespaces reflects the change.
+* {acm-short} is now available on the link:https://aws.amazon.com/marketplace/pp/prodview-s2nb6rqz2i3ru[AWS Marketplace], with either on-demand or annual pricing, and a simplified billing option for running {acm-short} on {rosa} clusters. {acm-short} on the AWS Marketplace offers billing consistency, flexible resource consumption, and cost efficiency. Purchase from the AWS Marketplace, then you can install `MultiClusterHub` resources and manage your clusters.
+
```
swopebe, line 39: only line 38 needs to be reviewed, edited. Writing/Reviewing checklist considered.

berenss: nice!
One additional sentence please
- To view the overall consumption navigate to console.redhat.com and view the Subscriptions > OpenShift usage, selecting Variant: Red Hat Advanced Cluster Management from the drop-down list.

berenss: URL to view the consumption usage is here: https://console.redhat.com/subscriptions/usage/openshift
I don't know if that URL might ever change, so it's fine with me if we do not include it.

I can also provide an example screenshot if that would be nice to have.

berenss (approved):

release_notes/acm_whats_new.adoc
swopebe: I am happy to add the URL, this announcement will only be for 2.14. If we want to place it elsewhere, like the install doc in the future, we can and we just have to keep an eye out for that link.

Screenshots we are not supposed to add, though we have one, because of accessibility. If we can avoid it, we should.

release_notes/acm_whats_new.adoc
```diff
@@ -35,7 +35,9 @@ Some features and components are identified and released as link:https://access.
 [#install-acm-whats-new]
 == Installation
 
-You can now change the name of your local cluster. By default, the hub cluster manages itself as the `local-cluster`. When the `disableHubSelfManagement` field is set to `true`, which disables the local cluster feature, you can change the name. Use 34 or fewer characters for the `<your-local-cluster-name>` value. The `local-cluster` resource and the namespaces reflects the change.
+* {acm-short} is now available on the link:https://aws.amazon.com/marketplace/pp/prodview-s2nb6rqz2i3ru[AWS Marketplace] with either on-demand or annual pricing and a simplified billing option for running {acm-short} on {rosa} clusters. {acm-short} on the AWS Marketplace offers billing consistency, flexible resource consumption, and cost efficiency. Purchase from the AWS Marketplace, then you can install `MultiClusterHub` resources and manage your clusters. To view the overall consumption, go to link:https://console.redhat.com/subscriptions/usage/openshift[console.redhat.com]. Click *Subscriptions Usage* > *OpenShift*. Select *Variant: Red Hat Advanced Cluster Management* from the drop-down menu.
```
dockerymick, line 38: The following statement seems like a task that should be in the install doc: "Click *Subscriptions Usage* > *OpenShift*. Select *Variant: Red Hat Advanced Cluster Management* from the drop-down menu." 

I agree with your comment about adding this to install. Is there a reason why we are describing the task in what's new?

berenss: Hi Mikela! 
it's not really an install task - it's something that comes up more at the procurement / senior leaders-org / accounting level rather than the technical person that is clicking the operator to install it. It's more about how the customer gets billed and charged for the usage of RHACM rather than a task they have to care about during install time. There is no change to the install experience itself. I guess if we had a Planning section of the install it could go there, but I don't recall seeing one since the IBM days.

swopebe: It is in the install section bc we don't have a general announcements section--I can look around to see how others announce these items, if they do. For now, we just place in install bc it is first and in the beginning of using ACM.

It's not really a task. It's a new place to go and buy ACM.

swopebe: The issue was to just announce it this release: https://issues.redhat.com/browse/ACM-20455

dockerymick (approved):
looks good to me! 
[Checklist](https://stolostron.github.io/rhacm-docs/checklist) considered

PR #7964, opened 2025-06-16
https://issues.redhat.com/browse/ACM-18946

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-18946

PR created by and reviewed with the Doc Checklist: https://stolostron.github.io/rhacm-docs/checklist

### DIFF
--- a/gitops/gitops_push_pull.adoc
+++ b/gitops/gitops_push_pull.adoc
@@ -76,7 +76,7 @@ See the following requirements to use the Argo CD _Pull_ model:
 
 - The required hub cluster {ocp-short} {gitops-short} operator must be version 1.9.0 or later.
 
-- The required managed clusters {ocp-short} {gitops-short} operator cannot be an older version than the hub cluster.
+- The version of the {gitops} Operator on the hub cluster must be equal to or more recent than the version of the operator on the managed clusters.
 
 - You need the _ApplicationSet controller_ to propagate the Argo CD application template for a managed cluster.
 
--- a/gitops/gitops_push_pull.adoc
+++ b/gitops/gitops_push_pull.adoc
@@ -76,7 +76,7 @@ See the following requirements to use the Argo CD _Pull_ model:
 
 - The required hub cluster {ocp-short} {gitops-short} operator must be version 1.9.0 or later.
 
-- The required managed clusters {ocp-short} {gitops-short} operator cannot be an older versino than the hub cluster.
+- The required managed clusters {ocp-short} {gitops-short} operator cannot be an older version than the hub cluster.
 
 - You need the _ApplicationSet controller_ to propagate the Argo CD application template for a managed cluster.
 
--- a/gitops/gitops_push_pull.adoc
+++ b/gitops/gitops_push_pull.adoc
@@ -77,7 +77,7 @@ See the following requirements to use the Argo CD _Pull_ model:
 
 - The required hub cluster {ocp-short} GitOps operator must be version 1.9.0 or later.
 
-- The required managed clusters {ocp-short} GitOps operator must be the same version as the hub cluster.
+- The required managed clusters {ocp-short} GitOps operator must be the same version as the hub cluster. You must only meet this requirement when you install the managed clusters and hub cluster. After you install, you do not need to have the same versions. 
 
 - You need the _ApplicationSet controller_ to propagate the Argo CD application template for a managed cluster.
 
--- a/gitops/gitops_push_pull.adoc
+++ b/gitops/gitops_push_pull.adoc
@@ -19,7 +19,7 @@ View the following prerequisites for the Argo CD Pull model:
  
 *Important:* 
 
-* If your `openshift-gitops-ArgoCD-application-controller` service account is _not_ assigned as a cluster administrator, the GitOps application controller might not deploy resources. The application status might send an error similar to the following error:
+* If your `openshift-gitops-ArgoCD-application-controller` service account is _not_ assigned as a cluster administrator, the {gitops} application controller might not deploy resources. The application status might send an error similar to the following error:
 
 ----
 cannot create resource "services" in API group "" in the namespace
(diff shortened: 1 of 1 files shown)


mikeshng (approved):

gitops/gitops_push_pull.adoc
```diff
@@ -77,7 +77,7 @@ See the following requirements to use the Argo CD _Pull_ model:
 
 - The required hub cluster {ocp-short} GitOps operator must be version 1.9.0 or later.
 
-- The required managed clusters {ocp-short} GitOps operator must be the same version as the hub cluster.
+- The required managed clusters {ocp-short} GitOps operator must be the same version as the hub cluster. You must only meet this requirement when you install the managed clusters and hub cluster. After you install, you do not need to have the same versions.
```
swopebe: 1. We have a conref for OCP Gitops, please check for that throughout.
2. You must only meet this requirement is a bit confusing. Are you trying to say you only need the hub cluster and managed cluster to match at install? @jc-berger ? 
3. Please try to avoid the "have" and such in doc. Change it to be concise. This is across the board: "You do not need the same versions."

Generally, i think this can be reduced quite a bit. I am not exactly sure what you are saying here:

The required managed clusters {ocp-short} GitOps operator must be the same version as the hub cluster.

The operator on the managed clusters? must be the same as the operator on the hub?? Is that the requirement? And only at install? (This confuses me, but maybe after install the versions can differ?)

Please check to understand what we are trying to say and maybe edit a bit?

swopebe (changes requested):

gitops/gitops_push_pull.adoc
```diff
@@ -73,18 +72,18 @@ metadata:
...
 See the following requirements to use the Argo CD _Pull_ model:
 
-- The GitOps operator must be installed on the hub cluster and the target managed clusters in the `openshift-gitops` namespace.
+- The {gitops} operator must be installed on the hub cluster and the target managed clusters in the `openshift-gitops` namespace.
 
-- The required hub cluster {ocp-short} GitOps operator must be version 1.9.0 or later.
+- The required hub cluster {ocp-short} {gitops-short} operator must be version 1.9.0 or later.
 
-- The required managed clusters {ocp-short} GitOps operator must be the same version as the hub cluster.
+- The required managed clusters {ocp-short} {gitops-short} operator cannot be an older version than the hub cluster.
```
jc-berger: @mikeshng based off our conversation, I made this change. I can write something this: 

```suggestion
- The required managed clusters {ocp-short} {gitops-short} operator must be the same version of the hub cluster or a newer version. 
```

Not sure if that's as clean though. What do you think? Is it still technically accurate to simply say it cannot be an older version of the hub cluster?

swopebe: The conref is this:

:gitops: Red Hat OpenShift GitOps
:gitops-short: OpenShift GitOps

Why do we have ocp in front of the conref after this change? I think we can remove that, unless there is something I am missing. ? (See suggestion below)

swopebe: I am trying to focus on what you mean by "version," which was written to suggest ACM version previously, but I think we mean _the operator?_ 

**the same version of the hub cluster** (suggests a version of the mch resource, and I don't think that is what we are going for?) 

**the same version on the hub cluster**? (that is something on the ACM hub)

Is this ok?
The required {gitops-short} operator on the managed clusters (plural on purpose?) must be the same version, or a more recent version, of the {gitops-short} operator on the hub cluster.
(?)

swopebe (changes requested):

gitops/gitops_push_pull.adoc
```diff
@@ -73,18 +72,18 @@ metadata:
 
 See the following requirements to use the Argo CD _Pull_ model:
 
-- The GitOps operator must be installed on the hub cluster and the target managed clusters in the `openshift-gitops` namespace.
+- The {gitops} operator must be installed on the hub cluster and the target managed clusters in the `openshift-gitops` namespace.
 
-- The required hub cluster {ocp-short} GitOps operator must be version 1.9.0 or later.
+- The required hub cluster {ocp-short} {gitops-short} operator must be version 1.9.0 or later.
```
swopebe, line 77: OCP in front of gitops.

gitops/gitops_push_pull.adoc
```diff
@@ -109,11 +108,11 @@ For both Push and Pull model, the _Argo CD ApplicationSet controller_ on the hub
 
 - Pull model can provide scalability relief compared to the push model by reducing stress on the controller in the hub cluster, but with more requests and status reporting required.
 
-- With Pull model, {ocp-short} GitOps _does not_ apply resources directly from a centralized hub cluster to the managed clusters. The Argo CD Application is propagated from the hub cluster to the managed clusters.
+- With Pull model, {ocp-short} {gitops-short} _does not_ apply resources directly from a centralized hub cluster to the managed clusters. The Argo CD Application is propagated from the hub cluster to the managed clusters.
```
swopebe, line 111: {ocp-short} {gitops-short} -- 

this is the conref:
:gitops: Red Hat OpenShift GitOps
:gitops-short: OpenShift GitOps

gitops/gitops_push_pull.adoc
mikeshng: > Is this ok?
The required {gitops-short} operator on the managed clusters (plural on purpose?) must be the same version, or a more recent version, of the {gitops-short} operator on the hub cluster.
(?)

The hub's GitOps operator version should be equal or more recent than the other ones.

swopebe: Ok, taking out the possessive, which we don't use, it sounds like that is what we are saying in proposal above.

**More concise:** The version of the {gitops} Operator on the hub cluster must be equal to or more recent than the version of the operator on the managed clusters.

dockerymick (approved):
This looks good to me! [Checklist](https://stolostron.github.io/rhacm-docs/checklist) considered

gitops/gitops_push_pull.adoc
swopebe: @jc-berger @dockerymick--see the comment above that I made when I reviewed. It seems it was merged without this change, and I didn't get to re-review it, which as we said is OK as long as the original reviewer's comments are addressed in the PR. (I may have been out on PTO.)

 **Please see line 62 here and look at the _General use_ column.** https://docs.google.com/spreadsheets/d/1DLS_lS3VKidgZIvcLmLp9BoiqptkvqHWfe1D5FD2kfk/edit?pli=1&gid=1375785039#gid=1375785039

It doesn't make much sense to use {ocp-short} here when that builds as the _long version_ of the conref. Since we are using the _short version_, which is approved, need to remove {ocp-short) This is correct in other places so we just need to remove here.

PR #7993, opened 2025-06-25
https://issues.redhat.com/browse/ACM-18103

dockerymick:


gparvin (approved):
Thanks, I don't know of other changes needed.

nelsonjean (approved):
Both pages appear to reference the same thing and I'm not sure which is the latest. However, gdpr.eu looks pretty good to me.

PR #7999, opened 2025-06-25
2.11 refresh 06/25

dockerymick:


swopebe (approved):

business_continuity/backup_restore/backup_schedule.adoc
```diff
@@ -24,14 +24,12 @@ metadata:
   name: schedule-acm
   namespace: open-cluster-management-backup
 spec:
-  veleroSchedule: "0 */2 * * *" <1>
-  veleroTtl: 120h <2>
-  useManagedServiceAccount: true <3>
+  veleroSchedule: "0 */2 * * *" # <1>
+  veleroTtl: 120h # <2>
+  useManagedServiceAccount: true # <3>
```
swopebe: Before we merge this @dockerymick @jc-berger 

We should **not** be adding comments to the YAML, remember we want the YAML to be clean. We remove comments in drafts. (I didn't see this for peer review or I would have stopped to ask why.) 

Not sure when/why it was added, but let's not add # to the YAML bc it is not part of the code. 

If this is to fix an error, we should actually find what is causing the error and not place a comment to patch it. **This can be done via an issue as tech preview for the next refresh** I also see some callouts in this file using () instead of <>. That might also cause an issue, but not sure.

swopebe: If the * causing the issue, we use quotation marks, remember? That is dev approved.

swopebe (approved):
approved but commented

business_continuity/backup_restore/backup_schedule.adoc
dockerymick: im not sure where the hashtags came from. i can remove them tho. We can check the git blame

dockerymick:
```suggestion
  veleroSchedule: "0 */2 * * *" <1>
  veleroTtl: 120h <2>
  useManagedServiceAccount: true <3>
```

dockerymick: I see. It happened when Dale was helping to fix the Linter errors: https://github.com/stolostron/rhacm-docs/commit/363e561f38fa91e030acf26073516708ad1f2e66

Apologies, I missed that. Great catch and it is updated. I added the quotation marks in a previous PR.

oafischer (approved):

PR #8010, opened 2025-06-27
https://issues.redhat.com/browse/ACM-21889--Add known issue for problem with deleting resources when performing IBBF

amolnar-gh:


release_notes/known_issues_install.adoc
```diff
@@ -70,3 +70,16 @@ oc annotate --overwrite AgentServiceConfig agent unsupported.agent-install.opens
 ----
 
 The agent appears in the inventory when the issue is resolved.
+
+[#install-siteconfig-ibbf]
+== Cannot remove `BareMetalHost` after a cluster reinstallation with Image-Based Break/Fix
+
+When you decommission a cluster and then reinstall the cluster with the Image-Based Break/Fix, the {sco} cannot delete the data image resources from the original cluster because the {sco} is unable to contact the original machine. To work around this issue, remove the finalizer from the data image resources before reinstalling a cluster with Image-Based Break/Fix.
```
trewest: Maybe it would be clearer if we call out that this is specifically related to the bmcAddress field used by the siteconfig operator (e.g. `/nodes/<node-index>/bmcAddress`).  If the user changes this field for performing IBBF and the old address is no longer reachable, they will run into this issue. Is that clearer?

amolnar-gh: Yes. I reworded it:
> If you change the `spec.nodes.<node-id>.bmcAddress` field while reinstalling a cluster with the Image-Based Break/Fix, the {sco} is unable to contact the original machine
and cannot delete the data image resources from the original cluster. To work around this issue, remove the finalizer from the data image resources before reinstalling a cluster with Image-Based Break/Fix.

trewest (approved):
/lgtm

release_notes/known_issues_install.adoc
```diff
@@ -70,3 +70,17 @@ oc annotate --overwrite AgentServiceConfig agent unsupported.agent-install.opens
 ----
 
 The agent appears in the inventory when the issue is resolved.
+
+[#install-siteconfig-ibbf]
+== Cannot remove data image resources after a cluster reinstallation with Image-Based Break/Fix
+
+If you change the `spec.nodes.<node-id>.bmcAddress` field while reinstalling a cluster with the Image-Based Break/Fix, the {sco} is unable to contact the original machine
+and cannot delete the data image resources from the original cluster. To work around this issue, remove the finalizer from the data image resources before reinstalling a cluster with Image-Based Break/Fix.
```
oafischer: There might be an unintentional line break here

release_notes/known_issues_install.adoc
```diff
@@ -70,3 +70,17 @@ oc annotate --overwrite AgentServiceConfig agent unsupported.agent-install.opens
...
+
+* To remove the finalizer from the data image resources, run the following command:
+
++
+----
+$ oc patch dataimages.metal3.io -n target-0 target-0-0 --patch '{"metadata":{"finalizers":[]}}' --type merge 
+dataimage.metal3.io/target-0-0 patched
+----
```
oafischer:
```suggestion
+
[source,bash]
----
oc patch dataimages.metal3.io -n target-0 target-0-0 --patch '{"metadata":{"finalizers":[]}}' --type merge 
dataimage.metal3.io/target-0-0 patched
----
```

oafischer (approved):
Some formatting suggestions, otherwise approved!

PR #8020, opened 2025-07-02
https://issues.redhat.com/browse/ACM-21150 Update Policy CLI doc

oafischer:
2.14 enhancement

[Checklist](https://stolostron.github.io/rhacm-docs/checklist) considered

### DIFF
--- a/governance/manage_policies.adoc
+++ b/governance/manage_policies.adoc
@@ -344,7 +344,7 @@ To resolve a policy with hub cluster templates, you must provide the `--cluster-
 If the input policy uses the `.ObjectName` or `.ObjectNamespace` template variables, you need to set the values with the `--object-name` and `--object-namespace` arguments.
 
 +
-The `template-resolver` connects to the managed cluster and hub cluster to retrieve information necessary to resolve the templates. You can view the resources it retrieved by providing paths via the `-save-resources` and `-save-hub-resources` arguments. The resulting files can be used as inputs by the `dryrun` subcommand.
+The `template-resolver` connects to the managed cluster and hub cluster to retrieve information necessary to resolve the templates. You can view the resources it retrieved by providing paths via the `--save-resources` and `--save-hub-resources` arguments. The resulting files can be used as inputs by the `dryrun` subcommand.
 
 `dryrun`:: The `dryrun` is a subcommand for `policytools` that runs a `ConfigurationPolicy` locally by using static input resources for the cluster state.
 
--- a/governance/manage_policies.adoc
+++ b/governance/manage_policies.adoc
@@ -352,7 +352,7 @@ The `template-resolver` connects to the managed cluster and hub cluster to retri
 To run a policy, add the path to the `ConfigurationPolicy` YAML by using the `--policy` flag. You can provide additional resource YAML files, which are used as the current cluster state. For example, if your policy defines a specific configuration for a namespace, provide the current namespace state to see what the policy does. Multiple resources can be in the same resource file.
 
 +
-By default, `dryrun` prints the resource diffs for any objects defined in the policu and the compliance messages that the `ConfigurationPolicy` releases. To save the compliance messages to a separate file, use the `--messages-path` argument. To save the full policy status, use the `--status-path` argument.
+By default, `dryrun` prints the resource diffs for any objects defined in the policy and the compliance messages that the `ConfigurationPolicy` releases. To save the compliance messages to a separate file, use the `--messages-path` argument. To save the full policy status, use the `--status-path` argument.
 
 +
 If the policy is `NonCompliant` based on the input, the tool returns a non-zero exit code. To compare the resulting status against a known status, use the `--desired-status` argument. For example, to verify that the policy reports a missing resource, provide a YAML file with the missing resource inside a `.relatedObjects` list. To verify that the policy is `NonCompliant`, create a YAML file with the `compliant: NonCompliant` line.
--- a/governance/manage_policies.adoc
+++ b/governance/manage_policies.adoc
@@ -14,7 +14,7 @@ View the following sections:
 [#creating-a-security-policy]
 == Creating a security policy
 
-You can create a security policy from the command line interface (CLI) or from the console.
+You can create a security policy from the command line or from the console.
 
 *Required access:* Cluster administrator
 
@@ -25,7 +25,7 @@ You can create a security policy from the command line interface (CLI) or from t
 [#creating-a-security-policy-from-the-command-line-interface]
 === Creating a security policy from the command line interface
 
-Complete the following steps to create a policy from the command line interface (CLI):
+Complete the following steps to create a policy from the command line:
 
 . Create a policy by running the following command:
 +
@@ -92,9 +92,9 @@ subjects:
 ----
 
 [#viewing-your-security-policy-from-the-cli]
-==== Viewing your security policy from the CLI
+==== Viewing your security policy from the command line
 
-Complete the following steps to view your security policy from the CLI:
+Complete the following steps to view your security policy from the command line:
 
 . View details for a specific security policy by running the following command:
 +
@@ -206,11 +206,11 @@ View any security policy and the status from the console.
 . Alternatively, select the _Policies_ tab to view the list of policies. Expand a policy row to view the _Description_, _Standards_, _Controls_, and _Categories_ details.
 
 [#creating-policy-sets-cli]
-=== Creating policy sets from the CLI
+=== Creating policy sets from the command line
 
 By default, the policy set is created with no policies or placements. You must create a placement for the policy set and have at least one policy that exists on your cluster. When you create a policy set, you can add numerous policies. 
 
-Run the following command to create a policy set from the CLI:
+Run the following command to create a policy set from the command line:
 
 ----
 oc apply -f <policyset-filename>
@@ -235,7 +235,7 @@ Your policy is listed from the policy table.
 Learn to update security policies.
 
 [#adding-policy-set-cli]
-=== Adding a policy to a policy set from the CLI
+=== Adding a policy to a policy set from the command line
 
 . Run the following command to edit your policy set: 
 
@@ -279,9 +279,9 @@ Click *Disable policy*. Your policy is disabled.
 [#deleting-a-security-policy]
 == Deleting a security policy
 
-Delete a security policy from the CLI or the console.
+Delete a security policy from the command line or the console.
 
-Use the following procedure to delete from the CLI:
+Use the following procedure to delete from the command line:
 
 . Delete a security policy by running the following command:
 
@@ -331,9 +331,9 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
 - As you disable the `config-policy-addon` resource on the managed cluster, the `pruneObjbectBehavior` is ignored. To automatically clean up the related resources on the policies, you must remove the policies from the managed cluster before the add-on is disabled.
 
 [#policy-cli-commands]
-== Policy command line interface
+== Policy command line
 
-With the `policytools` command line interface (CLI), you can interact with policies locally to help with creating and debugging.  The `policytools` CLI is available for download from the hub cluster console. See link:../console/console.adoc#command-line-tools[Command line tools].
+With the `policytools` command line, you can interact with policies locally to help with creating and debugging.  The `policytools` command line is available for download from the hub cluster console. See link:../console/console.adoc#command-line-tools[Command line tools].
 
 `template-resolver`:: The `template-resolver` is a subcommand for `policytools` that resolves managed cluster and hub cluster templates that are embedded in policies. The `template-resolver` reads from either a file, or from standard input.
 
--- a/governance/manage_policies.adoc
+++ b/governance/manage_policies.adoc
@@ -349,7 +349,7 @@ The `template-resolver` connects to the managed cluster and hub cluster to retri
 `dryrun`:: The `dryrun` is a subcommand for `policytools` that runs a `ConfigurationPolicy` locally by using static input resources for the cluster state.
 
 +
-To execute a policy, you must add the path to the `ConfigurationPolicy` YAML by using the `--policy` flag. You can provide additional resource YAML files, which are used as the current cluster state. For example, if your policy defines a specific configuration for a namespace, you provide the current state of the namespace to see what the policy does. Multiple resources can be in the same resource file.
+To run a policy, add the path to the `ConfigurationPolicy` YAML by using the `--policy` flag. You can provide additional resource YAML files, which are used as the current cluster state. For example, if your policy defines a specific configuration for a namespace, provide the current namespace state to see what the policy does. Multiple resources can be in the same resource file.
 
 +
 By default, `dryrun` prints the resource diffs for any objects defined in the policu and the compliance messages that the `ConfigurationPolicy` releases. To save the compliance messages to a separate file, use the `--messages-path` argument. To save the full policy status, use the `--status-path` argument.
--- a/governance/manage_policies.adoc
+++ b/governance/manage_policies.adoc
@@ -352,7 +352,7 @@ The `template-resolver` connects to the managed cluster and hub cluster to retri
 To run a policy, add the path to the `ConfigurationPolicy` YAML by using the `--policy` flag. You can provide additional resource YAML files, which are used as the current cluster state. For example, if your policy defines a specific configuration for a namespace, provide the current namespace state to see what the policy does. Multiple resources can be in the same resource file.
 
 +
-By default, `dryrun` prints the resource diffs for any objects defined in the policy and the compliance messages that the `ConfigurationPolicy` releases. To save the compliance messages to a separate file, use the `--messages-path` argument. To save the full policy status, use the `--status-path` argument.
+By default, `dryrun` prints the resource differences for any objects defined in the policy and the compliance messages that the `ConfigurationPolicy` releases. To save the compliance messages to a separate file, use the `--messages-path` argument. To save the full policy status, use the `--status-path` argument.
 
 +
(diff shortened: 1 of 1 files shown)


governance/manage_policies.adoc
```diff
@@ -333,14 +333,29 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
...
+With the `policytools` command line interface (CLI), you can interact with policies locally to help with creating and debugging.  The `policytools` CLI is available for download from the hub cluster console. See link:../console/console.adoc#command-line-tools[Command line tools].
 
 `template-resolver`:: The `template-resolver` is a subcommand for `policytools` that resolves managed cluster and hub cluster templates that are embedded in policies. The `template-resolver` reads from either a file, or from standard input.
 
 +
 To resolve a policy with hub cluster templates, you must provide the `--cluster-name` argument with the name of a managed cluster that is imported into {acm-short}, and you must provide the `--hub-kubeconfig` argument with the path to a `kubeconfig` file that references the hub cluster.
 
-The `policytools` CLI is available for download from the hub cluster console. See link:../console/console.adoc#command-line-tools[Command line tools].
++
+If the input policy uses the `.ObjectName` or `.ObjectNamespace` template variables, you need to set the values with the `-object-name` and `-object-namespace` arguments.
```
JustinKuli:
```suggestion
If the input policy uses the `.ObjectName` or `.ObjectNamespace` template variables, you need to set the values with the `--object-name` and `--object-namespace` arguments.
```

governance/manage_policies.adoc
```diff
@@ -333,14 +333,29 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
...
+
++
+The `template-resolver` connects to the managed cluster and hub cluster to retrieve information necessary to resolve the templates. You can view the resources it retrieved by providing paths via the `-save-resources` and `-save-hub-resources` arguments. The resulting files can be used as inputs by the `dryrun` subcommand.
```
JustinKuli:
```suggestion
The `template-resolver` connects to the managed cluster and hub cluster to retrieve information necessary to resolve the templates. You can view the resources it retrieved by providing paths via the `--save-resources` and `--save-hub-resources` arguments. The resulting files can be used as inputs by the `dryrun` subcommand.
```

governance/manage_policies.adoc
```diff
@@ -333,14 +333,29 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
...
+
+`dryrun`:: The `dryrun` is a subcommand for `policytools` that runs a `ConfigurationPolicy` locally by using static input resources for the cluster state.
+
++
+To execute a policy, you must add the path to the `ConfigurationPolicy` YAML by using the `--policy` flag. You can provide additional resource YAML files, which are used as the current cluster state. For example, if your policy defines a specific configuration for a namespace, you provide the current state of the namespace to see what the policy does. Multiple resources can be in the same resource file.
+
++
+By default, `dryrun` prints the resource diffs for any objects defined in the policu and the compliance messages that the `ConfigurationPolicy` releases. To save the compliance messages to a separate file, use the `-messages-path` argument. To save the full policy status, use the `-status-path` argument.
```
JustinKuli:
```suggestion
By default, `dryrun` prints the resource diffs for any objects defined in the policu and the compliance messages that the `ConfigurationPolicy` releases. To save the compliance messages to a separate file, use the `--messages-path` argument. To save the full policy status, use the `--status-path` argument.
```

JustinKuli (commented):
Looks good overall to me! All of the flags should start with double hyphens, but I think the info in Jira wasn't very well formatted :grimacing:

oafischer:
Updated, thanks @JustinKuli!

JustinKuli (approved):

drkho:
Overall, looks good to QE. There is a failed check though, fyi @oafischer

governance/manage_policies.adoc
```diff
@@ -333,14 +333,29 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
 [#policy-cli-commands]
 == Policy command line interface
 
-With the `policytools` command line interface (CLI), you can interact with policies locally to help with creating and debugging. 
+With the `policytools` command line interface (CLI), you can interact with policies locally to help with creating and debugging.  The `policytools` CLI is available for download from the hub cluster console. See link:../console/console.adoc#command-line-tools[Command line tools].
```
swopebe, line 337: Style guide:

Use with caution CLI noun
In most cases, use command line unless you need to make specific reference to the command-line interface, as opposed to the graphical interface.

swopebe: Change all occurrances.

governance/manage_policies.adoc
```diff
@@ -333,14 +333,29 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
...
 
 +
 To resolve a policy with hub cluster templates, you must provide the `--cluster-name` argument with the name of a managed cluster that is imported into {acm-short}, and you must provide the `--hub-kubeconfig` argument with the path to a `kubeconfig` file that references the hub cluster.
 
-The `policytools` CLI is available for download from the hub cluster console. See link:../console/console.adoc#command-line-tools[Command line tools].
++
+If the input policy uses the `.ObjectName` or `.ObjectNamespace` template variables, you need to set the values with the `--object-name` and `--object-namespace` arguments.
+
++
+The `template-resolver` connects to the managed cluster and hub cluster to retrieve information necessary to resolve the templates. You can view the resources it retrieved by providing paths via the `--save-resources` and `--save-hub-resources` arguments. The resulting files can be used as inputs by the `dryrun` subcommand.
```
swopebe:
```suggestion
The `template-resolver` connects to the managed cluster and hub cluster to retrieve information that is necessary to resolve the templates. You can view the resources it retrieved by providing paths from the `--save-resources` and `--save-hub-resources` arguments. The resulting files can be used as inputs by the `dryrun` subcommand.
```

swopebe: we don't write "via"...always change that.

swopebe: Confusing line here:
The resulting files can be used as inputs by the `dryrun` subcommand.

What are we trying to say? Make this active.

swopebe: Resulting from what?

Use the files from ? as `dryrun` subcommand inputs. (?) I am just not sure what this line means.

swopebe: Also this can be active and reduced:

You can view the resources it retrieved by providing paths from the `--save-resources` and `--save-hub-resources` arguments.

Not a good idea to use "it" -- name the thing you are talking about.

Provide paths from the X and Z resource arguments to view resources. (example)

The readability from this is a bit difficult and I think we need to understand what this means so we can write it as if we understand it.

governance/manage_policies.adoc
```diff
@@ -333,14 +333,29 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
...
+
+`dryrun`:: The `dryrun` is a subcommand for `policytools` that runs a `ConfigurationPolicy` locally by using static input resources for the cluster state.
```
swopebe: runs a `ConfigurationPolicy` _____ locally

governance/manage_policies.adoc
```diff
@@ -333,14 +333,29 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
...
+
++
+To execute a policy, you must add the path to the `ConfigurationPolicy` YAML by using the `--policy` flag. You can provide additional resource YAML files, which are used as the current cluster state. For example, if your policy defines a specific configuration for a namespace, you provide the current state of the namespace to see what the policy does. Multiple resources can be in the same resource file.
```
swopebe: please remove execute. Run is sufficient.

swopebe: Do we understand what 352 is saying and if so, can we reduce it a bit?

governance/manage_policies.adoc
```diff
@@ -333,14 +333,29 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
...
+
++
+By default, `dryrun` prints the resource diffs for any objects defined in the policu and the compliance messages that the `ConfigurationPolicy` releases. To save the compliance messages to a separate file, use the `--messages-path` argument. To save the full policy status, use the `--status-path` argument.
```
swopebe: diffs?

governance/manage_policies.adoc
```diff
@@ -333,14 +333,29 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
...
+
++
+If the policy is `NonCompliant` based on the input, the tool returns a non-zero exit code. To compare the resulting status against a known status, use the `--desired-status` argument. For example, to verify that the policy reports a missing resource, provide a YAML file with the missing resource inside a `.relatedObjects` list. To verify that the policy is `NonCompliant`, create a YAML file with the `compliant: NonCompliant` line.
```
swopebe, line 358: non-zero? first time i have seen that term.

swopebe (changes requested):
Some of the paragraphs are just a bit hard to understand and can be reduced or changed for clarity. CLI guidance added.

governance/manage_policies.adoc
JustinKuli: In this case, "via" is closer to "by way of", not "from".

Part of the idea is, our policies can have complicated templates inside of them, with multiple lookups of different resources on the clusters. The `template-resolver` resolves those templates by connecting to clusters in order to read those resources, and basically it then 'evaluates' those templates, resulting in simpler YAML.

The new thing we added was a way to inspect what the `template-resolver` read from the clusters: if the user sets a path with the `--save-resources` flag, then `template-resolver` will write a file to that location containing the YAML of the resources that it read in order to resolve the template. That resulting file is what can be used as an input to `dryrun`, which allows the user to basically repeat the evaluation without connecting to a cluster.

governance/manage_policies.adoc
oafischer: Updated

governance/manage_policies.adoc
oafischer: @swopebe @JustinKuli I updated this based on the suggestions and explanation, let me know if it looks good now. Thanks!

governance/manage_policies.adoc
oafischer: Updated

governance/manage_policies.adoc
oafischer: Updated and reduced.

governance/manage_policies.adoc
oafischer: Updated to differences

governance/manage_policies.adoc
```diff
@@ -331,16 +331,31 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
 - As you disable the `config-policy-addon` resource on the managed cluster, the `pruneObjbectBehavior` is ignored. To automatically clean up the related resources on the policies, you must remove the policies from the managed cluster before the add-on is disabled.
 
 [#policy-cli-commands]
-== Policy command line interface
+== Policy command line
 
-With the `policytools` command line interface (CLI), you can interact with policies locally to help with creating and debugging. 
+With the `policytools` command line, you can interact with policies locally to help with creating and debugging.  The `policytools` command line is available for download from the hub cluster console. See link:../console/console.adoc#command-line-tools[Command line tools].
```
JustinKuli:
```suggestion
With the `policytools` utility, you can interact with policies locally to help with creating and debugging.  The `policytools` utility is available for download from the hub cluster console. See link:../console/console.adoc#command-line-tools[Command line tools].
```

The other changes from "CLI" to "command line" seem fine, but this one feels different to me... the phrase "command line" seems more like the terminal itself or the shell (bash, zsh), but this `policytools` thing is something you run *on* or *in* the command line. Calling it the `policytools` command line seems confusing to me. I think "utility" or "tool" or something similar is more accurate.

By the way, I was looking at other similar documentation, just to see how other people have written about this sort of thing. `oc` seems to use "CLI" a lot, but it seems to be because the full name of it is "OpenShift CLI" (see [their docs](https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/cli_tools/openshift-cli-oc)). Another thing, `tkn`, uses "CLI" and "CLI tool" somewhat haphazardly (see the [OpenShift Pipelines tkn CLI docs](https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/cli_tools/pipelines-cli-tkn#op-configuring-tkn)).

governance/manage_policies.adoc
```diff
@@ -331,16 +331,31 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
...
 
 +
 To resolve a policy with hub cluster templates, you must provide the `--cluster-name` argument with the name of a managed cluster that is imported into {acm-short}, and you must provide the `--hub-kubeconfig` argument with the path to a `kubeconfig` file that references the hub cluster.
 
-The `policytools` CLI is available for download from the hub cluster console. See link:../console/console.adoc#command-line-tools[Command line tools].
++
+If the input policy uses the `.ObjectName` or `.ObjectNamespace` template variables, you need to set the values with the `--object-name` and `--object-namespace` arguments.
+
++
+The `template-resolver` connects to the managed cluster and hub cluster to retrieve information that is necessary to resolve the templates. Provide paths by using the `--save-resources` and `--save-hub-resources` arguments to view the resources that the `template-resolver` retrievs. Use the `template-resolver` files as inputs for the `dryrun` subcommand.
```
JustinKuli: This might be more precise:
```suggestion
The `template-resolver` subcommand connects to the managed cluster and hub cluster to retrieve information that is necessary to resolve the templates. Provide paths by using the `--save-resources` and `--save-hub-resources` arguments to save the resources that the `template-resolver` retrieves. Those saved files can be used as inputs for the `dryrun` subcommand.
```

governance/manage_policies.adoc
```diff
@@ -331,16 +331,31 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
...
+
+`dryrun`:: The `dryrun` is a subcommand for `policytools` that runs a `ConfigurationPolicy` resource locally by using static input resources for the cluster state.
```
JustinKuli:
```suggestion
`dryrun`:: The `dryrun` is a subcommand for `policytools` that evaluates a `ConfigurationPolicy` resource locally by using static input resources for the cluster state.
```

One of the previous review comments made me think that "run" was unclear - "evaluate" is possibly more accurate and descriptive.

governance/manage_policies.adoc
```diff
@@ -331,16 +331,31 @@ From the console, you can select the value in the _Prune Object Behavior_ sectio
...
+
++
+To run a policy, add the path to the `ConfigurationPolicy` YAML by using the `--policy` flag. You can provide additional resource YAML files, which are used as the current cluster state. For example, if your policy defines a specific configuration for a namespace, provide the current namespace state to see what the policy does. Multiple resources can be in the same resource file.
+
++
+By default, `dryrun` prints the resource differences for any objects defined in the policy and the compliance messages that the `ConfigurationPolicy` releases. To save the compliance messages to a separate file, use the `--messages-path` argument. To save the full policy status, use the `--status-path` argument.
```
JustinKuli:
```suggestion
By default, `dryrun` prints the differences between the desired states of objects defined in the policy and the current state of those resources, as well as compliance messages that the `ConfigurationPolicy` creates. To save the compliance messages to a separate file, use the `--messages-path` argument. To save the full policy status, including additional information about the related objects, use the `--status-path` argument.
```

Sorry to keep suggesting more details - I just keep finding little pieces that I want to clarify. At any point, I think I would consider the doc "good enough".

governance/manage_policies.adoc
JustinKuli: It's relatively standard among unix/linux things run on the command line: a "success" is represented by exit code 0, while any errors or failures use a non-zero code. Generally you don't see the exit code directly, but scripts which may run multiple commands/tools might stop in the middle of the script if something has a non-zero exit. For example, if a script uses `wget` to download a file from a url, but that url  doesn't actually exist, then `wget` uses a non-zero exit code to signal to the script that something went wrong.

So the idea here was that `policytools dryrun` would use a non-zero exit code to signal that the policy would not be compliant, and if that was run inside a user's CICD pipeline for changes to Policies, things would generally fail and the change wouldn't be merged. On the other hand, if the user wants to ensure that a ConfigurationPolicy would flag a hypothetical situation by being NonCompliant, they would want their CICD pipeline to fail if the policy was actually Compliant; in general the `--desired-status` allows users to specify any details of the status that they are interested in, not just Compliant/NonCompliant.

This paragraph gets into some technical details, and I expected that users wanting to use these features would already be well-versed in the ConfigurationPolicy and command line details that I omitted.

JustinKuli (commented):
Thanks for the updates - sorry I hadn't provided enough details before. I've got a few new suggestions and provided some extra background in some comments. I might still be missing some context I should be providing, so please feel free to ask for clarification on any remaining points.

governance/manage_policies.adoc
swopebe: You can also use "command-line interface" or "command-line tools" as the manual points out. Since you are interacting with it, that rule applies. My point is that we need to not use CLI and use the term that applies throughout.

swopebe (approved):
I am approving but see the comment to @JustinKuli about the option to use command-line tools -- and interface.

governance/manage_policies.adoc
oafischer: Updated to command-line tool

governance/manage_policies.adoc
oafischer: Updated

governance/manage_policies.adoc
oafischer: Updated

governance/manage_policies.adoc
oafischer: No worries, best to discuss this stuff in the PR and get it right here. Updated!

oafischer:
@JustinKuli Let me know if the latest updates look good. Thanks for the reviews everyone!

JustinKuli (approved):

PR #8016, opened 2025-06-30
https://issues.redhat.com/browse/ACM-20521

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-20521 

See draft PR: https://github.com/stolostron/rhacm-docs/pull/8014 

Consulted the Doc Peer Review Checklist: https://stolostron.github.io/rhacm-docs/checklist

PR #8030, opened 2025-07-09
added 2.14 changes to 2.13

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-21907 

See PR for 2.14: 

https://github.com/stolostron/rhacm-docs/pull/8016/files 
https://github.com/stolostron/rhacm-docs/pull/8029/files

### DIFF
--- a/release_notes/known_issues_global.adoc
+++ b/release_notes/known_issues_global.adoc
@@ -40,6 +40,42 @@ data-kafka-zookeeper-1                          Bound    pvc-1de22549-1b55-4cf5-
 data-kafka-zookeeper-2                          Bound    pvc-a9a3d47b-297d-4a2d-85d1-76a89b25b3c1   10Gi       RWO            gp3-csi        <unset>                 86s
 ----
 
+[#incomplete-data-dashboard]
+== Incomplete data display in Grafana dashboards 
+//2.13:20521
+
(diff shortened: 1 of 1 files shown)


release_notes/known_issues_global.adoc
```diff
@@ -40,6 +40,42 @@ data-kafka-zookeeper-1                          Bound    pvc-1de22549-1b55-4cf5-
...
+The Kafka message size limit might cause the Grafana dashboards to not display complete data. The default limit is `1 megabyte (MB)`. Verify if the total size of your policy or managed cluster data exceeds `1 MB`. If it does, increase the Kafka message size limit to meet your needs. 
+
+Modify the Kafka configuration by completing the following steps: 
+
+. To go to the Kafka container resource, run the following command: 
+
++
+[source,bash]
+----
+oc edit kafka kafka -n multicluster-global-hub
```
jc-berger, line 56: @clyang82 do we need the double "kafka" here? 

or should it be:

oc edit kafka -n multicluster-global-hub

swopebe: or is the second kafka supposed to be a variable?

clyang82: we need both. the first one is resource. the second one is name.
but it works for your command, because there is only one resource in that namespace.
I would like to use `oc edit kafka kafka -n multicluster-global-hub` which is more accurate.

PR #8064, opened 2025-07-21
https://issues.redhat.com/browse/ACM-20220 Fix content

oafischer:
2.14 only

Checklist considered


swopebe (approved):
I like these changes. If the engineers don't like "gathering" we can stick with accessing on those sub headers.

xuezhaojun:
Agree！
The PR looks good to me! Thanks for the update

PR #8092, opened 2025-07-30
[ACM-22742] GRC known issue: Misaligned `objectSelector`

dhaiducek:
Add known issue for GRC around the `objectSelector` and `namespaceSelector`.

ref: https://issues.redhat.com/browse/ACM-22742

### DIFF
--- a/release_notes/known_issues_governance.adoc
+++ b/release_notes/known_issues_governance.adoc
@@ -27,6 +27,12 @@ For your {ocp} cluster, see link:https://docs.redhat.com/documentation/en-us/ope
 
 For more about deprecations and removals, see xref:../release_notes/acm_deprecate_remove.adoc#deprecations-removals-acm[Deprecations and removals for {acm-short}].
 
+[#configurationpolicy-misinterprets-results]
+== _ConfigurationPolicy_ misinterprets results from the _objectSelector_ and _namespaceSelector_
+// 2.14:22676
+
+When you use both the `objectSelector` and the `namespaceSelector` fields in a `ConfigurationPolicy` resource, the objects that the `objectSelector` return get applied to all the namespaces that the `namespaceSelector` return. The `ConfigurationPolicy` misinterprets the results of these objects. To workaround this issue, apply the `object-templates-raw` field to iterate over the objects.
+
 [#config-policy-stuck]
 == Configuration policy listed complaint when namespace is stuck in _Terminating_ state
 //2.2:20715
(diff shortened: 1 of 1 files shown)


release_notes/known_issues_governance.adoc
```diff
@@ -27,6 +27,12 @@ For your {ocp} cluster, see link:https://docs.redhat.com/documentation/en-us/ope
 
 For more about deprecations and removals, see xref:../release_notes/acm_deprecate_remove.adoc#deprecations-removals-acm[Deprecations and removals for {acm-short}].
 
+[#configurationpolicy-misinterprets-results]
+== _ConfigurationPolicy_ misinterprets results from the _objectSelector_ and _namespaceSelector_
```
jc-berger: @dhaiducek does "misinterpret" capture the issue? I think that's a better verb choice than "misalign" 

Or is it more so that the configurationpolicy is unable to process the results from the objects because the objects are getting applied to all the namespaces? If it's more a matter of the configurationpolicy being unable to process the results, I can state that instead of "misinterpret" or "misalign" 

Thanks!

dhaiducek: Hmmm. I think either misinterprets or misaligns would be fine. Or even mishandles?

To illustrate what happens, if a `namespaceSelector` returns say `namespace1` through `namespace5` but the `objectSelector` returns a ConfigMap `my-stuff` that's only in `namespace1`, the controller still applies `my-stuff` to all namespaces `namespace1` through `namespace5` instead of only handling it in `namespace1` as it should.

swopebe: Hi, guys! Since these are both anthropomorphism, we may need to just reorder this header to try to avoid that.

some great examples and the standard:

https://www.ibm.com/docs/en/ibm-style?topic=grammar-anthropomorphism

release_notes/known_issues_governance.adoc
```diff
@@ -27,6 +27,12 @@ For your {ocp} cluster, see link:https://docs.redhat.com/documentation/en-us/ope
...
+// 2.14:22676
+
+When you use both the `objectSelector` and the `namespaceSelector` fields in a `ConfigurationPolicy` resource, the objects that the `objectSelector` return get applied to all the namespaces that the `namespaceSelector` return. The `ConfigurationPolicy` misinterprets the results of these objects. To workaround this issue, apply the `object-templates-raw` field to iterate over the objects.
```
swopebe:
```suggestion
When you use both the `objectSelector` and the `namespaceSelector` fields in a `ConfigurationPolicy` resource, the objects that the `objectSelector` return are applied to all the namespaces that the `namespaceSelector` return. The `ConfigurationPolicy` misinterprets the results of these objects. To workaround this issue, apply the `object-templates-raw` field to iterate over the objects.
```

swopebe: other than "misinterprets" -- let's see what we can do with that to change it to meet the standard.

swopebe (changes requested):
Peer review started

release_notes/known_issues_governance.adoc
dhaiducek: So something like this to make the actor and verb more specific?

> The `ConfigurationPolicy` controller incorrectly processes results from the _objectSelector_ and _namespaceSelector_

swopebe: The `ConfigurationPolicy` controller incorrectly processes `objectSelector` and `namespaceSelector` results.

?

only bc it is shorter, but I am ok if the first one is preferred.

swopebe (approved):

dockerymick (approved):

PR #8094, opened 2025-07-31
https://issues.redhat.com/browse/ACM-22763# Submariner note

swopebe:


release_notes/acm_new.adoc
```diff
@@ -9,6 +9,8 @@ Access the link:https://access.redhat.com/articles/7120842[{acm-short} Support M
 
 *Important:*
 
+- Submariner version 0.21.0 is not released at the same time as {acm-short} 2.14.
```
dfarrell07: Do you think we should add something like: "Users of Submariner with ACM should not update ACM until Submariner is released."

Hopefully not for long, but we want to make sure of quality with the new Konflux-built artifacts.

swopebe: Submariner version 0.21.0 is not released at the same time as {acm-short} 2.14. Do not update to {acm-short} version 2.14 until the latest Submariner version is released and announced.

dfarrell07 (approved):
Looks great, thanks!

oafischer (approved):
lgtm

PR #8105, opened 2025-08-06
https://issues.redhat.com/browse/ACM-17011

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-17011

### DIFF
--- a/global_hub/global_hub_requirements.adoc
+++ b/global_hub/global_hub_requirements.adoc
@@ -1,28 +1,23 @@
 [#global-hub-requirements]
 = {global-hub} requirements
 
-Learn about the {global-hub} requirements for installing and networking, and review the components and environments that are supported by {global-hub}. See the following requirements: 
-
-* <<gh-prerequisites,Prerequisites>>
-* <<gh-network-requirements,Networking requirements>>
-* <<gh-supported-components,Supported components>>
-
-[#gh-prerequisites]
-== Prerequisites
-
-To install {global-hub}, see the following prerequisites:
+See the following general {global-hub} requirements:
 
 * *Required access:* Cluster administrator
-* *{ocp-short} Dedicated environment required access:* You must have `cluster-admin` permissions. By default `dedicated-admin` role does not have the required permissions to create namespaces in the {ocp-short} Dedicated environment. 
-* {acm} must be installed and configured.
+* *{ocp-short} Dedicated environment required access:* You must have `cluster-admin` permissions. By default `dedicated-admin` role does not have the required permissions to create namespaces in the {ocp-short} dedicated environment. 
+* You must install and configure {acm}.
 
+Learn about the {global-hub} requirements for installing and networking, and review the components and environments that are supported by {global-hub}:
+
+* <<gh-network-requirements,Networking requirements>>
+* <<gh-supported-components,Supported components>>
 
 [#gh-network-requirements]
 == Networking requirements
 
 See the following networking requirements:
 
-* The managed hub is also a managed cluster of {global-hub} in {acm-short}. You must configure the network component in {acm-short}. For {acm-short} networking details, see link:../networking/networking_intro.adoc#networking[Networking]. 
+* The managed hub cluster is also a {global-hub} managed cluster in {acm-short}. You must configure the network component in {acm-short}. For {acm-short} networking details, see link:../networking/networking_intro.adoc#networking[Networking]. 
 
 * To review the {global-hub} network information, see the following table:
 
@@ -31,8 +26,8 @@ See the following networking requirements:
 | Direction | Protocol | Connection | Port that is specified | Source address |	Destination address |
     
 | Inbound from the browser | HTTPS | Access the Grafana dashboard | 443 | Browser|Grafana route IP address |
-| Outbound to Kafka Cluster | HTTPS | The {global-hub} manager needs to receive data from Kafka cluster | 443 | multicluster-global-hub-manager-xxx pod | Kafka route host |
-| Outbound to PostgreSQL database | HTTPS | The {global-hub} needs to contribute data to the PostgreSQL database | 443 | multicluster-global-hub-manager-xxx pod | PostgreSQL database IP address |
+| Outbound to Kafka Cluster | HTTPS | The {global-hub} manager needs to receive data from Kafka cluster | 443 | `multicluster-global-hub-manager-xxx` pod | Kafka route host |
+| Outbound to PostgreSQL database | HTTPS | The {global-hub} needs to contribute data to the PostgreSQL database | 443 | `multicluster-global-hub-manager-xxx` pod | PostgreSQL database IP address |
 |===
 +
 
@@ -42,7 +37,7 @@ See the following networking requirements:
 |===
 | Direction | Protocol | Connection | Port that is specified | Source address |	Destination address |
     
-| Outbound to Kafka Cluster | HTTPS | The {global-hub} agent needs to sync cluster and policy information to Kafka cluster | 443 | multicluster-global-hub-agent pod | Kafka route host |
+| Outbound to Kafka Cluster | HTTPS | The {global-hub} agent needs to sync cluster and policy information to Kafka cluster | 443 | `multicluster-global-hub-agent` pod | Kafka route host |
 |===
 +
 
--- a/global_hub/global_hub_requirements.adoc
+++ b/global_hub/global_hub_requirements.adoc
@@ -58,7 +58,7 @@ See the following components that {global-hub} supports:
 
 * Because they share an integrated console, the {global-hub} and {ocp-short} console share an integrated console, so they support the same browser. For information about supported browsers and versions, see link:https://docs.redhat.com/documentation/en-us/openshift_container_platform/4.17/html-single/web_console/index#web-console[Accessing the web console] in the {ocp} documentation
 
-* See following table for a list of platforms that are available for the supported {global-hub} cluster:
+* See the following table of supported platforms for the {global-hub} cluster:
 
 +
 |===
--- a/global_hub/global_hub_requirements.adoc
+++ b/global_hub/global_hub_requirements.adoc
@@ -1,31 +1,28 @@
 [#global-hub-requirements]
-= Global hub requirements
+= {global-hub} requirements
 
-Learn about the {global-hub} requirements for installing and networking, and review the components and environments that are supported by {global-hub}. Complete the following sections: 
+Learn about the {global-hub} requirements for installing and networking, and review the components and environments that are supported by {global-hub}. See the following requirements: 
 
-* <<gh-general-requirements,General requirements>>
+* <<gh-prerequisites,Prerequisites>>
 * <<gh-network-requirements,Networking requirements>>
 * <<gh-supported-components,Supported components>>
 
-[#gh-general-requirements]
-== General requirements
+[#gh-prerequisites]
+== Prerequisites
 
-To install Global hub, you need the following requirements:
+To install {global-hub}, see the following prerequisites:
 
 * *Required access:* Cluster administrator
-//this is normally at the beginning of a procedure and all of this as prereqs.
-
 * *{ocp-short} Dedicated environment required access:* You must have `cluster-admin` permissions. By default `dedicated-admin` role does not have the required permissions to create namespaces in the {ocp-short} Dedicated environment. 
-//same
 * {acm} must be installed and configured.
-//same
+
 
 [#gh-network-requirements]
 == Networking requirements
 
 See the following networking requirements:
 
-* The managed hub is also a managed cluster of {global-hub} in {acm-short}. You must configure the network component in {acm-short} is necessary. For {acm-short} networking details, see link:../networking/networking_intro.adoc#networking[Networking]. 
+* The managed hub is also a managed cluster of {global-hub} in {acm-short}. You must configure the network component in {acm-short}. For {acm-short} networking details, see link:../networking/networking_intro.adoc#networking[Networking]. 
 
 * To review the {global-hub} network information, see the following table:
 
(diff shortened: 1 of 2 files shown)


global_hub/global_hub_requirements.adoc
```diff
@@ -1,7 +1,7 @@
 [#global-hub-requirements]
 = Global hub requirements
```
jc-berger: Doc reviewers: I can use the conref for the title, but then would need to change all links to it. Didn't know how much we wanted to play around with xrefs and links at this moment.

global_hub/global_hub_requirements.adoc
```diff
@@ -1,7 +1,7 @@
...
 
-Learn about what is required for installation and networking, as well as supported components and environments.
+Learn about the {global-hub} requirements for installing and networking, and review the components and environments that are supported by {global-hub}. Complete the following sections:
```
jc-berger: Doc reviewers: I know this is self referential, but for usability and accessibility, I think we should introduce the bullet list of sections for the user to complete. Thoughts?

swopebe: I think this is not the "self-referential" that we try to avoid. We want to avoid:

Use this process. Run this command: Follow this procedure...

As much as possible...I think this is fine.

One thing we need to chat about though is the fact that this is a concept, but we are telling them to complete something. We need to either make this a procedure, or take out that language that there is a task within. This will be even more strict with Dita, if it is similar to what I learned, but it is also the modular concept.

swopebe (changes requested):

global_hub/global_hub_requirements.adoc
swopebe: Yes, we normally use conrefs in the title...you can see how big of a job to do that with a search in VS code and maybe save that for later. F and R is easiest in VS Code--moving to another editor, it may not be as easy.

global_hub/global_hub_requirements.adoc
swopebe: Since this is tech debt. I would use this to change things around a bit so it is not a blend of a task and concept. We have some examples of requirements topics that may help. It seems we have prereqs to install here.

global_hub/global_hub_requirements.adoc
```diff
@@ -12,47 +12,53 @@ Learn about what is required for installation and networking, as well as support
...
-* The following table lists the Global hub network information:
 +
 |===
-| Direction | Protocol | Connection | Port (if specified) | Source address |	Destination address |
+| Direction | Protocol | Connection | Port that is specified | Source address |	Destination address |
     
-| Inbound from browser of the user | HTTPS | User need to access the Grafana dashboard | 443 | Browser of the user | IP address of Grafana route |
-| Outbound to Kafka Cluster | HTTPS | Global hub manager need to get data from Kafka cluster | 443 | multicluster-global-hub-manager-xxx pod | Kafka route host |
-| Outbound to PostgreSQL database | HTTPS | Global hub manager need to persist data to PostgreSQL database | 443 | multicluster-global-hub-manager-xxx pod | IP address of the PostgreSQL database |
+| Inbound from browser of the user | HTTPS | The user needs to access the Grafana dashboard | 443 | Browser of the user | IP address of Grafana route |
```
swopebe: Inbound from browser **of the user**? We say "of the user" multiple times here, and that is likely due to the different languages of the engineer and writer, but we should always simplify and change that from the drafts. 

We typically use conversational tone; I would try to simplify this:

Inbound from the browser| HTTPS | Accesses the Grafana dashboard | Browser | Grafana route IP address

(removing the "of the" language where you can)

And will need to be mindful of that in the entire book bc it comes to us like that in drafts sometimes.

global_hub/global_hub_requirements.adoc
```diff
@@ -12,47 +12,53 @@ Learn about what is required for installation and networking, as well as support
...
+| Outbound to Kafka Cluster | HTTPS | The {global-hub} manager needs to get data from Kafka cluster | 443 | multicluster-global-hub-manager-xxx pod | Kafka route host |
+| Outbound to PostgreSQL database | HTTPS | The {global-hub} needs to persist data to PostgreSQL database | 443 | multicluster-global-hub-manager-xxx pod | IP address of the PostgreSQL database |
```
swopebe: Same suggestion to simplify line 38 if possible. Other cells are simplified and then some are not as concise in the table.

global_hub/global_hub_requirements.adoc
```diff
@@ -12,47 +12,53 @@ Learn about what is required for installation and networking, as well as support
...
 +
 |===
-| Direction | Protocol | Connection | Port (if specified) | Source address |	Destination address |
+| Direction | Protocol | Connection | Port that is specified | Source address |	Destination address |
     
-| Inbound from browser of the user | HTTPS | User need to access the Grafana dashboard | 443 | Browser of the user | IP address of Grafana route |
-| Outbound to Kafka Cluster | HTTPS | Global hub manager need to get data from Kafka cluster | 443 | multicluster-global-hub-manager-xxx pod | Kafka route host |
-| Outbound to PostgreSQL database | HTTPS | Global hub manager need to persist data to PostgreSQL database | 443 | multicluster-global-hub-manager-xxx pod | IP address of the PostgreSQL database |
+| Inbound from browser of the user | HTTPS | The user needs to access the Grafana dashboard | 443 | Browser of the user | IP address of Grafana route |
+| Outbound to Kafka Cluster | HTTPS | The {global-hub} manager needs to get data from Kafka cluster | 443 | multicluster-global-hub-manager-xxx pod | Kafka route host |
```
swopebe: simplify:
The {global-hub} manager needs to get data from Kafka cluster

receives data? May be more concise here.

swopebe (changes requested):

global_hub/global_hub_requirements.adoc
```diff
@@ -12,47 +12,53 @@ Learn about what is required for installation and networking, as well as support
...
+* For guidelines for sizing, see link:../install/cluster_size.adoc#sizing-your-cluster[Sizing your {acm-short} cluster].
 
 [#gh-supported-components]
 == Supported components
 
-Learn about supported platforms and components.
+See the following components that {global-hub} supports: 
 
-* Because they share an integrated console, the {global-hub} console supports the same browsers as the {ocp-short}. See link:https://docs.redhat.com/documentation/en-us/openshift_container_platform/4.17/html-single/web_console/index#web-console[Accessing the web console] in the {ocp} documentation for information about supported browsers and versions.
+* Because they share an integrated console, the {global-hub} and {ocp-short} console share an integrated console, so they support the same browser. For information about supported browsers and versions, see link:https://docs.redhat.com/documentation/en-us/openshift_container_platform/4.17/html-single/web_console/index#web-console[Accessing the web console] in the {ocp} documentation
```
swopebe: Reading this out loud, we say the same thing twice? and we use anthropomorphism:

Because they share an integrated console, the {global-hub} and {ocp-short} console share an integrated console, so they support the same browser. (???)

Why not just say

The integrated console for ocp and gh supports the same browser (??).  If we understand what the line is saying here, we can greatly reduce this. 

The OCP and ACM console is integrated and support the same browser. (Also a way to say that)

global_hub/global_hub_requirements.adoc
```diff
@@ -12,47 +12,53 @@ Learn about what is required for installation and networking, as well as support
...
+* To review the managed hub network information, see the following table: 
 
-* The following table lists the Managed hub network information:
 +
 |===
-| Direction | Protocol | Connection | Port (if specified) | Source address |	Destination address |
+| Direction | Protocol | Connection | Port that is specified | Source address |	Destination address |
     
-| Outbound to Kafka Cluster | HTTPS | Global hub agent need to sync cluster info and policy info to Kafka cluster | 443 | multicluster-global-hub-agent pod | Kafka route host |
+| Outbound to Kafka Cluster | HTTPS | The {global-hub} agent needs to sync cluster info and policy info to Kafka cluster | 443 | multicluster-global-hub-agent pod | Kafka route host |
```
swopebe: we don't write "info" so we need to spell that out and you can do that by reducing other words:

_The {global-hub} agent syncs cluster and policy information_

global_hub/global_hub_requirements.adoc
```diff
@@ -12,47 +12,53 @@ Learn about what is required for installation and networking, as well as support
...
 |===
++
 
-* You can see guidelines for sizing at link:../install/cluster_size.adoc#sizing-your-cluster[Sizing your {acm-short} cluster] in the product documentation.
+* For guidelines for sizing, see link:../install/cluster_size.adoc#sizing-your-cluster[Sizing your {acm-short} cluster].
```
swopebe:
```suggestion
* For sizing guidelines, see link:../install/cluster_size.adoc#sizing-your-cluster[Sizing your {acm-short} cluster].
```

global_hub/global_hub_requirements.adoc
```diff
@@ -12,47 +12,53 @@ Learn about what is required for installation and networking, as well as support
...
 
 [#gh-supported-components]
 == Supported components
 
-Learn about supported platforms and components.
+See the following components that {global-hub} supports:
```
swopebe:
```suggestion
See the following supported components: 
```

swopebe: (they know they are reading a gh document)

global_hub/global_hub_requirements.adoc
```diff
@@ -12,47 +12,53 @@ Learn about what is required for installation and networking, as well as support
...
 == Supported components
 
-Learn about supported platforms and components.
+See the following components that {global-hub} supports: 
 
-* Because they share an integrated console, the {global-hub} console supports the same browsers as the {ocp-short}. See link:https://docs.redhat.com/documentation/en-us/openshift_container_platform/4.17/html-single/web_console/index#web-console[Accessing the web console] in the {ocp} documentation for information about supported browsers and versions.
+* Because they share an integrated console, the {global-hub} and {ocp-short} console share an integrated console, so they support the same browser. For information about supported browsers and versions, see link:https://docs.redhat.com/documentation/en-us/openshift_container_platform/4.17/html-single/web_console/index#web-console[Accessing the web console] in the {ocp} documentation
 
-* The following table lists platforms that are available for the supported {global-hub} cluster:
+* See following table for a list of platforms that are available for the supported {global-hub} cluster:
```
swopebe:
```suggestion
* See the following table of supported platforms for the {global-hub} cluster:
```

swopebe: we do need to make sure the tables are titled...I have not checked all of them.

global_hub/global_hub_requirements.adoc
```diff
@@ -65,8 +71,9 @@ Learn about supported platforms and components.
 |{acm-short} on IBM Z | Yes | Yes
 |{acm-short} on IBM Power Systems | Yes | Yes
 |===
++
 
-* {global-hub} supports middleware, including: Kafka, PostgreSQL, and Grafana. The following table lists supported middleware and built-in versions:
+* {global-hub} supports middleware, including: Kafka, PostgreSQL, and Grafana. See the following table for a list of supported middleware and built-in versions:
```
swopebe:
```suggestion
* {global-hub} supports middleware, such as Kafka, PostgreSQL, and Grafana. See the following table for a list of supported middleware and built-in versions:
```

swopebe: including: 

Doesn't always orient the user that a list is following, because you just use a colon with no language. Good idea to avoid it.

ldpliu (approved):

global_hub/global_hub_requirements.adoc
```diff
@@ -1,58 +1,61 @@
...
 
-* <<gh-general-requirements,General requirements>>
+* <<gh-prerequisites,Prerequisites>>
 * <<gh-network-requirements,Networking requirements>>
 * <<gh-supported-components,Supported components>>
 
-[#gh-general-requirements]
-== General requirements
+[#gh-prerequisites]
+== Prerequisites
```
swopebe: As I mentioned, since this is not a task, we prob should not have prereqs here... these are prereqs to the install task.

global_hub/global_hub_requirements.adoc
```diff
@@ -1,58 +1,61 @@
...
 
-To install Global hub, you need the following requirements:
+To install {global-hub}, see the following prerequisites:
```
swopebe: Line 13, this is not the install task. Why do we have this here?

global_hub/global_hub_requirements.adoc
```diff
@@ -1,58 +1,61 @@
...
-
-* {acm} must be installed and configured. 
 
 [#gh-network-requirements]
 == Networking requirements
 
 See the following networking requirements:
 
-* The managed hub is also a managed cluster of {global-hub} in {acm-short}. The network configuration in {acm-short} is necessary. See link:../networking/networking_intro.adoc#networking[Networking]] for {acm-short} networking details.
+* The managed hub is also a managed cluster of {global-hub} in {acm-short}. You must configure the network component in {acm-short}. For {acm-short} networking details, see link:../networking/networking_intro.adoc#networking[Networking].
```
swopebe: Instances like this need to be reduced and turned around -- please scan the file for things like _of the_ which is placing the noun's adjective/modifier at the end of the sentence. 

Read this outloud:

**The managed hub is also a managed cluster of {global-hub} in {acm-short}.**

This is confusing. What are we trying to say here? And always use "hub cluster." Try:

The managed hub cluster is also a {global-hub} managed cluster in {acm-short}?

And should this be "local-cluster"? (managed hub cluster)

swopebe: The changes to the table look good. Just remember to code the elements within the table, such as the pod name.

global_hub/global_hub_requirements.adoc
swopebe: I think you can keep the bullet points, but remove the header. What we may run in to is that is typical of a task and this is not a task.

global_hub/global_hub_requirements.adoc
```diff
@@ -1,58 +1,56 @@
 [#global-hub-requirements]
-= Global hub requirements
+= {global-hub} requirements
 
-Learn about what is required for installation and networking, as well as supported components and environments.
+See the following general {global-hub} requirements:
```
jc-berger: Doc reviewers: 

please let me know what you think of this restructure to remove prereqs in this concept doc.

global_hub/global_hub_requirements.adoc
```diff
@@ -1,58 +1,54 @@
...
-
-* <<gh-general-requirements,General requirements>>
-* <<gh-network-requirements,Networking requirements>>
-* <<gh-supported-components,Supported components>>
-
-[#gh-general-requirements]
-== General requirements
-
-To install Global hub, you need the following requirements:
+Learn about the {global-hub} requirements for installing and networking, and review the components and environments that are supported by {global-hub}.
```
swopebe, line 5: we don't need this:
{acm} must be installed and configured.

That is prereq info for a task.

swopebe: This is not a task.

swopebe (approved):

PR #8107, opened 2025-08-06
https://issues.redhat.com/browse/ACM-21313 Add Subm note

oafischer:
2.14 only

Checklist considered

### DIFF
--- a/networking/submariner/subm_intro.adoc
+++ b/networking/submariner/subm_intro.adoc
@@ -5,6 +5,8 @@ Submariner is an open source tool that you can use with {acm} to provide direct 
 
 Submariner is not supported with all of the infrastructure providers that {acm-short} can manage. Refer to the link:https://access.redhat.com/articles/7120842[{acm-short} Support Matrix] for more details about the support levels of infrastructure providers, including which providers support xref:../submariner/deploy_subm_console.adoc#deploying-submariner-console[automated console deployments] or require xref:../submariner/deploy_subm_manual.adoc#deploying-submariner-manually[manual deployment].
 
+*Technology Preview:* Submariner without Globalnet supports dual-stack clusters with OVN-Kubernetes Container Network Interface.
+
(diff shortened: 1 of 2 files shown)


networking/submariner/subm_intro.adoc
```diff
@@ -5,6 +5,8 @@ Submariner is an open source tool that you can use with {acm} to provide direct
 
 Submariner is not supported with all of the infrastructure providers that {acm-short} can manage. Refer to the link:https://access.redhat.com/articles/7120842[{acm-short} Support Matrix] for more details about the support levels of infrastructure providers, including which providers support xref:../submariner/deploy_subm_console.adoc#deploying-submariner-console[automated console deployments] or require xref:../submariner/deploy_subm_manual.adoc#deploying-submariner-manually[manual deployment].
 
+*Technology Preview:* Submariner without Globalnet supports dual-stack clusters with OVN-Kubernetes Container Network Interface.
```
yboaron: Maybe , we should also specify that dual-stack is currently supported only with Libreswan cable driver . it should be something like :  

Technology Preview: Submariner, without Globalnet and using Libreswan (IPSec) cable driver, now supports dual-stack clusters with OVN-Kubernetes CNI.

oafischer: @yboaron Is this new for 2.14? If so, I'll add it to the what's new section.

yboaron: > @yboaron Is this new for 2.14? If so, I'll add it to the what's new section.

Yep, it's new for 2.14

oafischer: In that case I think I'll move this to the release notes section entirely. We usually don't have individual support statements in the doc, but it's OK if we mention newly supported things in the release notes.

yboaron (approved):

dockerymick (approved):
This looks good to me! Checklist considered

PR #8111, opened 2025-08-06
https://issues.redhat.com/browse/ACM-20316 Cancel host install doc

oafischer:
2.14

Checklist considered


clusters/support_troubleshooting/trouble_cancel_host.adoc
```diff
@@ -0,0 +1,124 @@
...
+----
++
+You can confirm that the host is unbound by running the following command:
++
+[source,bash]
+----
+oc get agent -n <namespace>
+----
++
+If the host is unbound, the output resembles the following example, where `CLUSTER` and `STAGE` are empty.
```
CrystalChun: Sorry, I forgot to add in the original issue description that the `STAGE` might not be empty and could be `unbinding-pending-user-action` or a related stage. Could we make a note of that around here too?

oafischer: Of course, just updated

CrystalChun (approved):
/lgtm

Thanks Oliver!

clusters/support_troubleshooting/trouble_cancel_host.adoc
CrystalChun: Awesome, thank you!

clusters/support_troubleshooting/trouble_cancel_host.adoc
```diff
@@ -0,0 +1,124 @@
...
++
+If your cluster is using {cim}, the following output with the `adding-hosts` `STATE` appears:
++
+[source,bash]
+----
+NAME                    CLUSTER                 STATE
+assisted-cluster        assisted-cluster        adding-hosts 
+----
+
+. Check if the host in installing. Run the following command:
```
lshilin-r: did you mean
```suggestion
. Check if the host is installing. Run the following command:
```

oafischer: Good catch, updated!

clusters/support_troubleshooting/trouble_cancel_host.adoc
```diff
@@ -0,0 +1,124 @@
+[#trouble-day2-host-stuck]
+= Production environment host becomes stuck during installation
```
lshilin-r: @CrystalChun I'm afraid the reason to use the cancellation of host installation is not fully clear in the chapter. It can be also for the case when a user by mistake began to bind not the right host to a cluster. I don't see it mentioned here. WDYT, should it be mentioned?

CrystalChun: Sure, good idea! 

Here are the use-cases that I think could fit:
1. Binding a host to an incorrect cluster
2. Wanting to change the configuration of a host after it starts installing
3. Host becomes stuck during installation (already in this doc currently)

What do you think? Are there other cases that should be added too?

oafischer: Yes we can add additional detail in the intro, I've added the ones mentioned here so far

CrystalChun: These look good to me, thanks! Wdyt @lshilin-r?

clusters/support_troubleshooting/trouble_cancel_host.adoc
```diff
@@ -0,0 +1,129 @@
...
+
+You might want to stop the host installation in one of the following situations:
+
+- A production environment host becomes stuck during installation
+- You bind the host to a wrong cluster
+- You want to change the host configuration after the installation starts
+- You want to quickly scale down the number of installed hosts on your hosted cluster
```
lshilin-r: This one is not clear for me. Scale down of installed hosts is not part of the procedure described here. I understand that the command `oc patch agent -n <namespace> <agent-name> --patch='[{"op":"remove","path":"/spec/clusterDeploymentName"}]' --type=json` also can be used for scale down, but this procedure describes the process of stopping installation, not scaling down already installed host. @CrystalChun do you agree?

CrystalChun: Yeah it's probably a better idea not to include this use-case.

oafischer: Removed the last bullet point. Let me know if it looks good now. Thanks!

clusters/release_notes/mce_new.adoc
```diff
@@ -46,4 +46,6 @@ Learn about new features and enhancements for Cluster lifecycle with {mce-short}
 
 - You can now use any file name for the `OSImageCACert` `ConfigMap` when enabling {cim}. To learn more, see xref:../assisted_installer/ai_enable_cim#enable-cim-disconnected[Enabling {cim} in disconnected environments].
 
+- Cancel host installations when using {cim} to change your host configuration or resolve installation issues. To learn more, see xref:../support_troubleshooting/trouble_rcancel_host.adoc#trouble-day2-host-stuck[Production environment host becomes stuck during installation].
```
swopebe: Why are we adding a troubleshooting topic to New features? This is not good practice.  New features are literally new features. Troubleshooting is support docs. 

T-shooting docs really should not be linked from anywhere--if they are linked from the main doc, they should become main doc and not t-shooting.

clusters/support_troubleshooting/trouble_cancel_host.adoc
```diff
@@ -0,0 +1,128 @@
+[#trouble-day2-host-stuck]
+= Production environment host becomes stuck during installation
+
```
swopebe: Right away from the title I can tell this needs to match format that we write in for t-shooting

clusters/support_troubleshooting/trouble_cancel_host.adoc
```diff
@@ -0,0 +1,128 @@
...
+You might want to stop the host installation in one of the following situations:
```
swopebe: Description about the problem??? That is usually the first thing. Please look at previous well-written troubleshooting and the internal doc:
https://stolostron.github.io/rhacm-docs/troubleshooting

swopebe (changes requested):
This looks more like a task, it is linked in new features, some other issues that I see with formatting.

clusters/support_troubleshooting/trouble_cancel_host.adoc
```diff
@@ -0,0 +1,128 @@
...
+. Check if the host is installing. Run the following command:
+
++
+[source,bash]
+----
+oc get agent -n <namespace>
+----
++
+If the host is installing, the output resembles the following example:
++
```
swopebe: format

clusters/support_troubleshooting/trouble_cancel_host.adoc
```diff
@@ -0,0 +1,128 @@
...
+
+. Unbind the host from the cluster by removing the `clusterDeploymentName` field in the `Agent` resource of the host that you want to stop installing. Run the following command:
+
++
+[source,bash]
+----
+oc patch agent -n <namespace> <agent-name> --patch='[{"op":"remove","path":"/spec/clusterDeploymentName"}]' --type=json
+----
++
+You can confirm that the host is unbound by running the following command:
```
swopebe: This is a verification step

clusters/support_troubleshooting/trouble_cancel_host.adoc
```diff
@@ -0,0 +1,128 @@
...
+. Confirm that the `Agent` resource is bound to the cluster. Run the following command:
+
++
+[source,bash]
+----
+oc get agent -n <namespace>
+----
++
+If the `Agent` resource is bound to the cluster, the output resembles the following example:
++
```
swopebe: format here

clusters/support_troubleshooting/trouble_cancel_host.adoc
```diff
@@ -0,0 +1,128 @@
...
+You can confirm that the host is unbound by running the following command:
++
+[source,bash]
+----
+oc get agent -n <namespace>
+----
++
+If the host is unbound, the output resembles the following example, where `CLUSTER` is empty and `STAGE` is either empty, or `unbinding-pending-user-action` or similar.
++
```
swopebe: format

clusters/support_troubleshooting/trouble_cancel_host.adoc
swopebe: anywhere commands and YAML are not slightly grayed out, remember from a few years ago....we create a space. I think we also put this in internal docs.

oafischer:
@swopebe I changed this to a procedure and addressed the comments, let me know if it looks better. Thanks!

clusters/assisted_installer/ai_cancel_host.adoc
```diff
@@ -0,0 +1,127 @@
...
+If the host is installing, the output resembles the following example:
+
++
+[source,bash]
+----
+NAME                                   CLUSTER                 APPROVED   ROLE          STAGE
+127877e3-2098-4bf8-9b44-588a78ffcccd   assisted-cluster        true       worker        Rebooting 
+----
+
+If your output resembles the previous example, you can continue cancelling the host installation. Complete the following steps to cancel the host installation:
```
oafischer, line 50: As per modular docs, we'd need to have the next section in a different file. Should I go ahead with that or keep the current format?

clusters/assisted_installer/ai_cancel_host.adoc
```diff
@@ -0,0 +1,127 @@
+[#cancel-host-install]
+= Cancelling host installations
+
+In some cases, a production environment host might become stuck during installation with the `installing-pending-user-action` status. You can cancel the installation to solve the issue. You might also want to cancel the host installation if you bind the host to the wrong cluster, or if you want to change the host configuration after the installation starts.
+
+.Prerequisites
```
CrystalChun, line 8: Sorry I forgot to mention this, this procedure only works for hosts that use "late-binding"
I don't think this concept has its own document... but it is mentioned here and there throughout the docs

[Import OCP Clusters](https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.14/html-single/clusters/index#import-ocp-cluster-steps) InfraEnv table in step 7

[Add nodes to OCP cluster](https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.14/html-single/clusters/index#add-nodes-cim-ocp-procedure) step 4
- Aside: I wonder if this current cancel host doc would work well under this one? Since the prerequisite is adding hosts to an installed cluster or an HCP cluster 

[Returning Non-BM Hosts to Cluster](https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.14/html-single/clusters/index#troubleshooting-return-non-bm-cluster)

clusters/assisted_installer/ai_cancel_host.adoc
```diff
@@ -0,0 +1,127 @@
+[#cancel-host-install]
+= Cancelling host installations
+
+In some cases, a production environment host might become stuck during installation with the `installing-pending-user-action` status. You can cancel the installation to solve the issue. You might also want to cancel the host installation if you bind the host to the wrong cluster, or if you want to change the host configuration after the installation starts.
```
swopebe: This is still very much written like a t-shooting file and not a new feature. Start with the benefit of host installation, which seems to be the new feature. Later in this paragraph you can go in to the problems it solves.


_You can cancel host installations to prevent incorrect cluster binding, change host configuration, or fix installation problems._ 

_For instance...._

swopebe (changes requested):

clusters/assisted_installer/ai_cancel_host.adoc
```diff
@@ -0,0 +1,127 @@
...
+
+.Prerequisites
+
+- You are using {cim}.
+- You have an installed cluster.
```
swopebe: Number 9 is a given, really so is number 8, or they would not be in the doc.

You have a {cim} installation on your cluster? Hub? Managed?

However, we need to be specific about the cluster.

You have an {acm-short} hub cluster? You have managed clusters? We typically would not just say you installed a cluster, that is a given.

clusters/assisted_installer/ai_cancel_host.adoc
```diff
@@ -0,0 +1,127 @@
...
+
+.Procedure
+
+Complete the following steps to check if you can cancel the host installation:
+
+. Check if your installed cluster is using {cim}. Run the following command:
```
swopebe: don't need "installed cluster", just cluster is fine....but is it a hub or managed?

clusters/assisted_installer/ai_cancel_host.adoc
swopebe: I think this is fine to have it as one procedure with all the steps together. Otherwise, you could break this out into sub sections--and that is more UX friendly but we already have enough of that to fix. I have been trying to either keep files separate or write as one procedure.

swopebe: The rest looks really good.

clusters/assisted_installer/ai_cancel_host.adoc
swopebe: Engineer will need to review.

clusters/assisted_installer/ai_cancel_host.adoc
oafischer: Thanks, I've added late binding as prereq. It's explained in CIM docs so I don't think we need to point to a dedicated source. The linked topic is troubleshooting again, where we cannot place this doc.

clusters/assisted_installer/ai_cancel_host.adoc
oafischer: Updated

clusters/assisted_installer/ai_cancel_host.adoc
oafischer: Updated

clusters/assisted_installer/ai_cancel_host.adoc
oafischer: OK thanks, I'm going to have to break this out during conversion anyway but I'll leave it for now.

clusters/assisted_installer/ai_cancel_host.adoc
oafischer: Managed, updated in PR

clusters/assisted_installer/ai_cancel_host.adoc
```diff
@@ -0,0 +1,126 @@
+[#cancel-host-install]
+= Cancelling host installations
+
+You can cancel host installations to prevent incorrect cluster binding, change host configurations, or fix installation problems.
+
+For example, you can cancel the installation if a production environment host becomes stuck during installation with the `installing-pending-user-action` status.
+
+.Prerequisites
```
dockerymick, line 8: Since there is only one prereq. I'm also not sure how ".Procedure" will render. Is this done in preparation of the CCS tool changes? 
```suggestion
.Prerequisite
```

clusters/assisted_installer/ai_cancel_host.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+- You are using the late binding feature on your managed cluster.
```
dockerymick:
```suggestion
You are using the late binding feature on your managed cluster.
```

clusters/assisted_installer/ai_cancel_host.adoc
```diff
@@ -0,0 +1,126 @@
...
+----
+
+.. If you are using the `BareMetalHost` resource to add a node, the host automatically returns to the discovery phase. Run the following command to confirm that the host is in the discovery phase:
++
+[source,bash]
+----
+oc get bmh -n <namespace>
+----
++
+If the host is in the discovery phase, the `BareMetalHost` resource status changes from deprovisioning to provisioning. It might take a few minutes for deprovisioning to complete. After the  `BareMetalHost` reaches the provisioning status, the host is in discovery mode.
```
dockerymick: is the status shown from the terminal? Should "deprovisioning" and "provisioning" be monospaced?

dockerymick (approved):
Left a couple of questions. Otherwise, this looks great!

clusters/assisted_installer/ai_cancel_host.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+. Check to see if the `Agent` resource is discovered and validated. Run the following command:
+
++
+[source,bash]
+----
+oc get agent -n <namespace> <agent-name> -ojsonpath='{"NAME: "}{.metadata.name}{"\n"}{"VALIDATIONS PASSING: "}{.status.conditions[?(@.type == "Validated")].status}{"\n"}'
+----
+
+. Bind the `Agent` resource to the managed cluster and reinstall the host. Run the following command:
```
CrystalChun:
```suggestion
. Bind the `Agent` resource to the managed cluster to reinstall the host. Run the following command:
```
Sorry very minor, but hopefully to prevent confusion. Binding the Agent again will reinstall the host and there's no additional step or action after that to reinstall

CrystalChun:
The rest looks good to me. Thanks Oliver!

clusters/assisted_installer/ai_cancel_host.adoc
oafischer: We agreed to use plural even when there is only one prereq, I think this was discussed when you were out.

And yeah we cannot have anchors on procedures for the conversion so we need to get rid of those. This was from the meeting with Jaromir.

clusters/assisted_installer/ai_cancel_host.adoc
oafischer: @CrystalChun Can you help here? Not sure what the output is supposed to look like

CrystalChun: Yeah this is what the output would look like when the host is still being reprovisioned by Ironic 
```sh
NAME         STATE            CONSUMER                  ONLINE   ERROR                AGE
host-1       deprovisioning                             true                         17h
```

And once it's ready to be added to a cluster again, it'll look like this
```sh
NAME         STATE            CONSUMER                  ONLINE   ERROR                AGE
host-1       provisioning                               true                         17h
```

oafischer: Thanks, updated!

clusters/assisted_installer/ai_cancel_host.adoc
swopebe: agreed

swopebe (approved):

clusters/assisted_installer/ai_cancel_host.adoc
dockerymick: Thanks for clarifying! I will be sure to also add that reminder to my tech debt issue

PR #8118, opened 2025-08-12
https://issues.redhat.com/browse/ACM-21718

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-21718

### DIFF
--- a/governance/adv_template_process.adoc
+++ b/governance/adv_template_process.adoc
@@ -23,7 +23,7 @@ If you need to manually initiate an update, use the special annotation, `policy.
 [#bypass-template-processing]
 == Bypassing template processing
 
-By default, {acm-short} processes all Go templates. You might create a policy that contains a template in `policy-templates` that is not intended to be processed by {acm-short}. To bypass template processing for a particular template, change `{{ template content }}` to `{{ `{{ template content }}` }}`. Then, the template returns a raw string to be processed by a subsequent templating engine.
+By default, {acm-short} processes all Go templates. You might create a policy that contains a template in `policy-templates` that is not intended to be processed by {acm-short}. To bypass template processing for a particular template, change `{{ template content }}` to `{{ `{{ template content }}` `}}`. Then, the template returns a raw string to be processed by a subsequent templating engine.
 
 If you want to completely bypass the template resolution for a configuration policy, add the `policy.open-cluster-management.io/disable-templates: "true"` annotation in the relevant `ConfigurationPolicy` that is contained in your `Policy` resource.
 
--- a/governance/adv_template_process.adoc
+++ b/governance/adv_template_process.adoc
@@ -23,7 +23,7 @@ If you need to manually initiate an update, use the special annotation, `policy.
 [#bypass-template-processing]
 == Bypassing template processing
 
-By default, {acm-short} processes all Go templates. You might create a policy that contains a policy in `policy-templates` that is not intended to be processed by {acm-short}. To bypass template processing for a particular template, change `{{ template content }}` to `{{ \`{{ template content }}\` }}`. Then, the template returns a raw string to be processed by a subsequent templating engine.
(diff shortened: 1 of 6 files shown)


governance/adv_template_process.adoc
```diff
@@ -23,7 +23,7 @@ If you need to manually initiate an update, use the special annotation, `policy.
 [#bypass-template-processing]
 == Bypassing template processing
 
-By default, {acm-short} processes all Go templates. You might create a policy that contains a policy in `policy-templates` that is not intended to be processed by {acm-short}. To bypass template processing for a particular template, change `{{ template content }}` to `+{{ `{{ template content }}` }}+`. Then, the template returns a raw string to be processed by a subsequent templating engine.
+By default, {acm-short} processes all Go templates. You might create a policy that contains a policy in `policy-templates` that is not intended to be processed by {acm-short}. To bypass template processing for a particular template, change `{{ template content }}` to `{{ `{{ template "content" }}` }}`. Then, the template returns a raw string to be processed by a subsequent templating engine.
```
JustinKuli: Is there a way to preview this content **_exactly how it will be prepared when it is published?_** Because these details seem very fiddly. 

I think we want it to look like this (from the [current source](https://github.com/stolostron/rhacm-docs/blob/2.13_prod/governance/adv_template_process.adoc)): 
> Screenshot From 2025-08-12 13-36-15

In the proposed change, it actually looks [incorrect in GitHub](https://github.com/stolostron/rhacm-docs/blob/d727414aaedadacec662ad7c40dd0c6b59a1ccda/governance/adv_template_process.adoc#bypassing-template-processing):
> Screenshot From 2025-08-12 13-38-33

But ultimately we need it to look correct on docs.redhat.com which it does not [right now](https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.13/html-single/governance/index#bypass-template-processing):

> Screenshot From 2025-08-12 13-41-26

So I'm asking how we can see what it will look like on docs.redhat.com, before we actually publish it.

JustinKuli: It seems like the previous version was using the correct ["literal monospace" style](https://docs.asciidoctor.org/asciidoc/latest/text/literal-monospace) so I don't know why it wasn't rendering correctly.

Maybe an alternative is to just escape the inner backticks:
```suggestion
By default, {acm-short} processes all Go templates. You might create a policy that contains a policy in `policy-templates` that is not intended to be processed by {acm-short}. To bypass template processing for a particular template, change `{{ template content }}` to `{{ \`{{ template content }}\` }}`. Then, the template returns a raw string to be processed by a subsequent templating engine.
```

But again, we need some way to test this...

luckylittle: Looks good - this is what it should look like

swopebe: This line a bit confusing:

`You might create a policy that contains a policy in `policy-templates` that is not intended to be processed by {acm-short}.`

You might create a policy that contains a policy FROM `policy-templates`?

swopebe: Not sure if this was answered, but yes, you can preview the link from stage, but that is after a merge. @jc-berger you can also build just your branch without merging, you just need to announce to the team that you are doing that so we don't override it while you check.

swopebe (approved):
approved but with a small comment and then need to check that code in the build

swopebe (approved):

governance/adv_template_process.adoc
JustinKuli: Good catch @swopebe! I think the intention was:
> You might create a policy that contains a ~policy~ _template_ in `policy-templates` that is not intended to be processed by {acm-short}.

jc-berger:
Closing this PR, please see the new PR for the correct code change: https://github.com/stolostron/rhacm-docs/pull/8136/files

PR #8119, opened 2025-08-12
https://issues.redhat.com/browse/ACM-22652

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-22652

### DIFF
--- a/networking/submariner/deploy_subm_console.adoc
+++ b/networking/submariner/deploy_subm_console.adoc
@@ -17,7 +17,7 @@ Before you deploy Submariner with {acm}, you must prepare the clusters on the ho
 
 *Notes:*
 
-* Only non-NSX deployments are supported on VMware vSphere.
+* You cannot deploy Submariner on a network virtualization and security platform (NSX) while getting support from VMware or VSphere. 
 
 * You must install the link:https://docs.tigera.io/calico/latest/operations/install-apiserver#install-the-api-server[Calico API server] on your cluster if you are using Red Hat OpenShift on IBM Cloud. Alternatively, you can manually create the IP pools required for cross-cluster communication by following the link:https://submariner.io/operations/deployment/calico/[CALICO CNI] topic in the Submariner upstream documentation.
 
--- a/networking/submariner/deploy_subm_console.adoc
+++ b/networking/submariner/deploy_subm_console.adoc
@@ -17,7 +17,7 @@ Before you deploy Submariner with {acm}, you must prepare the clusters on the ho
 
 *Notes:*
 
-* You cannot deploy Submariner on a network virtualization and security platform (NSX) while getting support from VMware or VSphere. 
+* Submariner deployments on VMware and vSphere environments that have VMware NSX enabled are not currently supported by {acm-short}.
 
 * You must install the link:https://docs.tigera.io/calico/latest/operations/install-apiserver#install-the-api-server[Calico API server] on your cluster if you are using Red Hat OpenShift on IBM Cloud. Alternatively, you can manually create the IP pools required for cross-cluster communication by following the link:https://submariner.io/operations/deployment/calico/[CALICO CNI] topic in the Submariner upstream documentation.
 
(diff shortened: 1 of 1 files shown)


networking/submariner/deploy_subm_console.adoc
```diff
@@ -17,7 +17,7 @@ Before you deploy Submariner with {acm}, you must prepare the clusters on the ho
 
 *Notes:*
 
-* Only non-NSX deployments are supported on VMware vSphere.
+* You cannot deploy Submariner on a network virtualization and security platform (NSX) while getting support from VMware or VSphere.
```
Jaanki: This reads to me like the support is given by the VMWare company.

What the sentence originally says is "Submariner is designed to work on VMware's vSphere platform only when you're using its standard, basic networking features. It is not supported if you are using VMware's advanced networking and security product, NSX."

Jaanki: Some suggestions from Gemini

1. Submariner deployments on VMware vSphere environments with VMware NSX enabled are not currently supported.
2. When deploying Submariner on VMware vSphere, it is required to use standard vSphere networking. Deployments in environments managed by VMware NSX are not supported at this time due to networking conflicts.

I personally like the 1st one.

jc-berger: @Jaanki thank you! I implemented your feedback (using the first one). I did provide context that ACM does not support this. Is that correct? Originally, I wasn't sure what wasn't providing the support.

Jaanki (approved):
Approving assuming "{acm-short}" is a place holder for ACM version.

networking/submariner/deploy_subm_console.adoc
Jaanki: its more like Submariner doesn't support this environment but it should suffice because Submariner is part of ACM and the line does say that submariner deployments are not supported.

networking/submariner/deploy_subm_console.adoc
```diff
@@ -17,7 +17,7 @@ Before you deploy Submariner with {acm}, you must prepare the clusters on the ho
 
 *Notes:*
 
-* Only non-NSX deployments are supported on VMware vSphere.
+* Submariner deployments on VMware and vSphere environments that have a VMware network virtualization and security platform (NSX} enabled are not currently supported by {acm-short} {product-version}.
```
swopebe: Please just remove the `version.` We are in the ACM version of the doc, so we don't need to have that in here UNLESS there is a feature coming to support this. Otherwise it has to be changed each release.

Also remove _currently_. We just simply remove this line when it is supported and currently suggests a fix is coming.

(NSX} uses two different punctuations, (}-- not sure what is meant, prob ()? 

Also, doesn't that _modify_ the _security platform_ and name it? So should it not go in the front? and then why is in it parentheses? 

I am not 100% sure we are naming this correctly...
I think you have a lot of words to say what this is? Look in to that and let me know: https://www.vmware.com/products/cloud-infrastructure/nsx (I always go to the official doc before anything)

Seems to be: **VMware NSX**

I see you took suggestions from Gemini as we all discussed doing at times--that's great! Here it is still good to know what the target item is/how it is presented, as Gemini doesn't always give the most concise. It may just need to be edited after some understanding.

A prompt idea that MAY help--just an idea, you may have used something similar:

...use the proper name of the product in the output and make in concise. My audience is virtual machine users who don't have a lot of time to read through the documentation.

jc-berger: Hi, thanks! Yes, through research I gathered that NSX stands for network virtualization and security platform. However, that's more of a catch all as the abbreviation doesn't stand for anything that specific. Like you said, the actual product documentation doesn't define it. They simply say VMware NSX, so I think we're fine to refer to it that way here: https://www.vmware.com/products/cloud-infrastructure/nsx

swopebe: Ok, you could also say _VMware NSX virtualization and security platform._ But having it at the end isn't really that concise....In my experience, Gemini is not the best with product names...it called us openshift cluster management on time. (combined OCP and ACM)

swopebe (approved):

PR #8176, opened 2025-08-25
https://issues.redhat.com/browse/ACM-22342 Hosted KI

oafischer:
2.14 only

Checklist considered

### DIFF
--- a/clusters/release_notes/mce_known_issues.adoc
+++ b/clusters/release_notes/mce_known_issues.adoc
@@ -47,7 +47,7 @@ When enabling the `managed-serviceaccount` add-on, the `installNamespace` field 
 
 Learn about Known issues for Cluster lifecycle with {mce-short}, such as issues with creating, discovering, importing, and removing clusters, and more cluster management issues for {mce-short}. 
 
-[#hcp-capi-known-issue]
+[#create-hosted-ocp-4193]
 === Creating a hosted cluster with {ocp-short} with 4.19.3 and earlier fails
 //2.14:ACM-22342
 
--- a/clusters/release_notes/mce_known_issues.adoc
+++ b/clusters/release_notes/mce_known_issues.adoc
@@ -48,6 +48,20 @@ When enabling the `managed-serviceaccount` add-on, the `installNamespace` field 
 Learn about Known issues for Cluster lifecycle with {mce-short}, such as issues with creating, discovering, importing, and removing clusters, and more cluster management issues for {mce-short}. 
 
 [#hcp-capi-known-issue]
+=== Creating a hosted cluster with {ocp-short} with 4.19.3 and earlier fails
+//2.14:ACM-22342
+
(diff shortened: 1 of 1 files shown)


rokej (approved):

clusters/release_notes/mce_known_issues.adoc
```diff
@@ -47,6 +47,20 @@ When enabling the `managed-serviceaccount` add-on, the `installNamespace` field
 
 Learn about Known issues for Cluster lifecycle with {mce-short}, such as issues with creating, discovering, importing, and removing clusters, and more cluster management issues for {mce-short}. 
 
+[#create-hosted-ocp-4193]
+=== Creating a hosted cluster with {ocp-short} with 4.19.3 and earlier fails
+//2.14:ACM-22342
+
+If you create a hosted cluster on the BareMetal agent platform with {ocp-short} 4.19.3 and earlier, and use the MetalLB operator for the hosted cluster, the network cluster operator in the hosted cluster fails to start with the following error:
```
jc-berger, line 54: Is it "and" or "or"? That is does it fail if you do both actions or one or the other? I think you mean that it fails if you either created the cluster on BareMetal or if you create it on MetalB operator.

Also, when you say "use the MetalB operator for the hosed cluster" do you mean create the hosted cluster by using the MetalB operator? 

I like this content, but it's a little long. See my suggestion to make it multiple sentences, shorten it, and address some of my questions above: 

```suggestion
If you create a hosted cluster on the BareMetal agent platform with {ocp-short} 4.19.3 and earlier, or create it with the the MetalLB operator, the network cluster operator in the hosted cluster fails to start. You get the following error:
```

swopebe: Double check bare metal agent naming, but reminder--we have this in our doc and internally--that for OCP known issues, we do not doc them. If there is an OCP known issue that relates to ACM/MCE function, that needs to be clear in the content. HCP and OCP are mentioned in this known issue, so why did we document it? (It could be that we just need to make it clearer in the content)

@rokej is there a reason we are not sending this to OCP? We are trying to keep our known issues from containing OCP-specific problems unless it relates to something we doc within ACM or MCE.

oafischer: @jc-berger I do mean "and", I can try to rewrite it to make it shorter. I want to wait on @rokej to give some guidance before making big changes.

rokej: It is AND. @swopebe That is a good point. The bug was in RHACM but you are right. It is an OCP known issue.

oafischer: @rokej Would you agree to just not document it then?

swopebe: It could be that we need this to go to OCP. It also could be that we need to be clear about when the ACM use gets this issue.

oafischer: @rokej We can document this in OCP instead, or if there is a situation where this applies to ACM users, we can mention it here if we explain the specific situation.

oafischer:
Confirmed not to be documented, closing PR.

PR #8192, opened 2025-09-03
https://issues.redhat.com/browse/ACM-19805--gh must gather should loo…

swopebe:
…k similar to the other must gathers instead of a full t-shooting topic bc it is a tool/method, but not a troubleshooting topic within GH.


swopebe:
We needed to change this, as it is not a troubleshooting topic; there is nothing to troubleshoot here. This is a method/tool to troubleshoot. If you look at ACM and MCE, you can see how these were added to the support docs. I rearranged but tried to keep the meaning/goal the same. Made it a bit more concise, as well.

swopebe:
approved on slack: https://redhat-internal.slack.com/archives/C05DLMA6VBR/p1756893601096469

jc-berger (approved):
/lgtm

PR #8216, opened 2025-09-10
https://issues.redhat.com/browse/ACM-23768

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-23768 

Change for backup_active_restore.adoc intro changes.

### DIFF
--- a/business_continuity/backup_restore/backup_active_restore.adoc
+++ b/business_continuity/backup_restore/backup_active_restore.adoc
@@ -1,6 +1,5 @@
 [#keep-hub-active-restore]
-= Keeping the primary hub cluster active during a restore process
-
+= Keeping the primary hub cluster active during a restore process and keeping your hub data safe
 When you restore data to a new hub cluster while your primary hub cluster is still active, both hub clusters try to manage the same cluster fleet. If the primary hub cluster takes control of the managed clusters, it might override the policy or application changes that you made to your new hub cluster. 
 
 In a typical disaster recovery restore operation, the primary hub cluster is inactive, and there is no conflict with the restore hub cluster. You might need to keep the original hub cluster active during or after your restore operation, for example, you would typically need to do this if you run a disaster recovery simulation test.
--- a/business_continuity/backup_restore/backup_active_restore.adoc
+++ b/business_continuity/backup_restore/backup_active_restore.adoc
@@ -1,13 +1,11 @@
 [#keep-hub-active-restore]
 = Keeping the primary hub cluster active during a restore process
 
-Restore operations typically occur after a disaster that makes the primary hub cluster fail. When the restore operation starts, you cannot access the primary hub cluster. 
+When you restore data to a new hub cluster while your primary hub cluster is still active, both of these hubs try to manage the same cluster fleet. If the primary hub cluster takes control of the managed clusters, it could override the policy or application changes that you made to your new hub cluster. 
 
-If you want to run a disaster recovery test and do not want the primary hub cluster to stop during the test, keep the primary hub cluster active during the entire hub cluster restoration.
+In a typical disaster recovery restore operation, the primary hub cluster is inactive, and there is no conflict with the restore hub cluster. You might need to keep the original hub cluster active during or after your restore operation, for example, you would typically need to do this if you run a disaster recovery simulation test.
 
-If you keep the primary hub cluster active while restoring the hub cluster data on a different hub cluster, the initial hub cluster can access the managed clusters. In this case, both hub clusters try to take control of the managed clusters. If the initial hub cluster takes control of the managed clusters, you might lose policy or application changes that you made. 
-
-Keep the primary hub cluster active during and after a restore operation by completing the following tasks:
+To safely restore the hub data on your new hub cluster while keeping the primary hub cluster active, configure the primary hub cluster before you run the disaster recovery test. Complete the following tasks: 
 
 * <<keep-hub-active-restore-prepare,Preparing the primary hub cluster>>
 * <<keep-hub-active-restore-run,Running the restore operation on the new hub cluster>>
--- a/business_continuity/backup_restore/backup_active_restore.adoc
+++ b/business_continuity/backup_restore/backup_active_restore.adoc
@@ -1,6 +1,7 @@
 [#keep-hub-active-restore]
-= Keeping the primary hub cluster active during a restore process and keeping your hub data safe
-When you restore data to a new hub cluster while your primary hub cluster is still active, both hub clusters try to manage the same cluster fleet. If the primary hub cluster takes control of the managed clusters, it might override the policy or application changes that you made to your new hub cluster. 
+= Keeping the primary hub cluster active during a restore process
+
+You are able to keep your hub data safe even when you keep your primary hub cluster active during a restore process. When you restore data to a new hub cluster while your primary hub cluster is still active, both hub clusters try to manage the same cluster fleet. If the primary hub cluster takes control of the managed clusters, it might override the policy or application changes that you made to your new hub cluster. 
 
 In a typical disaster recovery restore operation, the primary hub cluster is inactive, and there is no conflict with the restore hub cluster. You might need to keep the original hub cluster active during or after your restore operation, for example, you would typically need to do this if you run a disaster recovery simulation test.
 
--- a/business_continuity/backup_restore/backup_active_restore.adoc
+++ b/business_continuity/backup_restore/backup_active_restore.adoc
@@ -1,7 +1,7 @@
 [#keep-hub-active-restore]
 = Running the restore operation while the primary hub cluster is active
 
-You are able to keep your hub data safe even when you keep your primary hub cluster active during a restore process. When you restore data to a new hub cluster while your primary hub cluster is still active, both hub clusters try to manage the same cluster fleet. If the primary hub cluster takes control of the managed clusters, it might override the policy or application changes that you made to your new hub cluster. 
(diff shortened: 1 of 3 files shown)


business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -1,13 +1,11 @@
 [#keep-hub-active-restore]
 = Keeping the primary hub cluster active during a restore process
 
-Restore operations typically occur after a disaster that makes the primary hub cluster fail. When the restore operation starts, you cannot access the primary hub cluster. 
+When you restore data to a new hub cluster while your primary hub cluster is still active, both of these hubs try to manage the same cluster fleet. If the primary hub cluster takes control of the managed clusters, it could override the policy or application changes that you made to your new hub cluster.
```
swopebe:
```suggestion
When you restore data to a new hub cluster while your primary hub cluster is still active, both hub clusters try to manage the same cluster fleet. If the primary hub cluster takes control of the managed clusters, it might override the policy or application changes that you made to your new hub cluster. 
```

swopebe: @jc-berger and @birsanv this is great, see above suggestions. There is one thing missing and that is something that relates to what this topic is for....though the problem is clear. The topic is to solve the problem?

_Keep the primary hub cluster active during a restore to avoid unwanted changes to your cluster fleet_. (something like this? Maybe at the beginning?)

swopebe: removing referential, modal, and adding hub _cluster_ in the new changes.

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -1,6 +1,5 @@
 [#keep-hub-active-restore]
-= Keeping the primary hub cluster active during a restore process
-
+= Keeping the primary hub cluster active during a restore process and keeping your hub data safe
```
birsanv: I find this confusing; we don't keep the data safe, we avoid issues with the primary hub cluster being active when the restore happens
Why do we need to change the title, with something even longer ? :) 

We explain in the section the issue ( when the primary is active ) and the solution : 
We explain the problem : When you restore data to a new hub cluster while your primary hub cluster is still active, both hub clusters try to manage the same cluster fleet. If the primary hub cluster takes control of the managed clusters, it might override the policy or application changes that you made to your new hub cluster...
We show how to address it ( if you keep the primary active ) : To safely restore the hub data on your new hub cluster while keeping the primary hub cluster active, configure the primary hub cluster before you run the disaster recovery test. Complete the following tasks:...

birsanv (changes requested):

business_continuity/backup_restore/backup_active_restore.adoc
swopebe: This title is a bit too long, recommending we simplify this.

business_continuity/backup_restore/backup_active_restore.adoc
swopebe: To be clear, the comment above was about the content, not the title.

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -1,13 +1,10 @@
 [#keep-hub-active-restore]
-= Keeping the primary hub cluster active during a restore process
+= Keeping the primary hub cluster active during a restore process and keeping your hub data safe
+When you restore data to a new hub cluster while your primary hub cluster is still active, both hub clusters try to manage the same cluster fleet. If the primary hub cluster takes control of the managed clusters, it might override the policy or application changes that you made to your new hub cluster.
```
swopebe: This is where we should in every short description discuss why the user wants to complete this process. The short description as it is presents the problem very well. 

This is taken directly from the Red Hat manual about short descriptions:

_Why users must complete an action. You must build from the information in the title. This content helps users understand why completing an action is important or beneficial to them._

I recommended just adding something related to the title:

Keep your primary hub cluster active to _____ (something like _avoid unexpected changes_)

Then you have solid content right after that describes why those changes might occur.

business_continuity/backup_restore/backup_active_restore.adoc
swopebe: I think the title can remain as it was. My suggestion was more about leading the description:

https://github.com/stolostron/rhacm-docs/pull/8216#discussion_r2349482540
Really just a few words at the beginning is all I think it needs to talk about the benefit before the problem.

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -1,13 +1,11 @@
 [#keep-hub-active-restore]
 = Keeping the primary hub cluster active during a restore process
 
-Restore operations typically occur after a disaster that makes the primary hub cluster fail. When the restore operation starts, you cannot access the primary hub cluster. 
+You are able to keep your hub data safe even when you keep your primary hub cluster active during a restore process. When you restore data to a new hub cluster while your primary hub cluster is still active, both hub clusters try to manage the same cluster fleet. If the primary hub cluster takes control of the managed clusters, it might override the policy or application changes that you made to your new hub cluster.
```
swopebe: We don't need "You are able" (wordy) and I am not sure that we need "keep your data safe" -- 

Is the line I suggested not good for some reason?
_Keep your primary hub cluster active during restore to avoid any unexpected changes._

Or something similar?

birsanv: I find this confusing as a first sentence (You are able to keep your hub data safe even when you keep your primary hub cluster active during a restore process). If I read this as is, I understand this is a statement: there is no problem if you keep your primary active

I suggest updating line 4+5 as following ( basically move 6 first and get rid of the first sentence):

When you restore data to a new hub cluster while your primary hub cluster is still active, both hub clusters try to manage the same cluster fleet. If the primary hub cluster takes control of the managed clusters, it might override the policy or application changes that you made to your new hub cluster. 

In a typical disaster recovery restore operation, the primary hub cluster is inactive, and there is no conflict with the restore hub cluster. You might need to keep the original hub cluster active during or after your restore operation, for example, you would typically need to do this if you run a disaster recovery simulation test.

birsanv (changes requested):

business_continuity/backup_restore/backup_active_restore.adoc
birsanv: I think the first sentence (When you restore data to a new hub cluster while your primary hub cluster is still active, both hub clusters try to manage the same cluster fleet.) explains the problem and why you need to worry about keeping the primary active

swopebe: It doesn't fit what a short description should do, which is tell the user early why they must complete the action.

**This is the reason for the problem or the process:** (I think?)

- When you restore data to a new hub cluster while your primary hub cluster is still active, both hub clusters try to manage the same cluster fleet. _(I cannot tell the way this is written if this is desired or not, and so that is why we need clarity, based on the title.)_ 

- If the primary hub cluster takes control of the managed clusters, it might override the policy or application changes that you made to your new hub cluster. _(This seems undesired and a result of the first line the way it is currently written.)_

If this was a known issue or a troubleshooting doc that leads with the problem, this would work--leading with the problem. If the first line is the benefit of the process, then we need to write it as such.

birsanv: This is how it works, nothing to do with a bug in the product so no, there is no known issue or troubleshooting

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -1,13 +1,11 @@
 [#keep-hub-active-restore]
 = Keeping the primary hub cluster active during a restore process
```
swopebe:
```suggestion
= Running the restore operation while the primary hub cluster is active
```

swopebe (changes requested):

business_continuity/backup_restore/backup_active_restore.adoc
swopebe: Changed to reflect what the topic is about. Upon further discussion and reading, we saw that this topic was not titled to reflect the actual action for the user, and now the short description makes better sense.

business_continuity/backup_restore/backup_active_restore.adoc
swopebe:
```suggestion
When you restore data to a new hub cluster while your primary hub cluster is still active, both hub clusters try to manage the same cluster fleet. If the primary hub cluster takes control of the managed clusters, it might override the policy or application changes that you made to your new hub cluster. 
```

business_continuity/backup_restore/backup_active_restore.adoc
```diff
@@ -1,13 +1,11 @@
...
 
-Restore operations typically occur after a disaster that makes the primary hub cluster fail. When the restore operation starts, you cannot access the primary hub cluster. 
+You are able to keep your hub data safe even when you keep your primary hub cluster active during a restore process. When you restore data to a new hub cluster while your primary hub cluster is still active, both hub clusters try to manage the same cluster fleet. If the primary hub cluster takes control of the managed clusters, it might override the policy or application changes that you made to your new hub cluster. 
 
-If you want to run a disaster recovery test and do not want the primary hub cluster to stop during the test, keep the primary hub cluster active during the entire hub cluster restoration.
+In a typical disaster recovery restore operation, the primary hub cluster is inactive, and there is no conflict with the restore hub cluster. You might need to keep the original hub cluster active during or after your restore operation, for example, you would typically need to do this if you run a disaster recovery simulation test.
```
swopebe, line 4: Line 6 is a better intro now that the topic is titled the way it is.

birsanv (approved):
lgtm

swopebe (approved):

PR #8220, opened 2025-09-15
https://issues.redhat.com/browse/ACM-22457 Add OCP cluster upgrade doc

oafischer:
2.15

Checklist considered

### DIFF
--- a/clusters/cluster_lifecycle/ansible_config_cluster.adoc
+++ b/clusters/cluster_lifecycle/ansible_config_cluster.adoc
@@ -11,6 +11,8 @@
 * <<ansible-status-job-cluster,Viewing the status of an Ansible job>>
 * <<push-cl-cr-job-pod,Pushing custom labels from the _ClusterCurator_ resource to the automation job pod>>
 * <<eus-upgrades, Using the _ClusterCurator_ for Extended Update Support (EUS) upgrades>>
+* <<eus-upgrades, Using the _ClusterCurator_ for Extended Update Support (EUS) upgrades>>
+
 
 [#prerequisites-for-ansible-integration-cluster]
 == Prerequisites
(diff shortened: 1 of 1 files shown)


clusters/cluster_lifecycle/ansible_config_cluster.adoc
```diff
@@ -241,3 +243,33 @@ spec:
...
 ----
+
+[#upgrade-ocp-cluster-curator]
+== Upgrading {ocp-short} clusters by using the _ClusterCurator_
+
+You might want to upgrade your {ocp-short} cluster to a specific newer version to test new features quickly.
+
+*Note:* Upgrading to a newer {ocp-short} cluster version can reduce stability.
+
+To upgrade to a newer to {ocp-short} cluster version, add the following annotation:
```
fxiang1: It's not clear what `newer version` means, I think we should just mention this is to upgrade OCP clusters to a non-recommended version to test new features faster.

oafischer: Updated to "version with known issues" to match the OCP UI.

fxiang1 (approved):

clusters/cluster_lifecycle/ansible_config_cluster.adoc
```diff
@@ -241,3 +243,33 @@ spec:
...
+  desiredCuration: upgrade
+  upgrade:
+    desiredUpdate: <4.x.x> <3>
+    monitorTimeout: 120
+----
+<1> Add your cluster name.
+<2> Add your cluster namespace.
+<3> Add the {ocp-short} cluster version with known issues that you want to upgrade to.
+
+The `ClusterCurator` checks the `desiredUpdate` image digest version at startup in the conditional updates list. If the image digest is not found, the `ClusterCurator` uses the image tag instead.
```
swopebe, line 288: typically a procedure like this needs a couple of steps.

1. Add the following file. Add the following annotation to...... (something like this)
2. Run the following command to apply the file. (?)
3. Something that verifies the procedure worked. (we don't always have this)

We want to get away with just providing a YAML sample without telling the user the procedure to get the YAML in or where to add it, and verify this. There are quite a few examples in the doc and we want to try to make all procedures like this look simlar.

dtthuynh (approved):

clusters/cluster_lifecycle/ansible_config_cluster.adoc
oafischer: Updated

swopebe (approved):

PR #8229, opened 2025-09-17
https://issues.redhat.com/browse/ACM-23766 Remove adv config

oafischer:
2.13, 2.14, 2.15

Checklist considered


business_continuity/backup_restore/backup_velero.adoc
```diff
@@ -0,0 +1,37 @@
...
+After installing Velero, the Velero pod is set to the default CPU and memory limits. If your cluster backs up a large number of resources, you might need to update the Velero resource requests and limits. If you do not update the values, the Velero pod might fail due to an out-of-memory error.
+ 
+.Procedure
+
+. Update the `DataProtectionApplication` resource and add the following `resourceAllocation` template. The following example works with 2000 clusters:
+
+[source,yaml]
+----
+apiVersion: oadp.openshift.io/v1alpha1
+kind: DataProtectionApplication
```
dockerymick: I suggest adding the plus sign to align the command with the step
```suggestion
. Update the `DataProtectionApplication` resource and add the following `resourceAllocation` template. The following example works with 2000 clusters:

+
[source,yaml]
----
apiVersion: oadp.openshift.io/v1alpha1
kind: DataProtectionApplication
```

business_continuity/backup_restore/backup_velero.adoc
```diff
@@ -0,0 +1,37 @@
...
+        resourceAllocations:
+          limits:
+            cpu: "2"
+            memory: 1Gi
+          requests:
+            cpu: 500m
+            memory: 256Mi
+----
+
+[role="_additional-resources"]
```
dockerymick: I'm not sure how this builds, but sure that it is for the conversion. should there be an underscore before additional-resources?

dockerymick (changes requested):

business_continuity/backup_restore/backup_velero.adoc
oafischer: This is the Red Hat modular docs template that we need for conversion, it would build similar to this:

image

The underscore should be there.

dockerymick (approved):
Thanks for clarifying ! LGTM! Checklist considered

business_continuity/backup_restore/backup_velero.adoc
swopebe: I didn't think we were tagging, this would not be across the doc and would only show in a few topics. This is why we said we were waiting on tagging that changes the build output. @dockerymick @oafischer

business_continuity/backup_restore/backup_velero.adoc
```diff
@@ -0,0 +1,38 @@
+[#config-velero-res-limit]
+= Configuring Velero resource requests and limits
+
+After installing Velero, the Velero pod is set to the default CPU and memory limits. If your cluster backs up a large number of resources, you might need to update the Velero resource requests and limits. If you do not update the values, the Velero pod might fail due to an out-of-memory error.
```
swopebe:
```suggestion
The Velero pod is set to the default CPU and memory limits. If your cluster backs up a large number of resources, you might need to update the Velero resource requests and limits. If you do not update the values, the Velero pod might fail due to an out-of-memory error.
```

business_continuity/backup_restore/backup_velero.adoc
oafischer: In that case we would need remove the .Procedure as well since it changes the build output. Should I go ahead and revert both to match the rest of the doc?

swopebe: procedure doesn't put a box around anything

swopebe: But I have not been doing that, either. The only change we started was prereqs to get rid of headers that are not being used in links or anything.

swopebe (approved):

dockerymick (approved):
Looks great! Thanks for your updates and effort!

business_continuity/backup_restore/backup_velero.adoc
```diff
@@ -0,0 +1,37 @@
+[#config-velero-res-limit]
```
swopebe, line 1: @birsanv still needs to approve

We have to move this out of adv config because we have to reduce the amount of topics in each file for conversion and easier Ai access, additionally adv config was becoming a catch all.

birsanv (approved):
lgtm
I understand all that changes is the location for the #resource-requests-and-limits and title to refer to Velero, which looks good

PR #8235, opened 2025-09-22
https://issues.redhat.com/browse/ACM-23768

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-23768

See Gdoc with draft content and dev and doc discussion: https://docs.google.com/document/d/1oQJZkPeV7Z0Na_4N1NwjpUgk8HtAjDD6P5Nlag8UZ0A/edit?usp=sharing


business_continuity/backup_restore/tag_resources.adoc
```diff
@@ -1,14 +1,12 @@
...
-
-When you use an existing hub cluster for a restore hub cluster configuration, tag any resource that you want to be part of the hub cluster backup with the `velero.io/backup-name: backupName` label. Tagging the resources with the Velero annotation marks them as being generated by an earlier restore.
+= Tagging user resources with the backup label 
 
+When you use an existing hub cluster for a restore hub cluster configuration, decide what user-created resource that you want to include in the hub cluster, then tag that resource with the `velero.io/backup-name: backupName` label. If you tag these resources with the Valero annotation, it looks like the resources were generated by an earlier restore. Because of the Valero annotation,  when you run a hub restore operation and set the `cleanupBeforeRestore` option to `CleanupRestored` on the restore resource, user resources that are not part of your current restored backup would get cleaned up. As a result, you can restore hub data on an existing cluster and produce a cluster that is the same as the initial hub.
```
birsanv: typo : Valero should be Velero

business_continuity/backup_restore/tag_resources.adoc
```diff
@@ -1,14 +1,12 @@
...
+ 
 [#prerequisite-resources]
 == Prerequisites
 
-* Ensure you understand the constraints of using an existing hub cluster as a restore hub cluster with resources you created before the restore operation. See xref:../backup_restore/use_existing_hub_cluster.adoc#using-existing-hub[Constraints for using an existing hub cluster as a restore hub cluster].
+* Ensure you understand the constraints of using an existing hub cluster as a restore hub cluster with resources you created before the restore operation. See xref:../backup_restore/use_existing_hub_cluster.adoc#using-existing-hub[Considerations when using an existing hub cluster as a restore hub cluster].
```
birsanv: should we use the word : consideration here ( instead of constraint ) : Ensure you understand the **constraints**

business_continuity/backup_restore/use_existing_hub_cluster.adoc
```diff
@@ -1,7 +1,7 @@
 [#using-existing-hub]
-= Constraints for using an existing hub cluster as a restore hub cluster
+= Considerations when using an existing hub cluster as a restore hub cluster
 
-If you use an existing hub cluster as a restore hub cluster with resources that you created before the restore operation, review the following constraints:
+If you use an existing hub cluster as a restore hub cluster with resources that you created before the restore operation, consider the following situations:
 
 * Existing managed cluster are detached during the restore process.
```
birsanv: I think we said : Existing managed cluster **may be** detached during the restore process. ( isnt't it ? )

business_continuity/backup_restore/use_existing_hub_cluster.adoc
```diff
@@ -1,7 +1,7 @@
...
 * There are more resources on the restore hub cluster than the backup hub cluster.
```
birsanv: Here I am not 100% sure but did we go with : There **may be** more resources on the restore hub cluster than the backup hub cluster.  ( since those extra resources would go away when we run with CleanupRestored, as we explain, if they have the velero label )

birsanv (approved):
lgtm

business_continuity/backup_restore/tag_resources.adoc
```diff
@@ -1,14 +1,12 @@
...
-
-When you use an existing hub cluster for a restore hub cluster configuration, tag any resource that you want to be part of the hub cluster backup with the `velero.io/backup-name: backupName` label. Tagging the resources with the Velero annotation marks them as being generated by an earlier restore.
+= Tagging user resources with the backup label 
 
+When you use an existing hub cluster for a restore hub cluster configuration, decide what user-created resource that you want to include in the hub cluster, then tag that resource with the `velero.io/backup-name: backupName` label. If you tag these resources with the Velero annotation, it looks like the resources were generated by an earlier restore. Because of the Valero annotation,  when you run a hub restore operation and set the `cleanupBeforeRestore` option to `CleanupRestored` on the restore resource, user resources that are not part of your current restored backup would get cleaned up. As a result, you can restore hub data on an existing cluster and produce a cluster that is the same as the initial hub.
```
oafischer: Paragraph is a little long, I recommend breaking it up in 2 or 3 pieces.

swopebe: **Few things:** 

I agree this can be two paragraphs, maybe "Because" on the next line. It will increase readability.

There is an extra space in this line after _annotation._

_it looks like_ -- not sure what "it" is supposed to be--want to try to avoid using it unless things are very clear.

We typically don't use "looks like."

_Looks_ can mean many things translated.

Appears, displays, something like that is better.

And again, try to avoid "woulds" or "shoulds" per modal guidance:

_user resources that are not part of your current restored backup **would get** cleaned up._

Always try to reword. You say "when you run" (present, that's good), so the correct tense to use is present in the predicate: "are" cleaned up--should work.

business_continuity/backup_restore/use_existing_hub_cluster.adoc
```diff
@@ -1,10 +1,10 @@
 [#using-existing-hub]
-= Constraints for using an existing hub cluster as a restore hub cluster
+= Considerations when using an existing hub cluster as a restore hub cluster
```
swopebe: Why are we changing this? We don't normally use this word in the doc, so wondering the reason for the change.

business_continuity/backup_restore/use_existing_hub_cluster.adoc
```diff
@@ -1,10 +1,10 @@
...
 
-If you use an existing hub cluster as a restore hub cluster with resources that you created before the restore operation, review the following constraints:
+If you use an existing hub cluster as a restore hub cluster with resources that you created before the restore operation, consider the following situations:
```
swopebe: I would say:

consider the following possible results: 

(situations seems very broad)

business_continuity/backup_restore/use_existing_hub_cluster.adoc
```diff
@@ -17,8 +17,10 @@ If the hub cluster you plan to use for the restore operation already manages clu
 [#restore-hub-resources]
 == Restore hub cluster has more resources than the backup hub cluster
 
-When you restore a hub cluster with earlier user-created data and the `cleanupBeforeRestore` option is set to `CleanupRestored` for the restore operation, the resources of your restore hub cluster do not match the resources of the primary hub cluster. The user data includes the following resources: `policy`, `application`, `credential`, or any custom resource defined by the user. The resources are not cleaned because the resources are created by the users and not created by an earlier backup.
+When you restore data on a hub cluster with earlier user-created data and the `cleanupBeforeRestore` option is set to `CleanupRestored` for the restore operation, user-created resources are not cleaned up. This is because they are created by the user and not by an earlier restore.
```
swopebe: Again, these sentences are very long. I have to read very slowly and carefully to try to understand what is going on here, and that is a sign that we need to 1. Understand what the draft is saying and 2. Try to write it more concisely. This is for the whole book.

When you restore data on a hub cluster with **existing** data, and the `cleanupBeforeRestore` option is set to `CleanupRestored` for the restore operation, that **existing** data is not cleaned up.

Or can we use "previously created" instead of "earlier user-created" which is a lot.

swopebe: same for below.

business_continuity/backup_restore/use_existing_hub_cluster.adoc
```diff
@@ -17,8 +17,10 @@ If the hub cluster you plan to use for the restore operation already manages clu
...
+When you restore data on a hub cluster with earlier user-created data and the `cleanupBeforeRestore` option is set to `CleanupRestored` for the restore operation, user-created resources are not cleaned up. This is because they are created by the user and not by an earlier restore. 
 
-If you intend to use an existing hub cluster for a restore hub cluster configuration and expect for the hub cluster to have the same content as the backup hub cluster after the restore is completed, tag any resource that you want to be part of a hub cluster backup. See xref:../backup_restore/tag_resources.adoc#tagging-resources[Tagging resources] for more information.
+Despite the user-created resource not getting cleaned up, you can still use an existing hub cluster for a restore hub cluster configuration while making this hub cluster have the same content as the backup hub cluster. To do this,  tag all user resources with the backup label so that they look like they are generated by an earlier restore.
```
swopebe: Same here:

Despite the user-created resource not getting cleaned up, you can still use an existing hub cluster for a restore hub cluster configuration while making this hub cluster have the same content as the backup hub cluster.

This line is very confusing, and can use some work for conciseness.

_making this hub cluster have the same content_ can definitely be reduced, for instance

swopebe: _keeping the same content on the current and backup hub clusters_? Instead of _while making this hub cluster have the same content_

business_continuity/backup_restore/use_existing_hub_cluster.adoc
birsanv: These are not results though. What we show here is what are the things to consider if you are in different situations ( situation = hub scenario , or context )

business_continuity/backup_restore/use_existing_hub_cluster.adoc
birsanv: Because we review scenarios ( situations ) where you want to use an existing hub as a restore cluster and list what are this things that you should pay attention to in this case.
So these are not constraints but things to consider when planning to use an existing hub

business_continuity/backup_restore/use_existing_hub_cluster.adoc
swopebe: Then can we say scenarios? I have never really used "situation" in doc to discuss user scenarios. Never seen it used, so that is why I am asking. I think it is a bit more specific, but not hard-lined on this.

business_continuity/backup_restore/use_existing_hub_cluster.adoc
swopebe: Ok, but it's a lot like "Recommendations" which we cannot use based on guidance for _Recommends:_


**Description: Avoid "recommends".** Instead of "Red Hat recommends", direct users to take the recommended action. This allows Red Hat to be more prescriptive in documentation and prevent any user uncertainty, and is easier for upstream or downstream coordinated efforts.

For example, instead of "Red Hat recommends using X package because", write "Use this package because" or "Use this package when".


I don't see another way to word this, though, if constraints no longer works.

business_continuity/backup_restore/use_existing_hub_cluster.adoc
```diff
@@ -1,10 +1,10 @@
 [#using-existing-hub]
-= Constraints for using an existing hub cluster as a restore hub cluster
+= Considerations when using an existing hub cluster as a restore hub cluster
 
-If you use an existing hub cluster as a restore hub cluster with resources that you created before the restore operation, review the following constraints:
+If you use an existing hub cluster as a restore hub cluster with resources that you created before the restore operation, consider the following situations:
 
-* Existing managed cluster are detached during the restore process.
-* There are more resources on the restore hub cluster than the backup hub cluster.
+* The existing managed cluster might be detached during the restore process.
```
swopebe: Might "be" detached or "might detach during"? It's not clear here bc of **during**.

Either we need to say: **might become detached** OR **might detach** (the more concise is better if it does not change the meaning) -- and this is to match the rest of the sentence.

business_continuity/backup_restore/use_existing_hub_cluster.adoc
swopebe: Ok, it's a lot like "recommendations" which we cannot use, but I understand why you are changing this and not sure there is another way.

business_continuity/backup_restore/use_existing_hub_cluster.adoc
```diff
@@ -17,8 +17,10 @@ If the hub cluster you plan to use for the restore operation already manages clu
 [#restore-hub-resources]
 == Restore hub cluster has more resources than the backup hub cluster
 
-When you restore a hub cluster with earlier user-created data and the `cleanupBeforeRestore` option is set to `CleanupRestored` for the restore operation, the resources of your restore hub cluster do not match the resources of the primary hub cluster. The user data includes the following resources: `policy`, `application`, `credential`, or any custom resource defined by the user. The resources are not cleaned because the resources are created by the users and not created by an earlier backup.
+When you restore data on a hub cluster that includes data that the user created, and the `cleanupBeforeRestore` option is set to `CleanupRestored` for the restore operation, that data is not cleaned up. This is because the data is created by the user and not by an earlier restore.
```
swopebe: This is still a lot of words and a bit confusing, but better. User-created was fine to use here, but it was not easy to understand what you meant by earlier created.

business_continuity/backup_restore/use_existing_hub_cluster.adoc
```diff
@@ -17,8 +17,10 @@ If the hub cluster you plan to use for the restore operation already manages clu
...
```
swopebe, line 21: The feedback for line 22 still remains:
_make this hub cluster have the same content as the backup hub cluster_

swopebe: _To do this_ -- we can remove that, it's self-referential.

swopebe: _making it look like an earlier restore generated these resources._
Did my comments previously not get submitted, maybe?

swopebe (changes requested):
previous comments may not have been seen on these items.

business_continuity/backup_restore/use_existing_hub_cluster.adoc
swopebe: Please see the above comments

business_continuity/backup_restore/use_existing_hub_cluster.adoc
swopebe: I just checked and those comments are above.

business_continuity/backup_restore/use_existing_hub_cluster.adoc
jc-berger: HI, I meant that the user-created data was made on the hub cluster before our user (reader of the doc) restores data on the hub cluster. I removed the whole "earlier" because I didn't think it was needed, and we were trying to cut down word count. 

I'll keep refining this sentence though, thanks!

business_continuity/backup_restore/use_existing_hub_cluster.adoc
birsanv: might detach - looks good

business_continuity/backup_restore/use_existing_hub_cluster.adoc
jc-berger: Thanks, I applied your suggestion, as it is much better to read. 

With that said, I think it's more of a matter of the user _making_ the content the same on both hubs rather than _keeping_ the content the same on both hubs. 

Maybe something like this, `You can make the backup and hub cluster have the same content.` 

@birsanv what do you think? Can we keep it as I have it now `You can keep the same content on the current and backup hub clusters` or something along the lines of `You can make the backup and hub cluster have the same content.`

business_continuity/backup_restore/use_existing_hub_cluster.adoc
jc-berger: @swopebe Did you have an issue or suggested rewrite for, `making it look like an earlier restore generated these resources.` ? 

@birsanv confirm if I'm correct, but the whole result of tagging resources with the backup label is so that all the resources have continuity by looking like they've all been generated by the same restore.

business_continuity/backup_restore/use_existing_hub_cluster.adoc
```diff
@@ -17,8 +17,10 @@ If the hub cluster you plan to use for the restore operation already manages clu
...
+When you restore data on a hub cluster that includes user-created data, and the `cleanupBeforeRestore` option is set to `CleanupRestored` for the restore operation, that data is not cleaned up. This is because the data is created by the user and not by an earlier restore. 
 
-If you intend to use an existing hub cluster for a restore hub cluster configuration and expect for the hub cluster to have the same content as the backup hub cluster after the restore is completed, tag any resource that you want to be part of a hub cluster backup. See xref:../backup_restore/tag_resources.adoc#tagging-resources[Tagging resources] for more information.
+Even when the user-created resources do not get cleaned up, you can still use an existing hub cluster for a restore hub cluster configuration. You can keep the same content on the current and backup hub cluster. Tag all user resources with the backup label, making it look like an earlier restore generated these resources.
```
swopebe: DO we need to tag `backup` label on line 22?

there are many ways you can say this, but I was looking to you to see about making it more concise based on expertise about the goal in line 22.

_making it look like_ (here "it" is not clear, so I don't know what to replace it with)
_like like_ can be "matching" "appear" .....

It seems maybe....

Tag all user resources with the `backup `label so that the resources on the current and back up cluster match.

...to ensure the resources on the current and back up cluster match.

...so that the current and back up cluster resources match

Just a few ideas. Mainly we need to be clear about what "it" is and try to use fewer words for that last line.

business_continuity/backup_restore/use_existing_hub_cluster.adoc
swopebe: We just try to get away from using **have** and other verbs that take away from active voice.

business_continuity/backup_restore/use_existing_hub_cluster.adoc
swopebe: it's fine, and better, though we cannot use "may" so we need to say might. 

Existing managed cluster are detached during the restore process. (If this is singular, we say is. If plural, we say are)

business_continuity/backup_restore/use_existing_hub_cluster.adoc
```diff
@@ -19,7 +19,7 @@ If the hub cluster you plan to use for the restore operation already manages clu
 
 When you restore data on a hub cluster that includes user-created data, and the `cleanupBeforeRestore` option is set to `CleanupRestored` for the restore operation, that data is not cleaned up. This is because the data is created by the user and not by an earlier restore. 
 
-Even when the user-created resources do not get cleaned up, you can still use an existing hub cluster for a restore hub cluster configuration. You can keep the same content on the current and backup hub cluster. Tag all user resources with the backup label, making it look like an earlier restore generated these resources. 
+Even when the user-created resources do not get cleaned up, you can still use an existing hub cluster for a restore hub cluster configuration. You can make the backup and hub cluster have the same content by tagging all user resources with the `backup` label. The `backup` label makes all the user-created resources look like they were generated by the same restore configuration.
```
birsanv: 1. use `cluster.open-cluster-management.io/backup` instead of `backup`
2. Change: The `backup` label makes all the user-created resources look like they were generated by the same restore configuration.
with: The `cluster.open-cluster-management.io/backup` label makes all the user-created resources look like they were generated by an an earlier restore operation.
3. Change : You can make the backup and hub cluster 
with : You can make the backup and restore hub cluster

business_continuity/backup_restore/use_existing_hub_cluster.adoc
```diff
@@ -17,8 +17,10 @@ If the hub cluster you plan to use for the restore operation already manages clu
...
+When you restore data on a hub cluster that includes user-created data, and the `cleanupBeforeRestore` option is set to `CleanupRestored` for the restore operation, that data is not cleaned up. This is because the data is created by the user and not by an earlier restore. 
 
-If you intend to use an existing hub cluster for a restore hub cluster configuration and expect for the hub cluster to have the same content as the backup hub cluster after the restore is completed, tag any resource that you want to be part of a hub cluster backup. See xref:../backup_restore/tag_resources.adoc#tagging-resources[Tagging resources] for more information.
+Even when the user-created resources do not get cleaned up, you can still use an existing hub cluster for a restore hub cluster configuration. You can make the backup and hub cluster have the same content by tagging all user resources with the `backup` label. The `backup` label makes all the user-created resources look like they were generated by the same restore configuration. 
+
```
swopebe, line 23: There is a lot going on on line 22.  

_You can make the backup and hub cluster have the same content by tagging all user resources with the `backup` label. The `backup` label makes all the user-created resources look like they were generated by the same restore configuration._

If you try to cut that down, you can come up with a more active way of saying this, with fewer lines. Maybe?

_Tag the resources with the ____ label to make all the resources on the restore hub cluster and the backup cluster appear the same, as if they were generated by the same restore operation._

Just an example. That covers the two lines.

swopebe: _by tagging all user resources with the `backup` label._ (this is passive, and we need to write in active voice per the style manual, so I am suggesting the change above^^^) Also merging the two thoughts and taking out extra wording.

swopebe (approved):

PR #8237, opened 2025-09-22
https://issues.redhat.com/browse/ACM-22664--update

swopebe:


release_notes/acm_deprecate_remove.adoc
```diff
@@ -95,7 +95,7 @@ A _removed_ item is typically function that was deprecated in previous releases
 | Importing {ocp-short} 3.11 clusters from the console
 | 2.14
 | Upgrade to a supported version of {ocp-short}. See the link:https://docs.redhat.com/en/documentation/openshift_container_platform/4.17[{ocp-short} documentation].
-| The support for the deprecated custom resource definition `v1beta1` APIs is removed. The `v1beta1` APIs were only needed for {ocp-short} 3.11. The import command for {ocp-short} 3.11 is no longer displayed and support for this version was previously removed from {acm-short}.
+| The support for the deprecated custom resource definition `v1beta1` APIs is removed. The `v1beta1` APIs were only needed for {ocp-short} 3.11.
```
swopebe, line 98: I have removed extra details from this because at this point, we don't need to carry that over. However, we usually keep something that has been removed located in the table for two releases. So we don't need to remove this for 2.15, but I will know to remove it for 2.16 (no need to send an issue, as it is part of my GA prep to remove old items from the remove table.

swopebe: @zhujian7 If this is OK, I think we can approve and close the issue, no need for QE or Peer review.

zhujian7: 👌🏻

PR #8242, opened 2025-09-23
https://issues.redhat.com/browse/ACM-18871 Scaling metrics ("sharding feature")

dockerymick:


observability/obs_scale.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+Scale the metrics collection for your clusters to increase or decrease the remote write requests that you receive. When you scale metrics collection, you are controlling the resource capacity for your cluster.
+
+*Required access:* Cluster administrator
+
+.Prerequisistes
+
+- You have the Observability service enabled on your hub and managed clusters.
+
+[inc-dec-metric]
```
dockerymick:
```suggestion
[#inc-dec-metric]
```

observability/obs_scale.adoc
```diff
@@ -0,0 +1,28 @@
+[#obs-scale-metrics]
+= Scaling metrics collection (Technology Preview)
+
+Scale the metrics collection for your clusters to increase or decrease the remote write requests that you receive. When you scale metrics collection, you are controlling the resource capacity for your cluster.
+
+*Required access:* Cluster administrator
+
+.Prerequisistes
+
+- You have the Observability service enabled on your hub and managed clusters.
```
dockerymick:
```suggestion
- You have the Observability service enabled on your hub and managed clusters. See xref:../observability/observability_enable.adoc#enabling-observability-service[Enabling the Observability service].
```

observability/obs_scale.adoc
```diff
@@ -0,0 +1,28 @@
...
++
+[source,bash]
+----
+need command if there is one the user can run
+----
+
+. *Optional:* If you want to override the `workers` parameter for specific clusters, add the `observability.open-cluster-management.io/addon-source: "override"` annotation to the `ObservabilityAddOn` specification of your managed cluster.
+.. To revert the override, use the `observability.open-cluster-management.io/addon-source: "mco"` annotation.
+
+. To verify the value of the `workers` parameter, check the arguments on the `metrics-collector` pods
```
dockerymick: @saswatamcode How does the user do this? Do they run a command for the info to be displayed from the terminal? 
```suggestion
. To verify the value of the `workers` parameter, check the arguments on the `metrics-collector` pods.
```

observability/obs_scale.adoc
```diff
@@ -0,0 +1,28 @@
+[#obs-scale-metrics]
+= Scaling metrics collection (Technology Preview)
+
+Scale the metrics collection for your clusters to increase or decrease the remote write requests that you receive. When you scale metrics collection, you are controlling the resource capacity for your cluster.
```
saswatamcode: Not exactly what this is doing.

A user would scale metrics collector workers, when they notice that metrics-collector pods in their spoke clusters are struggling with the load of the metrics present in their clusters.

Increasing the workers, would help metrics-collector split that load between goroutines and handle that load by sending multiple smaller remote write requests to Thanos, instead of one large one.

observability/obs_scale.adoc
saswatamcode: A user would set this on the MCO CR.

observability/obs_scale.adoc
```diff
@@ -0,0 +1,28 @@
...
+.Prerequisistes
+
+- You have the Observability service enabled on your hub and managed clusters. See xref:../observability/observability_enable.adoc#enabling-observability-service[Enabling the Observability service].
+
+[#inc-dec-metric]
+== Increasing and decreasing metrics collection
+
+To increase and decrease metrics collection on your clusters, you must edit the `workers` parameter in the `multicluster-observability-operator` resource. Complete the following steps:
+
+. If you want to use the same `workers` value for each of your managed clusters, set the value in your `MultiClusterObservability` custom resource definition. By default, the value is set to `1`. Run the following example command to edit your `workers` parameter:
```
saswatamcode: There is also a way to set an annotation on the ObservabilityAddon CR in spoke clusters, to override this value on single spoke cluster level

observability/obs_scale.adoc
dockerymick:
```suggestion
Scale the metrics collection for your clusters by increasing or decreasing the remote write requests that you receive. When you increase the value of the `workers` parameter, the amount of write request become smaller between goroutines. If you notice a backlog of metrics from the `metrics-collector` pod of your manage cluster, decrease the remote write requests.
```

observability/obs_scale.adoc
dockerymick:
```suggestion
. To verify the value of the `workers` parameter, check the values in the `MultiClusterObservability` custom resource.
```

observability/obs_scale.adoc
dockerymick: I listed it as an Optional step already

observability/obs_scale.adoc
```diff
@@ -0,0 +1,28 @@
...
+
++
+[source,bash]
+----
+need command if there is one the user can run
```
dockerymick: @subbarao-meduri is there a command that the user can run to change the `workers` value?

observability/observability_arch.adoc
```diff
@@ -123,6 +124,7 @@ When you install {acm-short} the following persistent volumes (PV) must be creat
 
 *Note:* The time series historical data is stored in object stores. Thanos uses object storage as the primary storage for metrics and metadata related to them. For more details about the object storage and downsampling, see _Enabling observability service_.
 
+//should there be a description of the metrics-collector here
```
dockerymick: Does metrics-collector need to be mentioned here, especially since it is mentioned in the diagram @subbarao-meduri

observability/observe_environments_intro.adoc
```diff
@@ -14,13 +14,5 @@ Read the following documentation for more details about the observability compon
 * xref:../observability/use_observability.adoc#using-observability[Using observability]
 * xref:../observability/observability_alerts.adoc#observability-alerts[Managing alerts]
 * xref:../observability/adv_config_obs.adoc#adv-config-obs[Observability advanced configuration]
-** xref:../observability/obs_metrics.adoc#adding-custom-metrics[Adding custom metrics]
```
jacobbaungard, line 17: Why are all these removed?

observability/observability_enable.adoc
```diff
@@ -597,11 +597,11 @@ spec:
   imagePullSecret: multiclusterhub-operator-pull-secret
   observabilityAddonSpec: <1>
     enableMetrics: false <2>
-  workers: <3> 
+  workers: 1 <3> 
 ----
 <1> Use the `observabilityAddonSpec` parameter to define the global settings for all managed clusters that have the Observability add-on enabled.
 <2> Use the `enableMetrics` parameter to indicate that the Observability add-on is enabled to push metrics to hub cluster server.
-<3> *Technology Preview*: Use the `workers` parameter to increase the number of internal workers in the metric collector process. The internal workers shard `/federate` endpoint requests made to Prometheus on your managed cluster and send separate remote-write requests to Thanos on your hub cluster.
+<3> *Technology Preview*: Use the `workers` parameter to increase the number of internal workers in the metric collector process. The default value is `1`.
```
jacobbaungard: Could we not delete this line completely? The amount of workers is not relevant in this section about "Disabling Observability on all clusters", and I think might cause confusion. Especially since we will now have another section about this feature.

observability/observe_environments_intro.adoc
dockerymick: We have to remove sub topics out of in-file table of contents. We are not actually removing the information. When topics are expanded from the navigation menu, the subtopics are also listed. So this part will help clean up the doc

Image

dockerymick: The topics are listed from the "Observability advanced configuration" topic as well:

Image

observability/obs_scale.adoc
```diff
@@ -0,0 +1,28 @@
+[#obs-scale-metrics]
+= Scaling metrics collection (Technology Preview)
+
+Scale the metrics collection for your clusters by increasing or decreasing the remote write requests that you receive. When you increase the value of the `workers` parameter, the amount of write request become smaller between goroutines. If you notice a backlog of metrics from the `metrics-collector` pod of your manage cluster, decrease the remote write requests.
```
jacobbaungard:
```suggestion
By default the metrics-collector federates all metrics from Prometheus and write them to Thanos in single requests. This can be problematic in larger environments, as single requests might become too large to efficiently handle. By increasing the worker count beyond 1 (default), the metric collector will split the metrics to collect between multiple workers (goroutines), resulting in more, but smaller requests being made.
```
I would suggest we write it something like this. It quickly gives the reader an understanding of the default, the situation where the default may be inappropriate, and then explain how this setting can solve the problem.

jacobbaungard: Another suggestion after discussion with Thibault:

```suggestion
To enhance performance in high-scale deployments, the metrics-collector supports parallel metric collection. The default configuration uses a single worker to federate metrics. By increasing the number of workers, you enable a sharded collection strategy where the workload is divided among them.
```

observability/observability_arch.adoc
jacobbaungard: No. This section is about "Persistent stores used in the observability service", the metrics-collector doesn't have persistent storage.

observability/obs_scale.adoc
jacobbaungard: They need to edit the resource created in this step https://github.com/stolostron/rhacm-docs/blob/56b003e69d78c57cf77b7cb8f63e2d08768bef9a/observability/observability_enable.adoc#creating-the-multiclusterobservability-custom-resource

For example, to change to 4 workers, one need to update the `workers` parameter under the `observabilityAddonSpec` to 4.

```
apiVersion: observability.open-cluster-management.io/v1beta2
kind: MultiClusterObservability
metadata:
  name: observability
spec:
  observabilityAddonSpec:
    enableMetrics: true
    workers: 4
....
```

observability/observability_arch.adoc
dockerymick: Ok, what about if it should be mentioned after the arch diagram ?

observability/obs_scale.adoc
dockerymick: Thanks!

observability/observability_enable.adoc
dockerymick: That's a great point, thank you! I'll remove it

observability/observability_arch.adoc
jacobbaungard: I don't think it's needed. We don't have "mentions" of every single component explicitly in the text.

observability/observability_arch.adoc
```diff
@@ -56,15 +56,16 @@ See the following list to learn which component versions Oberservability uses in
...
+The following diagram shows the components of Observability:
 
-image:../images/observability-arch-29.png[Multicluster observability architecture]
+image:../images/observability-arch-29.png[Multicluster Observability architecture]
 
-The components of the observability architecture include the following items:
+The components of the Observability architecture include the following items:
 
 - The multicluster hub operator, also known as the `multiclusterhub-operator` pod, deploys the `multicluster-observability-operator` pod. It is the root component that deploys resources for the {acm-short} Observability service, such as the metrics store on the hub cluster and collectors on managed clusters through the generation of `ManifestWorks` resources.
+** *Technology Preview*: Use the `workers` parameter to increase the number of internal workers in the metric collector process. The internal workers shard `/federate` endpoint requests made to Prometheus on your managed cluster and send smaller,separate remote-write requests to Thanos on your hub cluster.
```
jacobbaungard: I don't think we need this here. You can incorporate what is written here in the `obs_scale` doc if needed.

dockerymick:
@jacobbaungard thank you so much for reviewing and providing suggestions! When you have a chance, please re-review. @subbarao-meduri will you also review to double-check the content?

observability/observability_enable.adoc
```diff
@@ -597,11 +597,10 @@ spec:
   imagePullSecret: multiclusterhub-operator-pull-secret
   observabilityAddonSpec: <1>
     enableMetrics: false <2>
-  workers: <3> 
+  workers: 1 
 ----
 <1> Use the `observabilityAddonSpec` parameter to define the global settings for all managed clusters that have the Observability add-on enabled.
 <2> Use the `enableMetrics` parameter to indicate that the Observability add-on is enabled to push metrics to hub cluster server.
-<3> *Technology Preview*: Use the `workers` parameter to increase the number of internal workers in the metric collector process. The internal workers shard `/federate` endpoint requests made to Prometheus on your managed cluster and send separate remote-write requests to Thanos on your hub cluster.
```
swopebe, line 604: I am reviewing this as a new feature doc. What this part of that new feature? Meaning was this part of the ACM-18871? I am asking bc we moved a tech preview item from this list.

observability/obs_scale.adoc
```diff
@@ -0,0 +1,35 @@
+[#obs-scale-metrics]
+= Scaling metrics collection (Technology Preview)
+
+To enhance performance in high-scale deployments, configure the `metrics-collector` to collect data from multiple sources. The default configuration uses a single worker to federate metrics. By increasing the number of workers in the metric collection process, you enable internal workers to shard `/federate` endpoint requests made to Prometheus on your managed cluster and send smaller, separate remote-write requests to Thanos on your hub cluster.
```
swopebe:
```suggestion
To enhance performance in high-scale deployments, configure the `metrics-collector` to collect data from multiple sources. The default configuration uses a single worker to federate metrics. By increasing the number of workers in the metric collection process, you enable internal workers to shard `/federate` endpoint requests that are made to Prometheus on your managed cluster. The internal workers also send smaller, separate remote-write requests to Thanos on your hub cluster.
```

swopebe: Can we make this a new sentence?

and send smaller, separate remote-write requests to Thanos on your hub cluster.

observability/obs_scale.adoc
```diff
@@ -0,0 +1,35 @@
...
+.Prerequisistes
+
+- You have the Observability service enabled on your hub and managed clusters. See xref:../observability/observability_enable.adoc#enabling-observability-service[Enabling the Observability service].
+
+[#inc-dec-metric]
+== Increasing and decreasing metrics collection
+
+To increase and decrease metrics collection on your clusters, you must edit the `workers` parameter in the `multicluster-observability-operator` resource. Complete the following steps:
+
+. If you want to use the same `workers` value for each of your managed clusters, set the value in your `MultiClusterObservability` custom resource definition. By default, the value is set to `1`. For example, change the `workers` parameter value to `4`. Your YAML might resemble the following resource:
```
swopebe:
```suggestion
. If you want to use the same `workers` value for each of your managed clusters, set the value in your `MultiClusterObservability` custom resource definition. By default, the value is set to `1`. For example, change the `workers` parameter value to `4`. Your YAML file might resemble the following resource: 
```

observability/observability_enable.adoc
jacobbaungard: The feature was added back in 2.12, seems it was only documented at this place in the doc until now . However it makes no sense to document it here, in a section about disabling observability.

swopebe (changes requested):

observability/obs_scale.adoc
jacobbaungard: The first sentence is now inaccurate. We do not collect metrics from multiple sources, we collect metrics from the same source in parallel (i.e from multiple workers). I recommend the wording from my [previous suggestion](https://github.com/stolostron/rhacm-docs/pull/8242#discussion_r2420427525).

In general I would say, this paragraph is probably better to keep out the internal implementation specifics, as we did in the earlier suggestion. While the rest of the paragraph is accurate, I think it's easy to misunderstand not knowing the implementation details in advance. I am not sure how to re-write it better.

observability/obs_scale.adoc
```diff
@@ -0,0 +1,35 @@
+[#obs-scale-metrics]
+= Scaling metrics collection (Technology Preview)
+
+To enhance performance in high-scale deployments, configure the `metrics-collector` to collect data from multiple sources. The default configuration uses a single worker to federate metrics. By increasing the number of workers in the metric collection process, you enable internal workers to shard `/federate` endpoint requests made to Prometheus on your managed cluster and send smaller, separate remote-write requests to Thanos on your hub cluster.
+
+*Required access:* Cluster administrator
+
+.Prerequisistes
```
oafischer:
```suggestion
.Prerequisites
```

observability/obs_scale.adoc
```diff
@@ -0,0 +1,35 @@
...
+
+- You have the Observability service enabled on your hub and managed clusters. See xref:../observability/observability_enable.adoc#enabling-observability-service[Enabling the Observability service].
+
+[#inc-dec-metric]
+== Increasing and decreasing metrics collection
+
+To increase and decrease metrics collection on your clusters, you must edit the `workers` parameter in the `multicluster-observability-operator` resource. Complete the following steps:
```
oafischer:
```suggestion
To increase and decrease metrics collection on your clusters, edit the `workers` parameter in the `multicluster-observability-operator` resource. Complete the following steps:
```

Not sure if need to say "must" here

observability/obs_scale.adoc
oafischer:
```suggestion
. If you want to use the same `workers` value for each of your managed clusters, set the value in your `MultiClusterObservability` custom resource definition. The default value is `1`. For example, change the `workers` parameter value to `4`. Your YAML might resemble the following resource: 
```

This is a bit shorter, might help with the long step

oafischer (approved):
Some small suggestions, otherwise looks great. Thanks!

observability/obs_scale.adoc
dockerymick: Yup thanks for the suggestion. I updated your suggestion so that I can commit your original suggestion and separate the statement.

dockerymick: Got it, thanks Jacob. I was attempting to use a different word from "parallel". Your explanation helps!

dockerymick:
Made a few more updates. @swopebe @jacobbaungard please re-review when you are available

swopebe (approved):

harshil-codes (approved):

observability/obs_scale.adoc
```diff
@@ -0,0 +1,35 @@
+[#obs-scale-metrics]
+= Scaling metrics collection (Technology Preview)
+
+To enhance performance in high-scale deployments, configure the `metrics-collector` to collect metrics from the same source. The default configuration uses a single worker to federate metrics. By increasing the number of workers in the metric collection process, you enable internal workers to shard `/federate` endpoint requests that are made to Prometheus on your managed cluster. The internal workers also send smaller, separate remote-write requests to Thanos on your hub cluster.
```
jacobbaungard: We've gone over this paragraph a lot of times now. It is still no good. Maybe we can try another approach to this. Can you explain what is wrong with the paragraph we provided here: https://github.com/stolostron/rhacm-docs/pull/8242#discussion_r2420427525 ?

> To enhance performance in high-scale deployments, configure the `metrics-collector` to collect metrics from the same source. 

We _always_ collect the metrics from the same source. The worker argument doesn't change that. When changing the worker argument you start to collect metrics in _parallel_ from the same source with multiple workers. The metrics to collect is split between each worker. This _can_ in some cases, improve performance in high-scale environments. Further, the sentence here is very "assertive" (i.e why not at least "To enhance performance, you _can_ configure" or "it is possible to...", or "the metric-collect supports...") ?

> By increasing the number of workers in the metric collection process, you enable internal workers to shard `/federate` endpoint requests that are made to Prometheus on your managed cluster.

This is true

> The internal workers also send smaller, separate remote-write requests to Thanos on your hub cluster.|

This is confusing, and probably hard to understand not knowing the implementation details in advanced. This is what each worker do:

1) Collect metrics from the Prometheus `/federate` endpoint
2) remote-write these metrics to the hub

When there is 1 worker, this single worker collects all the metrics in one go, and then send all of these metrics, again in one go, to the hub, over the remote write protocol.

When there are 2 workers, worker 1 collect half the metrics, and send this to the hub over the remote-write protocol. Worker 2 collects the other half of the metrics, and send this to the hub over the remote-write protocol. These workers do this independently and in [parallel](https://en.wikipedia.org/wiki/Parallel_computing).

When you enable sharding, the result is that we have more remote-write requests to the hub, but each request is smaller. I don't think this is communicated well as it is here.

I believe we are providing too many implementation specifics here, which are 1) hard to explain well without fully explaining how the code works 2) not necessary for the customer to understand.

jacobbaungard (changes requested):

observability/obs_scale.adoc
dockerymick: @jacobbaungard thank you for taking the time to explain why you made your initial suggestion. I value your perspective and appreciate you reviewing. 

It is not my intention to be "assertive", but to maintain an active tone and be concise. I do understand your point though, meaning that configuring the metrics-collector is not the only way to enhance the performance in high-scale deployments. I avoided saying "metrics-collector supports...", to adhere to the IBM Style Guide. When I read the statement, I interrupted it as anthropomorphism: https://www.ibm.com/docs/en/ibm-style?topic=grammar-anthropomorphism 

At first i didn't use "parallel", because it is directional language. However, after I talked to the team, we searched for the word from the Word Usage page, and it did not appear as a term that we could not use. We would avoid using "parallel", if we were describing a shape. Thanks for providing the link that has details about parallel computing

```suggestion
To enhance performance in high-scale deployments, you can configure the `metrics-collector` to collect metrics in parallel with multiple workers. The default configuration uses a single worker to federate metrics. By increasing the number of workers in the metric collection process, you enable internal workers to shard `/federate` endpoint requests that are made to Prometheus on your managed cluster.
```

jacobbaungard: This looks acceptable just one final thing, can we remove `from the same source` ? I think it doesn't make much sense to talk about the source in the context here.

dockerymick: Thanks! Updated the suggestion and going to commit

jacobbaungard (approved):
/lgtm

dockerymick:
Thanks everyone for helping with this PR!

PR #8259, opened 2025-09-26
https://issues.redhat.com/browse/ACM-23878

oafischer:
2.15

Checklist considered

### DIFF
--- a/clusters/cluster_lifecycle/import_gui.adoc
+++ b/clusters/cluster_lifecycle/import_gui.adoc
@@ -80,7 +80,7 @@ To configure a cluster API address, see xref:../cluster_lifecycle/import_gui.ado
 To configure your managed cluster klusterlet to run on specific nodes, see xref:../cluster_lifecycle/import_gui.adoc#import-configuring-nodeselector-tolerations[Optional: Configuring the klusterlet to run on specific nodes].
 
 [#import-configuring-cluster-api]
-== Optional: Configuring the cluster API address
+=== Optional: Configuring the cluster API address
 
 Complete the following steps to optionally configure the *Cluster API address* that is on the cluster details page by configuring the URL that is displayed in the table when you run the `oc get managedcluster` command:
 
@@ -108,7 +108,7 @@ spec:
 <1> Replace the value of the URL with the URL that provides external access to the managed cluster that you are importing.
 
 [#import-configuring-nodeselector-tolerations]
-== Optional: Configuring the klusterlet to run on specific nodes
+=== Optional: Configuring the klusterlet to run on specific nodes
 
 You can specify which nodes you want the managed cluster klusterlet to run on by configuring the `nodeSelector` and `tolerations` annotation for the managed cluster. Complete the following steps to configure these settings: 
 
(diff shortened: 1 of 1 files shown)


zhiweiyin318:
LGTM

clusters/cluster_lifecycle/import_gui.adoc
```diff
@@ -80,7 +80,7 @@ To configure a cluster API address, see xref:../cluster_lifecycle/import_gui.ado
 To configure your managed cluster klusterlet to run on specific nodes, see xref:../cluster_lifecycle/import_gui.adoc#import-configuring-nodeselector-tolerations[Optional: Configuring the klusterlet to run on specific nodes].
 
 [#import-configuring-cluster-api]
-=== Optional: Configuring the cluster API address
+== Optional: Configuring the cluster API address
```
swopebe: Is seems these were part of import, so I would not have them at h2 as a peer to import, if that is the case. In situations like this, I am making these steps with the procedure under them sub steps--so consider that.

I think one of these optionals is at the beginning as a TOC?

clusters/cluster_lifecycle/import_gui.adoc
```diff
@@ -174,6 +174,17 @@ oc -n open-cluster-management delete -l name=multiclusterhub-operator pods
...
 You can change the value of the hub cluster to not import automatically by changing the `disableHubSelfManagement` value to `true`. For more information, see the _disableHubSelfManagement_ topic.
 
+[#retain-namespace-managed-cluster]
+== Retaining the namespace after removing a managed cluster
+
+By default, the cluster namespace is deleted after you detach a managed cluster.
+
+To prevent {mce-short} from deleting the namespace after detaching a managed cluster, add the `open-cluster-management.io/retain-namespace` annotation to the managed cluster namespace.
+
+If you remove the `open-cluster-management.io/retain-namespace` annotation after detaching a managed cluster, {mce-short} deletes the cluster namespace.
```
swopebe, line 183: Should this line be before the action?

swopebe (changes requested):
See comments

clusters/cluster_lifecycle/import_gui.adoc
oafischer: If it's a little more work I'll defer this to my tech debt issue and a separate PR if you don't mind. I'll remove the header changes from the PR for now to keep it simple.

clusters/cluster_lifecycle/import_gui.adoc
oafischer: I think it makes more sense to have it after the action since the user would not have that annotation unless they complete the action first. Let me know what you think. Maybe I can also reword it, if that helps.

clusters/cluster_lifecycle/import_gui.adoc
swopebe: ok, that is fine. Having it h2 is technically not correct because it is part of the higher header, if we are looking at the original set up.

clusters/cluster_lifecycle/import_gui.adoc
swopebe: Ok, I see.

clusters/cluster_lifecycle/import_gui.adoc
```diff
@@ -174,6 +174,17 @@ oc -n open-cluster-management delete -l name=multiclusterhub-operator pods
 
 You can change the value of the hub cluster to not import automatically by changing the `disableHubSelfManagement` value to `true`. For more information, see the _disableHubSelfManagement_ topic.
 
+[#retain-namespace-managed-cluster]
```
swopebe: You don't need a header here, it is just part of Detach/Remove. If you were to break that out as a file, this would go along with it. I know we aren't going through that right now, but think of it as part of detaching....you can get rid of the header and just have a step:

By default the ---- 

. Add the -- annotation to prevent....

swopebe: Please add that to the tech debt if you don't have time, but really this is just makeing it a step.

swopebe (approved):

clusters/cluster_lifecycle/import_gui.adoc
oafischer: I'll remove the header, initially I wanted to add it to the previous topic but thought it would be too long then.

PR #8260, opened 2025-09-26
https://issues.redhat.com/browse/ACM-21429 ClusterInstance field updates

dockerymick:


apis/clusterinstance.json.adoc
```diff
@@ -419,10 +419,16 @@ __optional__|Specify the public Secure Shell (SSH) key to provide access to inst
 This key is added to the host to allow SSH access.|string
 |*nodes* +
 __required__|Specify the configuration parameters for each node.|<<_rhacm-docs_apis_clusterinstance_jsoncluster_nodes,nodes>> array
+|*reference* +
+_required_|Specify dependencies or related objects in different namespaces.|array
+|*namespace* +
+_required_|Specify the namespace of the reference object.|string
```
sakhoury: Can you share where you got these fields (viz. `reference` and `namespace`) from?

apis/clusterinstance.json.adoc
```diff
@@ -419,10 +419,16 @@ __optional__|Specify the public Secure Shell (SSH) key to provide access to inst
...
 |*templateRefs* +
 __required__|Specify the list of the references to cluster-level templates.
 A cluster-level template consists of a `ConfigMap` object, in which the keys of the data field represent the kind of the installation manifests.
 Cluster-level templates are instantiated once per cluster in the `ClusterInstance` CR.|array
+|*hostRef* +
+_optional_|Specify a refence to a `BareMetalHost` resource that is in a different resource.|
```
sakhoury: I feel like this description is missing key elements from the description in the ClusterInstance CRD:
```
// HostRef is a reference to a BareMetalHost node located in another namespace.
// It is used to link a resource to a specific BareMetalHost instance.
```

sakhoury (changes requested):
Thanks @dockerymick! Looks like we are missing the `cpuArchitecture` field defined here https://github.com/stolostron/siteconfig/commit/269de14ce5661d77a4af603152321c2fd43ba981.

apis/clusterinstance.json.adoc
dockerymick: @sakhoury thanks for your patience! I was on PTO and catching up. Here is the link: https://github.com/stolostron/siteconfig/commit/a804ef8a4b8265f1469861c30a1c57a72489d413@

apis/clusterinstance.json.adoc
dockerymick:
```suggestion
_optional_|Specify a reference to a `BareMetalHost` nodes that are in different namespaces. Use `hostRef` to link a resource to a specific `BareMetalHost` instance.|
```

apis/clusterinstance.json.adoc
sakhoury: Thanks @dockerymick! Those are internal source code field names. The user facing fields are described in the CRD. See https://github.com/stolostron/siteconfig/blob/main/config/crd/bases/siteconfig.open-cluster-management.io_clusterinstances.yaml#L321-L337.

sakhoury: So for example,
```yaml
apiVersion: siteconfig.open-cluster-management.io/v1alpha1
kind: ClusterInstance
metadata:
  name: "my-clusterinstance"
  namespace: "my-clusterinstance"
spec:
  ...
  nodes:
    - hostName: "somehost.redhat.com"
      hostRef:
         name: "some-nodename"
         namespace: "some-namespace"
      ...
```

sakhoury:
@dockerymick please also include the `cpuArchitecture` field. See https://github.com/stolostron/siteconfig/blob/main/config/crd/bases/siteconfig.open-cluster-management.io_clusterinstances.yaml#L127-L135 and https://github.com/stolostron/siteconfig/blob/main/config/crd/bases/siteconfig.open-cluster-management.io_clusterinstances.yaml#L294-L301

apis/clusterinstance.json.adoc
```diff
@@ -356,6 +356,8 @@ The following values are supported:
 
 `SNO`:: {sno}
 `HighlyAvailable`:: Multi-node OpenShift|string
+|*cpuArchitecture* +
```
fontivan, line 359: cpuArchitecture should be present on both ClusterInstanceSpec (https://github.com/stolostron/siteconfig/blob/release-2.15/api/v1alpha1/clusterinstance_types.go#L420) and NodeSpec (https://github.com/stolostron/siteconfig/blob/release-2.15/api/v1alpha1/clusterinstance_types.go#L192).

If it is not defined at the Node level, then it is inheirited from the ClusterInstance level. If it is not defined at the ClusterInstance level, then it defaults to x86_64

I would also expect the valid values to be present at the ClusterInstance level (https://github.com/stolostron/siteconfig/blob/release-2.15/api/v1alpha1/clusterinstance_types.go#L95), similar to the bootMode doc

dockerymick:
@sakhoury @fontivan thanks for yall's reviews. I made a couple of updates. Please re-review when you have an opportunity.

mce_acm_integration/siteconfig/siteconfig_install_clusters.adoc
```diff
@@ -206,9 +211,12 @@ spec:
 <3> Reference the `name` of one or more extra manifests `ConfigMap` objects.
 <4> Reference the `name` of your pull secret.
 <5> Ensure that the value of the `clusterName` field in the `ClusterInstance` custom resource matches the value of the `namespace` field.
-<6> Reference the `name` of the cluster-level templates in the `spec.templateRefs` field. If you are using a default installation template, the `namespace` must match the namespace where the Operator is installed.
-<7> Reference the `name` of the BMC secret.
-<8> Reference the `name` of the node-level templates in the `spec.nodes.templateRefs` field. If you are using a default installation template, the `namespace` must match the namespace where the Operator is installed.
+<6> Specify dependencies or related objects in different namespaces.
+<7> Specify the namespace of the reference object.
+<8> Reference the `name` of the cluster-level templates in the `spec.templateRefs` field. If you are using a default installation template, the `namespace` must match the namespace where the Operator is installed.
+<9> Specify a refence to a `BareMetalHost` resource that is in a different resource.
```
sakhoury:
```suggestion
<9> Specify a refence to a `BareMetalHost` resource that is in a different Namespace.
```

sakhoury: Looks like we are missing the `name` and `namespace` fields under the `hostRef` specification here. Look at how it's done for `templateRefs`.

apis/clusterinstance.json.adoc
sakhoury: @fontivan does `cpuArchitecture` need to be defined at both ClusterInstanceSpec and NodeSpec? If we specify it at the ClusterInstanceSpec and not in the NodeSpec, then the node inherits the ClusterInstanceSpec value right? Setting at the NodeSpec is an override for the specific node. So, it is not required to set both unless there is a mixed architecture deployment consisting of multiple nodes in which case it makes sense to set both. I feel like we are not explaining this clearly here.

sakhoury (changes requested):

apis/clusterinstance.json.adoc
fontivan: I still have a few issues with this

- cpuArchitecture is actually optional on both `ClusterInstanceSpec` and `NodeSpec`
- I don't see cpuArchitecture added to the `nodes` section
- I don't see the allowed values added to either clusterinstance (multi, x86, arm) or nodes (x86, arm) section
- I think the verbage around the default value could be better, specifically:
```
The architecture value from `ClusterInstanceSpec` is applied as a default value to nodes with undefined software architectures.
```
and dropping the mention of "level" in favor of:
```
If the `cpuArchitecture` parameter is not defined in `ClusterInstanceSpec`, the default value is `x86_64`
```

dockerymick:
@sakhoury @fontivan thanks again for providing your feedback. i made a couple more updates. Please re-review when you have a moment

apis/clusterinstance.json.adoc
```diff
@@ -475,6 +481,10 @@ extraLabels:
 |object
 |*hostName* +
 __required__|Define the host name.|string
+|*cpuArchitecture* +
+__optional__|Specify the type of software architecture to use for nodes that do not have an architecture defined. The following software architectures are supported: `multi`, `x86_64`, and `aarch64`. The architecture from the `ClusterInstanceSpec` is applied to nodes with undefined software architectures.
```
fontivan: I think this should say 
```Specify the type of software architecture to use for a specific node.```

If I remember right `multi` is only valid on the cluster, not on the node

Comment (dockerymick) [inline apis/clusterinstance.json.adoc]:
```suggestion
__optional__|Specify the type of software architecture to use for a specific node that does not have an architecture defined. The `x86_64` and `aarch64` software architectures are supported. The architecture from the `ClusterInstanceSpec` is applied to nodes with undefined software architectures.
```

dockerymick: @fontivan thanks Steven! Updated

apis/clusterinstance.json.adoc
```diff
@@ -356,6 +356,10 @@ The following values are supported:
 
 `SNO`:: {sno}
 `HighlyAvailable`:: Multi-node OpenShift|string
+|*cpuArchitecture* +
+__optional__|Specify the type of software architecture to use for nodes that do not have an architecture defined. The following software architectures are supported: `multi`, `x86_64`, and `aarch64`. The architecture from the `ClusterInstanceSpec` is applied to nodes with undefined software architectures.
```
sakhoury: Can we address the repetition here: `for nodes that do not have an architecture defined` and later `applied to nodes with undefined software architectures`.

apis/clusterinstance.json.adoc
```diff
@@ -423,6 +427,8 @@ __required__|Specify the configuration parameters for each node.|<<_rhacm-docs_a
 __required__|Specify the list of the references to cluster-level templates.
 A cluster-level template consists of a `ConfigMap` object, in which the keys of the data field represent the kind of the installation manifests.
 Cluster-level templates are instantiated once per cluster in the `ClusterInstance` CR.|array
+|*hostRef* +
+_optional_|Specify a reference to a `BareMetalHost` nodes that are in different namespaces. Use `hostRef` to link a resource to a specific `BareMetalHost` instance.|
```
sakhoury:
```suggestion
_optional_|Specify a reference to a `BareMetalHost` node that is in a different namespace. Use `hostRef` to link a resource to a specific `BareMetalHost` instance.|
```

apis/clusterinstance.json.adoc
```diff
@@ -475,6 +481,10 @@ extraLabels:
 |object
 |*hostName* +
 __required__|Define the host name.|string
+|*cpuArchitecture* +
+__optional__|Specify the type of software architecture to use for a specific node that does not have an architecture defined. The `x86_64` and `aarch64` software architectures are supported. The architecture from the `ClusterInstanceSpec` is applied to nodes with undefined software architectures.
```
sakhoury: This description here should be tailored for the node and does not need to mention the `ClusterInstanceSpec`.

apis/clusterinstance.json.adoc
dockerymick:
```suggestion
__optional__|Specify the type of software architecture to use for nodes that do not have an architecture defined. The following software architectures are supported: `multi`, `x86_64`, and `aarch64`.
```

apis/clusterinstance.json.adoc
dockerymick:
```suggestion
__optional__|Specify the type of software architecture to use for a specific node that does not have an architecture defined. The `x86_64` and `aarch64` software architectures are supported.
```

dockerymick:
@sakhoury thanks so much, updated the PR

apis/clusterinstance.json.adoc
```diff
@@ -475,6 +481,10 @@ extraLabels:
 |object
 |*hostName* +
 __required__|Define the host name.|string
+|*cpuArchitecture* +
+__optional__|Specify the type of software architecture to use for a specific node that does not have an architecture defined. The `x86_64` and `aarch64` software architectures are supported.
+
+If the `cpuArchitecture` parameter is not defined in `ClusterInstanceSpec`, the default value is `x86_64`.|string
```
fontivan: This reads a bit weird

The default value for nodes, if not defined, is the value in clusterinstance.

The default for clusterinstance, if not defined, is x86.

dockerymick:
```suggestion


If the `cpuArchitecture` parameter is not defined for your nodes, the value in the `ClusterInstance` specification. The default value for the `ClusterInstance` specification is `x86_64`.|string
```

apis/clusterinstance.json.adoc
```diff
@@ -475,6 +481,12 @@ extraLabels:
 |object
 |*hostName* +
 __required__|Define the host name.|string
+|*cpuArchitecture* +
+__optional__|Specify the type of software architecture to use for a specific node that does not have an architecture defined. The `x86_64` and `aarch64` software architectures are supported.
+
+
+
+If the `cpuArchitecture` parameter is not defined for your nodes, the value in the `ClusterInstance` specification. The default value for the `ClusterInstance` specification is `x86_64`.|string
```
fontivan: The sentence seems cut off here. I would imagine it should read "the value in the `ClusterInstance` specification **is used**"

dockerymick:
```suggestion
If the `cpuArchitecture` parameter is not defined for your nodes, the value in the `ClusterInstance` specification is used. The default value for the `ClusterInstance` specification is `x86_64`.|string
```

jc-berger (approved):
/lgtm

sakhoury (approved):
/lgtm

PR #8262, opened 2025-09-29
https://issues.redhat.com/browse/ACM-24676

jc-berger:
See JIra: https://issues.redhat.com/browse/ACM-24676 

See draft PR: https://github.com/ldpliu/rhacm-docs/pull/4/files

### DIFF
--- a/global_hub/global_hub_migrate.adoc
+++ b/global_hub/global_hub_migrate.adoc
@@ -165,7 +165,7 @@ spec:
 <2> The `Placement` resource that has the name `production-clusters` and that defines which clusters get migrated. You must define the `Placement` resource in the {global-hub} agent namespace of the source hub cluster. For example, you can name the `Placement` resource `multicluster-global-hub` or `multicluster-global-hub-agent`.
 <3> The `target` hub cluster is `hub2`.
 
-. Specify the `includedManagedClusters` or `includedManagedClustersPlacementRef` cluster to include in your migration. Include only one of these clusters in your migration because they cannot exist together. 
+. Specify the `includedManagedClusters` or `includedManagedClustersPlacementRef` cluster type to include in your migration. Include only one of these cluster types in your migration because they cannot exist together. 
 
 [#tracking-migration-status]
 == Tracking the migration status
--- a/global_hub/global_hub_migrate.adoc
+++ b/global_hub/global_hub_migrate.adoc
@@ -120,7 +120,7 @@ After you configure your {global-hub} migration environment, migrate your manage
 
 You can create a `ManagedClusterMigration` resource in the {global-hub} migration environment by directly selecting your managed clusters by name or by using a `Placement` resource. 
 
-To migrate your specific managed clusters by name, apply the following YAML file: 
+To migrate your specific managed clusters by name, go to your {global-hub} namespace and apply the following YAML file: 
 
 +
 [source,yaml]
@@ -158,14 +158,14 @@ spec:
 <2> The `Placement` resource that has the name `production-clusters` and that defines which clusters get migrated. You must define the `Placement` resource in the {global-hub} agent namespace of the source hub cluster. For example, you can name the `Placement` resource `multicluster-global-hub` or `multicluster-global-hub-agent`.
 <3> The `target` hub cluster is `hub2`.
 
-*Note:* After you migrate your managed clusters with the `Placement` resource, you must select the `includedManagedClusters` or `includedManagedClustersPlacementRef` cluster in your migration specification. Select only one of these clusters because they cannot exist together. 
+*Note:* Select the `includedManagedClusters` or `includedManagedClustersPlacementRef` cluster in your migration specification. Select only one of these clusters because they cannot exist together. 
 
 [#migrating-status-tracking]
 == Migrating status tracking 
 
 During the migration process, you can track the statuses of your managed clusters by using the {global-hub} `ConfigMap` or the migration customer resource. 
 
-{global-hub} automatically creates a `ConfigMap` to track the migration status of individual clusters. This `ConfigMap` provides detailed information about which clusters have successfully migrated or failed. 
+{global-hub} automatically creates a `ConfigMap` which gives you a detailed cluster list, telling you if the cluster migration fails or succeeds. The migration customer resources gives you the status of the entire migration process. 
 
 To view the migration status `ConfigMap`, run the following command:
 
--- a/global_hub/global_hub_migrate.adoc
+++ b/global_hub/global_hub_migrate.adoc
@@ -16,6 +16,7 @@ To fully migrate your managed clusters, complete the following sections:
 . <<migration-process, Managed cluster migration process>> 
 . <<preparing-migration-environment, Preparing the migration environment>> 
 . <<migrating-managed-clusters, Migrating managed clusters>> 
+. <<migrating-status-tracking, Migrating status tracking>> 
 
 [#prerequisites]
 == Prerequisites 
@@ -117,7 +118,9 @@ global-hub.open-cluster-management.io/deploy-mode=hosted
 
 After you configure your {global-hub} migration environment, migrate your managed clusters. Complete the following steps to migrate the `cluster1` sample and the associated resources from `hub1` to `hub2` hub clusters:
 
-. Create a `ManagedClusterMigration` resource in the {global-hub} migration environment. Apply the following YAML file: 
+You can create a `ManagedClusterMigration` resource in the {global-hub} migration environment by directly selecting your managed clusters by name or by using a `Placement` resource. 
+
+To migrate your specific managed clusters by name, apply the following YAML file: 
 
 +
 [source,yaml]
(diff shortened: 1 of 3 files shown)


global_hub/global_hub_migrate.adoc
```diff
@@ -117,7 +118,9 @@ global-hub.open-cluster-management.io/deploy-mode=hosted
 
 After you configure your {global-hub} migration environment, migrate your managed clusters. Complete the following steps to migrate the `cluster1` sample and the associated resources from `hub1` to `hub2` hub clusters:
 
-. Create a `ManagedClusterMigration` resource in the {global-hub} migration environment. Apply the following YAML file: 
+You can create a `ManagedClusterMigration` resource in the {global-hub} migration environment by directly selecting your managed clusters by name or by using a `Placement` resource. 
+
+To migrate your specific managed clusters by name, apply the following YAML file:
```
jc-berger: Is there a benefit to migrating managed clusters by name? I notice we give a benefit to migrating them by using the `Placement` resource, so I wondered if there's a benefit we can state for migrating managed clusters directly by their name/

global_hub/global_hub_migrate.adoc
```diff
@@ -130,24 +133,69 @@ spec:
...
+  from: local-cluster <1>
+  includedManagedClustersPlacementRef: production-clusters <2>
+  to: hub2 <3>
 ----
-. In the `status` resource section, verify that your migration status is `True` and resembles the following status: 
+<1> The `source` hub cluster is a `local-cluster` that {global-hub} installed in `hub1` hub cluster.
+<2> The `Placement` resource that has the name `production-clusters` and that defines which clusters get migrated. You must define the `Placement` resource in the {global-hub} agent namespace of the source hub cluster. For example, you can name the `Placement` resource `multicluster-global-hub` or `multicluster-global-hub-agent`.
+<3> The `target` hub cluster is `hub2`.
+
+*Note:* After you migrate your managed clusters with the `Placement` resource, you must select the `includedManagedClusters` or `includedManagedClustersPlacementRef` cluster in your migration specification. Select only one of these clusters because they cannot exist together.
```
jc-berger: Am I correct that the `includeManagedClusters` and `includeManagedClustersPlacementRef` are both types of clusters that the users need to select in the migration specification? 

And am I correct that users need to only select one due to the two not being able to exist simultaneously together?

global_hub/global_hub_migrate.adoc
```diff
@@ -130,24 +133,69 @@ spec:
...
+<1> The `source` hub cluster is a `local-cluster` that {global-hub} installed in `hub1` hub cluster.
+<2> The `Placement` resource that has the name `production-clusters` and that defines which clusters get migrated. You must define the `Placement` resource in the {global-hub} agent namespace of the source hub cluster. For example, you can name the `Placement` resource `multicluster-global-hub` or `multicluster-global-hub-agent`.
+<3> The `target` hub cluster is `hub2`.
+
+*Note:* After you migrate your managed clusters with the `Placement` resource, you must select the `includedManagedClusters` or `includedManagedClustersPlacementRef` cluster in your migration specification. Select only one of these clusters because they cannot exist together. 
+
+[#migrating-status-tracking]
+== Migrating status tracking 
+
+During the migration process, you can track the statuses of your managed clusters by using the {global-hub} `ConfigMap` or the migration customer resource.
```
jc-berger, line 173: What do you think of structuring this section similar to the "Managing managed clusters" section? 

Basically, giving the users two options and explaining how to use these two options.

global_hub/global_hub_migrate.adoc
jc-berger: Where are users applying this YAML file? Within the managed cluster?

global_hub/global_hub_migrate.adoc
```diff
@@ -130,24 +133,69 @@ spec:
...
+metadata:
+  name: migration-sample
+  namespace: multicluster-global-hub
+data:
+  success: '["cluster1","cluster3"]'
+  failure: '["cluster2"]'
+----
+
+
+To view the tracking status from the migration customer resource, verify that its migration status is `True` and that it resembles the following sample:
```
jc-berger: We say the benefit of using the ConfigMap to track the migration status. Is there a benefit for the user if they use the customer resource to track the status?

global_hub/global_hub_migrate.adoc
ldpliu: I think it's no need to give benefit here.

ldpliu: Within Globalhub cluster, and Globalhub namespace

global_hub/global_hub_migrate.adoc
ldpliu: Yes, use must specify `includeManagedClusters` or `includeManagedClustersPlacementRef` in migration specification. and  they can not exist simultaneously together. 
Should also remove `After you migrate your managed clusters with the `Placement` resource` ,  user must set the `includedManagedClusters` or `includedManagedClustersPlacementRef` when define the migration specification.

global_hub/global_hub_migrate.adoc
ldpliu: Both configmap and migration status show the migration process.
For configmap it show the detail cluster list and if the specific cluster migration success or fail.
For migration status, it show the migration whole process status. 
User could see both of them to trace the migration.

ldpliu (approved):

yahliu:
Looks good to me.

global_hub/global_hub_migrate.adoc
```diff
@@ -130,24 +132,65 @@ spec:
...
+
+To view the migration status `ConfigMap`, run the following command:
 
-+
 [source,bash]
 ----
+kubectl get configmap <migration-name> -n <global-hub-namespace> -o yaml
+----
+
+Ensure that the  `ConfigMap` that you see resembles the following sample:
```
swopebe:
```suggestion
Ensure that the `ConfigMap` that you see resembles the following sample: 
```

global_hub/global_hub_migrate.adoc
```diff
@@ -130,24 +132,65 @@ spec:
...
 ----
-. In the `status` resource section, verify that your migration status is `True` and resembles the following status: 
+<1> The `source` hub cluster is a `local-cluster` that {global-hub} installed in `hub1` hub cluster.
+<2> The `Placement` resource that has the name `production-clusters` and that defines which clusters get migrated. You must define the `Placement` resource in the {global-hub} agent namespace of the source hub cluster. For example, you can name the `Placement` resource `multicluster-global-hub` or `multicluster-global-hub-agent`.
+<3> The `target` hub cluster is `hub2`.
+
+*Note:* Select the `includedManagedClusters` or `includedManagedClustersPlacementRef` cluster in your migration specification. Select only one of these clusters because they cannot exist together. 
+
+[#migrating-status-tracking]
+== Migrating status tracking
```
swopebe: Since this is a procedure, we should start with the verb that applies. Here it seems to be "Tracking" migration status?

global_hub/global_hub_migrate.adoc
```diff
@@ -130,24 +132,65 @@ spec:
...
+  from: local-cluster <1>
+  includedManagedClustersPlacementRef: production-clusters <2>
+  to: hub2 <3>
 ----
-. In the `status` resource section, verify that your migration status is `True` and resembles the following status: 
+<1> The `source` hub cluster is a `local-cluster` that {global-hub} installed in `hub1` hub cluster.
+<2> The `Placement` resource that has the name `production-clusters` and that defines which clusters get migrated. You must define the `Placement` resource in the {global-hub} agent namespace of the source hub cluster. For example, you can name the `Placement` resource `multicluster-global-hub` or `multicluster-global-hub-agent`.
+<3> The `target` hub cluster is `hub2`.
+
+*Note:* Select the `includedManagedClusters` or `includedManagedClustersPlacementRef` cluster in your migration specification. Select only one of these clusters because they cannot exist together.
```
swopebe: Select the `includedManagedClusters` or `includedManagedClustersPlacementRef` _clusters_?

I would like to understand what these values are, as it seems here you are naming these "clusters" but they seem more like statuses or values for the configuration? If these are not the cluster names (which would seem odd), then we need to refer to them another way.

Select the `includedManagedClusters` _cluster status/cluster placement_ or the `includedManagedClustersPlacementRef` _cluster status_  (????)

Something like that.^^^^

global_hub/global_hub_migrate.adoc
```diff
@@ -117,9 +118,10 @@ global-hub.open-cluster-management.io/deploy-mode=hosted
 
 After you configure your {global-hub} migration environment, migrate your managed clusters. Complete the following steps to migrate the `cluster1` sample and the associated resources from `hub1` to `hub2` hub clusters:
 
-. Create a `ManagedClusterMigration` resource in the {global-hub} migration environment. Apply the following YAML file: 
+You can create a `ManagedClusterMigration` resource in the {global-hub} migration environment by directly selecting your managed clusters by name or by using a `Placement` resource.
```
swopebe, line 121: These seem like they need to be written as steps, considering this is a procedure for migrating?

swopebe: . Create a `ManagedClusterMigration`...
. Migrate your.....
. Optional: If you use the Placement resource selection, you can select.
. Migrate managed clusters.

This is just a brief outline not to be taken literally, but to show that for some reason here in this section, this is not written as a procedure.

global_hub/global_hub_migrate.adoc
swopebe: Line 159 is also a step.

swopebe (changes requested):
This is a procedure, so the steps should be written as such, but the other things I mention are small.

global_hub/global_hub_migrate.adoc
```diff
@@ -130,24 +136,73 @@ spec:
...
+  from: local-cluster <1>
+  includedManagedClustersPlacementRef: production-clusters <2>
+  to: hub2 <3>
 ----
-. In the `status` resource section, verify that your migration status is `True` and resembles the following status: 
+<1> The `source` hub cluster is a `local-cluster` that {global-hub} installed in `hub1` hub cluster.
+<2> The `Placement` resource that has the name `production-clusters` and that defines which clusters get migrated. You must define the `Placement` resource in the {global-hub} agent namespace of the source hub cluster. For example, you can name the `Placement` resource `multicluster-global-hub` or `multicluster-global-hub-agent`.
+<3> The `target` hub cluster is `hub2`.
+
+. Specify the `includedManagedClusters` or `includedManagedClustersPlacementRef` cluster in your migration. Specify only one of these clusters because they cannot exist together.
```
jc-berger: @swopebe thanks, so the original content is to specify the cluster, however I didn't want to be redundant and say "specify the cluster in the migration cluster specification" 

however, after seeing your comment, I see how "select" is a little confusing, like are we selecting a value/property, etc. Let me know if you think this is clearer, or I can keep reworking, thanks.

swopebe: @jc-berger the issue here is not the verb, it's what the _thing_ is.

the `includedManagedClusters` or `includedManagedClustersPlacementRef` **clusters**? That implies these are cluster names. So what are these? That was my question.

swopebe: includedManagedClusters or includedManagedClustersPlacementRef  

these don't look like cluster names, are these actually "clusters" as they are identified in line 168?

It seems to be a qualifier for either including or not including a cluster in migration, so the line would need to reflect that if that is the case.

swopebe (changes requested):

global_hub/global_hub_migrate.adoc
```diff
@@ -130,24 +136,73 @@ spec:
...
+  from: local-cluster <1>
+  includedManagedClustersPlacementRef: production-clusters <2>
+  to: hub2 <3>
 ----
-. In the `status` resource section, verify that your migration status is `True` and resembles the following status: 
+<1> The `source` hub cluster is a `local-cluster` that {global-hub} installed in `hub1` hub cluster.
+<2> The `Placement` resource that has the name `production-clusters` and that defines which clusters get migrated. You must define the `Placement` resource in the {global-hub} agent namespace of the source hub cluster. For example, you can name the `Placement` resource `multicluster-global-hub` or `multicluster-global-hub-agent`.
+<3> The `target` hub cluster is `hub2`.
+
+. Specify the `includedManagedClusters` or `includedManagedClustersPlacementRef` cluster to include in your migration. Include only one of these clusters in your migration because they cannot exist together.
```
jc-berger: @swopebe please let me know what you think :)

swopebe (approved):

PR #8278, opened 2025-10-05
Adding step to get the RHEM URL (do not merge, need more info in procedure first)

alokstech:
https://issues.redhat.com/browse/ACM-24936 Adding step to get the RHEM URL

### DIFF
--- a/edge_manager/edge_mgr_build_images.adoc
+++ b/edge_manager/edge_mgr_build_images.adoc
@@ -52,12 +52,22 @@ sudo dnf install flightctl
 
 If you want to include an agent configuration in the image, complete the following steps:
 
+. Run the following command to get the {rhem} API server URL:
+
++
+[source,bash]
+----
+export RHEM_API_SERVER_URL=$(oc get route -n open-cluster-management flightctl-api-route -o json | jq -r .spec.host)
+----
+
++
+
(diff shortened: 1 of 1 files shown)


swopebe:
@alokstech RHEM is not resourced from our team right now, but we will help maintain it--thanks for opening proper issues for us. 

That said, we need to keep the RHEM doc up to standard with the rest of our doc, so we don't (should not) have multiple steps in commands. 

We will ask for these to be broken out. (I realize some of the doc had that but that should not have passed review from our team and we plan to clean that up as we go.)

We will need the steps to be written out as such (just an example):

**. Run the following command to export** ____ (why are we telling the user to do this, that is what we would place here):

`command`

**. Log in with your credentials to** (why are we telling the user to do this, that is what we would place here). **Run the following command:**

`command`

Example of how we write a procedure: https://github.com/stolostron/rhacm-docs/blob/2.15_stage/mce_acm_integration/discover_hosted/acm_discover_hosted.adoc

swopebe:
It seems there is info in the issue to help with this.

alokstech:
@swopebe I have updated the commit. Please take a look and suggest for any further modifications if any required.

jc-berger:
Closing as Dev and Doc used the following PR to collaborate and merge the necessary changes: https://github.com/stolostron/rhacm-docs/pull/8288/files

PR #8317, opened 2025-10-22
Swope getting started 1

swopebe:


clusters/about/mce_hcp_overview.adoc
```diff
@@ -0,0 +1,49 @@
+[#mce-hcp-overview]
```
swopebe, line 1: This is the new file for review

clusters/about/mce_hcp_overview.adoc
```diff
@@ -0,0 +1,49 @@
...
+= Learn about hosted control planes with {mce-short}
+
+With hosted control planes, you create control planes as pods on a management cluster without the need for dedicated virtual or physical machines for each control plane. {ocp-short} clusters can use standalone clusters or hosted control plane clusters.
```
rokej: I think we can write something like this and replace lines 4 ~ 8.

Hosted control planes represent a deployment model for OpenShift clusters that provides significant infrastructure efficiency. Instead of requiring dedicated virtual or physical machines for each cluster's control plane, hosted control planes run as pods in a single namespace on a centralized management cluster.

### Standalone Clusters

Traditional OpenShift clusters that use dedicated virtual or physical machines for the control plane components. Each cluster requires its own infrastructure for etcd, API server, controller managers, and scheduler.

### Hosted Control Plane Clusters

A more efficient model where control planes run as pods on a shared management cluster. This approach:
- Reduces infrastructure overhead and costs
- Enables faster cluster provisioning
- Simplifies control plane management and updates
- Allows for better resource utilization

clusters/about/mce_hcp_overview.adoc
```diff
@@ -0,0 +1,54 @@
...
+
+Standalone:: {ocp-short} clusters that use dedicated virtual or physical machines for the control plane components. Each cluster requires individual infrastructure for etcd, API servers, controller managers, and schedulers.
+
+Hosted:: A more efficient model where control planes run as pods on a shared management cluster. The hosted model offers the following advantages:
+
+- Reduces infrastructure overhead and costs
+- Enables faster cluster provisioning
+- Simplifies control plane management and updates
+- Allows for better resource utilization
+
```
swopebe, line 14: engineer review comments added in 4-13

rokej (approved):
Looks good to me. Thank you so much.

oafischer (approved):
lgtm

PR #8320, opened 2025-10-23
https://issues.redhat.com/browse/ACM-24014 Add manage hub doc

oafischer:
2.12 to 2.15

Checklist considered


clusters/assisted_installer/ai_manage_hub.adoc
```diff
@@ -0,0 +1,36 @@
+[#manage-hub-cim]
+= Managing a hub cluster by using {cim}
+
+Importing and managing a hub cluster by using {cim} can simplify your cluster management by providing a single point of control. You can also configure additional options, such as the hub cluster Kube API server verification strategy.
```
CrystalChun: > hub cluster Kube API server verification strategy.

Sorry, what is this? 😅 Is this something the user can do? I'm not familiar with it so just curious!

CrystalChun:
Small question, but looks good otherwise!

clusters/assisted_installer/ai_manage_hub.adoc
oafischer: We needed some info on why the user would manage a hub cluster by using CIM. I pieced together some stuff from our doc but clearly I got it wrong. :D

If you have a suggestion of what we want to highlight here (i.e. why the user would want to enable this feature), I would appreciate it. Thanks!

CrystalChun: Ah gotcha! I just personally do not know what Kube API server verification is, so if you have a link to that and know that it can only be used when you add the hub cluster as managed, then I think it's fine to keep it!

As far as why a user would want this feature, from my experience, most customers want it so that they can scale their hub cluster by adding more hosts to it. Essentially what you wrote below

> add and manage additional hosts on the hub cluster by using {cim}

oafischer: Ah I see, this is our topic on the verification strategy: https://github.com/stolostron/rhacm-docs/blob/2.15_stage/clusters/cluster_lifecycle/adv_config_cluster.adoc#configuring-the-hub-cluster-kubeapiserver-verification-strategy

Let me know if it makes sense. Otherwise I'll just remove the intro we have right now. Thanks!

CrystalChun: Thank you for the link! 

Apologies, but I don't see how this feature is used when a hub cluster is imported into itself 😅 FWIU the kube apiserver verification feature details how managed cluster to hub communication can be solved, but it doesn't seem like this feature would be used or needed if it's the same cluster (hub imported as managed cluster)? 

Let me know if I'm totally misunderstanding!

oafischer: You are right, I'll just go ahead and remove it, thanks for checking!

CrystalChun (approved):
/lgtm

Thanks Oliver!

vkolodny (approved):
LGTM

clusters/assisted_installer/ai_manage_hub.adoc
```diff
@@ -0,0 +1,36 @@
+[#manage-hub-cim]
+= Managing a hub cluster by using {cim}
+
+Importing and managing a hub cluster by using {cim} can simplify your cluster management by providing a single point of control.
+
+By default, {cim} does not automatically manage your hub cluster. Enable the feature to add and manage additional hosts on the hub cluster by using {cim}.
```
dockerymick:
```suggestion
By default, {cim} does not automatically manage your hub cluster. Enable the feature to add and manage additional hosts on your hub cluster by using {cim}.
```

clusters/assisted_installer/ai_manage_hub.adoc
```diff
@@ -0,0 +1,36 @@
...
+----
+oc annotate agentserviceconfig agent agent-install.openshift.io/enable-local-cluster-import=true 
+----
++
+Verify that the feature is enabled by checking if an `InfraEnv` custom resource is created for the `local-cluster`. Run the following command:
++
+[source,bash]
+----
+oc get infraenv -n multicluster-engine 
+----
```
dockerymick, line 31: I recommend that you make "Verify that the feature is enabled..." as a step

```suggestion
. Annotate the `AgentServiceConfig` custom resource to enable `local-cluster` import. Run the following command:

+
[source,bash]
----
oc annotate agentserviceconfig agent agent-install.openshift.io/enable-local-cluster-import=true 
----

. Verify that the feature is enabled by checking if an `InfraEnv` custom resource is created for the `local-cluster`. Run the following command:

+
[source,bash]
----
oc get infraenv -n multicluster-engine 
----
```

dockerymick (approved):
Great work ! This looks good to me. I left a suggestion

PR #8321, opened 2025-10-23
https://issues.redhat.com/browse/ACM-25011

mletalie:
Have considered the checklist.
Issue: https://issues.redhat.com/browse/ACM-25011

### DIFF
--- a/edge_manager/edge_mgr_build_images_platform.adoc
+++ b/edge_manager/edge_mgr_build_images_platform.adoc
@@ -15,6 +15,8 @@ When building operating system images and disk images for {ocp-virt}, you can fo
 * Add the `open-vm-tools` guest tools to the image.
 * Build a disk image of type `qcow2` instead of `iso`.
 
+*Note:* You must build the operating system image on a Red Hat Enterprise Linux host that has the required entitlement for the specified `rhacm` repository.
+
 Complete the generic steps with changes to the following steps:
 
 . Build an operating system image that is based on RHEL 9 that includes the {rhem} agent and virtual machine guest tools, but excludes the agent configuration.
@@ -26,7 +28,7 @@ Complete the generic steps with changes to the following steps:
 ----
 FROM registry.redhat.io/rhel9/bootc-image-builder:latest
 
-RUN subscription-manager repos --enable rhacm-<2.x>-for-rhel-<version>-$(uname -m)-rpms && \
+RUN dnf config-manager --set-enabled rhacm-<2.x>-for-rhel-<version>-$(uname -m)-rpms && \
     dnf -y install flightctl-agent && \
     dnf -y clean all && \
     systemctl enable flightctl-agent.service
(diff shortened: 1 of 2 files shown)


mletalie:
/cherrypick 2.14_stage

(Ran this with Oliver to see if it would be recognized).

mletalie:
@fzdarsky, Could you please review this PR when you get a moment? Want to make sure the note I added (based on your comment in the JIRA ticket tied to this PR) as well as the command change is correct. Thanks!

fzdarsky (approved):

edge_manager/edge_mgr_build_images_platform.adoc
```diff
@@ -26,7 +28,7 @@ Complete the generic steps with changes to the following steps:
 ----
 FROM registry.redhat.io/rhel9/bootc-image-builder:latest
 
-RUN subscription-manager repos --enable rhacm-<2.x>-for-rhel-<version>-$(uname -m)-rpms && \
+RUN dnf config-manager --set-enabled rhacm-2.14-for-rhel-9-x86_64-rpms && \
```
siserafin: maybe here we could keep the variables rhacm-<2.x> and rhel-<version>-$(uname -m)

edge_manager/edge_mgr_intro.adoc
```diff
@@ -8,7 +8,7 @@ By defining the required state of your edge devices, which includes your operati
 By using the {rhem} on {acm}, you can manage non-Kubernetes workloads and your operating system configuration on a Red Hat Enterprise Linux machine consistently with how you manage it on {ocp}.
 
 See the following topics to learn more about using {rhem} on {acm-short}. All features are in *Technology Preview* status:
-
+//ML comment
```
oafischer: Don't forget to remove this

governance/template_functions.adoc
```diff
@@ -167,10 +167,10 @@ spec:
   object-templates:
   - objectDefinition:
       ...
-      data: <1>
+      data: # <1>
```
oafischer: We don't usually have comments in code snippets, looks like this might have been added by accident? (since there aren't any comments)

edge_manager/edge_mgr_intro.adoc
mletalie: Good catch, thanks!

governance/template_functions.adoc
mletalie: Actually, this was added on purpose b/c I was getting a lint break on call out #2 (although I didnt add that callout). I added it to call out #1 for conformity within the code, even though it wasn't necessary ( I could of added just to call out # 2). However, it had to be added to #2 b/c  YAML parsers expect a newline immediately after a block scalar indicator (like >), or just whitespace and a comment.  Adding a callout after `>` seems to produce a lint break. We can discuss further if needed. Thanks.

oafischer: I see, thanks for explaining. In that case, I'd maybe even consider not using callouts since we have to move away from them anyway for conversion. Let's bring this up to the team and see what the others say. Maybe @swopebe has some guidance here?

oafischer (approved):
lgtm

PR #8323, opened 2025-10-27
Exclude stolostron/backlog links from checks

dhaiducek:
These links are inaccessible to the public.


dockerymick (approved):
Thanks so much Dale! Also adding link to doc issue: https://issues.redhat.com/browse/ACM-25592

PR #8328, opened 2025-10-28
[ACM-25673] Document Gatekeeper 3.20 features

dhaiducek:
Also removes version stipulations now that 3.17 is the oldest supported version.

ref: https://issues.redhat.com/browse/ACM-25673

### DIFF
--- a/governance/gatekeeper_operator/config_gk_operator.adoc
+++ b/governance/gatekeeper_operator/config_gk_operator.adoc
@@ -7,7 +7,7 @@ Install the {gate} from the {olm} catalog to install Gatekeeper on your cluster.
 == Prerequisites
 
 - *Required access*: Cluster administrator. 
-- Understand how to use the Operator Lifecycle Manager (OLM) and the OperatorHub by completing the _Adding Operators to a cluster_ and the _Additional resources_ section in the link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.17/html/operators/administrator-tasks#olm-adding-operators-to-a-cluster[{ocp-short} documentation].
+- Understand how to use the Operator Lifecycle Manager and the OperatorHub by completing the _Adding Operators to a cluster_ and the _Additional resources_ section in the link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.18/html/operators/administrator-tasks#olm-adding-operators-to-a-cluster[{ocp-short} documentation].
 
 [#gatekeeper-custom-resource]
 == Gatekeeper custom resource sample
@@ -22,14 +22,11 @@ metadata:
   name: gatekeeper
 spec:
   audit:
-    replicas: 1
+    auditChunkSize: 66
     auditEventsInvolvedNamespace: Enabled <1>
-    logLevel: DEBUG
+    auditFromCache: Enabled
     auditInterval: 10s
     constraintViolationLimit: 55
-    auditFromCache: Enabled
-    auditChunkSize: 66
-    emitAuditEvents: Enabled
     containerArguments: <4>
     - name: ""
       value: ""
@@ -40,22 +37,26 @@ spec:
       requests:
         cpu: 500m
         memory: 130Mi
+    emitAuditEvents: Enabled
+    logLevel: DEBUG
+    podAnnotations: {} <2>
+    replicas: 1
   validatingWebhook: Enabled
   mutatingWebhook: Enabled
   webhook:
-    replicas: 3
-    emitAdmissionEvents: Enabled
-    admissionEventsInvolvedNamespace: Enabled <2>
-    disabledBuiltins:
-     - http.send
-    operations: <3>
-     - "CREATE"
-     - "UPDATE"
-     - "CONNECT"
-    failurePolicy: Fail
+    admissionEventsInvolvedNamespace: Enabled <3>
     containerArguments: <4>
     - name: ""
       value: ""
+    disabledBuiltins:
+     - http.send
+    emitAdmissionEvents: Enabled
+    failurePolicy: Fail
+    operations: <5>
+     - CREATE
+     - UPDATE
+     - CONNECT
+    replicas: 3
     resources:
       limits:
         cpu: 480m
@@ -63,6 +64,25 @@ spec:
       requests:
         cpu: 400m
         memory: 120Mi
+    rules: <6>
+    - operations: []
+      resources: []
+    timeoutSeconds: 3 <7>
+  mutatingWebhookConfig: <8>
+    failurePolicy: ""
+    logMutations: Disabled
+    mutationAnnotations: Disabled
+    namespaceSelector: {}
+    operations: []
+    rules: <6>
+    - operations: []
+      resources: []
+    timeoutSeconds: 1 <7>
+  config: <9>
+    matches: 
+      - excludedNamespaces: ["test-*", "my-namespace"] 
+        processes: ["*"] 
+    disableDefaultMatches: false <10>
   nodeSelector:
     region: "EMEA"
   affinity:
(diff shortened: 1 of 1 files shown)


governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -7,7 +7,7 @@ Install the {gate} from the {olm} catalog to install Gatekeeper on your cluster.
 == Prerequisites
 
 - *Required access*: Cluster administrator. 
-- Understand how to use the Operator Lifecycle Manager (OLM) and the OperatorHub by completing the _Adding Operators to a cluster_ and the _Additional resources_ section in the link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.17/html/operators/administrator-tasks#olm-adding-operators-to-a-cluster[{ocp-short} documentation].
+- Understand how to use the Operator Lifecycle Manager (OLM) and the OperatorHub by completing the _Adding Operators to a cluster_ and the _Additional resources_ section in the link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.20/html/operators/administrator-tasks#olm-adding-operators-to-a-cluster[{ocp-short} documentation].
```
JustinKuli: https://access.redhat.com/documentation/en-us/openshift_container_platform/latest/html/operators/administrator-tasks#olm-adding-operators-to-a-cluster (using "latest") also seems like a valid link. But actually, it might be better to keep it static in case they re-organize their docs. Just FYI.

governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -97,13 +112,17 @@ spec:
...
 - `tls-min-version`
-<5> Use the `config` section to exclude namespaces from certain processes for all constraints on your hub cluster.
-<6> The `disableDefaultMatches` parameter is a boolean parameter that disables appending the default exempt namespaces provided by the Gatekeeper operator. The default exempt namespaces are {ocp-short} or Kubernetes system namespaces. By default, this parameter is set to `false` to allow the default namespaces to be appended.
+<5> Use `operations` to manage the operations for which the webhook is invoked. For example, `CREATE`, `UPDATE`, `CONNECT`, and `DELETE`.
+<6> For version 3.20 and later, specify `rules` to overwrite the rules configuration in the webhook configuration. When set, this field takes precedence over the Operations field.
+<7> For version 3.20 and later, specify the timeout for the webhook configuration.
+<8> For version 3.20 and later, use the `mutatingWebhookConfig` section to configure the mutating webhook with configurations that override the values from `spec.webhook` or are specific to the mutating webhook.
+<9> Use the `config` section to exclude namespaces from certain processes for all constraints on your hub cluster.
```
JustinKuli: I know you only re-numbered this, but "on your hub cluster" doesn't seem right? Should it be "managed cluster"?

governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -40,29 +37,52 @@ spec:
...
         cpu: 480m
         memory: 140Mi
       requests:
         cpu: 400m
         memory: 120Mi
+    rules: <6>
+    - operations: []
+      resources: []
+    timeoutSeconds: 3 <7>
+mutatingWebhookConfig: <8>
```
JustinKuli: Indent 2 spaces
```suggestion
  mutatingWebhookConfig: <8>
```

JustinKuli (commented):
New content looks good overall to me!

governance/gatekeeper_operator/config_gk_operator.adoc
dhaiducek: I like using the `latest` tag, but I think the preference in doc is to keep it versioned. I'll leave this conversation open to get doc feedback on it.

governance/gatekeeper_operator/config_gk_operator.adoc
dhaiducek: Good catch. There's a pattern here to just use "cluster", so I'll go with that.

jan-law (commented):
The edits LGTM!

governance/gatekeeper_operator/config_gk_operator.adoc
swopebe: FYI, for most of the doc, we link to the bottom of n-2, I have an issue to change all of these to 4.18. We do this (pm suggestion) to keep from suggesting that everything is aligned with the latest OCP. We would have to check each function to send everything to 4.20. Is that ok to do here? If not, can we chat about that here?

dhaiducek: Makes sense--that works for me! I'll update it.

governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -97,13 +112,17 @@ spec:
 - `disable-cert-rotation`
 - `client-cert-name`
 - `tls-min-version`
-<5> Use the `config` section to exclude namespaces from certain processes for all constraints on your hub cluster.
-<6> The `disableDefaultMatches` parameter is a boolean parameter that disables appending the default exempt namespaces provided by the Gatekeeper operator. The default exempt namespaces are {ocp-short} or Kubernetes system namespaces. By default, this parameter is set to `false` to allow the default namespaces to be appended.
+<5> Use `operations` to manage the operations for which the webhook is invoked. For example, `CREATE`, `UPDATE`, `CONNECT`, and `DELETE`.
+<6> For version 3.20 and later, specify `rules` to overwrite the rules configuration in the webhook configuration. When set, this field takes precedence over the Operations field.
+<7> For version 3.20 and later, specify the timeout for the webhook configuration.
```
swopebe, line 117: If we can, we need to stop putting "for versions __ and later" -- This is understood because the doc moves forward as of 3.20 and later. When we add this type of version info to the doc, we have to remember to go back and change it. It's rare that this is needed in the main documentation.

swopebe (changes requested):
see comments

governance/gatekeeper_operator/config_gk_operator.adoc
dhaiducek: Since Gatekeeper versions and its lifecycle don't explicitly align with ACM, the plan was to backport this content to all supported ACM versions for visibility, hence the version stipulations here.

governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -79,16 +99,11 @@ spec:
...
-<3> For version 3.14 and later, to manage your webhook operations, use the following values for the `operations` parameter, `"CREATE"`, `"UPDATE"`, `"CONNECT"`, and `"DELETE"`.
-<4> For version 3.17 and later, specify `containerArguments` by providing a list of argument names and values to pass to the container. Omit leading dashes from the argument name. An omitted value is treated as `true`. Arguments that you provide are ignored if the argument is set previously by the operator or configurations from other fields. See the following list of flags that are deny-listed and are not currently supported:
+<1> Enable the `auditEventsInvolvedNamespace` parameter to manage the namespace audit event you want to create. When you enable this parameter, the Gatekeeper controller deployment runs with the following argument: `--audit-events-involved-namespace=true`.
+<2> For version 3.20 and later, specify `audit.podAnnotations` specific to the audit pod by providing a list of annotation names and values to add to the pod. An omitted value is treated as `true`.
```
swopebe, line 104: As I mentioned above, do we need _For version 3.20 and later_?

governance/gatekeeper_operator/config_gk_operator.adoc
swopebe: Ok, we just need to remember to update it.

dhaiducek: >Ok, we just need to remember to update it.

Indeed. I'll issue the `/cherry-pick` commands now. 🙂 I think it should work since we overwrote it all previously.

governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -7,7 +7,7 @@ Install the {gate} from the {olm} catalog to install Gatekeeper on your cluster.
 == Prerequisites
 
 - *Required access*: Cluster administrator. 
-- Understand how to use the Operator Lifecycle Manager (OLM) and the OperatorHub by completing the _Adding Operators to a cluster_ and the _Additional resources_ section in the link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.17/html/operators/administrator-tasks#olm-adding-operators-to-a-cluster[{ocp-short} documentation].
+- Understand how to use the Operator Lifecycle Manager (OLM) and the OperatorHub by completing the _Adding Operators to a cluster_ and the _Additional resources_ section in the link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.18/html/operators/administrator-tasks#olm-adding-operators-to-a-cluster[{ocp-short} documentation].
```
swopebe:
```suggestion
- Understand how to use the Operator Lifecycle Manager and the OperatorHub by completing the _Adding Operators to a cluster_ and the _Additional resources_ section in the link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.18/html/operators/administrator-tasks#olm-adding-operators-to-a-cluster[{ocp-short} documentation].
```

swopebe (approved):

swopebe:
Approved. @dhaiducek @jc-berger we do want to add these to release notes or skip that bc you add it to GK release notes?

dhaiducek:
> do we want to add these to release notes or skip that bc you add it to GK release notes?

I think I like the idea of adding something to the release notes for 2.15. Since this PR is getting cherry-picked, it should be in a new PR.

Maybe something like this?
>Gatekeeper v3.20 is released with new operator configurations including separate configuration options for mutating webhooks.

swopebe:
We are sometimes asked about agnostic operator function, so I think that makes sense. Then maybe link to the advisory? @jc-berger, can you help with that?

PR #8330, opened 2025-10-30
https://issues.redhat.com/browse/ACM-25672

jc-berger:
See Jira: https://issues.redhat.com/browse/ACM-25672

### DIFF
--- a/governance/template_functions.adoc
+++ b/governance/template_functions.adoc
@@ -140,6 +140,29 @@ spec:
 <2> Configuration values can be set as key-value properties.
 <3> The value for the `platform` data key is a template that retrieves the value of the `platform.open-cluster-management.io` cluster claim. Similarly, it retrieves values for `product` and `version` from the `ClusterClaim` resource.
 
+[#lookupclusterclaim-func]
+=== _lookupClusterClaim_
+
+The `lookupClusterClaim` function is similar to the `fromClusterClaim` function. The only difference is that when the `lookupClusterClaim` function cannot find the `ClusterClaim` resource, `lookupClusterClaim` gives you an empty string. You can fill this empty string with other templating functions when the `ClusterClaim` does not exist. 
+
+View the following example of the configuration policy that enforces a Kubernetes resource on the target managed cluster:
+
+[source,yaml]
+----
+apiVersion: policy.open-cluster-management.io/v1
+kind: ConfigurationPolicy
+metadata:
+  name: demo-clusterclaims (1)
+spec:
+  object-templates:
+  - objectDefinition:
+      ...
+      kind: ConfigMap
+      data:
+        something: '{{ default "empty" (lookupClusterClaim "possibly.nonexistent") }}'
+----
+<1>  When the `possibly.nonexistent` `ClusterClaim` is found, its value gets used in the ConfigMap. If the `ClusterClaim` does not exist, `empty` gets used instead through the `default` function that is provided by Sprig. 
+
 [#lookup-func]
 === _lookup_
 
--- a/governance/template_functions.adoc
+++ b/governance/template_functions.adoc
@@ -143,7 +143,7 @@ spec:
 [#lookupclusterclaim-func]
 === _lookupClusterClaim_
 
-The `lookupClusterClaim` function is similar to the `fromClusterClaim` function. The only difference is that when the `lookupClusterClaim` function cannot find the `ClusterClaim` resource, `lookupClusterClaim` gives you an empty string. You can use your preferred templating function to detect when the response is an empty string and then adjust the output. For example, when the `ClusterClaim` does not exist, you can define a default value or skip the object. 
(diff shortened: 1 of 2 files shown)


governance/template_functions.adoc
```diff
@@ -140,6 +140,29 @@ spec:
 <2> Configuration values can be set as key-value properties.
 <3> The value for the `platform` data key is a template that retrieves the value of the `platform.open-cluster-management.io` cluster claim. Similarly, it retrieves values for `product` and `version` from the `ClusterClaim` resource.
 
+[#lookupclusterclaim-func]
+=== _lookupClusterClaim_
+
+The `lookupClusterClaim` function is similar to the `fromClusterClaim` function. The only difference is that when the `lookupClusterClaim` function cannot find the `ClusterClaim` resource, `lookupClusterClaim` gives you an empty string. You can fill this empty string with other templating functions when the `ClusterClaim` does not exist.
```
jc-berger: Seems to me that the point / benefit is that the lookupclusterclaim gives users an empty string which users can then fill in with their preferred templating functions. Please correct me if I'm wrong, thank you!

JustinKuli: I would just tweak the wording slightly - "you can _handle_ this empty string" seems more accurate to me.

With `fromClusterClaim`, if the resource doesn't exist, the template totally stops and the policy author can't do anything about it.

JustinKuli (commented):
Good overall to me, just one possible tweak (although, it would be fine as-is)

JustinKuli (approved):
LGTM! Thanks!

release_notes/acm_new.adoc
```diff
@@ -87,6 +87,8 @@ See link:../observability/observe_environments_intro.adoc#observing-environments
 [#governance-new-features]
 == Governance
 
+- You can use the `lookupClusterClaim` function to get an empty string which you can handle with your preferred templating functions. For more information, see link:../governance/template_functions.adoc#lookupclusterclaim-func[_lookupClusterClaim_].
```
swopebe:
```suggestion
- You can use the `lookupClusterClaim` function to get an empty string, which you can handle with your preferred templating functions. For more information, see link:../governance/template_functions.adoc#lookupclusterclaim-func[_lookupClusterClaim_]. 
```

governance/template_functions.adoc
```diff
@@ -140,6 +140,29 @@ spec:
 <2> Configuration values can be set as key-value properties.
 <3> The value for the `platform` data key is a template that retrieves the value of the `platform.open-cluster-management.io` cluster claim. Similarly, it retrieves values for `product` and `version` from the `ClusterClaim` resource.
 
+[#lookupclusterclaim-func]
+=== _lookupClusterClaim_
+
+The `lookupClusterClaim` function is similar to the `fromClusterClaim` function. The only difference is that when the `lookupClusterClaim` function cannot find the `ClusterClaim` resource, `lookupClusterClaim` gives you an empty string. You can handle this empty string with other templating functions when the `ClusterClaim` does not exist.
```
swopebe: Can we be a bit more specific about what you mean by "handle this empty string"? Other than that, it looks pretty good.

governance/template_functions.adoc
```diff
@@ -140,6 +140,29 @@ spec:
 <2> Configuration values can be set as key-value properties.
 <3> The value for the `platform` data key is a template that retrieves the value of the `platform.open-cluster-management.io` cluster claim. Similarly, it retrieves values for `product` and `version` from the `ClusterClaim` resource.
 
+[#lookupclusterclaim-func]
+=== _lookupClusterClaim_
+
+The `lookupClusterClaim` function is similar to the `fromClusterClaim` function. The only difference is that when the `lookupClusterClaim` function cannot find the `ClusterClaim` resource, `lookupClusterClaim` gives you an empty string. You can handle this empty string by applying your preferred templating functions when the `ClusterClaim` does not exist.
```
swopebe: again, I think "handle" is not quite the verb we would use. Technically all the processes could be called "handling". Can we just shorten this line to what they are doing and take out the extra verb?

JustinKuli: Could say something like: You can use other template functions to detect when the response is an empty string, and adjust the rest of the output. For example, when the `ClusterClaim` does not exist, you might define a default value to be used, or you might skip the object entirely.

swopebe (approved):

governance/template_functions.adoc
```diff
@@ -140,6 +140,29 @@ spec:
 <2> Configuration values can be set as key-value properties.
 <3> The value for the `platform` data key is a template that retrieves the value of the `platform.open-cluster-management.io` cluster claim. Similarly, it retrieves values for `product` and `version` from the `ClusterClaim` resource.
 
+[#lookupclusterclaim-func]
+=== _lookupClusterClaim_
+
+The `lookupClusterClaim` function is similar to the `fromClusterClaim` function. The only difference is that when the `lookupClusterClaim` function cannot find the `ClusterClaim` resource, `lookupClusterClaim` gives you an empty string. You can use your preferred templating function to detect when the response is an empty string and then adjust the output. For example, when the `ClusterClaim` does not exist, you can define a default value or skip the object.
```
swopebe:
```suggestion
The `lookupClusterClaim` function is similar to the `fromClusterClaim` function. The only difference is that when the `lookupClusterClaim` function cannot find the `ClusterClaim` resource, `lookupClusterClaim` gives you an empty string. You can use your preferred templating function to detect when the response is an empty string, and then adjust the output. For example, when the `ClusterClaim` does not exist, you can define a default value or skip the object. 
```

PR #7843, opened 2025-05-19
Update Sprig list

dhaiducek:
Also fixes OCP install links identified by the link checker.

ref: https://issues.redhat.com/browse/ACM-20892

PR #8379, opened 2025-11-18
https://issues.redhat.com/browse/ACM-26150

jc-berger:
See Jiras: 

- https://issues.redhat.com/browse/ACM-26150 
- https://issues.redhat.com/browse/ACM-25626 

Checklist considered


fxiang1 (approved):
LGTM! Thanks for fixing these Jake!

release_notes/known_issues_application.adoc
```diff
@@ -124,10 +124,8 @@ All the resources listed in the `MulticlusterApplicationSetReport` are actually
 
 link:https://issues.redhat.com/browse/ACM-17819[ACM-17819]
 
-
 [#argo-pull-model-large]
 == Resource allocation might take several minutes
-//2.8.0: 17819
```
swopebe, line 129: no longer needed, removed in a commit.

swopebe (approved):
This looks much better! Thank you.

**There is one potential issue I found in the file:**

These entries use the same link: ACM-17819

If you can just confirm that this is correct, we can move forward. I am approving with understanding of the need for confirmation of that inquiry.

---

[#argo-pull-model-controller]
== Argo CD controller and the propagation controller might reconcile simultaneously

Both the Argo CD controller and the propagation controller might reconcile on the same application resource and cause the duplicate instances of application deployment on the managed clusters, but from the different deployment models.

For deploying applications by using the pull model, the Argo CD controllers ignore these application resources when the Argo CD `argocd.argoproj.io/skip-reconcile` annotation is added to the template section of the `ApplicationSet`. 

The `argocd.argoproj.io/skip-reconcile` annotation is only available in the GitOps operator version 1.9.0, or later. To prevent conflicts, wait until the hub cluster and all the managed clusters are upgraded to GitOps operator version 1.9.0 before implementing the pull model. 

link:https://issues.redhat.com/browse/ACM-17819[ACM-17819]

[#argo-pull-model-resource]
== Resource fails to deploy

All the resources listed in the `MulticlusterApplicationSetReport` are actually deployed on the managed clusters. If a resource fails to deploy, the resource is not included in the resource list, but the cause is listed in the error message.

link:https://issues.redhat.com/browse/ACM-17819[ACM-17819]

release_notes/known_issues_application.adoc
```diff
@@ -122,7 +122,7 @@ link:https://issues.redhat.com/browse/ACM-3910[ACM-3910]
 
 All the resources listed in the `MulticlusterApplicationSetReport` are actually deployed on the managed clusters. If a resource fails to deploy, the resource is not included in the resource list, but the cause is listed in the error message.
 
-link:https://issues.redhat.com/browse/ACM-3910[ACM-3910]
+link:https://issues.redhat.com/browse/ACM-17819[ACM-17819]
```
swopebe: This is not the right issue, and not what was in the 2.14 link:

From 2.14:

```
//2.8.0: 3910
//2.9:ACM-7843
```

It seems this should be:

https://issues.redhat.com/browse/ACM-7843

I am not sure where #17818 came from.

swopebe (changes requested):
discovered a couple issues. 

1. Some of the other ones were pointing to ACM and not GH, I fixed those.

2. A couple of these, I am not sure where they came from. Either the writer did not add the correct # in previous sources or the recent link that was added was not correct.

I fixed what was clear. A couple others need research.

release_notes/known_issues_application.adoc
swopebe: This is the one that is currently linked that is not correct:

https://issues.redhat.com/browse/ACM-17819

#3910 is not correct either, so let's use #7843, unless there is a valid reason to use this one.

swopebe:
@fxiang1 a couple of these were still off, please check again. Also, we are carrying forward known issues that go back to 2.5, 2.4, and even 1.0....are you sure we cannot remove some of these? @xiangjingli

xiangjingli:
@swopebrandi1  @jc-berger  FYI. 

I double checked the links for each known issue and listed the updates here

https://issues.redhat.com/browse/ACM-25626?focusedId=28493474&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-28493474

fxiang1:
> @fxiang1 a couple of these were still off, please check again. Also, we are carrying forward known issues that go back to 2.5, 2.4, and even 1.0....are you sure we cannot remove some of these? @xiangjingli

@swopebe The only one we can remove is:
`Application topology displays invalid expression`

It seems this was fix in ACM 2.13 so we can remove from the 2.13 docs and above.

release_notes/known_issues_application.adoc
```diff
@@ -239,7 +239,7 @@ spec:
 
 After applying both, you should see the Ansible instance created in your hub cluster.
 
-link:https://github.com/stolostron/backlog/issues/8036[ACM-8036]
+link:https://bugzilla.redhat.com/show_bug.cgi?id=1908691[ACM-1908691]
```
jc-berger: @xiangjingli I linked here to the Bugzilla that was linked to the Github issue, 8036, I think it's easier for users to read the github issues, so since the bugzilla is linked there, I'm leaning towards keeping it as a link to 8036, but let me know what you think!

release_notes/known_issues_application.adoc
```diff
@@ -115,14 +115,14 @@ For deploying applications by using the pull model, the Argo CD controllers igno
...
-link:https://issues.redhat.com/browse/ACM-3910[ACM-3910]
+link:https://issues.redhat.com/browse/ACM-3404[ACM-3404]
 
 [#argo-pull-model-resource]
 == Resource fails to deploy
 
 All the resources listed in the `MulticlusterApplicationSetReport` are actually deployed on the managed clusters. If a resource fails to deploy, the resource is not included in the resource list, but the cause is listed in the error message.
 
-link:https://issues.redhat.com/browse/ACM-3910[ACM-3910]
+link:https://issues.redhat.com/browse/ACM-7843[7843]
```
xiangjingli: https://issues.redhat.com/browse/ACM-7843  is not really related to the known issue.  

The 7843 is related to this known issue
- Local cluster is excluded as a managed cluster for pull model

For this known issue
- Resource fails to deploy

it is also related to the ArgoCD pull model, so we may link it to https://issues.redhat.com/browse/ACM-3404 as well, just like what you did for this known issue
- Argo CD controller and the propagation controller might reconcile simultaneously

release_notes/known_issues_application.adoc
xiangjingli: yeah, It looks better to keep the link to the github issue 8036. users can find the bugzilla link there for more details

xiangjingli:
@jc-berger   Thanks Jake. All bug links for each ALC known issue looks good now :-)

PR #8392, opened 2025-11-20
Note that BMHs are not backed up by default

carbonin:
BMHs used to add hosts in CIM are not backed up by default. Add a note about this in the DR known issues and document a workaround.

https://issues.redhat.com/browse/ACM-26586

cc @swopebe

### DIFF
--- a/release_notes/known_issues_continuity.adoc
+++ b/release_notes/known_issues_continuity.adoc
@@ -93,6 +93,15 @@ oc delete appliedmanifestwork <the-left-appliedmanifestwork-name>
 
 link:https://github.com/stolostron/backlog/issues/27129[ACM-27129]
 
+[#baremetalhost-not-backed-up]
+=== The BareMetalHost resources are not backed up by default
+
+BareMetalHosts used to add hosts to a host inventory when using Central Infrastructure Management are not included in backups by default.
+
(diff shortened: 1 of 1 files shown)


swopebe:
For this, we can use the known issue task that I sent you guys, which needs to be reviewed by tomorrow to make the final build. https://issues.redhat.com/browse/ACM-25642 

Or you can open a doc task for us.

For future, for all changes outside of a typo or something small, we need a doc task of some kind.

carbonin:
I don't think this is really an infra operator known issue. It more came up because of https://issues.redhat.com/browse/ACM-23254

I guess the goal would be to document all unexpected backup/restore behavior in the backup/restore known issues even though that behavior is related to separate components.

In any case I updated the link in the commit and description to the doc task.

swopebe:
Great, we have to link either to that task or a bug, as discussed, so it's good to have it. :-)

carbonin:
> Great, we have to link either to that task or a bug, as discussed, so it's good to have it. :-)

I put it in the commit and description

jc-berger:
SeeDoc PR, implementing this awesome change! 
https://github.com/stolostron/rhacm-docs/pull/8446

PR #8413, opened 2025-11-25
https://issues.redhat.com/browse/ACM-24933 part 2

dockerymick:
Goal is to have this merged before GA. Today is pencils down, but we expect to complete another build to catch missed changes

### DIFF
--- a/secure_clusters/cluster_permission_intro.adoc
+++ b/secure_clusters/cluster_permission_intro.adoc
@@ -8,6 +8,6 @@ With the cluster permssions API, `clusterpermissions.rbac.open-cluster-managemen
 Learn about how to create and manage cluster permissions in the following topics:
 
 * xref:../secure_clusters/create_cluster_perm.adoc#creating-cluster-permission[Creating a cluster permission]
-* xref:../secure_clusters/cluster_permissions/create_cluster_perm_exist.adoc#creating-cluster-permission-exist[Creating a cluster permission using an existing role]
+* xref:../secure_clusters/cluster_permissions/create_cluster_perm_exist.adoc#creating-cluster-permission-exist[Creating a cluster permission by using an existing role]
 * xref:../secure_clusters/cluster_permissions/create_cluster_perm_ref.adoc#creating-cluster-permission-reference[Creating a cluster permission resource to reference subjects]
 * xref:../secure_clusters/cluster_permissions/cluster_permission_validate.adoc#enable-cluster-permission-validate[Enabling validation for cluster permissions]
--- a/secure_clusters/cluster_permission_intro.adoc
+++ b/secure_clusters/cluster_permission_intro.adoc
@@ -7,9 +7,7 @@ With the cluster permssions API, `clusterpermissions.rbac.open-cluster-managemen
 
 Learn about how to create and manage cluster permissions in the following topics:
 
-//* xref:../secure_clusters/create_cluster_perm.adoc#creating-cluster-permission[Creating a cluster permission]
-
-//* xref:../secure_clusters/cluster_permissions/create_cluster_perm_exist.adoc#creating-cluster-permission-exist[Creating a cluster permission using an existing role]
-//* xref:../secure_clusters/cluster_permissions/create_cluster_perm_ref.adoc#creating-cluster-permission-reference[Creating a cluster permission resource to reference subjects]
-
+* xref:../secure_clusters/create_cluster_perm.adoc#creating-cluster-permission[Creating a cluster permission]
+* xref:../secure_clusters/cluster_permissions/create_cluster_perm_exist.adoc#creating-cluster-permission-exist[Creating a cluster permission using an existing role]
+* xref:../secure_clusters/cluster_permissions/create_cluster_perm_ref.adoc#creating-cluster-permission-reference[Creating a cluster permission resource to reference subjects]
 * xref:../secure_clusters/cluster_permissions/cluster_permission_validate.adoc#enable-cluster-permission-validate[Enabling validation for cluster permissions]
--- a/secure_clusters/create_cluster_permission.adoc
+++ b/secure_clusters/create_cluster_permission.adoc
@@ -1,7 +1,7 @@
 [#create-cluster-permission]
 = Creating a cluster permission 
 
-When the service account is created, it does not have any permission associated to it. To grant permissions to the new service account, use the `ClusterPermission` resource. Create the `ClusterPermission` resource in the managed cluster namespace on the hub cluster. Use the `ClusterPermission` resource to create roles, cluster roles resources on the managed clusters, and bind them to a service account through a `RoleBinding` or `ClusterRoleBinding` resource.
+When the service account is created there are no associated permissions. To grant permissions to the new service account, use the `ClusterPermission` resource. Create the `ClusterPermission` resource in the managed cluster namespace on the hub cluster. Use the `ClusterPermission` resource to create roles, cluster roles resources on the managed clusters, and bind them to a service account through a `RoleBinding` or `ClusterRoleBinding` resource.
 
 *Required access:* Cluster administrator
 
--- a/secure_clusters/create_cluster_permission.adoc
+++ b/secure_clusters/create_cluster_permission.adoc
@@ -7,7 +7,7 @@ When the service account is created, it does not have any permission associated 
 
 .Prerequisites
 
-- are there any prereqs for this? 
+- If you want to bind to the `ManagedServiceAccount` resource, you must have the `ManagedServiceAccount` add-on. For more information about enabling the `ManagedServiceAccount` add-on, see link:../add-ons/configure_klusterlet_addons.adoc[Configuring klusterlet add-ons]. 
 
 .Procedure 
 
@@ -71,8 +71,7 @@ spec:
       name: <managed-sa-sample>
 ----
 
-. Save the YAML file and name the file, `cluster-permission.yaml`.
-//should the name match the name in the previous example?
+. Save your YAML file.
 
 . Apply your YAML file by running the following command:
 
--- a/secure_clusters/create_cluster_permission.adoc
+++ b/secure_clusters/create_cluster_permission.adoc
@@ -1,0 +1,95 @@
+[#create-cluster-permission]
+= Creating a cluster permission 
+
+When the service account is created, it does not have any permission associated to it. To grant permissions to the new service account, use the `ClusterPermission` resource. Create the `ClusterPermission` resource in the managed cluster namespace on the hub cluster. Use the `ClusterPermission` resource to create roles, cluster roles resources on the managed clusters, and bind them to a service account through a `RoleBinding` or `ClusterRoleBinding` resource.
+
+*Required access:* Cluster administrator
+
+.Prerequisites
+
+- are there any prereqs for this? 
+
+.Procedure 
+
+Complete the following steps to create a `ClusterPermission` resource that grants permissions to a managed service account:
+
+. To grant the `<managed-sa-sample>` service account permission to the sample mortgage application that is deployed to the `mortgage` namespace on `<managed-cluster>`, create a YAML file with the following content: 
+
++
+[source,yaml]
+----
+apiVersion: rbac.open-cluster-management.io/v1alpha1
+kind: ClusterPermission
+metadata:
+  name: <clusterpermission-msa-subject-sample>
+  namespace: <managed-cluster>
+spec:
+  roles:
+  - namespace: default
+    rules:
+    - apiGroups: ["apps"]
+      resources: ["deployments"]
+      verbs: ["get", "list", "create", "update", "delete", "patch"]
+    - apiGroups: [""]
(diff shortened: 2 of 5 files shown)


secure_clusters/create_cluster_permission.adoc
```diff
@@ -0,0 +1,95 @@
+[#create-cluster-permission]
+= Creating a cluster permission 
+
+When the service account is created, it does not have any permission associated to it. To grant permissions to the new service account, use the `ClusterPermission` resource. Create the `ClusterPermission` resource in the managed cluster namespace on the hub cluster. Use the `ClusterPermission` resource to create roles, cluster roles resources on the managed clusters, and bind them to a service account through a `RoleBinding` or `ClusterRoleBinding` resource.
+
+*Required access:* Cluster administrator
+
+.Prerequisites
+
+- are there any prereqs for this?
```
fxiang1: Yes, we need to add the following:

- If the you wish to bind to the `ManagedServiceAccount` then the ManagedServiceAccount addon needs to be enabled. For help enabling the addon, see https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.14/html-single/add-ons/index#configure-klusterlet-addons .

secure_clusters/create_cluster_permission.adoc
```diff
@@ -0,0 +1,95 @@
...
+      name: <managed-sa-sample>
+  clusterRoleBinding:
+    subject:
+      apiGroup: authentication.open-cluster-management.io
+      kind: ManagedServiceAccount
+      name: <managed-sa-sample>
+----
+
+. Save the YAML file and name the file, `cluster-permission.yaml`.
+//should the name match the name in the previous example?
```
fxiang1: The user can use any name, it doesn't have to match.

secure_clusters/create_cluster_permission.adoc
fxiang1: Looks like this prereq applies to create_cluster_permission_exist.adoc and create_cluster_permission_ref.adoc as well. I wonder if we can have an overall prereq section for all the cluster permission docs?

dockerymick: @fxiang1

secure_clusters/create_cluster_permission.adoc
```diff
@@ -0,0 +1,95 @@
...
+
```
dockerymick, line 97: @fxiang1 ^

secure_clusters/create_cluster_permission_exist.adoc
```diff
@@ -0,0 +1,76 @@
+[#create-cluster-permission-exist]
+= Creating a cluster permission by using an existing role 
+
+When you create a `ClusterPermission` resource, you automatically create new roles. You can still use the `ClusterPermission` resource with an existing role. 
+
+*Required access:* Cluster administrator
+
+.Prerequisites
+
+- are there any prereqs for this?
```
dockerymick: @fxiang1

secure_clusters/create_cluster_permission_ref.adoc
```diff
@@ -0,0 +1,82 @@
+[#creating-cluster-permission-reference]
+= Creating a cluster permission resource to reference subjects 
+
+If you create a `ClusterPermission` resource to reference different subjects, you can work with different subjects at the same time instead of only working on one subject at a time. 
+
+*Required access:* Cluster administrator
+
+.Prerequisites
+
+- are there any prereqs for this?
```
dockerymick: @fxiang1

secure_clusters/create_cluster_permission.adoc
fxiang1: Yes, it's the same prereq as above. That's why I'm wondering instead of repeating the same prereq three times, should we just have one prereq?
image

secure_clusters/create_cluster_permission_exist.adoc
fxiang1: Yes, the same prereq as above
image

secure_clusters/create_cluster_permission_ref.adoc
fxiang1: Yes, the same prereq as above
image

secure_clusters/create_cluster_permission.adoc
fxiang1: See my comment above.
image

secure_clusters/create_cluster_permission.adoc
dockerymick: I like your suggestions. It is required to list the the prereqs for each Task file (procedure focused), according to the [Red Hat Style Guide](https://redhat-documentation.github.io/modular-docs/#con-creating-procedure-modules_writing-mod-docs). However, we can discuss this at a future Doc guild

secure_clusters/create_cluster_permission.adoc
dockerymick: Ok, I will remove the instruction about what to name the file.

fxiang1 (approved):

secure_clusters/cluster_permission_intro.adoc
```diff
@@ -7,9 +7,7 @@ With the cluster permssions API, `clusterpermissions.rbac.open-cluster-managemen
 
 Learn about how to create and manage cluster permissions in the following topics:
 
-//* xref:../secure_clusters/create_cluster_perm.adoc#creating-cluster-permission[Creating a cluster permission]
-
-//* xref:../secure_clusters/cluster_permissions/create_cluster_perm_exist.adoc#creating-cluster-permission-exist[Creating a cluster permission using an existing role]
-//* xref:../secure_clusters/cluster_permissions/create_cluster_perm_ref.adoc#creating-cluster-permission-reference[Creating a cluster permission resource to reference subjects]
-
+* xref:../secure_clusters/create_cluster_perm.adoc#creating-cluster-permission[Creating a cluster permission]
+* xref:../secure_clusters/cluster_permissions/create_cluster_perm_exist.adoc#creating-cluster-permission-exist[Creating a cluster permission using an existing role]
```
swopebe:
```suggestion
* xref:../secure_clusters/cluster_permissions/create_cluster_perm_exist.adoc#creating-cluster-permission-exist[Creating a cluster permission by using an existing role]
```

secure_clusters/create_cluster_permission.adoc
```diff
@@ -0,0 +1,94 @@
+[#create-cluster-permission]
+= Creating a cluster permission 
+
+When the service account is created, it does not have any permission associated to it. To grant permissions to the new service account, use the `ClusterPermission` resource. Create the `ClusterPermission` resource in the managed cluster namespace on the hub cluster. Use the `ClusterPermission` resource to create roles, cluster roles resources on the managed clusters, and bind them to a service account through a `RoleBinding` or `ClusterRoleBinding` resource.
```
swopebe: We should not be using "it" so much. I recommend:

By default, a service account with created without an associated permission.

secure_clusters/create_cluster_permission.adoc
```diff
@@ -0,0 +1,94 @@
...
+.Prerequisites
+
+- If you want to bind to the `ManagedServiceAccount` resource, you must have the `ManagedServiceAccount` add-on. For more information about enabling the `ManagedServiceAccount` add-on, see link:../add-ons/configure_klusterlet_addons.adoc[Configuring klusterlet add-ons]. 
+
+.Procedure 
+
+Complete the following steps to create a `ClusterPermission` resource that grants permissions to a managed service account:
+
+. To grant the `<managed-sa-sample>` service account permission to the sample mortgage application that is deployed to the `mortgage` namespace on `<managed-cluster>`, create a YAML file with the following content: 
+
```
swopebe, line 18: We need to introduce this YAML better. I would rather not have most of the file a YAML file with no explanation. 

We need to introduce the kind, then we should have separate steps about the roles, role bindings, etc... It's not good practice to give a large about of YAML in one step and say Create this YAML.

swopebe: See here that the bindings were created in a step in 2.14 and pieces of the YAML were defined in a step or the intro: https://github.com/stolostron/rhacm-docs/blob/2.14_stage/secure_clusters/fine_grain_rbac_cli.adoc

We got away from that for some reason here?

secure_clusters/create_cluster_permission_exist.adoc
```diff
@@ -0,0 +1,76 @@
...
+      name: crb-cluster1-argo-app-con-3 <3>
+      roleRef: <4> 
+        apiGroup: rbac.authorization.k8s.io
+        kind: ClusterRole
+        name: argocd-application-controller-3
+      subject:
+        apiGroup: rbac.authorization.k8s.io
+        kind: Group
+        name: group1 
+----
```
swopebe, line 84: Again, we need to introduce this large amount of YAML better. For new doc, we said we are not using call outs. These should be steps. Is there a reason we went with new callouts? (conversion is the reason)

secure_clusters/create_cluster_permission_ref.adoc
```diff
@@ -0,0 +1,82 @@
...
+. Run the following command to apply your `ClusterPermission` resource that references different subjects:
+
++
+[source,bash]
+----
+oc apply -f <clusterpermission-users-groups-yaml-file>.yaml
+----
+
++
+Your `ClusterPermission` resource might resemble the following YAML file with the specified `subjects` field:
```
swopebe: such as ___ for ___ and ___ for ___...


We have this large amount of YAML and we need to tell the user what to look for.

swopebe (changes requested):

secure_clusters/create_cluster_permission.adoc
dockerymick: This is old content. I just didn't focus on updating all of the content because we were up against time.

secure_clusters/create_cluster_permission_exist.adoc
dockerymick: Just didn't focus on that during the PR bc the content is old and needed to be moved.

secure_clusters/create_cluster_permission.adoc
dockerymick:
```suggestion
When the service account is created there are no associated permissions. To grant permissions to the new service account, use the `ClusterPermission` resource. Create the `ClusterPermission` resource in the managed cluster namespace on the hub cluster. Use the `ClusterPermission` resource to create roles, cluster roles resources on the managed clusters, and bind them to a service account through a `RoleBinding` or `ClusterRoleBinding` resource.
```

secure_clusters/create_cluster_permission.adoc
dockerymick: @fxiang1

fxiang1: Right I agree with @dockerymick , I thought we were just moving this from the ALC Gitops section. This is not something new we've introduced recently.

swopebe: We have gotten away from pushing large YAML samples without directions for them, but now I understand that this is previous content. I suggest we introduce the pieces that make the cluster permission, without reorganizing. Going forward the team agrees that this is not best practice and that important pieces of YAML need to be implemented as part of a larger process for understanding.

fxiang1: Yep understood, going through our docs I see we have many large YAML samples without explanation. Maybe we need to create a separate story/task for each of the component teams to just check and add explanation steps for these?

dockerymick: @fxiang1 yup that's a great idea. I will improve the examples in this PR and then focus on other areas in the tech debt/refresh doc issue

secure_clusters/create_cluster_permission_ref.adoc
dockerymick:
```suggestion
Your `ClusterPermission` resource might resemble the following YAML file where the `subjects` specification is added for your `roleBindings` and `clusterRoleBinding` specification: 
```

dockerymick:
@swopebe @fxiang1 just made updates to break up the long YAML. Let me know what yall think.

secure_clusters/create_cluster_permission.adoc
swopebe: Wow, these changes are more than I even expected here for this PR. I thought we could just add some intro info and be good but you took it even further to improve it. Great job.

swopebe (approved):

secure_clusters/create_cluster_permission.adoc
swopebe: @fxiang1 I think for the newer info that is being used a lot, yes, we can improve that YAML presentation. We have as a team mostly been doing that in new content. I don't think we will have time to fix everything, but certainly highly used doc we can improve.

secure_clusters/create_cluster_permission_exist.adoc
swopebe: I see.

dockerymick:
Received approval from @xiangjingli through Slack.

PR #8425, opened 2025-12-02
https://issues.redhat.com/browse/ACM-25316 Add TNF AI docs

oafischer:
2.15

Checklist considered


clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,181 @@
...
+metadata:
+  name: assisted-service-custom
+  namespace: multicluster-engine
+----
+
+.. Reference the `ConfigMap` resource in an annotation on the `AgentServiceConfig` resource. Run the following command:
++
+[source,bash]
+----
+oc annotate agentserviceconfig agent unsupported.agent-install.openshift.io/assisted-service-configmap=assisted-service-custom
```
carbonin, line 35: Is this here because it's a tech preview feature? Maybe we should mention here that's why this is necessary?

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,181 @@
...
+    servingCertificates: {}
+  platform:
+    agentBareMetal:
+      agentSelector:
+        cluster: tnf
+  pullSecretRef:
+    name: pull-secret
+----
+
+.. Create a `ClusterDeployment` resource. Use the following YAML sample. Replace values where needed:
```
carbonin: This should be `InfraEnv`

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,181 @@
...
+  networking:
+    clusterNetwork:
+    - cidr: 10.128.0.0/14
+      hostPrefix: 23
+    machineNetwork:
+    - cidr: 192.168.122.0/24
+    serviceNetwork:
+    - 172.30.0.0/16
+  provisionRequirements:
+    controlPlaneAgents: 2
```
carbonin, line 90: https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.14/html/clusters/cluster_mce_overview#create-agentclusterinstall-cr this lists the options for this field as 1 or 3. Maybe we should just roll TNF into the existing doc? The only differences are this field, and the BMC secret, right?

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,181 @@
...
+- Set an annotation on the Agent `BareMetalHost` resource to create the secret from the `BareMetalHost` BMC configuration. Then, set the Agent `fencingCredentialsSecretRef` to the secret name, which is `<your-agent-name>-fencing-credentials`. Use the following annotation key:
++
+[source,bash]
+----
+bmac.agent-install.openshift.io/create-fencing-credentials-secret
+----
++
+*Note:* You can use any value.
+
+- Manually create the secret. Then, manually set the Agent zfencingCredentialsSecretRef` field to the secret name.
```
carbonin:
```suggestion
- Manually create the secret. Then, manually set the Agent `fencingCredentialsSecretRef` field to the secret name.
```

I think this is a typo?

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,181 @@
+[#install-tnf-cluster-ai]
+= Installing a two nodes with fencing cluster with {ai} (Technology Preview)
+
+A two nodes with fencing (TNF) cluster is an {ocp} cluster that has two control plane nodes. Use TNF clusters to ensure redundancy at the edge and limit hardware costs.
+
+.Prerequisites
+
+- You are using the `baremetal` or `none` platform.
+- You have 2 control plane nodes configured.
```
carbonin: I don't think this is a prereq since these hosts will be added using the infraenv we create in this procedure, right?

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,181 @@
...
+- You have access to the baseboard management controller (BMC) of the platform nodes.
+- Your {ocp-short} version is 4.20 or later.
```
carbonin: This should probably be more clear that it means that the cluster we're _installing_ needs to be 4.20 or later (as opposed to the version MCE is running on)

carbonin (commented):
@giladravid16 Can you also take a look?

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: It is, good catch!

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,181 @@
...
+- Manually create the secret and add an annotation on the Agent `BareMetalHost` resource that references the secret. Use the following annotation key:
++
+[source,bash]
+----
+bmac.agent-install.openshift.io/fencing-credentials-secret-name
+----
++
+*Note:* The {ai} sets the Agent `fencingCredentialsSecretRef` field to the value of the annotation.
+
+- Set an annotation on the Agent `BareMetalHost` resource to create the secret from the `BareMetalHost` BMC configuration. Then, set the Agent `fencingCredentialsSecretRef` to the secret name, which is `<your-agent-name>-fencing-credentials`. Use the following annotation key:
```
giladravid16: In this method the users don't need to update the Agent themselves, Assisted does it for them

clusters/assisted_installer/ai_install_tnf.adoc
giladravid16: Yes all of the prerequisites are for the spoke cluster

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,181 @@
...
+spec:
+  clusterRef:
+    name: tnf-cluster
+    namespace: tnf-cluster
+  pullSecretRef:
+    name: pull-secret
+  sshAuthorizedKey: <your-public-ssh-key>
+----
+
+. Create a fencing credentials secret. Use the following YAML sample. Replace values where needed:
```
giladravid16: This should be done for each node, but this step should be marked as optional because of the methods mentioned bellow (one of them doesn't require users to create these secrets)

clusters/assisted_installer/ai_install_tnf.adoc
giladravid16: It's mentioned at the top that it's tech preview, maybe we should also mention it here and say that's why it's disabled by default?

clusters/assisted_installer/ai_install_tnf.adoc
giladravid16: The differences are this field, the secrets at the end of this doc, and the prerequisites

oafischer: I'm OK with moving the entire doc set if it makes sense, right now I don't see a strong preference for either options.

@carbonin @giladravid16 What's the consensus?

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: Fixed!

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: If tech preview is mentioned at the beginning, it applies to the entire section. We wouldn't repeat it again in the procedure as per our current doc guidelines.

If it's too confusing for customers we can maybe revisit this or add a note if necessary.

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: I went ahead and removed it

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: Updated!

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: Removed

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: Updated!

oafischer:
Thanks for the reviews @carbonin and @giladravid16, I addressed both of your suggestions if you don't mind taking a look again.

giladravid16:
@oafischer It looks good to me

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,180 @@
...
+  namespace: tnf-cluster
+type: Opaque
+----
+- *Optional:* The `certificateVerification` field default value is `Enabled`.
++
+*Note:* Fencing credentials for an Agent are stored as a Kubernetes secret in the Agent namespace that the Agent references in the `agent.spec.fencingCredentialsSecretRef` field.
+
+. Set the fencing credentials to approve two Agents. Use one of the following methods:
+
+- Manually create the secret and add an annotation on the Agent `BareMetalHost` resource that references the secret. Use the following annotation key:
```
oafischer: Not sure if a bullet list is the best way to indicate different options, let me know what you think @swopebe 

Mostly concerned since each bullet point is still a step/part of the procedure

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,180 @@
...
+  certificateVerification: Disabled
+  username: <username>
+  password: <password>
+kind: Secret
+metadata:
+  name: node-0-fencing-credentials
+  namespace: tnf-cluster
+type: Opaque
+----
+- *Optional:* The `certificateVerification` field default value is `Enabled`.
```
oafischer: Red Hat style guide tells us to do call outs like this. Do we want to follow that?

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,180 @@
+[#install-tnf-cluster-ai]
+= Installing a two nodes with fencing cluster with {ai} (Technology Preview)
+
+A two nodes with fencing (TNF) cluster is an {ocp} cluster that has two control plane nodes. Use TNF clusters to ensure redundancy at the edge and limit hardware costs.
+
+.Prerequisites
+
+- You are using the `baremetal` or `none` platform.
+- You have access to the baseboard management controller (BMC) of the platform nodes.
+- Your {ocp-short} managed cluster version is 4.20 or later.
```
swopebe: We need to remove this: (or later) 

bc there is only 4.20. 

You can say 4.20, but the user could be stopped here and asking where is "later" releases and how do they get to them. We just will have to edit it next release.

oafischer: I'll remove it, I'll make a note to add it back for 2.16

clusters/assisted_installer/ai_install_tnf.adoc
swopebe: If you mention here that this is tech preview, you suggest to the user that the rest of the procedure is not tech preview, because you highlight one step as tech preview. That is why we put it in the title. 

We should also put it in the short description:

**Technology Preview:** short description....

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,180 @@
...
++
+[source,yaml]
+----
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: tnf-cluster
+----
+
+.. Create a pull secret. Use the following YAML sample. Replace values where needed:
```
swopebe: Replace values where needed: (we should be specific). I have seen this entering our doc recently, but we need to tell them what to replace.

For instance:

Replace `<your-base64-pull-secret>` with the name of your pull secret.

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,180 @@
...
+kind: Secret
+type: kubernetes.io/dockerconfigjson
+metadata:
+  name: pull-secret
+  namespace: tnf-cluster
+data:
+  .dockerconfigjson: <your-base64-pull-secret>
+----
+
+.. Create a `AgentClusterInstall` resource. Use the following YAML sample. Replace values where needed:
```
swopebe: Same here

clusters/assisted_installer/ai_install_tnf.adoc
swopebe: Keep in mind our docs have to be chunked up into smaller files going forward, with one or MAYBE two concepts per file. This is for JTBD and for conversion.

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,180 @@
...
+    - cidr: 192.168.122.0/24
+    serviceNetwork:
+    - 172.30.0.0/16
+  provisionRequirements:
+    controlPlaneAgents: 2
+  sshPublicKey: <your-public-ssh-key>
+----
+
+.. Create a `ClusterDeployment` resource. Use the following YAML sample. Replace values where needed:
++
```
swopebe, line 95: Same here with replace.

clusters/assisted_installer/ai_install_tnf.adoc
swopebe: you can but you can also make it part of the step. Not sure why we have optional here?

clusters/assisted_installer/ai_install_tnf.adoc
swopebe: I think the bulleted list works. What is better I think is to provide the reasons to pick one or the other. I am not sure why there are two options at this point, is what I am saying. Something like

. Set the ____ . Use one of the two following steps, depending on __:

.. If ___, manually create...

.. If ___, set an annotation...

You can also do that with the bullets, we may just have to change it later for conversion to DITA-- not sure. We may not be able to create bullets under a step. Don't remember exactly from my days in DITA.

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,180 @@
...
+- *Optional:* The `certificateVerification` field default value is `Enabled`.
++
+*Note:* Fencing credentials for an Agent are stored as a Kubernetes secret in the Agent namespace that the Agent references in the `agent.spec.fencingCredentialsSecretRef` field.
+
+. Set the fencing credentials to approve two Agents. Use one of the following methods:
+
+- Manually create the secret and add an annotation on the Agent `BareMetalHost` resource that references the secret. Use the following annotation key:
++
+[source,bash]
+----
```
swopebe, line 165: watch format here

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,180 @@
...
+bmac.agent-install.openshift.io/fencing-credentials-secret-name
+----
++
+*Note:* The {ai} sets the Agent `fencingCredentialsSecretRef` field to the value of the annotation.
+
+- Set an annotation on the Agent `BareMetalHost` resource to create the secret from the `BareMetalHost` BMC configuration. Use the following annotation key:
++
```
swopebe, line 173: format

clusters/release_notes/mce_new.adoc
```diff
@@ -33,5 +33,6 @@ Learn about new features and enhancements for Cluster lifecycle with {mce-short}
...
 - *Technology Preview:* Provision and manage the lifecycle of a {rosa-hcp} cluster with Cluster API so that you can use a GitOps approach to cluster liffecycle at scale. See xref:../cluster_api/create_capi_rosa.adoc#create-rosa-hcp-capi-cluster[Creating a {rosa-hcp} cluster with Cluster API (Technology Preview)] to learn more.
 - You can now define managed namespaces for `ManagedClusterSet` resources to manage and scale clusters efficiently, or to apply configurations consistently across your fleet. See xref:../cluster_lifecycle/set_managed_ns.adoc#set-managed-ns[Setting managed namespaces for _ManagedClusterSets_].
+- You can now install a two nodes with fencing cluster to ensure redundancy at the edge and limit hardware costs. See xref:../assisted_installer/ai_install_tnf.adoc#install-tnf-cluster-ai[Installing a two nodes with fencing cluster with {ai} (Technology Preview)] to learn more.
```
swopebe: can we put TNF here because reading this outloud, and how it will be translated or read with a screen reader, it is confusing:

"You can now install a two nodes with fencing cluster" appears at first like the "a" should not be there, for instance.

swopebe:
```suggestion
- You can now install a two nodes with fencing (TNF) cluster to ensure redundancy at the edge and limit hardware costs. See xref:../assisted_installer/ai_install_tnf.adoc#install-tnf-cluster-ai[Installing a two nodes with fencing cluster with {ai} (Technology Preview)] to learn more.
```

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,180 @@
+[#install-tnf-cluster-ai]
+= Installing a two nodes with fencing cluster with {ai} (Technology Preview)
```
swopebe: can we put (TNF) here because reading this out loud, and how it will be translated or read with a screen reader, it is confusing:

"You can now install a two nodes with fencing cluster" appears at first like the "a" should not be there, for instance.

swopebe (commented):
The content itself is very clean. Couple things:

Replaceable values should be specifically guided for the user. So that will take a bit of time to fix.

Couple format items.

Small content changes.

This won't likely make today's refresh, but let me know if you want to try to get this in today. If not, we can refresh again tomorrow.

swopebe (changes requested):
The content itself is very clean. Couple things:

Replaceable values should be specifically guided for the user. So that will take a bit of time to fix.

Couple format items.

Small content changes.

This won't likely make today's refresh, but let me know if you want to try to get this in today. If not, we can refresh again tomorrow.

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: I'll update it across the file

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: I added it to the description

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: I'll leave this as is for now

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: Updated

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: Looks OK on my end, what exactly is off?

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: It's listed in order of priority, I can add that info to the step

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: I'll move it to the step

clusters/release_notes/mce_new.adoc
oafischer: I'll change it to TNF, you can see what it stands for in the link anyway so that makes sense

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: Added

oafischer: OCP Docs uses `two-node with fencing cluster`. Do we want to follow them or go our own way with TNF?

clusters/assisted_installer/ai_install_tnf.adoc
swopebe: these should be grayed out.

swopebe: I just pressed enter and it worked.

clusters/assisted_installer/ai_install_tnf.adoc
swopebe: did the same here.

swopebe (approved):

clusters/assisted_installer/ai_install_tnf.adoc
oafischer: The code is part of the bullet list, so yellow is correct for these.

The format currently with the space is off:

image

oafischer:
Dev approval received via Slack DM.

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,183 @@
...
+kind: Secret
+metadata:
+  name: node-0-fencing-credentials
+  namespace: tnf-cluster
+type: Opaque
+----
++
+*Note:* Fencing credentials for an Agent are stored as a Kubernetes secret in the Agent namespace that the Agent references in the `agent.spec.fencingCredentialsSecretRef` field.
+
+. Set the fencing credentials to approve two Agents. Use one of the following methods. The methods are listed in descending order of priority:
```
carbonin: Do the credentials have anything to do with approving agents?
This first sentence is a bit strange to me. Maybe just
```suggestion
. Use one of the following methods to set Agent fencing credentials. The methods are listed in descending order of priority:
```

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,183 @@
...
+
+- Manually create the secret and add an annotation on the Agent `BareMetalHost` resource that references the secret. Use the following annotation key:
```
carbonin: `Agent` and `BareMetalHost` are two different CRs that represent very similar concepts. Maybe just drop "Agent" here for clarity?
```suggestion
- Manually create the secret and add an annotation on the `BareMetalHost` resource that references the secret. Use the following annotation key:
```

clusters/assisted_installer/ai_install_tnf.adoc
```diff
@@ -0,0 +1,183 @@
...
+
++
+[source,bash]
+----
+bmac.agent-install.openshift.io/fencing-credentials-secret-name
+----
++
+*Note:* The {ai} sets the Agent `fencingCredentialsSecretRef` field to the value of the annotation.
+
+- Set an annotation on the Agent `BareMetalHost` resource to create the secret from the `BareMetalHost` BMC configuration. Use the following annotation key:
```
carbonin, line 171: Can this use any value? @giladravid16 ?

carbonin (approved):

PR #8428, opened 2025-12-03
https://issues.redhat.com/browse/ACM-23758 forwarding alerts mcoa

dockerymick:


observability/obs_alert_forward.adoc
```diff
@@ -0,0 +1,247 @@
...
+                config.yaml: |
+                  {{ (merge $cy $mangedConfig)| toYaml |autoindent }}
+              kind: ConfigMap
+              metadata:
+                name: cluster-monitoring-config
+                namespace: openshift-monitoring
+            recordDiff: InStatus
+  remediationAction: inforce
+----
+//you mention `inform` mode, but I only see `enforce` for the remediation action. Please clarify. Do we need to mention anything about `inform` mode here?
```
swopebe: comment is added here, just in case.

-----

For these YAML samples, I could be wrong, but I think we have to add something to the YAML to show it is a part of a larger. See the following example where we add `...`

```
Complete the following steps to add a worker node to the managed cluster:

. Define a new node object in the existing `ClusterInstance` custom resource:

+
[source,yaml]
----
spec:
  ...
  nodes:
    - hostName: "<host_name>"
      role: "worker"
      templateRefs:
        - name: ai-node-templates-v1
          namespace: rhacm
      bmcAddress: "<bmc_address>"
      bmcCredentialsName:
        name: "<bmc_credentials_name>"
      bootMACAddress: "<boot_mac_address>"
...
----
```

observability/obs_alert_forward_mcoa.adoc
```diff
@@ -0,0 +1,247 @@
+[#forward-alert-mcoa]
+= Forwarding alerts for multicluster observability add-on
```
swopebe, line 2: I had this question on the others:

multicluster observability add-on (is this coded?) If not, this isn't a title?

observability/obs_alert_forward.adoc
```diff
@@ -0,0 +1,247 @@
...
+ 
+          ## Get the current ConfigMap and read the value as a yaml structure
+          {{- $cmo := (lookup "v1" "ConfigMap" "openshift-user-workload-monitoring" "user-workload-monitoring-config") }}
+          {{- $cy := dict }}
+          {{- if and $cmo $cmo.data }}
+            {{- $cy = (index $cmo "data" "config.yaml") | fromYaml }}
+          {{- end }}
+----
+
+.. Define what you want your configuration to include. Your YAML file might resemble the following partial example:
```
swopebe: What is the reason for breaking this out in to a step? Not sure what the goal is, that is why I am asking.

observability/obs_alert_forward_mcoa.adoc
```diff
@@ -0,0 +1,247 @@
...
+                    key: service-ca.crt
+                    name: hub-alertmanager-router-ca-%[1]s
+                  insecureSkipVerify: false
+              externalLabels:
+                managed_cluster: %[3]s
+          ` }}
+----
+
+.. Merge all of the configurations into one. Your YAML file might resemble the following partial example:
+
```
swopebe, line 183: The step is to merge? What do we mean by this? What is the reason for breaking this part out?

observability/obs_alert_forward_mcoa.adoc
```diff
@@ -0,0 +1,247 @@
...
+  apiGroup: cluster.open-cluster-management.io
+  kind: Placement
+  name: global
+subjects:
+- apiGroup: policy.open-cluster-management.io
+  kind: Policy
+  name: mcoa-alert-forward-platform
+----
+
+.. Apply your YAML file. If you make custom changes to the `additionalAlertmanagerConfigs` specification of the config map, be sure to add the changes directly into the policy, before applying the policy. Run the following command:
```
swopebe: there is a command to apply usually.

observability/obs_alert_forward.adoc
```diff
@@ -0,0 +1,247 @@
+[#forward-alert-mcoa]
+= Forwarding alerts for multicluster observability add-on
+
+Configure alert forwarding for the multicluster observability add-on to send alerts to external alert management systems from your managed clusters. By default, multicluster observability add-on is not configured to send alerts from Prometheus on your managed clusters to the hub cluster Alertmanager.
```
jacobbaungard: It is specifically for sending alerts to the Alertmanager with provide with the Observability stack on the Hub.

observability/obs_alert_forward.adoc
```diff
@@ -0,0 +1,247 @@
...
+
+.Prerequisites
+
+- You must have the multicluster observability add-on is enabled. For more details, see xref:../observability/obs_mcoa_intro.adoc#enabling-mcoa[Enabling the multicluster observability add-on].
+
+.Procedure
+
+Complete the following steps to create a policy to configure alert forwarding for the multicluster observability add-on:
+
+. To configure alert forwarding for platform metrics, create a YAML file named `mcoa-alert-forward-platform`.
```
jacobbaungard:
```suggestion
. To configure alert forwarding for platform metrics, create a YAML file named `mcoa-alert-forward-platform.yaml`. 
```

observability/obs_alert_forward.adoc
```diff
@@ -0,0 +1,247 @@
...
+          {{ $hubBaseDomain:= "INPUT_BASE_DOMAIN" }}
+          {{ $hubName := (split "." $hubBaseDomain)._0 }}
+          {{- $cmo := (lookup "v1" "ConfigMap" "openshift-monitoring" "cluster-monitoring-config") }}
+          {{- $cy := dict }}
+          {{- if and $cmo $cmo.data }}
+            {{- $cy = (index $cmo "data" "config.yaml") | fromYaml }}
+          {{- end }}
+----
+
+.. Define what you want your configuration to include. Your YAML file might resemble the following partial example:
```
jacobbaungard: I don't expect people really changes this configuration much. Only thing they have to change is the "INPUT_BASE_DOMAIN" above.

observability/obs_alert_forward.adoc
```diff
@@ -0,0 +1,247 @@
...
+                  ca:
+                    key: service-ca.crt
+                    name: hub-alertmanager-router-ca-%[1]s
+                  insecureSkipVerify: false
+              externalLabels:
+                managed_cluster: %[3]s
+          ` }}
+----
+
+.. Merge all of the configurations into one. Your YAML file might resemble the following partial example:
```
jacobbaungard: I think we should avoid splitting the yaml configuration as that just adds confusion in my opinion. If you need it to be smaller the placementBinding can be separate as you already did.

If it's not possible to keep as one, is there no other way to fix it? Can we provide the file to download instead of rendering it directly on the docs site?

observability/obs_alert_forward.adoc
```diff
@@ -0,0 +1,247 @@
...
+.Prerequisites
+
+- You must have the multicluster observability add-on is enabled. For more details, see xref:../observability/obs_mcoa_intro.adoc#enabling-mcoa[Enabling the multicluster observability add-on].
+
+.Procedure
+
+Complete the following steps to create a policy to configure alert forwarding for the multicluster observability add-on:
+
+. To configure alert forwarding for platform metrics, create a YAML file named `mcoa-alert-forward-platform`. 
+.. Add the platform config map details to the `objects-templates-raw` specification. Be sure to add the base domain for your hub cluster in the `$hubBaseDomain` parameter. Your YAML file might resemble the following example:
```
jacobbaungard:
```suggestion
.. Your YAML file should look like below. Make sure to add the base domain for your hub cluster in the `$hubBaseDomain` parameter. 
```

observability/obs_alert_forward.adoc
```diff
@@ -0,0 +1,247 @@
...
+.. Apply your YAML file. If you make custom changes to the `additionalAlertmanagerConfigs` specification of the config map, be sure to add the changes directly into the policy, before applying the policy. Run the following command:
+
++
+[source,bash]
+----
+oc apply -f ./forward-platform-alerts.yaml
+----
+
++
+*Note:* If you have a managed cluster other than a {ocp} cluster, you must create a `Placement` resource.
```
jacobbaungard:
```suggestion
*Note:* If you have managed clusters other than a {ocp} cluster, you must create the appropriate `PlacementBinding` resource targeting the managed clusters.
```

observability/obs_alert_forward.adoc
```diff
@@ -0,0 +1,247 @@
...
+
+.. To apply the policy changes across all managed clusters and enable alert forwarding from your managed clusters to the hub cluster, change the `remediationAction` parameter value to `enforce`.
```
jacobbaungard:
```suggestion
.. The policy presented here is in inform mode. This means you can validate the changes to be applied using the governance user interface in the OCP console, to ensure you are satisfied with the change. The policy will show up to be as non-complient, and you can then verify the diff the policy would apply if enforced. To apply the policy changes across all managed clusters and enable alert forwarding from your managed clusters to the hub cluster, change the `remediationAction` parameter value to `enforce`. 
```

Anything we can do to highlight this, so it is not missed, would be good.

observability/obs_alert_forward.adoc
```diff
@@ -0,0 +1,247 @@
...
+              apiVersion: v1
+              data:
+                config.yaml: |
+                  {{ (merge $cy $mangedConfig)| toYaml |autoindent }}
+              kind: ConfigMap
+              metadata:
+                name: cluster-monitoring-config
+                namespace: openshift-monitoring
+            recordDiff: InStatus
+  remediationAction: inforce
```
jacobbaungard:
```suggestion
  remediationAction: inform
```

observability/obs_alert_forward.adoc
```diff
@@ -0,0 +1,247 @@
...
+              apiVersion: v1
+              data:
+                config.yaml: |
+                  {{ (merge $cy $mangedConfig )| toYaml |autoindent }}
+              kind: ConfigMap
+              metadata:
+                name: user-workload-monitoring-config
+                namespace: openshift-user-workload-monitoring
+            recordDiff: InStatus
+  remediationAction: enforce
```
jacobbaungard:
```suggestion
  remediationAction: inform
```

observability/obs_alert_forward.adoc
jacobbaungard: Other places we have collapsible examples like here, would that not be acceptable? https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.4/html-single/governance/index#policy-sample-file

observability/obs_alert_forward.adoc
jacobbaungard: Sorry, I meant to put them in "inform" in the example. I added suggestions to fix it.

observability/obs_alert_forward.adoc
swopebe: We cannot use directional language or "look like" -- this is why Mikela wrote it the way that it is. Also we cannot use the word "should". Most of this writing changes to fit the IBM Style and Red Hat Style company guidance.

swopebe: We can play with that line a little bit but typically we say:

See the following example:

Your YAML file might resemble the following example:

Something like that.

observability/obs_alert_forward.adoc
swopebe: I would just merge this with the YAML above. We only need to pull out important specs or other YAML values (if there are many) that need to be changed or built a certain way for part of the procedure. We don't typically pull this info out .

observability/obs_alert_forward.adoc
swopebe: @jacobbaungard we are no longer (for a while now) dumping large YAML samples in the doc without explanation or procedures, that is why she did this for this PR, but in this case, these YAML breakouts are not necessary.

2.4 doc is very old, not good to use bc we were not doing peer reviews and we were pointing to links we should not have.

I will go over this with the team again: When to break YAML up for a procedure, when not to break it up. In this case, the config info here does not need to be separated, where if the user is adding specs to YAML, it likely does.

We are no longer allowed to use callouts and with the move to DITA, you can expect we will need to build procedures that way.

observability/obs_alert_forward.adoc
swopebe: We also cannot say "presented here" -- as it is self-demonstrative. We try to avoid "here". Same with "this means"

observability/obs_alert_forward.adoc
swopebe: Same feedback, we don't need to break this out.

swopebe:
This needs a bit more work and we will push this for an immediate refresh. It will not make GA. Mikela had to back me up while as I was on PTO and the team is dealing with resource issues. 

That said, it seems MCOA was **much** bigger than we sized.

observability/obs_alert_forward.adoc
jacobbaungard: Right, copied the wrong link, it is also there in [2.14](https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.14/html-single/governance/index#policy-sample-file).

observability/obs_alert_forward.adoc
jacobbaungard: I reacted mostly on "might". I would prefer language that more strongly say it _has_ to look like this. Anyway that is no big deal, and feel free to adjust so it matches guidelines. Most important is removing the first sentence there.

observability/obs_alert_forward.adoc
jacobbaungard: sure, just adjust it to correct wording. Most important is the information in the additional couple of sentences are conveyed as that info was dropped from the original text.

observability/obs_alert_forward_mcoa.adoc
dockerymick: When you say coded are you referring to it being added to the conref file?

observability/obs_alert_forward_mcoa.adoc
dockerymick: It was a hidden comment in the original issue. The YAML is really long so I pulled explanation and used it as the purpose for specific parts of the YAML

observability/obs_alert_forward.adoc
dockerymick:
```suggestion
Configure alert forwarding for the multicluster observability add-on to send alerts to the Alertmanager of the Observability stack from your managed clusters. By default, multicluster observability add-on is not configured to send alerts from Prometheus on your managed clusters to the hub cluster Alertmanager.
```

observability/obs_alert_forward.adoc
dockerymick: Thanks Brandi. Yup, this is just my attempt to break up the long YAML a bit.

observability/obs_alert_forward_mcoa.adoc
dockerymick: There is a command on line 125-128. At first the following statement was a note: "If you make custom changes to the `additionalAlertmanagerConfigs` specification of the config map, be sure to add the changes directly into the policy, before applying the policy."

If the note is something the user needs to consider before running the command, it should be included in the step. Or i can just add as a note like before. let me know your thoughts

observability/obs_alert_forward.adoc
dockerymick: Heads up, i updated your suggestion

observability/obs_alert_forward.adoc
dockerymick: ack

observability/obs_alert_forward.adoc
dockerymick: similar to my response. This was my attempt to break up the YAML. The hidden comment from the original YAML mentioned  "Defining what you want your configuration to include.."

observability/obs_alert_forward.adoc
dockerymick: ack, will update this soon

observability/obs_alert_forward.adoc
dockerymick: ack, updating

observability/obs_alert_forward.adoc
jacobbaungard: Btw @swopebe , I really appreciate the explanation of why it doesn't match the guidelines - that is very helpful so we can learn along the way and try to provide the docs in a way that closer matches the guidelines next time :1st_place_medal:

observability/obs_alert_forward_mcoa.adoc
```diff
@@ -0,0 +1,225 @@
+[#forward-alert-mcoa]
+= Forwarding alerts for multicluster observability add-on
+
+Configure alert forwarding for the multicluster observability add-on to send alerts to the Alertmanager of the Observability stack from your managed clusters. By default, multicluster observability add-on is not configured to send alerts from Prometheus on your managed clusters to the hub cluster Alertmanager.
+
+Create a policy to configure alert forwarding for the multicluster observability add-on.
```
jacobbaungard:
```suggestion
To setup alert forwarding when using the multicluster observability add-on, you can create a policy.
```

Would prefer it like this, but it's a small detail, and maybe it better matches guidelines as is, so I'll leave it to you to decide.

observability/obs_alert_forward_mcoa.adoc
jacobbaungard: I think perhaps we should move this paragraph a bit. Either before the initial yaml at L22 or just after the yaml at L114

jacobbaungard (commented):
Just a couple of minor comments, looks good generally!

Also just to confirm, in terms of where this actually shows up in the documentation, it's directly after the remainder of the MCOA documentation right?

observability/obs_alert_forward_mcoa.adoc
swopebe: no, is the add on name supposed to be coded in the titles and such...I am not sure bc I didn't review this work.

observability/obs_alert_forward_mcoa.adoc
jacobbaungard: To clarify

> If you make custom changes to the `additionalAlertmanagerConfigs` specification of the config map, be sure to add the changes directly into the policy, before applying the policy.

Could be moved, but not the part about applying the yaml / running the command.

observability/obs_alert_forward_mcoa.adoc
swopebe: The way it is written, yes it matches the guidelines for conciseness and imperative language. The point is to start with the action first and not start with a conditional phrase or any other phrase. 

Your way is not wrong, we are just encouraged to lead with the action first. @jacobbaungard -- so I am not against it 100%--it's just not leading with the verb "Create," which is the point of the step.

Does this work?

_Create a policy to configure alert forwarding when you use the multicluster observability add-on._

jacobbaungard: Yes that is also fine :+1:

observability/obs_alert_forward_mcoa.adoc
swopebe: The step itself is "apply" and the user needs to know how to do that right away as part of the step. The info is important and can be presented before or after the step. But what is the actual step here? It is "Apply".

dockerymick:
@jacobbaungard thanks Jacob, i made updates.

observability/obs_alert_forward_mcoa.adoc
swopebe: Long YAML is OK if it doesn't require many steps/procedures in the YAML. We just need to pull out what requires a procedure:

Example:

. Add the ___ specification for the ___ to your __ resource. Change the ___ value to your specific namespace. See the following YAML sample:

```
insert that piece of the yaml
```
. Run the following command to apply your YAML file.

```
oc apply ___
```

This is just an example of how/why we would break out a large YAML sample to build a procedure.

observability/obs_alert_forward.adoc
swopebe: Sure thing! We try to get mutual understanding where there is time.

I agree that _might_ is not always the best way to show an example so we can say:

**See the following example:**

That makes it imperative and takes out room for interpretation.

observability/obs_alert_forward.adoc
swopebe: ok, we have a little work to do here to change without changing the meaning, but I think we can do it pretty easily.

dockerymick:
> Just a couple of minor comments, looks good generally!
> 
> Also just to confirm, in terms of where this actually shows up in the documentation, it's directly after the remainder of the MCOA documentation right?

Yes you are correct

observability/obs_alert_forward_mcoa.adoc
```diff
@@ -0,0 +1,228 @@
...
+            recordDiff: InStatus
+  remediationAction: inform
+----
+
++
+*Note:* The `remediationAction` of the previous policy is set to `inform` mode. From the console, validate your changes by viewing the policy from the *Governance* page. Your policy appears as non-compliant.
+
+.. To apply the policy changes across all managed clusters and enable alert forwarding from your managed clusters to the hub cluster, change the `remediationAction` parameter value to `enforce`. 
+
+.. Create a `PlacementBinding` resource to bind the policy to the `global` placement. Your YAML file might resemble the following sample:
```
jacobbaungard: I'm a bit unsure, if it's better to move the placement binding steps, directly above in L92.

The reason being:
- We add the content to the existing file (maybe we can clarify that also)
- The actions the note on line L94 will only happen after both the policy and the placementbinding is applied

Would that make sense ?

dockerymick: Yea that makes sense. I will move the note. it will be after the PlacementBinding step

jacobbaungard (approved):
Thanks for the effort with this, looks great!

/lgtm

thibaultmg (approved):
/lgtm

swopebe (approved):

swopebe:
This looks great. Approved. This can make an immediate refresh.

swopebe (approved):

observability/obs_alert_forward_mcoa.adoc
dockerymick: no it shouldn't be coded or monospaced

dockerymick:
Thanks everyone for your diligence and effort during this review!

PR #8436, opened 2025-12-05
https://issues.redhat.com/browse/ACM-27270

swopebe:
Added known issues related to Submariner installation and source IP retention for specific OCP versions.

### DIFF
--- a/release_notes/known_issues_network.adoc
+++ b/release_notes/known_issues_network.adoc
@@ -15,7 +15,7 @@ See the following known issues and limitations that might occur while using netw
 [#ocp-acm-submariner-fails]
 === Submariner installation with {ocp-short} on IBM Power Systems fails
 
-If you are using {ocp-short} version 4.20 and {acm-short} 2.15, you cannot install Submariner on IBM systems, as it is currently not supported for these versions.
+If you are using {acm-short} 2.15, you cannot install Submariner on IBM systems, as it is currently not supported for these versions.
 
 link:https://issues.redhat.com/browse/ACM-27270[ACM-27270]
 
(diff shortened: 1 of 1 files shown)


dockerymick (approved):
Looks good thanks! Checklist considered

release_notes/known_issues_network.adoc
```diff
@@ -12,6 +12,13 @@ For more about deprecations and removals, see xref:../release_notes/acm_deprecat
 
 See the following known issues and limitations that might occur while using networking features.
 
+[#ocp-acm-submariner-fails]
+=== Submariner installation with {ocp-short} on IBM Power Systems fails
```
vthapar:
```suggestion
=== Submariner installation on IBM Power Systems fails
```

release_notes/known_issues_network.adoc
```diff
@@ -12,6 +12,13 @@ For more about deprecations and removals, see xref:../release_notes/acm_deprecat
...
+
+If you are using {ocp-short} version 4.20 and {acm-short} 2.15, you cannot install Submariner on IBM systems, as it is currently not supported for these versions.
```
vthapar:
```suggestion
If you are using {acm-short} 2.15, you cannot install Submariner on IBM systems, as it is currently not supported for these versions.
```

vthapar (changes requested):
Since it is an arch issue with builds, it will fail for all OCP versions. IBM team just tested with 4.20 only so that was mentioned in discussion and Jira so far.

release_notes/known_issues_network.adoc
bkhadars: We need to add IBM Z as well

vthapar (approved):

PR #8442, opened 2025-12-09
https://issues.redhat.com/browse/ACM-24626 Customer bug

oafischer:
2.12 to 2.15

Checklist considered


rokej (commented):
https://github.com/stolostron/rhacm-docs/pull/8442 is not needed for ACM 2.15 but needed for ACM 2.14 ~ 2.12. For 2.15, it is not needed because of https://issues.redhat.com/browse/ACM-24952 that is going to automate this additional step.

PR #8472, opened 2025-12-16
https://issues.redhat.com/browse/ACM-24786\#

dockerymick:


release_notes/known_issues_network.adoc
```diff
@@ -38,6 +38,30 @@ See the following known issues and limitations that might occur while using netw
 
 If you are using {ocp-short} versions between 4.18 and versions earlier than 4.19.5 with OVN-Kubernetes for Submariner, the source IP is not retained when packets reach the destination pod. As a result, applications that rely on the source IP, such as `NetworkPolicy`, might not work correctly.
 
+To preserve your source IP for hub cluster traffic on the OVN Kubernetes versions, enable the Submariner SNAT handler. Complete the following steps:
+
+. Create a `ConfigMap` resource named `submariner-global` in the `submariner-operator` namespace with the following configuration. Create the following resource to configure Submariner to manage SNAT and override the default behavior of OVN-Kubernetes:
```
oafischer: Is there a way to not use create twice in the same step? At first glance it might seem like these are two steps

oafischer (approved):
One comment, otherwise approved!

release_notes/known_issues_network.adoc
```diff
@@ -38,6 +38,30 @@ See the following known issues and limitations that might occur while using netw
 
 If you are using {ocp-short} versions between 4.18 and versions earlier than 4.19.5 with OVN-Kubernetes for Submariner, the source IP is not retained when packets reach the destination pod. As a result, applications that rely on the source IP, such as `NetworkPolicy`, might not work correctly.
 
+To preserve your source IP for hub cluster traffic on the OVN Kubernetes versions, enable the Submariner SNAT handler. Complete the following steps:
```
yboaron: The workaround is for traffic between managed clusters and not for " hub cluster traffic" , 
so it should be something like : " To preserve your source IP for inter-cluster traffic on OVN-Kubernetes, enable the Submariner Avoid SNAT handler on all managed clusters."

release_notes/known_issues_network.adoc
```diff
@@ -38,6 +38,30 @@ See the following known issues and limitations that might occur while using netw
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+ {{ name: submariner-global}}
+  namespace: submariner-operator
+data:
+  {{ enable-snat-handler: "true"}}
+----
+
+. Restart the `route-agent` pods apply the `ConfigMap` resources changes. Run the following command:
```
yboaron: Maybe something like: "restart of the route-agent pods is required for the new ConfigMap configuration to take effect. Run the ...."

release_notes/known_issues_network.adoc
dockerymick: what does "inter-cluster" mean?

release_notes/known_issues_network.adoc
dockerymick: Thanks. I recommend that we not say "is required". This helps to reduce wordiness. Since we are starting the step with a direct verb, it indicates that the user must complete this part of the procedure. If a step is optional, we add the "Optional" label in the step.

release_notes/known_issues_network.adoc
dockerymick: I googled it and searched for it in the IBM style guide. "inter-cluster" is not listed so I added per your request.

yboaron (approved):

PR #8506, opened 2026-01-12
https://issues.redhat.com/browse/ACM-26413 Console KI

oafischer:
Also for 2.14 and 2.13

### DIFF
--- a/release_notes/known_issues_console.adoc
+++ b/release_notes/known_issues_console.adoc
@@ -7,10 +7,17 @@ For your {ocp} cluster, see link:https://docs.redhat.com/en/documentation/opensh
 
 For more about deprecations and removals, see xref:../release_notes/acm_deprecate_remove.adoc#deprecations-removals-acm[Deprecations and removals for {acm-short}].
 
-[#follow-external-link-wrong-page]
-== Following an external link opens the wrong page
+[#external-link-console-wrong-page]
+== External links in certain versions of the integrated console lead to incorrect pages
 
-If you follow an external link that leads to the console, open a link in a new tab, enter a URL manually, or refresh a page, part of the URL might be lost. As a result, you see the wrong console page. For example, a link to the *Governance Policies* tab might open the *Governance Overview* tab instead.
+If your {ocp} version is 4.18.29 or older, 4.19.20 or older, or 4.20.7 or older, completing one of the following actions causes part of the URL to be lost:
+
+* Opening a link that leads to the console
+* Opening a console link in a new tab
+* Entering a console URL manually
+* Refreshing a console page
+
+As a result, you see the wrong console page. For example, a link to the *Governance Policies* tab might open the *Governance Overview* tab instead.
 
 To resolve the issue, upgrade to {ocp} version 4.18.30, 4.19.21, or 4.20.8 or later.
 
--- a/release_notes/known_issues_console.adoc
+++ b/release_notes/known_issues_console.adoc
@@ -21,7 +21,7 @@ As a result, you see the wrong console page. For example, a link to the *Governa
 
 To resolve the issue, upgrade to {ocp} version 4.18.30, 4.19.21, or 4.20.8 or later.
 
-Alternatively, you might be able to access the correct page if you reload the page. You can also navigate to the correct page from the console by following the appropriate links in the navigation menu, or by using the search function.
+Alternatively, you can reload to try to access the correct page. You can also navigate to the correct page from the links in the console navigation menu, or use the search function.
 
 link:https://issues.redhat.com/browse/ACM-24689[ACM-24689]
 
--- a/release_notes/known_issues_console.adoc
+++ b/release_notes/known_issues_console.adoc
@@ -7,7 +7,7 @@ For your {ocp} cluster, see link:https://docs.redhat.com/en/documentation/opensh
 
 For more about deprecations and removals, see xref:../release_notes/acm_deprecate_remove.adoc#deprecations-removals-acm[Deprecations and removals for {acm-short}].
 
-[#external-link-console-wrong-page]
+[#external-link-wrong-page]
 == External links in certain versions of the integrated console lead to incorrect pages
 
 If your {ocp} version is 4.18.29 or older, 4.19.20 or older, or 4.20.7 or older, completing one of the following actions causes part of the URL to be lost:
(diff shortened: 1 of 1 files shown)


release_notes/known_issues_console.adoc
```diff
@@ -7,6 +7,17 @@ For your {ocp} cluster, see link:https://docs.redhat.com/en/documentation/opensh
 
 For more about deprecations and removals, see xref:../release_notes/acm_deprecate_remove.adoc#deprecations-removals-acm[Deprecations and removals for {acm-short}].
 
+[#follow-external-link-wrong-page]
+== Following an external link opens the wrong page
+
+If you follow an external link that leads to the console, open a link in a new tab, enter a URL manually, or refresh a page, part of the URL might be lost. As a result, you see the wrong console page. For example, a link to the *Governance Policies* tab might open the *Governance Overview* tab instead.
+
+To resolve the issue, upgrade to {ocp} version 4.20.8 or later.
```
KevinFCormier: It's fixed in earlier releases than this... I would list all 3 versions.

oafischer: This is for 2.15 doc, we need separate PRs for the other versions since they are in other branches. I'll list the other OCP versions in each of those.

KevinFCormier: ACM supports OCP n-2, and customers are often not on the latest OCP release. So ACM 2.15 is supported with OCP 4.18. I don't want to give the impression that they need OCP 4.20 to get this fix.

oafischer: I see, thanks for explaining. I'll add the other versions!

KevinFCormier (approved):

KevinFCormier (approved):

Randy424 (approved):
/lgtm

release_notes/known_issues_console.adoc
```diff
@@ -7,6 +7,17 @@ For your {ocp} cluster, see link:https://docs.redhat.com/en/documentation/opensh
 
 For more about deprecations and removals, see xref:../release_notes/acm_deprecate_remove.adoc#deprecations-removals-acm[Deprecations and removals for {acm-short}].
 
+[#follow-external-link-wrong-page]
+== Following an external link opens the wrong page
```
swopebe:
```suggestion
== External links to certain versions of the integrated console lead to incorrect pages
```

swopebe: This is a bit confusing to me. Following an external link.... any external link, and external link from where???

release_notes/known_issues_console.adoc
```diff
@@ -7,6 +7,17 @@ For your {ocp} cluster, see link:https://docs.redhat.com/en/documentation/opensh
...
+
+If you follow an external link that leads to the console, open a link in a new tab, enter a URL manually, or refresh a page, part of the URL might be lost. As a result, you see the wrong console page. For example, a link to the *Governance Policies* tab might open the *Governance Overview* tab instead.
```
swopebe: Do you mean "open a new tab" then enter the URL?

Open an link in a new tab... what link???

release_notes/known_issues_console.adoc
oafischer: Should be any external link that leads to the console. Could be from anywhere.

Can you confirm @KevinFCormier ?

swopebe: Now that I read the example, I am not sure where the "external" comes from.

But it seems we need to make it clearer 1. The OCP Version and 2. The the tabs in the console are what open to incorrect pages.

I find this quite confusing as of now.

release_notes/known_issues_console.adoc
oafischer: Yes, opening a new tab and entering the URL.

The link is any link that leads to the console, not a specific link.

release_notes/known_issues_console.adoc
oafischer: The internal links should work I think. Meaning the ones in the console navigation.

I'll add the OCP versions that are affected.

The tabs in the console are just one of several ways to reach incorrect pages from what I understand.

release_notes/known_issues_console.adoc
swopebe: "An external link to the integrated console leads to incorrect links within the console"?

An external link that leads to the console -- this seems broad. 

And I think we need to say early that it is only certain versions.

release_notes/known_issues_console.adoc
swopebe: Can we shorten this:

[#external-link-wrong-page]

release_notes/known_issues_console.adoc
swopebe: When reading that first line, I felt it was a series of steps. 

When you do this, then this, then this. Some changes, maybe?

_____

If you follow an external link that leads to **certain versions** of the console, open an external link in a new tab, enter a URL **to the console manually**, or refresh a page t**o the console,** part of the URL might be lost.

____

It doesn't have to be this but seeing if we can make it clearer. This call also just be bullets.

release_notes/known_issues_console.adoc
```diff
@@ -7,6 +7,17 @@ For your {ocp} cluster, see link:https://docs.redhat.com/en/documentation/opensh
...
+
+To resolve the issue, upgrade to {ocp} version 4.18.30, 4.19.21, or 4.20.8 or later.
+
+Alternatively, you might be able to access the correct page if you reload the page. You can also navigate to the correct page from the console by following the appropriate links in the navigation menu, or by using the search function.
```
swopebe:
```suggestion
Alternatively, you can reload to try to access the correct page. You can also navigate to the correct page from the links in the console navigation menu, or use the search function.
```

swopebe: I think Search might be capitalized?

release_notes/known_issues_console.adoc
oafischer: I changed it to bullets and did something similar to your suggestion. I think it would be too long without the bullets. Let me know what you think.

release_notes/known_issues_console.adoc
oafischer: I've seen both in the doc. Not sure tbh. I applied the other suggestions.

I think if we refer to the search function as a generic function, I'd probably leave it lower case. But if we refer to the component/feature, perhaps we should capitalize?

release_notes/known_issues_console.adoc
oafischer: I added the specific OCP versions that are affected

swopebe (approved):

PR #8509, opened 2026-01-13
ACM-28519: Provide additional context for `objectSelector`

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-28519


dockerymick (approved):
Seems like you only added `object-templates`. This looks good to me!
this was only an update to yaml examples. From my perspective, this does not require a peer review.

PR #8550, opened 2026-02-02
https://issues.redhat.com/browse/ACM-28533 Update APIs

oafischer:
2.16

### DIFF
--- a/apis/clustermanagementaddon.json.adoc
+++ b/apis/clustermanagementaddon.json.adoc
@@ -1,5 +1,5 @@
 [#clustermanagementaddon-api]
-= ClusterManagementAddOn API (v1alpha1)
+= ClusterManagementAddOn API (v1beta1)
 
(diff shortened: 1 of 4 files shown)


apis/clustermanagementaddon.json.adoc
```diff
@@ -1,5 +1,5 @@
 [#clustermanagementaddon-api]
-= ClusterManagementAddOn API (v1alpha1)
+= ClusterManagementAddOn API (v1beta1)
```
haoqing0110, line 2: The v1alpha1 and v1beta1 are both served in 2.16, can we add both v1alpha1 and v1beta1 API in doc?

oafischer: @haoqing0110 Is there a JIRA or some background info available for this? We have never had the situation where both versions of the API are served. Usually we have to deprecate alpha in the doc, which I actually forgot to do.

Can you confirm that v1alpha1 is NOT deprecated and NOT removed?

PR #8552, opened 2026-02-03
https://issues.redhat.com/browse/ACM-27966 Auto import doc update

oafischer:
2.14, 2.15, 2.16


birsanv:
thank you for the update @oafischer 

I had to refresh my memory on the issue so I went back to the document.
The things I think we need to cover and are still missing are : 
1. What happens if you the set ImportOnly option ( what are the consequences )
2. Why would you want to set the ImportOnly 

For 1, this is what I have in the document
With this option, once a managed cluster joins the hub, the import-controller stops applying the manifests and **an auto import operation is skipped.**  @elgnay I see the ImportOnly option is explained but I couldn't get from that description that the auto import option will be skipped .. and this was the issue we encountered when we run the restore

For 2, I understand we can't mention primary or secondary but we still need to explain why I would think of using the ImportOnly vs ImportAndSync

@elgnay what about , updating the doc content  : 
The reason for this change is to support situations when the primary hub goes down uncontrolled and comes back again after the clusters have moved to another hub. In this case the default auto import strategy stops the primary hub to try to recover the managed clusters.
to : 
_The reason for this change is to support the situation when a hub goes down uncontrolled and comes back again after the managed clusters have moved to another hub. In this case the ImportOnly strategy stops this hub from trying to recover the managed clusters. This is a common scenario for a hub disaster recovery when managed clusters are moved to a new hub once the current hub goes down unexpectedly._

elgnay:
@birsanv 

For 1, we can replace 
```
After the cluster joins the hub cluster, the hub controller stops applying klusterlet manifests. 
```
with
```
Once the managed cluster successfully joins the hub cluster and the `ManagedClusterImportSucceeded` condition is set to `True`, the hub controller stops applying klusterlet manifests and skips any subsequent automatic import operations for this cluster.
```
to explicitly state that auto-import is skipped

And for 2, your proposed change looks good to me.

birsanv:
hi @elgnay your proposed change here looks good, thanks https://github.com/stolostron/rhacm-docs/pull/8552#issuecomment-3845105694

oafischer:
Thanks @birsanv @elgnay, I've updated the PR with both of your suggestions. Let me know if it looks good now.

birsanv:
hi @oafischer part 1 is good
For part 2, I know you are trying to reduce the number of words but the result is not clear to me. Your doc says : 

_The ImportOnly strategy is useful for a hub cluster disaster recovery scenario, where managed clusters **move to a new hub cluster after the current hub cluster fails and restarts unexpectedly**, by preventing the hub cluster from recovering the managed clusters._

How I read it : 
ImportOnly is useful when managed clusters move to a new hub after the current one fails and restarts unexpectedly

It should say :   
The ImportOnly strategy is useful for a hub cluster disaster recovery scenario to prevent the initial hub cluster from recovering the managed clusters, if the initial hub cluster restarts unexpectedly after the managed clusters were moved to another hub 

something on this line ( I actually prefer more the version from https://github.com/stolostron/rhacm-docs/pull/8552#issuecomment-3842393327 because here we start using the initial hub cluster notion which I am not sure is clear enough in this short sentence )

One more comment : 
I see you put this info in the same paragraph with the upgrade info ( If you are using {mce} 2.9 or later, the default automatic import strategy is ImportOnly ... ) 
Has nothing to do with that, I would either move it to the ImportOnly section or make a new paragraph for it

birsanv (approved):
lgtm

oafischer:
@birsanv Thanks, I moved that line and used your suggestion. @elgnay Let me know if it looks good to you as well.

elgnay:
It looks good to me as well. Thanks!

clusters/cluster_lifecycle/custom_import_strat.adoc
```diff
@@ -5,12 +5,14 @@
...
+If a managed cluster is missing the `ManagedClusterImportSucceeded` condition, or the condition is not set to `True`, the hub cluster controller initiates the import by applying the klusterlet manifests to the managed cluster. After the managed cluster joins the hub cluster and the `ManagedClusterImportSucceeded` condition is set to `True`, the hub cluster controller stops applying klusterlet manifests and skips automatic import operations for the cluster. Now, only the klusterlet agent on the managed cluster keeps the hub configuration synchronized. If the klusterlet agent stops working, the manifests might become outdated, and might cause the managed cluster to appear in an `Unknown` state on the hub cluster. You must then xref:../cluster_lifecycle/import_cli.adoc#importing-cluster-manual[import the cluster manually]. The `ImportOnly` strategy is useful for a hub cluster disaster recovery scenario where you want to prevent the initial hub cluster from recovering the managed clusters, if the initial hub cluster restarts unexpectedly after the managed clusters were moved to a different hub cluster.
```
dockerymick: Suggesting to break this up for readability. I also suggest to be more concise when you explain the boolean condition. Also made suggestions related to maintaining an active voice and being direct. Let me know your thoughts.

```suggestion
If a managed cluster is missing the `ManagedClusterImportSucceeded` condition, or the condition is set to `False`, the hub cluster controller applies the klusterlet manifests to the managed cluster to initiate the import.

After the managed cluster joins the hub cluster and the `ManagedClusterImportSucceeded` condition is set to `True`, the hub cluster controller stops applying klusterlet manifests and skips automatic import operations for the cluster. From this point, only the klusterlet agent on the managed cluster synchronizes the hub cluster configuration.

If the klusterlet agent fails, the manifests can become outdated, and might cause the managed cluster to appear in an `Unknown` state on the hub cluster. You must then xref:../cluster_lifecycle/import_cli.adoc#importing-cluster-manual[import the cluster manually].

Use the `ImportOnly` strategy for a hub cluster disaster recovery scenario where you want to prevent the initial hub cluster from recovering the managed clusters, if the initial hub cluster restarts unexpectedly after the managed clusters were moved to a different hub cluster.
```

dockerymick (changes requested):

oafischer:
Thanks @dockerymick, I applied your suggestion!

dockerymick (approved):
Looks good to me! Thanks for accepting my suggestions, checklist considered

PR #8557, opened 2026-02-04
https://issues.redhat.com/browse/ACM-25617 for 2.16

dockerymick:
@birsanv Here is more info from the issue description: "On the restore hub cluster, install the operator at the same version as the hub cluster where the backup was created". Which Operator do you refer to? Is it the OADP operator?"

From my perspective, the ACM operator is what we were referring to. Should we refer to a different operator?


birsanv:
@dockerymick yes, we are referring to the ACM operator

birsanv (approved):
lgtm

oafischer (approved):

PR #8559, opened 2026-02-05
https://issues.redhat.com/browse/ACM-27984--release note

swopebe:


clusters/release_notes/mce_errata.adoc
```diff
@@ -7,6 +7,17 @@ If no release notes are listed, the product does not have an Errata release at t
...
 *Important:* For reference, Jira links and Jira numbers might be added to the content and used internally. Links that require access might not be available for the user.
 
+== Errata 2.8.4
+
+* Delivers updates to one or more product container images.
+
+* Fixes a YAML generation problem that prevented the default network attachment from disabling during hosted clusters deployment on {ocp-virt-short}.
+link:https://issues.redhat.com/browse/ACM-24960[ACM-24960]
+
+* Improves performance and increases timeouts that caused by a `CrashLoopBackOff` condition on the pods during deployment. link:https://issues.redhat.com/browse/ACM-24721[ACM-24721]
```
KevinFCormier: I had provided the following release note text:
> Reduces the likelihood of the console becoming unavailable due to a CrashLoopBackOff condition on pods of the deployment

The "during deployment" phrase is not accurate, and it is not clear which pods we are talking about with the current text.

swopebe: Ok--we cannot write "reduces the likelihood" -- the is exactly why I changed it.

Further, we need to avoid "of the" as much as we can. 

We can say "deployment pods" if that works. @KevinFCormier

swopebe: Improves performance and increases timeouts to prevent.... also an option?

KevinFCormier: @swopebe OK, how about the following. (I don't think we need to get into the technical details of what was changed, but more on the end result for the user.)

> Addresses an issue that might cause the console to become unavailable because of a CrashLoopBackOff condition

swopebe: It's due to feedback from PM, and I believe Deutsche bank, if I remember correctly. They wanted more specifics on these. In many cases the bugs do not offer this info at first glance.

KevinFCormier: Well, slightly more detailed if needed:
> Addresses an issue that might cause the console to become unavailable because of a CrashLoopBackOff condition by improving performance and increasing probe timeouts on the console-mce-console deployment

clusters/release_notes/mce_errata.adoc
```diff
@@ -7,6 +7,17 @@ If no release notes are listed, the product does not have an Errata release at t
 
 *Important:* For reference, Jira links and Jira numbers might be added to the content and used internally. Links that require access might not be available for the user.
 
+== Errata 2.8.4
+
+* Delivers updates to one or more product container images.
+
+* Fixes a YAML generation problem that prevented the default network attachment from disabling during hosted clusters deployment on {ocp-virt-short}.
+link:https://issues.redhat.com/browse/ACM-24960[ACM-24960]
```
KevinFCormier: This looks great.

clusters/release_notes/mce_errata.adoc
swopebe: even better!

KevinFCormier (approved):

clusters/release_notes/mce_errata.adoc
```diff
@@ -7,6 +7,17 @@ If no release notes are listed, the product does not have an Errata release at t
...
+
+* Addresses an issue that might cause the console to become unavailable because of a `CrashLoopBackOff` condition by improving performance and increasing probe timeouts on the `console-mce-console` deployment. link:https://issues.redhat.com/browse/ACM-24721[ACM-24721]
+
+* Corrects the problem with the operator installation window missing the _Repository_ and _Source Code_ link for this version of {mce-short}. The expected repository link is link:https://github.com/open-cluster-management/backplane-operator[backplane-operator]. link:https://issues.redhat.com/browse/ACM-22364[ACM-22367]
```
ngraham20, line 19: lgtm!

dockerymick (approved):

PR #8596, opened 2026-02-16
https://issues.redhat.com/browse/ACM-30068--to ensure the dev team is…

swopebe:
… looking at older GH issues, we needed to name them this first release that we changed the linking.


dockerymick (approved):
added a commit to fix the linter. This looks good to me

PR #8600, opened 2026-02-19
https://issues.redhat.com/browse/ACM-29060

oafischer:
2.16

### DIFF
--- a/clusters/cluster_lifecycle/ansible_config_hosted_cluster.adoc
+++ b/clusters/cluster_lifecycle/ansible_config_hosted_cluster.adoc
@@ -5,13 +5,11 @@
 
 *Required access:* Cluster administrator
 
-* <<prerequisites-for-ansible-integration-hosted-cluster,Prerequisites>>
 * <<ansible-hosted-cluster-install,Running an {aap-short} job to install a hosted cluster>>
 * <<ansible-hosted-cluster-update,Running an {aap-short} job to update a hosted cluster>>
 * <<ansible-hosted-cluster-delete,Running an {aap-short} job to delete a hosted cluster>>
 
-[#prerequisites-for-ansible-integration-hosted-cluster]
-== Prerequisites
+.Prerequisites
 
 You must meet the following prerequisites to run Automation templates on your clusters:
 
--- a/clusters/cluster_lifecycle/ansible_config_hosted_cluster.adoc
+++ b/clusters/cluster_lifecycle/ansible_config_hosted_cluster.adoc
@@ -145,26 +145,6 @@ spec:
       - my-nodepool-2
 ----
 
-== Running an {aap-short} job to update the control plane and specific node pools
-
-To run an {aap-short} job that updates the control plane and specific node pools, edit the `nodePoolNames` field in your `ClusterCurator` resource of the hosted cluster that you want to update. See the following example:
-
-[source,yaml]
-----
-apiVersion: cluster.open-cluster-management.io/v1beta1
-kind: ClusterCurator
-metadata:
-  name: my-hosted-cluster
-  namespace: clusters
-spec:
-  desiredCuration: upgrade
-  upgrade:
-    desiredUpdate: "4.17.1"
-    nodePoolNames:
-      - my-nodepool-1
-      - my-nodepool-2
-----
-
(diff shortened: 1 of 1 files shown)


clusters/cluster_lifecycle/ansible_config_hosted_cluster.adoc
```diff
@@ -127,7 +125,158 @@ spec:
...
+  desiredCuration: upgrade
+  upgrade:
+    desiredUpdate: "4.17.1"
+    upgradeType: NodePools
+    nodePoolNames:
+      - my-nodepool-1
+      - my-nodepool-2
+----
+
+== Running an {aap-short} job to update specific node pools
```
rokej: Isn't it a duplicate of the previous section?

clusters/cluster_lifecycle/ansible_config_hosted_cluster.adoc
```diff
@@ -127,7 +125,158 @@ spec:
...
+spec:
+  desiredCuration: upgrade
+  upgrade:
+    desiredUpdate: "4.17.1"
+    nodePoolNames:
+      - my-nodepool-1
+      - my-nodepool-2
+----
+
+== Running an {aap-short} job to update the control plane and specific node pools
```
rokej: Isn't it a duplicate of the previous block?

rokej (changes requested):

clusters/cluster_lifecycle/ansible_config_hosted_cluster.adoc
oafischer: Yes must have been copy paste error, fixed both instances!

rokej (approved):

clusters/cluster_lifecycle/ansible_config_hosted_cluster.adoc
```diff
@@ -127,7 +125,117 @@ spec:
...
+apiVersion: cluster.open-cluster-management.io/v1beta1
+kind: ClusterCurator
+metadata:
+  name: my-hosted-cluster
+  namespace: clusters
+spec:
+  desiredCuration: upgrade
+  upgrade:
+    desiredUpdate: "4.17.1"
+    channel: "fast-4.17"
```
dockerymick, line 237: Based on how the file is organized, this looks fine to me. However, it seems like the sections can be steps. Making them steps will reduce the number of H2 headers. Let me know what you think. It could just be legacy content and maybe something that we tackle as we prep for migration.

dockerymick (changes requested):

clusters/cluster_lifecycle/ansible_config_hosted_cluster.adoc
swopebe: I agree with this, as we need to try to get away from adding headers where we can. You can merge the header and the lines before to make a step and some substeps--or if there is only one step, just merge the header two with the intro paragraph: 

**example:**

```
. Run an {aap-short} job to update the cluster version and channel.
.. Add the cluster version to the channel to the `desiredUpdate` field.
.. Add the cluster channel to the `channel` field.

+
See the following example:
```

oafischer: Yeah I'd probably reorganize everything here once we start converting this book. Definitely a bunch of legacy content here.

oafischer: Or should I start converting this file now?

dockerymick: I'm not sure. If it is short go for it. I would leave as is until we are ready to convert

oafischer: I have some other higher priority items than need attending first, I could probably get this in this week or next though. @swopebe What do you think?

swopebe: Push it, we cannot reorg this right now, noted as one we have to change.

dockerymick (approved):

swopebe (approved):

PR #8612, opened 2026-02-25
https://issues.redhat.com/browse/ACM-29692

dockerymick:
Removed the following 

- anthropomorphism: "Controllers watching... interpret" gives the software a human thought process. 

- Passive-adjacent/Wordy: "clusters are selected by..." 

- "May" and "Can": we need to describe the behavior as a fact when it is a known issue


clusters/release_notes/mce_known_issues.adoc
```diff
@@ -37,6 +37,14 @@ link:https://issues.redhat.com/browse/ACM-7523[ACM-7523]
 
 Learn about Known issues for Cluster lifecycle with {mce-short}, such as issues with creating, discovering, importing, and removing clusters, and more cluster management issues for {mce-short}.
 
+=== Resource deletion during cluster distribution updates
+
+When cluster redistribution occurs across multiple `PlacementDecisions` resources, adding or removing clusters results in a shift between decisions. Because, the update mechanism is sequential, a cluster is temporarily not exist in any `PlacementDecision` resources during the update.
```
haoqing0110: When clusters are distributed across multiple PlacementDecision resources, adding or removing clusters can trigger a redistribution. Due to the sequential update mechanism, a cluster may temporarily not exist in any PlacementDecision during this shift.
（The logic is adding/removing clusters -> triggers cluster redistribution -> the sequential update mechanism has an issue)

clusters/release_notes/mce_known_issues.adoc
```diff
@@ -37,6 +37,14 @@ link:https://issues.redhat.com/browse/ACM-7523[ACM-7523]
 
 Learn about Known issues for Cluster lifecycle with {mce-short}, such as issues with creating, discovering, importing, and removing clusters, and more cluster management issues for {mce-short}.
 
+=== Resource deletion during cluster distribution updates
+
+When you add or remove clusters, cluster redistribution occurs across multiple `PlacementDecisions` resources, resulting in a shift between decisions. Because the update mechanism is sequential, a cluster temporarily does not exist in any `PlacementDecision` resources during the update.
```
haoqing0110: On second thought, this part still needs a bit more work.
It should be:
When clusters are added or removed from `PlacementDecisions`, cluster redistribution occurs across multiple `PlacementDecisions` resources ...

----
The current wording implies that the user manually added or deleted clusters and caused the issue. Actually, this is triggered by changes in the clusters selected by the Placement. For example, if a cluster is tainted, it gets filtered out of the placement decision, causing redistribution across multiple PlacementDecisions.

dockerymick: Thanks for that clarity! In my following suggestion, I am ensuring that we use an active tone and make sure we focus on describing the status of cluster and resources. 

When a change occurs in the cluster because of the `Placement` resource, cluster redistribution occurs across multiple `PlacementDecisions` resources. During the sequential update, a cluster is temporarily missing from all `PlacementDecisions` resources.

oafischer (approved):

PR #8615, opened 2026-02-26
https://issues.redhat.com/browse/ACM-29846 for 2.16

dockerymick:


search/search_klusterlet_addon.adoc
```diff
@@ -26,7 +26,45 @@ metadata:
...
+    resources:
+      limits:
+        memory: 1099Mi
+      requests:
+        memory: 199Mi
+----
 
-*Note:* You can discover all resources defined in your managed cluster by using the API Explorer in the console. Alternatively, you can discover all resources by running the following command: `oc api-resources`
\ No newline at end of file
+.. Add the `backup` label to include the `search-addon-resources` config map to the cluster backups. Run the following command:
```
smcavey: It is a kind `AddOnDeploymentConfig`, not a ConfigMap

search/search_klusterlet_addon.adoc
```diff
@@ -26,7 +26,45 @@ metadata:
...
+  name: search-addon-resources
+  namespace: your-managed-ns
+spec:
+  resourceRequirements:
+  - containerID: deployments:klusterlet-addon-search:collector
+    resources:
+      limits:
+        memory: 1099Mi
+      requests:
+        memory: 199Mi
```
smcavey: Maybe we should give better numbers for examples like 1024Mi and 512Mi for limits and requests, respectively, these numbers came from using intentionally odd values to make it very clear this functionality worked when developing

smcavey:
Lastly, do you think we should mention what happens when annotations and resourcerequirements are both set? If both are set, the annotation values have priority and are applied.

search/search_klusterlet_addon.adoc
dockerymick:
```suggestion
        memory: 1024Mi
      requests:
        memory: 512Mi
```

search/search_klusterlet_addon.adoc
dockerymick:
```suggestion
.. Add the `backup` label to include the `AddOnDeploymentConfig` resource to the cluster backups. Run the following command:
```

search/search_klusterlet_addon.adoc
```diff
@@ -15,7 +15,7 @@ Resource requirements for the `klusterlet-addon-search` pod in a managed cluster
 oc edit managedclusteraddon search-collector -n xyz
 ----
 
-. Append the resource requirements as annotations. View the following example:
+. Append the resource requirements as annotations. The annotation overrides the resource requirements on the managed clusters and automatically restarts the pod with new resource requirements. View the following example:
```
dockerymick: Based on your [question](https://github.com/stolostron/rhacm-docs/pull/8615#issuecomment-3984395258) about mentioning the priority, I think it can be helpful to possibly mention it here. I does make the step longer though. Maybe we can add that info as an additional note. Alternatively, I mention the priority in the step and move the following statement to the Notes: 

"The annotation overrides the resource requirements on the managed clusters and automatically restarts the pod with new resource requirements."

smcavey: I see, that could do the trick since that will be read first prior to both the methods of adjusting resource requirements. I can run with this 👍

dockerymick:
```suggestion
. Append the resource requirements as annotations. When you define both `annotations` and `resourceRequirements`, the `annotations` value overrides the `resourceRequirements` value. View the following example:
```

search/search_klusterlet_addon.adoc
```diff
@@ -26,7 +26,45 @@ metadata:
   annotations: addon.open-cluster-management.io/search_memory_limit: 2048Mi
   addon.open-cluster-management.io/search_memory_request: 512Mi
 ----
+*Note:* You can discover all resources defined in your managed cluster by using the API Explorer in the console. Alternatively, you can discover all resources by running the following command:
++
+[source,bash]
+----
+oc api-resources
+----
+
```
dockerymick:
```suggestion
+
*Notes:* 

- The annotation overrides the resource requirements on the managed clusters and automatically restarts the pod with new resource requirements.

- You can discover all resources defined in your managed cluster by using the API Explorer in the console. Alternatively, you can discover all resources by running the following command:
+
[source,bash]
----
oc api-resources
----

```

search/search_klusterlet_addon.adoc
```diff
@@ -26,7 +26,50 @@ metadata:
...
+- The annotation overrides the resource requirements on the managed clusters and automatically restarts the pod with new resource requirements.
+
+- You can discover all resources defined in your managed cluster by using the API Explorer in the console. Alternatively, you can discover all resources by running the following command:
++
+[source,bash]
+----
+oc api-resources
+----
+
+. Create an `AddonDeploymentConfig` resource and update the `klusterlet-addon-search` deployment resource requirements and limits. Your `AddonDeploymentConfig` resource might resemble the following example:
```
swopebe: Not fully sure what you want me to notice about this example bc there is no: `klusterlet-addon-search`

Is seems like we want them to notice:

 name: search-addon-resources
  namespace: your-managed-ns

So we should define that in the intro, at least the name?

search/search_klusterlet_addon.adoc
```diff
@@ -26,7 +26,50 @@ metadata:
...
+  resourceRequirements:
+  - containerID: deployments:klusterlet-addon-search:collector
+    resources:
+      limits:
+        memory: 1024Mi
+      requests:
+        memory: 512Mi
+----
+
+.. Add the `backup` label to include the `AddOnDeploymentConfig` resource to the cluster backups. Run the following command:
```
swopebe: Why is this a substep?

search/search_klusterlet_addon.adoc
```diff
@@ -26,7 +26,50 @@ metadata:
...
+
++
+[source,bash]
+----
+oc label addondeploymentconfig search-addon-resources -n <namespace-for-search-addon-resources> cluster.open-cluster-management.io/backup=""
+----
 
-The annotation overrides the resource requirements on the managed clusters and automatically restarts the pod with new resource requirements.
+.. Update the `klusterlet-addon-search` deployment to use the new `AddonDeploymentConfig` resource. Run the following command:
```
swopebe: these are simply just steps after the create step, it seems, so these don't beed to be substeps.

. Create X
. Edit X to ___ and __ as in the following example
. Change x to y
. Apply the __ file 

(this is normally the format for creating a resource--unless there is a specific to go with substeps?)

swopebe: I think we need the _apply_ step as well; there are some good examples in the doc to scan.

swopebe (changes requested):

search/search_klusterlet_addon.adoc
dockerymick: `klusterlet-addon-search` is a part of the  `resourceRequirements` parameter section on starting at line 52. I can see where it can be confusing though. What are your thoughts about me saying: 

```suggestion
. Create an `AddonDeploymentConfig` resource and update the limit in the `resourceRequirements` field for the `klusterlet-addon-search` deployment . Your `AddonDeploymentConfig` resource might resemble the following example:
```

search/search_klusterlet_addon.adoc
dockerymick: In another PR where I added the same step and it was a substep. It was a mistake. Good catch, thank you !

search/search_klusterlet_addon.adoc
dockerymick: @smcavey Does the user need to apply the AddOnDeploymentConfig? When the user saves the YAML are changes automatically applied? The step would resemble the following step: 

. Apply the `search-addon-resources` `AddOnDeploymentConfig` file to your klusterlet add-on by running the following command:

+
[source,bash]
----
oc apply -f search-addon-resources.yaml
----

smcavey: The `AddOnDeploymentConfig` for the doesn't exist for the klusterlet-addon-search pod in a fresh install. So the resource must be created. In the Jira comment Jorge had an [example](https://issues.redhat.com/browse/ACM-19020?focusedId=27445928&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-27445928) which creates the resource and applies it in the same command. But one could create the resource .yaml file first and then `oc apply -f _file.yaml`_

swopebe (approved):

search/search_klusterlet_addon.adoc
```diff
@@ -26,7 +26,59 @@ metadata:
   annotations: addon.open-cluster-management.io/search_memory_limit: 2048Mi
   addon.open-cluster-management.io/search_memory_request: 512Mi
 ----
++
+*Notes:* 
+
+- The annotation overrides the resource requirements on the managed clusters and automatically restarts the pod with new resource requirements.
+
+- You can discover all resources defined in your managed cluster by using the API Explorer in the console. Alternatively, you can discover all resources by running the following command:
++
```
swopebe, line 36: Need a space on line 35

search/search_klusterlet_addon.adoc
```diff
@@ -26,7 +26,59 @@ metadata:
...
+[source,bash]
+----
+oc api-resources
+----
```
dockerymick, line 40:
```suggestion
- You can discover all resources defined in your managed cluster by using the API Explorer in the console. Alternatively, you can discover all resources by running the following command:

+
[source,bash]
----
oc api-resources
----
```

PR #8668, opened 2026-03-11
https://issues.redhat.com/browse/ACM-30210\# for 2.16

dockerymick:


fxiang1 (approved):
Thanks @dockerymick ! The change looks good!

gitops/gitops_console.adoc
```diff
@@ -9,19 +9,13 @@ Learn more about integrated {ocp-short} GitOps console features. Create and view
 
 *Important:* Available actions are based on your assigned role. Learn about access requirements from the link:../secure_clusters/rbac.adoc#role-based-access-control[Role-based access control] documentation.
 
-[#gitops-console-prereqs]
-== Prerequisites
+.Prerequisites
 
-See the following prerequisites and requirements:
+See the following prerequisites and requirements to view *Argo server* options in the console and create an `ApplicationSet`:
```
swopebe: Is this clickable? If not we should not bold it.

gitops/gitops_console.adoc
```diff
@@ -9,19 +9,13 @@ Learn more about integrated {ocp-short} GitOps console features. Create and view
...
 
-- For an Argo CD `ApplicationSet` to be created, you need to enable `Automatically sync when cluster state changes` from the `Sync policy`.
-
-- For Flux with the `kustomization` controller, find Kubernetes resources with the label `kustomize.toolkit.fluxcd.io/name=<app_name>`.
-    
-- For Flux with the `helm` controller, find Kubernetes resources with the label `helm.toolkit.fluxcd.io/name=<app_name>`.
+- An Argo CD `ApplicationSet` resource must be created. From the console, verify that the *Automatically sync when cluster state changes* option is selected in the *Sync policy* step.
```
swopebe: I am unsure why this is bold.

gitops/gitops_console.adoc
swopebe:
```suggestion
See the following prerequisites and requirements to view the *Argo server* options in the console and create an `ApplicationSet`:
```

gitops/gitops_console.adoc
swopebe: **First line on 16:**
This is passive, so remember we need to focus on active voice as much as we can: 

- Create an Argo CD `ApplicationSet` resource. **or..**.
- You need to create a resource?

**2nd line:** Why is this not an active step? I am unsure about how this is written.

_...verify that the _Automatically sync when cluster state changes_ option is selected in the _Sync policy_ step._
(also not sure why it is bold)

_You need to select the ___ in the ____? 

(Is there a reason we are not writing it in active and imperative voice here)

(Is it because this is done previously from another topic, and they just need to verify? (usually I would just say: Complete the procedure to ___. See ___. (as a prereq)

If we can be more active here and ensure proper use of bold, then let's change. If not, we can chat about it later and I can approve.

gitops/gitops_console.adoc
fxiang1: I think "Argo server" is referring to this part of the UI. It's a dropdown that allow users to select the Argo server.

Image

gitops/gitops_console.adoc
swopebe: For bold, we have tried to stick to the philosphy from DITA and that is `<ui-control>` bolds clickable items and `<wintitle> `italicizes windows, fields, pages.... So if it is something they have to click, go ahead and bold, but don't bold everything for emphasis. It is an issue with accessibility. A screen reader does not know why something is bold and they user listening to it can get lost in `<b>` tags.

swopebe (changes requested):

gitops/gitops_console.adoc
swopebe: Then we definitely need to call it a drop-down menu. @dockerymick for this view, for instance, I would write:
"...to view the **this can be bold** drop-down menu and choose `Select the Argo server`.

swopebe: so now I know this is a drop-down menu--it can be bold but we need to call it that, as well, so it is understood as a clickable item.

gitops/gitops_console.adoc
```diff
@@ -9,19 +9,13 @@ Learn more about integrated {ocp-short} GitOps console features. Create and view
 
 *Important:* Available actions are based on your assigned role. Learn about access requirements from the link:../secure_clusters/rbac.adoc#role-based-access-control[Role-based access control] documentation.
 
-[#gitops-console-prereqs]
-== Prerequisites
+.Prerequisites
 
-See the following prerequisites and requirements:
+See the following prerequisites and requirements to view the *Argo server* options in the console and create an `ApplicationSet`:
```
dockerymick:
```suggestion
See the following prerequisites and requirements to access the *Argo server* drop-down menu in the console and create an `ApplicationSet` resource:
```

gitops/gitops_console.adoc
dockerymick: Since it was listed in prereqs, I used past tense since it seems like this is something the user has to do before moving forward. 
"*Automatically sync when cluster state changes*" is an option during creation steps. The Sync policy step is within the installation prompt. Here's a screen capture

Image

swopebe: You can use past tense in prereqs but it is better to use present. Bold is correct here, then. TY!

swopebe (approved):

gitops/gitops_console.adoc
```diff
@@ -9,19 +9,13 @@ Learn more about integrated {ocp-short} GitOps console features. Create and view
...
 
-See the following prerequisites and requirements:
+See the following prerequisites and requirements to access the *Argo server* drop-down menu in the console and create an `ApplicationSet` resource:
 
-- For an Argo CD `ApplicationSet` to be created, you need to enable `Automatically sync when cluster state changes` from the `Sync policy`.
-
-- For Flux with the `kustomization` controller, find Kubernetes resources with the label `kustomize.toolkit.fluxcd.io/name=<app_name>`.
-    
-- For Flux with the `helm` controller, find Kubernetes resources with the label `helm.toolkit.fluxcd.io/name=<app_name>`.
+- An Argo CD `ApplicationSet` resource must be created. From the console, verify that the *Automatically sync when cluster state changes* option is selected in the *Sync policy* step.
```
dockerymick:
```suggestion
- Create an Argo CD `ApplicationSet` resource. Click the *Automatically sync when cluster state changes* option in the *Sync policy* step.
```

PR #8683, opened 2026-03-17
https://redhat.atlassian.net/browse/ACM-30705 CAPM3 release note

oafischer:
2.16

### DIFF
--- a/clusters/release_notes/mce_known_issues.adoc
+++ b/clusters/release_notes/mce_known_issues.adoc
@@ -37,6 +37,42 @@ link:https://issues.redhat.com/browse/ACM-7523[ACM-7523]
 
 Learn about Known issues for Cluster lifecycle with {mce-short}, such as issues with creating, discovering, importing, and removing clusters, and more cluster management issues for {mce-short}.
 
+== Cluster API Provider Metal3 certificate generation fails after upgrade
+
+After you upgrade from {mce} version 2.10 or 2.10.1 to version 2.11, the automatic generation of certificates for the Cluster API Provider Metal3 webhooks fail. As a result, webhooks do not validate.
+
+To resolve the issue, manually delete the existing secret and the renamed service. The operator then regenerates the correct secret.
+
+Complete the following steps on your managed cluster:
+
+. Delete the webhook service. Run the following command:
+
++
+[source,bash]
+----
+oc delete capm3-webhook-service -n multicluster-engine
+----
+
+. Delete the Transport Layer Security (TLS) secret. Run the following command:
+
++
+[source,bash]
+----
+oc delete secret capm3-webhook-service-cert -n multicluster-engine
+----
+
+. Clear the error annotations. Run the following command:
+
++
+[source,bash]
+----
+oc annotate service mce-capm3-webhook-service -n multicluster-engine \
+service.alpha.openshift.io/serving-cert-generation-error- \
+service.alpha.openshift.io/serving-cert-generation-error-num- \
+service.beta.openshift.io/serving-cert-generation-error- \
+service.beta.openshift.io/serving-cert-generation-error-num-
+----
+
 === Resource deletion during cluster distribution updates
 
 When a change occurs in the cluster because of the `Placement` resource, cluster redistribution occurs across multiple `PlacementDecisions` resources. During the sequential update, a cluster is temporarily missing from all `PlacementDecisions` resources.
--- a/clusters/release_notes/mce_known_issues.adoc
+++ b/clusters/release_notes/mce_known_issues.adoc
@@ -50,7 +50,7 @@ Complete the following steps on your managed cluster:
 +
 [source,bash]
 ----
-oc delete capm3-webhook-service -n multicluster-engine
+oc delete service capm3-webhook-service -n multicluster-engine
 ----
 
 . Delete the Transport Layer Security (TLS) secret. Run the following command:
--- a/clusters/release_notes/mce_known_issues.adoc
+++ b/clusters/release_notes/mce_known_issues.adoc
@@ -39,7 +39,7 @@ Learn about Known issues for Cluster lifecycle with {mce-short}, such as issues 
 
 === Cluster API Provider Metal3 certificate generation fails after upgrade
 
-After you upgrade from {mce} version 2.10 or 2.10.1 to version 2.11, the automatic generation of certificates for the Cluster API Provider Metal3 webhooks fail. As a result, webhooks do not validate.
+After you upgrade from {mce} version 2.10 or 2.10.1 to version 2.11, the automatic generation of certificates for the Cluster API Provider Metal3 webhooks fail due to a naming conflict in the target resources. As a result, webhooks do not validate.
 
 To resolve the issue, manually delete the existing secret and the renamed service. The operator then regenerates the correct secret.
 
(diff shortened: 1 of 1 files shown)


clusters/release_notes/mce_known_issues.adoc
```diff
@@ -37,6 +37,42 @@ link:https://issues.redhat.com/browse/ACM-7523[ACM-7523]
...
+To resolve the issue, manually delete the existing secret and the renamed service. The operator then regenerates the correct secret.
+
+Complete the following steps on your managed cluster:
+
+. Delete the webhook service. Run the following command:
+
++
+[source,bash]
+----
+oc delete capm3-webhook-service -n multicluster-engine
```
frajamomo: missing "**service**" in the command

```
oc delete capm3-webhook-service -n multicluster-engine
error: the server doesn't have a resource type "capm3-webhook-service"

oc delete service capm3-webhook-service -n multicluster-engine
service "capm3-webhook-service" deleted from multicluster-engine namespace
```

clusters/release_notes/mce_known_issues.adoc
```diff
@@ -37,6 +37,42 @@ link:https://issues.redhat.com/browse/ACM-7523[ACM-7523]
...
+=== Cluster API Provider Metal3 certificate generation fails after upgrade
+
+After you upgrade from {mce} version 2.10 or 2.10.1 to version 2.11, the automatic generation of certificates for the Cluster API Provider Metal3 webhooks fail. As a result, webhooks do not validate.
+
+To resolve the issue, manually delete the existing secret and the renamed service. The operator then regenerates the correct secret.
+
+Complete the following steps on your managed cluster:
+
+. Delete the webhook service. Run the following command:
+
```
rccrdpccl, line 49: @oafischer sorry if it wasnt' specified in the original issue, but do you think it's worth mentioning it's due to a conflict in names? It would make a bit clearer the reason why we are deleting the webhook service (because there's another one named mce-capm3-webhook-service)

rccrdpccl:
just a small question, otherwise LGTM

clusters/release_notes/mce_known_issues.adoc
```diff
@@ -37,6 +37,42 @@ link:https://issues.redhat.com/browse/ACM-7523[ACM-7523]
...
+----
+oc delete capm3-webhook-service -n multicluster-engine
+----
+
+. Delete the Transport Layer Security (TLS) secret. Run the following command:
+
++
+[source,bash]
+----
+oc delete secret capm3-webhook-service-cert -n multicluster-engine
```
frajamomo, line 61: in my test the secret capm3-webhook-service-cert no longer exists — it was likely already cleaned up or regenerated during the upgrade

clusters/release_notes/mce_known_issues.adoc
```diff
@@ -37,6 +37,42 @@ link:https://issues.redhat.com/browse/ACM-7523[ACM-7523]
...
++
+[source,bash]
+----
+oc annotate service mce-capm3-webhook-service -n multicluster-engine \
+service.alpha.openshift.io/serving-cert-generation-error- \
+service.alpha.openshift.io/serving-cert-generation-error-num- \
+service.beta.openshift.io/serving-cert-generation-error- \
+service.beta.openshift.io/serving-cert-generation-error-num-
+----
+
```
frajamomo, line 75: After this point, the old capm3-mutating-webhook-configuration is still registered and trying to call the now-deleted capm3-webhook-service. The stale webhook configurations need to be deleted too.
After that the TLS cert still has the old service name. The certificate needs to be regenerated

rccrdpccl:
@frajamomo thanks for your feedback - could you please update the JIRA issue describing the problem, with the right commands for resolution?

clusters/release_notes/mce_known_issues.adoc
frajamomo: this is the sequence I followed:
- Delete the old webhook service
`oc delete service capm3-webhook-service -n multicluster-engine`
- Delete the old TLS secret
`oc delete secret capm3-webhook-service-cert -n multicluster-engine`
- Clear error annotations on the new service
```
oc annotate service mce-capm3-webhook-service -n multicluster-engine \                                                                       
service.alpha.openshift.io/serving-cert-generation-error- \                                                                                
service.alpha.openshift.io/serving-cert-generation-error-num- \                                                                            
service.beta.openshift.io/serving-cert-generation-error- \                                                                                 
service.beta.openshift.io/serving-cert-generation-error-num-
```
- Delete the stale webhook configurations
```
oc delete mutatingwebhookconfiguration capm3-mutating-webhook-configuration                                                                  
oc delete validatingwebhookconfiguration capm3-validating-webhook-configuration
```
- Delete the new TLS secret and restart the pod to regenerate the certificate
```
oc delete secret mce-capm3-webhook-service-cert -n multicluster-engine
oc delete pod mce-capm3-controller-manager-<hash> -n multicluster-engine
```
- Delete the stale old controller-manager deployment
`oc delete deployment capm3-controller-manager -n multicluster-engine`

clusters/release_notes/mce_known_issues.adoc
oafischer: Updated

clusters/release_notes/mce_known_issues.adoc
oafischer: Should I remove this step or leave it for now?

clusters/release_notes/mce_known_issues.adoc
oafischer: Is there a way we can add that to the first paragraph? It is best to provide context in that introductory paragraph.

clusters/release_notes/mce_known_issues.adoc
oafischer: I've added the additional steps, thanks!

oafischer:
I've updated the topic with the new commands, let me know if they look good now @rccrdpccl @frajamomo. Thanks!

clusters/release_notes/mce_known_issues.adoc
rccrdpccl:
```
After you upgrade from {mce} version 2.10 or 2.10.1 to version 2.11, the automatic generation of certificates for the Cluster API Provider Metal3 webhooks fail due to a conflict in the target resources naming. As a result, webhooks do not validate.
```
@oafischer WDYT?

In any case this is just a nit, in my opinion it can go live as it is as all the necessary info are there

clusters/release_notes/mce_known_issues.adoc
rccrdpccl: I would leave it in case it's not always the case

rccrdpccl:
/lgtm

thanks @oafischer !

clusters/release_notes/mce_known_issues.adoc
oafischer: @rccrdpccl 

That works, I added that real quick, thanks!

clusters/release_notes/mce_known_issues.adoc
```diff
@@ -37,6 +37,77 @@ link:https://issues.redhat.com/browse/ACM-7523[ACM-7523]
...
+. Delete the stale webhook configurations. Run the following commands:
+
++
+[source,bash]
+----
+oc delete mutatingwebhookconfiguration capm3-mutating-webhook-configuration
+oc delete validatingwebhookconfiguration capm3-validating-webhook-configuration
+----
```
dockerymick, line 88: I would add a line break to separate the commands

dockerymick (approved):
one minor suggestion, but it looks good to me!

PR #8724, opened 2026-04-01
https://redhat.atlassian.net/browse/ACM-31142

oafischer:
2.16 and 2.15

### DIFF
--- a/gitops/gitops_deploy_argocd_appset.adoc
+++ b/gitops/gitops_deploy_argocd_appset.adoc
@@ -32,7 +32,7 @@ To deploy the `ApplicationSet` resource for standard configurations, complete th
 apiVersion: cluster.open-cluster-management.io/v1beta1
 kind: Placement
 metadata:
-  name: all-openshift-clusters
+  name: all-openshift-clusters-except-local
   namespace: openshift-gitops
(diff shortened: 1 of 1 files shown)


mikeshng (commented):
Thanks for the PR. Could you use the following example instead? I just realized in 2.16 the local-cluster is not that well supported in some cases so it's better the example version to exclude it :

```
apiVersion: cluster.open-cluster-management.io/v1beta1
kind: Placement
metadata:
  name: all-openshift-clusters-except-local
  namespace: openshift-gitops
spec:
  tolerations:
  - key: cluster.open-cluster-management.io/unreachable
    operator: Exists
  - key: cluster.open-cluster-management.io/unavailable
    operator: Exists
  predicates:
  - requiredClusterSelector:
      labelSelector:
        matchExpressions:
        - key: vendor
          operator: "In"
          values:
          - OpenShift
        - key: local-cluster
          operator: DoesNotExist
```

Thanks again!

oafischer:
Of course, I updated the sample @mikeshng. Thanks for the review!

mikeshng (approved):
/lgtm

dockerymick (approved):

PR #8776, opened 2026-04-27
Resolve conversion-blocking AsciiDocDITA warnings and errors 2.17--use for info only

jhradilek:
This pull request resolves all conversion blocking [AsciiDocDITA](https://github.com/jhradilek/asciidoctor-dita-vale) warnings in the `clusters/install_upgrade/` directory:

- [ ] AsciiDocDITA.CalloutList
- [ ] AsciiDocDITA.ConceptLink
- [x] AsciiDocDITA.ContentType
- [x] AsciiDocDITA.NestedSection
- [x] AsciiDocDITA.RelatedLinks
- [x] AsciiDocDITA.ShortDescription
- [x] AsciiDocDITA.TaskContents
- [x] AsciiDocDITA.TaskSection
- [x] AsciiDocDITA.TaskStep

Note that I chose the less invasive approach to unblock test conversion. The correct approach would be to:

- split `uninstall.adoc` in two distinct procedures rather than combining two approaches in one.
- split `upgrade_cluster_disconnected_policies.adoc` into multiple concept and procedure modules and use correct templates for each.

I am going to address `AsciiDocDITA.CalloutList` in a separate pull request.

### DIFF
--- a/clusters/install_upgrade/install_connected.adoc
+++ b/clusters/install_upgrade/install_connected.adoc
@@ -8,18 +8,17 @@ The {mce-short} is installed with {olm}, which manages the installation, upgrade
 
 *Required access:* Cluster administrator
 
+After you prepare, you can install from the {ocp-catalog} console, or from the {ocp-short} command-line interface.
+
 .Prerequisites
 
 Before you start the installation procedure, you must complete the preparation process. See xref:./prepare_install_mce_connected.adoc#preparing-to-install-mce[Preparing to install {mce-short}].
 
 .Procedure
 
-After you prepare, you can install from the {ocp-catalog} console, or from the {ocp-short} command-line interface.
+* *Installing from the {ocp-catalog} console:*
 
-== Installing from the {ocp-catalog} console
-
-From the _Administrator_ view in your {ocp-short} navigation, install the {ocp-catalog} console that is provided with {ocp-short}. See the following steps:
-
+. From the _Administrator_ view in your {ocp-short} navigation, install the {ocp-catalog} console that is provided with {ocp-short}. See the following steps:
 . Go to *Ecosystem* > *Software Catalog*.
 . Choose the Operator that you want to install.
 . On the _Operator Installation_ page, select the options for your installation:
@@ -68,9 +67,7 @@ spec: {}
 +
 After the `MultiClusterEngine` resource is created, the status for the resource is `Available` on the `MultiCluster Engine` tab.
 
-== Installing from the {ocp-short} command-line interface
-
-See the following steps to install from the {ocp-short} command-line interface:
+* *Installing from the {ocp-short} command-line interface:*
 
 . Create a {mce-short} namespace where the operator requirements are contained. Run the following command, where `<namespace>` is the name for your {mce-short} namespace. The value for `<namespace>` might be referred to as _Project_ in the {ocp-short} environment:
 
@@ -190,7 +187,7 @@ oc delete validatingwebhookconfiguration multiclusterengines.multicluster.opensh
 oc delete mce --all 
 ----
 
-*Notes:*
+.Results
 
 - A `ServiceAccount` with a `ClusterRoleBinding` automatically gives cluster administrator privileges to 
 {mce-short} and to any user credentials with access to the namespace where you install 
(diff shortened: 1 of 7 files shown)


clusters/install_upgrade/install_connected.adoc
```diff
@@ -190,7 +187,7 @@ oc delete validatingwebhookconfiguration multiclusterengines.multicluster.opensh
 oc delete mce --all 
 ----
 
-*Notes:*
+.Results
```
swopebe, line 190: Here, this change does not work.

This cannot be results. We write our notes differently than the rest of the writers here and that is due to the history/move from IBM and PM reaction to the large/colorful/invasive boxes that showed up when we moved here (which are not accessible, not sure how red hat is getting away with those boxes--IBM had to be accessible for government contracts) so the notes are not results through out the doc. There is a dita tag (special tagging) that we will use when we move to DITA, those are usually (from memory) "note", "important", etc... @jhradilek 

We need to find a way to work around this, if possible. All of our notes are like this. We try not to use them much (not like other products who use them a lot) but they are built like the output of DITA.

clusters/install_upgrade/install_disconnected.adoc
```diff
@@ -22,8 +22,8 @@ You must meet the following requirements before you install the {mce-short}:
 
 .Procedure
 
-Confirm your {ocp-short} installation and then install in a disconnected environment.
-
+. Confirm your {ocp-short} installation and then install in a disconnected environment.
```
swopebe, line 25: This is not a step, this is a description. This change will not work, as step one is line 29. The information before it is conceptual. This can move above the tag.

clusters/install_upgrade/uninstall.adoc
```diff
@@ -27,9 +29,7 @@ For more information about detaching clusters, see the _Removing a cluster from
 
 .Procedure
 
-You can use the terminal or the console to remove {mce-short}.
-
-. Uninstall {mce-short} from the terminal.
+* Uninstall {mce-short} from the terminal.
```
swopebe, line 32: this is now a bullet with a sub step, which doesn't make sense.

clusters/install_upgrade/uninstall.adoc
```diff
@@ -73,7 +73,7 @@ oc delete sub multicluster-engine
 +
 The CSV version might be different.
 
-. Uninstall {mce-short} from the console. When you use the console to uninstall, you remove the operator. 
+* Uninstall {mce-short} from the console. When you use the console to uninstall, you remove the operator.
```
swopebe, line 77: same, a bullet with a sub step, you need a step before a sub step.

clusters/install_upgrade/uninstall.adoc
jhradilek: I can revert it to numbered procedures, but you effectively have a choice of two methods that are not performed both in the same sequence, where unordered list is the appropriate choice.

clusters/install_upgrade/upgrade_mce.adoc
```diff
@@ -3,6 +3,7 @@
 [#upgrading-mce]
 = Upgrading {mce} 
 
+[role="_abstract"]
```
swopebe, line 6: this was changed, just realized we are looking at 2.17. I meant for us to look at 2.16 :-/

clusters/install_upgrade/upgrade_cluster_disconnected_policies.adoc
```diff
@@ -543,6 +546,6 @@ Complete the following steps to upgrade:
 If your cluster upgrade fails, the Operator generally retries the upgrade a few times, stops, and reports the status of the failing component. In some cases, the upgrade process continues to cycle through attempts to complete the process. Rolling your cluster back to a previous version after a failed upgrade is not supported. Contact Red Hat support for assistance if your cluster upgrade fails.
 
 [#add-resources-upgrade-service]
-=== Additional resources
+== Additional resources
```
swopebe, line 550: reaching out to get confirmation on what we should be doing here.

clusters/install_upgrade/uninstall.adoc
swopebe: Ok, we can do that. I could not remember if we could use bullets within the procedures like this, so thanks for that!!

PR #8778, opened 2026-04-27
[2.16 Stage] Correct outstanding AsciiDocDITA style violations

jhradilek:
This pull request resolves all conversion blocking [AsciiDocDITA](https://github.com/jhradilek/asciidoctor-dita-vale) warnings in the clusters/install_upgrade/ directory:

- [x] AsciiDocDITA.TaskStep
- [x] AsciiDocDITA.RelatedLinks
- [ ] AsciiDocDITA.ConceptLink
- [x] AsciiDocDITA.TaskSection
- [x] AsciiDocDITA.TaskContents

### DIFF
--- a/clusters/install_upgrade/build_graph_init_update_service.adoc
+++ b/clusters/install_upgrade/build_graph_init_update_service.adoc
@@ -8,8 +8,6 @@
 
 .Procedure
 
-See the following procedure to build the graph data init container:
-
 . Clone the _graph data_ Git repository by running the following command:
 
 +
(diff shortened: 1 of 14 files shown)


swopebe:
One thing here: we do not use .**Troubelshooting steps** 

Reason: We actually have troubleshooting section that support will be looking for in our doc. We should not have something titled **troubleshooting** in the doc--the rule is if you need t-shooting in the doc, you should be changing the procedure to avoid it.

clusters/install_upgrade/build_graph_init_update_service.adoc
```diff
@@ -84,4 +82,4 @@ Follow the steps to xref:../install_upgrade/deploy_update_service_instance.adoc#
 
 .Additional resources
 
-Return to xref:../install_upgrade/upgrade_cluster_overview.adoc#upgrade-clusters-mce-overview[Upgrading disconnected clusters by using policies] for an overview.
```
swopebe, line 87: I am wondering, is this a DITA limitation that calls for use of only related links here and no complete sentences? Sometimes we use this section to orient the user back to the beginning.

swopebe (approved):

PR #8783, opened 2026-04-27
Fixing pantheon book

dockerymick:


observability/obs_metrics.adoc
```diff
@@ -208,4 +208,4 @@ The data is only displayed in Grafana when the rule is initiated.
 
 .Additional resources
 
-- See xref:../observability/use_observability.adoc#exporting-metrics-to-external-endpoints[Exporting metrics to external endpoints]. 
+- See xref:../observability/use_observability.adoc[Exporting metrics to external endpoints].
```
swopebe: Not sure if you want to address this here or later in conversion...

But we were told to just make these links. (not sentences) 

I am not sure the reason, as I don't remember what DITA requires for this section. I will approve knowing we have to change this anyway.

swopebe (approved):

observability/obs_metrics.adoc
dockerymick: I don't mind changing it now

PR #8784, opened 2026-04-28
https://redhat.atlassian.net/browse/ACM-30664 for search postgrsql 2.17

dockerymick:


search/search_console.adoc
```diff
@@ -60,7 +60,7 @@ Provides access to all cluster data in the `search-indexer` through GraphQL and
 | `search-postgres`
 | 
 |
-| Stores collected data from all managed clusters in an instance of the PostgreSQL database.
+| Stores collected data from all managed clusters in an instance of the PostgreSQL database. Use the PostgreSQL database to improve the performance of the search service. The database is a cache and does not contain any unrecoverable information. If you recreate the database, consider the time it takes to rebuild the cache.
```
smcavey: https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.16/html/search/acm-search 1.b already mentions the performance gains by setting up a PVC.
I'd omit `Use the PostgreSQL database to improve the performance of the search service`, also because using Postgres with Search is non-negotiable.

Not sure about the particular verbage on saying the database `The database is a cache` vs `The database acts like a cache as it contains no historical data.`

When the database restarts, even if storage persisted, none of the data gets reused, Search assumes that it's all stale and overwrites it on first update

dockerymick: It's best to not say "acts like" because it is anthropomorphic.  How about this rewrite?
```suggestion
| Stores collected data from all managed clusters in an instance of the PostgreSQL database. The database contains no historical data. When the database restarts, the data is overridden. If you recreate the database, consider the time it takes to rebuild the cache.
```

smcavey: That sounds good to me then

swopebe (approved):

PR #8791, opened 2026-04-29
https://redhat.atlassian.net/browse/ACM-31310--mce 2.15 install modul…

swopebe:
…arization


dockerymick (approved):
Are the DS stores required? I recall that we didn't need them. If I am remembering correctly, please delete the DS_Store files. Otherwise, this looks good to me!

PR #8797, opened 2026-04-30
https://redhat.atlassian.net/browse/ACM-32499

oafischer:
2.17 and 2.16


mshort55:
I think this is good, but I would like @yiraeChristineKim to check also.

yiraeChristineKim:
actually We already set  decentralizedLiveMigration":true already by default. What about this.. if you don;t see the vir console then we can use this  oc -n openshift-cnv patch hco kubevirt-hyperconverged --type='merge' -p '{"spec":{"featureGates":{"decentralizedLiveMigration":true}}}'

oafischer:
@yiraeChristineKim That works, I updated the PR, thanks!

yiraeChristineKim (approved):

swopebe (approved):

PR #8818, opened 2026-05-08
https://redhat.atlassian.net/browse/ACM-33198--z-stream drafts, thoug…

swopebe:
need more info

### DIFF
--- a/clusters/release_notes/mce_fixed_issues.adoc
+++ b/clusters/release_notes/mce_fixed_issues.adoc
@@ -4,3 +4,7 @@
 By default, fixed issues for _errata_ releases are automatically applied when released. The details are published here when the release is available. If no release notes are listed, the product does not have an errata release at this time.
 
 *Important:* For reference, Jira links and Jira numbers might be added to the content and used internally. Links that require access might not be available for the user. Learn about the types of link:https://access.redhat.com/security/updates/advisory[errata releases] from Red Hat.
+
+== Errata 2.11.1
+
+* Delivers updates to one or more product container images.
--- a/release_notes/acm_fixed_issues.adoc
+++ b/release_notes/acm_fixed_issues.adoc
@@ -9,16 +9,10 @@ See link:../install/upgrade_hub.adoc#upgrading-by-using-the-operator[Upgrading b
 
 *Important:* Cluster lifecycle components and features are within the {mce-short}, which is a software operator that enhances cluster fleet management. Release notes for {mce-short}-specific features are found in at link:../clusters/release_notes/mce_release_notes.adoc#mce-release-notes[Release notes for Cluster lifecycle with {mce-short}].
 
-== Errata 2.16.1
+== Errata 2.16.1 (Deprecated)
 
 * Delivers updates to one or more product container images.
 
-* Fixes ____ Thanos compact now deletes empty folders on S3, which could materialize on some S3 vendors following block deletion.
-//this needs work, not sure if this is the fix or the problem. I sent examples in the issue.
+* Fixes an issue on some S3 vendors that left empty folders existing on object storage after Thanos Compact deletes blocks.
 
 link:https://redhat.atlassian.net/browse/ACM-32930[ACM-32930]
-
-* Fixes a problem with a required placements field in the `MulticlusterRoleAssignment` resource that caused the Role Assignments page to fail when upgrading... to 2.16.0 from 2.15.0 or earlier releases. Now the .....
-//this needs work, I sent examples in the issue.
-
-link:https://redhat.atlassian.net/browse/ACM-33108[ACM-33108]
--- a/release_notes/acm_fixed_issues.adoc
+++ b/release_notes/acm_fixed_issues.adoc
@@ -8,3 +8,17 @@ By default, fixed issues for _errata_ releases are automatically applied when re
 See link:../install/upgrade_hub.adoc#upgrading-by-using-the-operator[Upgrading by using the operator] for more information about upgrades.
 
 *Important:* Cluster lifecycle components and features are within the {mce-short}, which is a software operator that enhances cluster fleet management. Release notes for {mce-short}-specific features are found in at link:../clusters/release_notes/mce_release_notes.adoc#mce-release-notes[Release notes for Cluster lifecycle with {mce-short}].
+
+== Errata 2.16.1
+
+* Delivers updates to one or more product container images.
+
+* Fixes ____ Thanos compact now deletes empty folders on S3, which could materialize on some S3 vendors following block deletion.
+//this needs work, not sure if this is the fix or the problem. I sent examples in the issue.
+
+link:https://redhat.atlassian.net/browse/ACM-32930[ACM-32930]
+
+* Fixes a problem with a required placements field in the `MulticlusterRoleAssignment` resource that caused the Role Assignments page to fail when upgrading... to 2.16.0 from 2.15.0 or earlier releases. Now the .....
+//this needs work, I sent examples in the issue.
+
+link:https://redhat.atlassian.net/browse/ACM-33108[ACM-33108]


clusters/release_notes/mce_fixed_issues.adoc
```diff
@@ -3,4 +3,8 @@
 
 By default, fixed issues for _errata_ releases are automatically applied when released. The details are published here when the release is available. If no release notes are listed, the product does not have an errata release at this time.
 
-*Important:* For reference, Jira links and Jira numbers might be added to the content and used internally. Links that require access might not be available for the user. Learn about the types of link:https://access.redhat.com/security/updates/advisory[errata releases] from Red Hat.
\ No newline at end of file
+*Important:* For reference, Jira links and Jira numbers might be added to the content and used internally. Links that require access might not be available for the user. Learn about the types of link:https://access.redhat.com/security/updates/advisory[errata releases] from Red Hat.
+
+== Errata 2.11.1
```
swopebe, line 8: this one is good.

release_notes/acm_fixed_issues.adoc
```diff
@@ -7,4 +7,18 @@ By default, fixed issues for _errata_ releases are automatically applied when re
...
-*Important:* Cluster lifecycle components and features are within the {mce-short}, which is a software operator that enhances cluster fleet management. Release notes for {mce-short}-specific features are found in at link:../clusters/release_notes/mce_release_notes.adoc#mce-release-notes[Release notes for Cluster lifecycle with {mce-short}].
\ No newline at end of file
+*Important:* Cluster lifecycle components and features are within the {mce-short}, which is a software operator that enhances cluster fleet management. Release notes for {mce-short}-specific features are found in at link:../clusters/release_notes/mce_release_notes.adoc#mce-release-notes[Release notes for Cluster lifecycle with {mce-short}].
+
+== Errata 2.16.1
+
+* Delivers updates to one or more product container images.
+
+* Fixes ____ Thanos compact now deletes empty folders on S3, which could materialize on some S3 vendors following block deletion.
+//this needs work, not sure if this is the fix or the problem. I sent examples in the issue.
```
swopebe: @jacobbaungard can we work in this here.

release_notes/acm_fixed_issues.adoc
```diff
@@ -7,4 +7,18 @@ By default, fixed issues for _errata_ releases are automatically applied when re
...
+
+link:https://redhat.atlassian.net/browse/ACM-32930[ACM-32930]
+
+* Fixes a problem with a required placements field in the `MulticlusterRoleAssignment` resource that caused the Role Assignments page to fail when upgrading... to 2.16.0 from 2.15.0 or earlier releases. Now the .....
+//this needs work, I sent examples in the issue.
```
swopebe: @oksanabaza can we work on this here?

release_notes/acm_fixed_issues.adoc
jacobbaungard: > Thanos compact now deletes empty folders on S3, which could materialize on some S3 vendors following block deletion.

Is correct, and should make sense to our user-base. Do we have to have "fixes" in front? It makes it much harder to create a reasonable sentence in this scenario.

The change is a workaround for for an issue seen some "S3 compatible" vendors that are doing things off-spec/incorrectly.

I do not like the "fixes" in front of this, because we are specifically going out of our way to workaround an issue which actually is not an issue with _our_ software but with some S3 vendors doing things incorrectly.

If we really have to have "fixes" in front, which imo makes the note more unclear and less correct in this case has you can write:

> Fixes an issue on some S3 vendors that would leave empty folders existing on object storage after Thanos Compact deletes blocks.

swopebe: All of them lead with a verb, that cannot change. 

See all other release notes to see how it fits in, but basically we need something like...

The z-stream...

**Adds, Resolves, Fixes,** (insert the problem) and then the behavior as a result is usually after, but we don't always have that. Customers have complained about these not saying enough so I try to add a little more.

We have been using this method for a long time and it works.

This is better:

 Fixes an issue on some S3 vendors that left empty folders existing on object storage after Thanos Compact deletes blocks.

release_notes/acm_fixed_issues.adoc
oksanabaza: > or earlier releases

I think we can drop this part since RBAC UI was only introduced in 2.15, so earlier releases would not be affected

oksanabaza: In ACM 2.16 and higher versions, the `placements` field became mandatory. Existing `MulticlusterRoleAssignment` resources created in ACM 2.15 do not contain `placements` field and are incompatible with the updated schema.

The UI now prevents the page from crashing and still displays affected role assignments. However, these role assignments remain non-functional. The only supported resolution is to delete and recreate the affected `MulticlusterRoleAssignment` resources.

oksanabaza: @mshort55 could you please double check

release_notes/acm_fixed_issues.adoc
jacobbaungard: Alright, go for that then.

release_notes/acm_fixed_issues.adoc
oksanabaza: also would like to mention that it should be under known issues, not fixed

swopebe: There should be a separate issue for known issues, but my concern is that we published (meaning the ACM product team) as a fix. Will comment in the issue.

oksanabaza: Hi @swopebe, sorry for the confusion, I may have created the issue incorrectly. We do not actually have a fix for this issue. My change only prevents the page from crashing, it doesn't resolve the issue itself. Since the RBAC UI in 2.15 was a Tech Preview feature that we do not officially support, we are not planning to fix this in the future. We need a small known issue note for cases where users upgrade a cluster from 2.15 to 2.16 while having existing MRAs on the cluster. I'll create a ticket shortly

mshort55: I agree with you Oksana. This should be in the known issues and limitations section. We should just mention that if customers created MulticlusterRoleAssignments in 2.15, and then upgraded to 2.16, they need to delete/recreate all their MulticlusterRoleAssignments. This is because 2.15 was TP, and we did not need to support a better migration method. This doc note is just a curtesy notice that Scott Berens wanted added.

swopebe: If this did not go into the release, then there is a bigger problem. The advisory just sent out that this was fixed. Please get in touch with @tsze-redhat -- they use the query to pull in the bug fixes. It seems this was incorrectly announced as a fix and will need to be removed.

I can remove it easily from here, this isn't me complaining, but we need to do better about making sure the bugs in the advisories were actually released--and that comes from the Jira query that they rely on.

swopebe: Link to the Jira that will be covered by @dockerymick --ty for taking this.

https://redhat.atlassian.net/browse/ACM-33974

swopebe: This is removed.

release_notes/acm_fixed_issues.adoc
```diff
@@ -7,4 +7,12 @@ By default, fixed issues for _errata_ releases are automatically applied when re
 
 See link:../install/upgrade_hub.adoc#upgrading-by-using-the-operator[Upgrading by using the operator] for more information about upgrades.
 
-*Important:* Cluster lifecycle components and features are within the {mce-short}, which is a software operator that enhances cluster fleet management. Release notes for {mce-short}-specific features are found in at link:../clusters/release_notes/mce_release_notes.adoc#mce-release-notes[Release notes for Cluster lifecycle with {mce-short}].
\ No newline at end of file
+*Important:* Cluster lifecycle components and features are within the {mce-short}, which is a software operator that enhances cluster fleet management. Release notes for {mce-short}-specific features are found in at link:../clusters/release_notes/mce_release_notes.adoc#mce-release-notes[Release notes for Cluster lifecycle with {mce-short}].
+
+== Errata 2.16.1 (Deprecated)
```
swopebe, line 12: this was reviewed by the engineer, however, it does not matter now bc this z-stream release is deprecated. We do not need to do anything further than this.

dockerymick (approved):

PR #8891, opened 2026-05-28
https://redhat.atlassian.net/browse/ACM-34544 Modularize GRC 2.17

dockerymick:


governance/disable_template_process.adoc
oafischer, line 1: One warning for this file. Line 19 needs to be a step or similar, currently it is a standalone paragraph

Image

governance/enable_grc_standalone_template.adoc.adoc
oafischer, line 1: One warning in this file. Line 72 needs to be a step or similar

Image

governance/filter_object_selector.adoc
oafischer, line 1: I believe the issue here is that you cannot have procedures in Reference files. I'm not 100% sure though.

Image

governance/disable_template_process.adoc
dockerymick: thanks!

governance/disable_template_process.adoc
```diff
@@ -0,0 +1,32 @@
...
+.Procedure
+
+. To disable template processing for a particular template, add the `pass:` label in the raw string. See the following example where `{{ template content }}` is ignored:
+
++
+----
+{{ pass:[`{{ template content }}`] }}
+----
+
+ Then, the template returns a raw string to be processed by a subsequent templating engine.
```
dockerymick:
```suggestion
Then, the template returns a raw string to be processed by a subsequent templating engine.
```

governance/enable_grc_standalone_template.adoc.adoc
```diff
@@ -0,0 +1,75 @@
...
+
++
+[source,yaml]
+----
+oc label role cm-reader -n <managed_cluster_name> cluster.open-cluster-management.io/backup=""
+
+oc label rolebinding cm-reader-binding -n <managed_cluster_name> cluster.open-cluster-management.io/backup=""
+---- 
+
+After you add these resources, the add-on can match the hub cluster templates, and the state of the policy gets saved in a secret on the managed clusters. The secret prevents interruptions if the hub cluster becomes temporarily unavailable to the managed cluster.
```
dockerymick:
```suggestion
. The add-on matches the hub cluster templates and the state of the policy gets saved in a secret on the managed clusters. The secret prevents interruptions if the hub cluster becomes temporarily unavailable to the managed cluster.
```

governance/disable_template_process.adoc
```diff
@@ -0,0 +1,32 @@
...
+.Procedure
+
+. To disable template processing for a particular template, add the `pass:` label in the raw string. See the following example where `{{ template content }}` is ignored:
+
++
+----
+{{ pass:[`{{ template content }}`] }}
+----
+
+Then, the template returns a raw string to be processed by a subsequent templating engine.
```
dockerymick:
```suggestion
.Procedure

. To disable template processing for a particular template, add the `pass:` label in the raw string. The template returns a raw string to be processed by a subsequent templating engine. See the following example where `{{ template content }}` is ignored:

+
----
{{ pass:[`{{ template content }}`] }}
----
```

governance/filter_object_selector.adoc
dockerymick: Thanks ! I originally assigned the REFERENCE tag for this file because it is more of an example instead of a task. However, the way it is named and my language reads as a task. I added the file content and error message into Gemini. Here is the response: 


**To resolve the DITA validation error for this file, you need to wrap your code block inside an AsciiDoc Example Block (====) and attach the title directly to it.

When your conversion tool processes a standard [source,yaml] block with a title text above it, it attempts to generate an invalid DITA structure (assigning a title to a raw code block). By wrapping it in ====, the converter creates a valid DITA <example> container with a <title> element, which perfectly satisfies the DITA schema constraints.**

I think we don't need to do that, i just thought it was interesting and decided to share. Let me know if it continues to appear as broken

governance/disable_template_process.adoc
dockerymick: Moved the line to step one

oafischer:
There is still a warning for `governance/filter_object_selector.adoc`, the rest looks good though!

Screenshot 2026-05-29 at 10 38 07

governance/process_raw_templates.adoc
```diff
@@ -0,0 +1,122 @@
+:_mod-docs-content-type: REFERENCE
```
dockerymick:
```suggestion
:_mod-docs-content-type: PROCEDURE
```

dockerymick:
> There is still a warning for `governance/filter_object_selector.adoc`, the rest looks good though!
> 
> Screenshot 2026-05-29 at 10 38 07

Thanks Oliver! I will close this PR and refer to it in a new PR

PR #8905, opened 2026-06-01
https://redhat.atlassian.net/browse/ACM-34544 mod changes for grc 2.17

dockerymick:


governance/ansible_grc.adoc
```diff
@@ -1,91 +1,33 @@
...
-* Install the {aap-short} Resource Operator from the {olm}. In the _Update Channel_ section, select `stable-2.x-cluster-scoped`. Select the *All namespaces on the cluster (default)* installation mode.
+. Log in to the {ocp-short} cluster console.
 
-* You can install the {aap-short} Resource Operator on a different cluster than the hub cluster. If you install the {aap-short} Resource Operator on the hub cluster, the operator must be the same version as the {aap-short} Tower.
-+
-*Note:* Ensure that the {aap-short} job template is idempotent when you run it. If you do not have {aap-short} Resource Operator, you can find it from the {ocp-catalog}. 
+. Navigate to the {olm} and select the *{aap-short} Resource Operator*. 
 
-For more information about installing and configuring {aap}, see link:../applications/ansible_config.adoc#setting-up-ansible[Setting up Ansible tasks].
+. In the _Update Channel_ section, select *`stable-2.x-cluster-scoped`*.
```
swopebe:
```suggestion
. In the _Update Channel_ section, select *stable-2.x-cluster-scoped*.
```

swopebe: You won't be able to code block and bold something in DITA from what I can remember, here we can just bold. The point is to highlight something that is clickable.

governance/object_selector_raw_templates.adoc
```diff
@@ -0,0 +1,154 @@
...
+* Raw template processing
+
++
+The optional `object-templates-raw` parameter supports advanced templating use cases, such as the `if` conditional function and the `range` loop function. The `object-templates-raw` parameter accepts a string containing Go templates that must resolve to an `object-templates` array.
+
++
+*Note:* You must set exactly one of either `spec.object-templates` or `spec.object-templates-raw`.
+
++
+.Raw template configuration example with conditional labels
```
swopebe: @dockerymick these should not be headers. Stick to the headers we already are using (.Procedure, .Prerequisites); don't turn H2s to this. You can make these bullets, which I think is what we discussed in the meeting,

swopebe: If that builds out to too much, then we need to do another file for these.

swopebe (changes requested):

swopebe (approved):

PR #8907, opened 2026-06-02
https://redhat.atlassian.net/browse/ACM-34130--GCP 2.17 new files and…

swopebe:
… link changes, clean up


clusters/credentials/credential_google_api.adoc
```diff
@@ -0,0 +1,46 @@
...
+.Prerequisites
+
+* A deployed {mce-short} hub cluster
+* Internet access for your {mce-short} hub cluster so it can create the Kubernetes cluster on GCP
+* GCP login credentials, which include user Google Cloud Platform Project ID and Google Cloud Platform service account JSON key. See _Creating and managing projects_.
+* Account permissions that allow installing clusters on GCP. See _Configuring a GCP project_ for instructions on how to configure an account.
+
+.Procedure
+
+. Create an opaque secret for by using the API. Apply YAML content that is similar to the following example:
```
dockerymick: If this is the only step, let's use a bullet point instead (unordered)

clusters/credentials/credential_google_api.adoc
```diff
@@ -0,0 +1,46 @@
...
+----
+kind: Secret
+metadata:
+    name: <managed-cluster-name>-gcp-creds
+    namespace: <managed-cluster-namespace>
+type: Opaque
+data:
+    osServiceAccount.json: $(base64 -w0 "${GCP_CRED_JSON}")
+----
++
```
dockerymick, line 33: the plus sign can be removed

Image

clusters/credentials/credential_google_ui.adoc
```diff
@@ -0,0 +1,48 @@
...
+
+.Procedure
+
+. From the {mce-short} console, click *Credentials* to choose from existing credential options.
+
+. Create a namespace specifically to host your credentials, for both convenience and security.
+
+. *Optional:* Add a _Base DNS domain_ for your credential. If you add the base DNS domain to the credential, it is automatically populated in the correct field when you create a cluster with this credential. See the following steps:
+
+. Add your _Google Cloud Platform project ID_ for your GCP account. See _Log in to GCP_ to retrieve your settings.
```
dockerymick: are you intending to start substeps? If not, let's remove "See the following steps:" If yes, it is not clear that these are substeps

dockerymick (changes requested):

clusters/credentials/credential_google_api.adoc
swopebe: Yes, meant to change that! TY--good catch.

clusters/credentials/credential_google_api.adoc
swopebe: I don't think it can because of the requirement to embed notes with the previous step @dockerymick -- we were dinged on that when we did the test with Jaromir. I will play with it and reformat it, we cannot have a note hanging out there alone, basically.

clusters/credentials/credential_google_ui.adoc
swopebe: Yes, these wasn't even correct in the former content and I checked the UI wondering why it was there, we can remove--I just forgot.

clusters/credentials/credential_google_api.adoc
swopebe: ok, I think the trick is that the step above it has to also be indented to indent the note.

dockerymick (approved):

PR #8923, opened 2026-06-09
doc: Add prominent note about default openshift-* and kube-* namespace exclusions in Gatekeeper intro 

anasalli:
Fixes: ACM-34959
## What changed
- Added IMPORTANT callout in gk_operator_overview.adoc (intro page) stating that openshift-* and kube-* namespaces are excluded from Gatekeeper constraint enforcement by default.
- Added NOTE callout in config_gk_operator.adoc alongside the config.matches step to surface the same information on the configuration page.
## Why
Customer feedback: Assign mutations and constraints created without specifying disableDefaultMatches silently had no effect in openshift-* namespaces. The behavior is documented in the config section but is not discoverable from the chapter introduction where first-time users land. Customer spent hours before finding the fix.
## Versions needing backport
This fix applies to 2.11 through 2.16 (the gap was introduced when disableDefaultMatches was added but not called out in the overview). Please also apply to 2.17_stage.

### DIFF
--- a/governance/gatekeeper_operator/config_gk_operator.adoc
+++ b/governance/gatekeeper_operator/config_gk_operator.adoc
@@ -345,7 +345,7 @@ oc explain gatekeeper.spec.audit.auditFromCache
 
 .Additional resources 
 
-- For more information, see link:https://open-policy-agent.github.io/gatekeeper/website/docs/audit/#configuring-audit[Configuring Audit] in the Gatekeeper documentation.
-- For more information about configuring fail-open and fail-closed behavior, see link:https://open-policy-agent.github.io/gatekeeper/website/docs/failing-closed[Gatekeeper documentation].
-- For a list of arguments to use for the `containerArguments` parameter, see link:https://open-policy-agent.github.io/gatekeeper/website/docs/runtime-flags[Runtime Flags] in the Gatekeeper documentation.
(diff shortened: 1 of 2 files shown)


dockerymick (approved):

swopebe:
place in 2.15+

JustinKuli (approved):

governance/gatekeeper_operator/config_gk_operator.adoc
```diff
@@ -164,6 +164,8 @@ spec:
 
 . Use the `config` section to exclude namespaces from certain processes for all constraints on your cluster using the Gatekeeper `Config` custom resource. See the following example:
 
+*Note:* The {gate} automatically appends {ocp-short} system namespaces (`openshift-*`) and Kubernetes system namespaces (`kube-*`) to the exclusion list by default. As a result, Gatekeeper constraints and mutations are _not_ enforced in those namespaces unless you set `disableDefaultMatches: true`.
```
dockerymick, line 167: Image

Just realized that the namespaces are not rendering correctly

JustinKuli: That might just be in GitHub, I'm not sure it necessarily means it will be incorrect in the published docs? (But it could be wrong there too.)

We could remove the `*` from the code sections here and I don't feel like it would be confusing.

dockerymick: @anasalli @JustinKuli is it necessary to have the example prefix for the namespaces?

PR #8956, opened 2026-06-22
https://redhat.atlassian.net/browse/ACM-34544 split manage policies

dockerymick:


governance/create_policy_console.adoc
```diff
@@ -0,0 +1,92 @@
+:_mod-docs-content-type: PROCEDURE
+
+[#create-policy-console]
+= Creating a cluster security policy from the console
+
+[role="_abstract"]
+
```
dockerymick:
```suggestion
Create a cluster security policy from the console with placement resources.
```

governance/create_policy_set.adoc
```diff
@@ -0,0 +1,25 @@
+:_mod-docs-content-type: PROCEDURE
+
+[#create-policy-set]
+= Creating policy sets from the command line
+
+[role="_abstract"]
+By default, the policy set is created with no policies or placements. You must create a placement for the policy set and have at least one policy that exists on your cluster. When you create a policy set, you can add numerous policies. 
+
```
dockerymick:
```suggestion
Create a policy set from the CLI or console and assign policies. By default, the policy set is created with no policies or placements. You must create a placement for the policy set and have at least one policy that exists on your cluster. When you create a policy set, you can add numerous policies. 

```

governance/policy_cli_tool.adoc
```diff
@@ -0,0 +1,45 @@
+:_mod-docs-content-type: REFERENCE
+
+[#policy-cli-tool]
+= Policy command-line tool
+
+[role="_abstract"]
+With the `policytools` command-line tool, you can interact with policies locally to help with creating and debugging. The `policytools` command-line tool is available for download from the hub cluster console.
```
dockerymick:
```suggestion
Review the `policytools` command-line tool. You can interact with policies locally to help with creating and debugging. The `policytools` command-line tool is available for download from the hub cluster console.
```

governance/cleanup_resource_policy.adoc
```diff
@@ -0,0 +1,23 @@
+:_mod-docs-content-type: CONCEPT
```
swopebe, line 1: We start with a gerund in the title/verb in the description, if this is a true concept, that needs to change. However, it seems this should be rewritten as a procedure because the user has to complete steps....

so here we need to read the content/repurpose; where AI may let this pass, we have to write for UX and do the clean up.

governance/delete_policy.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+[#delete-policy]
+= Deleting a security policy
+
+[role="_abstract"]
+Delete a security policy from the command line or the console.
+
+.Procedure
+
+. Delete a security policy from the command line interface:
```
swopebe: these are two different options that should be under an unordered list, since it is one or the other

swopebe: then you don't need the sub steps

governance/policy_cli_tool.adoc
```diff
@@ -0,0 +1,45 @@
+:_mod-docs-content-type: REFERENCE
+
+[#policy-cli-tool]
+= Policy command-line tool
```
swopebe: let's title this better to say it's an overview of the `policytools` 

Thinking 

Command line `policytools` overview (or something line this)

swopebe (changes requested):
Most of this looks great--awesome work, there are just a few things to look at in this large PR. Great job!

governance/cleanup_resource_policy.adoc
dockerymick: I wouldn't change this to a procedure. There might be one step, but it is doc'ed to let the user know what they had access to do. We are more so describing the parameters and some scenarios. From my POV, it is best to rename the title to "Resource cleanup for policies".  I will retitle it

governance/cleanup_resource_policy.adoc
```diff
@@ -0,0 +1,23 @@
...
+
+[#cleanup-resources]
+= Cleaning up resources that are created by policies
```
dockerymick:
```suggestion
= Resource clean up for policies
```

governance/delete_policy.adoc
dockerymick: I'm a bit confused about the unordered list suggestion. I understand that I have two options listed. I created steps w/substeps to avoid creating another file

governance/policy_cli_tool.adoc
dockerymick:
```suggestion
= Command line _policytools_ overview
```

swopebe (approved):

swopebe:
approving contingent on the suggested improvements

governance/delete_policy.adoc
swopebe: @dockerymick that will build as step one and step two, which is incorrect UX because the user doesn't have 2 steps. You either must create a bullet with steps or a new file. This was also brought up when Jarmoir ran the test, but it is just better UX. In the past we could have two headers but now we have to have two bullets. I think that will pass, as I had to do that with one of my files.

dockerymick: Gotchu, thanks for the reminder. I will make it an unordered list as you suggested. Thanks!

PR #9007, opened 2026-07-07
https://redhat.atlassian.net/browse/ACM-34132--script changes 2.17

swopebe:


dockerymick (approved):
Similar comment from the previous PR. Add the mod content type at the beginning of the file. Approving

PR #9031, opened 2026-07-13
 modularize gh4 (undoing some problematic content formatting in disconnected) (2.17)

swopebe:


global_hub/global_hub_import_hub_default.adoc
```diff
@@ -11,15 +11,15 @@ Import an existing hub cluster as a managed hub cluster to help you control your
 . Disable the cluster self-management setting in the existing hub cluster by setting the `disableHubSelfManagement` setting to `true` in the `multiclusterhub` custom resource. This setting disables the automatic import of the hub cluster as a managed cluster.
 . Import the managed hub cluster by completing the steps in link:../clusters/cluster_lifecycle/import_intro.adoc#import-intro[Cluster import introduction].
 
-== Importing a managed hub cluster 
-
-To import an existing hub cluster as a managed hub cluster, complete the following steps:
+.Procedure
```
dockerymick, line 14: Based on Red Hat style guide, we cannot have an ordered list in prereqs. Can you either make it an unordered list or consider moving this info after the Procedure tag?

dockerymick (approved):
One comment that should be addressed but will not hold up the build. Thanks for your patience!

global_hub/global_hub_import_hub_default.adoc
swopebe: hmm, I know that, wondering what happened here. Good catch @dockerymick .

swopebe: Fixed, turned to a bullet.

PR #9035, opened 2026-07-14
https://redhat.atlassian.net/browse/ACM-34132--more global hub (2.16)

swopebe:


global_hub/global_hub_ztp_monitoring.adoc
oafischer: This paragraph throws an error as a standalone line. I'd suggest attaching it to the previous step as a quick fix.

oafischer (approved):
One comment that needs addressing, otherwise approved

swopebe:
moved that line up one.

PR #9036, opened 2026-07-14
https://redhat.atlassian.net/browse/ACM-34132--more global hub (2.17)

swopebe:


global_hub/global_hub_ztp_apply_policy.adoc
```diff
@@ -40,9 +40,9 @@ oc get application -n openshift-gitops -l app=policies
...
 oc get policy -A
 ----
-
-Wait for all policy applications to show a healthy state before you proceed to the next migration process. This process might take one or two minutes.
-
++
 *Important:* Do not create the migration resource until the policy applications are fully synchronized. If policies are missing during migration, the clusters might fail to reach a compliant state.
 
+.Next steps
+
```
dockerymick, line 47: I'm not as familiar here yet. But the agent response is that this should not be a full sentence

global_hub/global_hub_ztp_migration_prep.adoc
```diff
@@ -11,8 +11,6 @@ To migrate managed clusters between hub clusters, you must configure the target
 
 .Prerequisites
 
-See the following prerequisites before you begin:
-
 * Ensure all managed clusters are connected by network to the target hub cluster.
 * Verify version compatibility. See xref:../global_hub/global_hub_ztp_migration_prep.adoc#ztp-migration-overview[Zero Touch Provisioning migration overview] for supported versions of {acm-short} and {ocp-short}.
```
dockerymick: I know this was not a part of your changes, but it appeared as an error. Use this link instead: 
```suggestion
* Verify version compatibility. See xref:../global_hub/global_hub_ztp_migration.adoc#ztp-migration-overview[Zero Touch Provisioning migration overview] for supported versions of {acm-short} and {ocp-short}.

```

dockerymick (approved):
More comments but nothing to hold up the PR. Great work!

global_hub/global_hub_ztp_apply_policy.adoc
swopebe: Ok. **.Next steps** is a tag we can use like **.Additional links.** It will be a dita tag. I have already used it in spots like this: https://github.com/stolostron/rhacm-docs/blob/2.17_stage/clusters/credentials/credential_hosted.adoc

And remember, credentials passed the test for modularization. So I am using it because it applies here and if we were in DITA, that is what we would use. I don't use it unless we have to send them do something next.

PR #9045, opened 2026-07-16
No issue: Fix leftover CIM conversion problems

oafischer:


clusters/assisted_installer/ai_create_console_ref.adoc
```diff
@@ -6,7 +6,7 @@
 [role="_abstract"]
 When you create a host inventory (infrastructure environment), you can configure various settings in the console. See the following definition list to learn more:
 
-.List of host inventory console settings
+== List of host inventory console settings
```
swopebe: Remember we said we were getting rid of all header twos. I have either been changing them to new files or bullets.

clusters/assisted_installer/ai_manage.adoc
```diff
@@ -10,7 +10,7 @@ Each host that you successfully boot with the Discovery ISO appears as a row in
 
 If you manage your host inventory by using the command line interface, an `Agent` resource represents each host. See the following defintion list to learn about the properties in an `Agent` resource:
 
-.List of `Agent` resource properties
+== List of `Agent` resource properties
```
swopebe: Same here, we are not supposed to be using **h2**, at least we agreed on that a while ago....

nor this method: 
**.List of `Agent` resource properties**

Wondering if bullets here are sufficient--that is what I have been using for lists and such that do not require a new file.

swopebe (approved):

PR #9055, opened 2026-07-31
No issue, clean up Obs conversion

oafischer:
Might need to be backported


swopebe (approved):

dockerymick (approved):
Thank you for fixing these Oliver! Teamwork!

PR #9061, opened 2026-08-03
https://redhat.atlassian.net/browse/ACM-34544: modularize manage policies procedures for 2.16

dockerymick:
Split console, CLI view, policy set, disable, and delete procedures out of manage_policies so the 2.16 governance book matches the 2.17 module structure.

Phase 1/12


governance/delete_policy.adoc
```diff
@@ -0,0 +1,37 @@
+:_mod-docs-content-type: PROCEDURE
+
+[#delete-policy]
+= Deleting a security policy
+
+[role="_abstract"]
+Delete a security policy from the command line or the console.
```
swopebe: Since you are introducing two procedures here, place that info in the short description so the use knows there are two here. You can ___ , or you can ___.

governance/delete_policy_set.adoc
```diff
@@ -0,0 +1,13 @@
+:_mod-docs-content-type: PROCEDURE
```
swopebe, line 1: it's odd that this one is standing alone and the other  Deleting a security policy options are not, can this not merge to the one above it? If not, that is fine.

governance/manage_policies.adoc
```diff
@@ -369,11 +144,17 @@ To save the compliance messages to a separate file, use the `--messages-path` ar
...
+* xref:../governance/view_policy_console.adoc#view-policy-console[Viewing your security policy from the console]
+* xref:../governance/create_policy_set.adoc#create-policy-set[Creating a policy set]
+* xref:../governance/add_policy_policy_set.adoc#add-policy-policy-set[Adding a policy to a policy set]
+* xref:../governance/disable_policy.adoc#disable-policy[Disabling a security policy]
+* xref:../governance/delete_policy.adoc#delete-policy[Deleting a security policy]
+* xref:../governance/delete_policy_set.adoc#delete-policy-sets[Deleting a policy set]
+* xref:../governance/hub_policy_framework.adoc#hub-policy-framework[Hub cluster policy framework]
+* link:https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements[Resources that support support set-based requirements]
+* link:https://github.com/open-cluster-management-io/policy-collection/tree/main/policygenerator/policy-sets/stable[PolicySets-- Stable]
+* xref:../governance/grc_intro.adoc#governance[Governance]
```
swopebe, line 160: This list should only be what is needed for this topic. We don't need to link to the intro and the whole TOC, for instance. If we do that, we would probably do that for every topic.

The Additional resources section is usually things that you mentioned instead of inline links and things the user might need.

swopebe (changes requested):
Few comments

governance/delete_policy.adoc
swopebe: The other changes I think can wait, if they are to be implemented. This change is necessary because it has multiple procedures that are not introduced in the description.

dockerymick: Ok, I have to also add that to 2.17.

dockerymick:
```suggestion
You can delete a security policy from the command line, or from the console.
```

governance/manage_policies.adoc
dockerymick: My intention was to link to the sections that were created into individual files. I understand not linking to the intro. To help with accuracy, it would be best to rename the file to `create_security_policy.adoc` and remove the links. Thoughts?

swopebe (approved):

governance/manage_policies.adoc
swopebe: yes, but since this is for 2.16, I would just leave it as-is and do that for 2.17 @dockerymick

governance/delete_policy_set.adoc
dockerymick: It was because the earlier file was specifically for individual policies. I see how and why it makes sense to add deleting a policy set to the topic, deleting a security policy. I will leave this as-is like you mentioned. I have taken note of the residual changes and will address them shortly

PR #9093, opened 2026-08-18
https://redhat.atlassian.net/browse/ACM-34217--z-stream (2.16.2 and 2.11.2)

swopebe:


release_notes/acm_fixed_issues.adoc
```diff
@@ -9,6 +9,14 @@ See link:../install/upgrade_hub.adoc#upgrading-by-using-the-operator[Upgrading b
 
 *Important:* Cluster lifecycle components and features are within the {mce-short}, which is a software operator that enhances cluster fleet management. Release notes for {mce-short}-specific features are found in at link:../clusters/release_notes/mce_release_notes.adoc#mce-release-notes[Release notes for Cluster lifecycle with {mce-short}].
 
+== Errata 2.16.2
+
+* Delivers updates to one or more product container images.
+
+* Fixes a problem with the multicluster observability add-on node exporter pods, which failed to start on managed Kubernetes platforms such as Amazon EKS and Azure Kubernetes Service (AKS) when the `kube-rbac-proxy` container did not bind to port `19100` because the cloud provider already ran a node exporter on that port. The add-on now uses a separate internal listener port for node exporter, which prevents the port conflict.
```
thibaultmg: I would clarify it a bit adding some precisions about the possible failures and the possible use of the `nodeExporterHostPort` variable. Here is a new generated version with the correct details. Feel free to adapt:

* Fixes a problem where the multicluster observability add-on node exporter pods failed to start on managed Kubernetes platforms (such as Amazon EKS and Azure Kubernetes Service) because the cloud provider already ran a node exporter on port `19100`. 
  
  Additionally, some container runtimes were reporting a port clash when both the HostPort used by the `node-exporter` container and the internal port used by the `kube-rbac-proxy` container were configured to the same number. 

  To resolve these issues, the add-on now defines a different default port number for each container (using `19100` for the host port and `19101` for the internal listener port). If the default host port `19100` is already in use by the managed platform, users can customize these values by setting the `nodeExporterHostPort` and `nodeExporterInternalPort` variables in the `AddOnDeploymentConfig`:

  ```yaml
  apiVersion: addon.open-cluster-management.io/v1beta1
  kind: AddOnDeploymentConfig
  metadata:
    name: multicluster-observability-addon
    namespace: open-cluster-management-observability
  spec:
    customizedVariables:
    - name: nodeExporterHostPort
      value: "19200" # Change this if port 19100 is already in use on the host
    - name: nodeExporterInternalPort
      value: "19201"
  ```

swopebe: @thibaultmg I can add some of this:

`Additionally, some container runtimes were reporting a port clash when both the HostPort used by the node-exporter container and the internal port used by the kube-rbac-proxy container were configured to the same number.`

For the resolution, we don't get in to that on these simple release notes. If that is something you think the user should know, we would need a Jira for an update to either the main documentation or to place in a known issue/limitation.

For reference and understanding of the roles of this doc, it would help to see the documentation as it is published:

Release notes (Fixed issues are z-streams), then you can see the known issues, as well:

https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_management_for_kubernetes/2.16/html/release_notes/acm-release-notes

thibaultmg: ok sounds good if we can add the detail about the 2 ways the ports can clash. 
I have several updates needed for the documents regarding MCOA, I'll add the details there for the variables to use.

swopebe: Ok, I made that update.

swopebe: From Slack:

@thibaultmg wrote:

I would change it slightly:

Fixes a problem with the multicluster observability add-on node exporter pods, which failed to start on managed Kubernetes platforms such as Amazon EKS and Azure Kubernetes Service (AKS) when port `19100` was already occupied by the cloud provider's own node exporter. To resolve this, users can now configure a custom host port for `kube-rbac-proxy` using the `nodeExporterHostPort` variable in the `AddonDeploymentConfig`. Additionally, to prevent container runtime port clashes when both the host port (used by `kube-rbac-proxy`) and the internal port (used by `node-exporter`) are identical, the add-on now uses a separate, dedicated internal listener port for the node exporter container.

swopebe: Reply:

Fixes a problem with the multicluster observability add-on node exporter pods, which failed to start on managed Kubernetes platforms, such as Amazon EKS and Azure Kubernetes Service (AKS), when port `19100` was already occupied by the node exporter for that cloud provider. You can now configure a custom host port for the `kube-rbac-proxy` container by using the `nodeExporterHostPort` variable in the `AddonDeploymentConfig` resource. Additionally, to prevent container runtime port clashes when both the host port, which is used by the `kube-rbac-proxy` container, and the internal port, which is used by the `node-exporter` container, are identical, the add-on now uses a separate, dedicated internal listener port for the node exporter container.

(removes possessive, adds relative pronouns, reduces translation issues with word choice changes)

swopebe: See the above discussion. Given the time differences, the final review will be later but this can go in to peer review.

release_notes/acm_fixed_issues.adoc
```diff
@@ -9,10 +9,18 @@ See link:../install/upgrade_hub.adoc#upgrading-by-using-the-operator[Upgrading b
 
 *Important:* Cluster lifecycle components and features are within the {mce-short}, which is a software operator that enhances cluster fleet management. Release notes for {mce-short}-specific features are found in at link:../clusters/release_notes/mce_release_notes.adoc#mce-release-notes[Release notes for Cluster lifecycle with {mce-short}].
 
+== Errata 2.16.2
+
+* Delivers updates to one or more product container images.
+
+* Fixes a problem with the multicluster observability add-on node exporter pods, which failed to start on managed Kubernetes platforms, such as Amazon EKS and Azure Kubernetes Service (AKS), when port `19100` was already occupied by the node exporter for that cloud provider. You can now configure a custom host port for the `kube-rbac-proxy` container by using the `nodeExporterHostPort` variable in the `AddonDeploymentConfig` resource. Additionally, to prevent container runtime port clashes when both the host port, which is used by the `kube-rbac-proxy` container, and the internal port, which is used by the `node-exporter` container, are identical, the add-on now uses a separate, dedicated internal listener port for the node exporter container.
```
dockerymick: I got lost in the length of the sentences at the beginning and the nouns

I wrote EKS to remain consistent 

Suggest to remove "now" because I would usually see it in New features. That change is optional though. 

Reduced repetitiveness
```suggestion
* Fixes a problem that prevented node exporter pods for the multicluster observability add-on from starting on managed Kubernetes platforms, such as Amazon Elastic Kubernetes Service (EKS) and Azure Kubernetes Service (AKS), when port `19100` was already in use by the node exporter on the cloud platform. You can configure a custom host port for the `kube-rbac-proxy` container with the `nodeExporterHostPort` variable in the `AddonDeploymentConfig` resource. To prevent container runtime port conflicts when the host port for kube-rbac-proxy and the internal port for node-exporter are the same, the add-on now assigns a dedicated internal listener port to the node exporter container.
```

dockerymick (approved):

thibaultmg (approved):
/lgtm

release_notes/acm_fixed_issues.adoc
swopebe: We should keep _now_ here -- I have used it in these release notes--it's fine. I prefer "failed to start" -- I will consider the other changes but I think the paragraph is not incorrect in the first draft...we went back and forth alot and I scanned for style.

swopebe: by using the `nodeExporterHostPort` variable (the developer gave me "using" so I am going to keep that instead of with), I made a few small changes, but the paragraph suggested is a bit long, as well, and I think we can keep much of the first draft--small changes added that do not change the meaning.

PR #9096, opened 2026-08-19
https://redhat.atlassian.net/browse/ACM-39118 backport secure_cluster

dockerymick:
Backport RBAC split, Cluster permissions, MCA, Fine-grain intro/enable


secure_clusters/cluster_permission_validate.adoc
swopebe: take a look throughout the PR to check to see if .Procedure is added here needed, you can remove line 11.

secure_clusters/create_cluster_permission_ref.adoc
swopebe: Remember, we cannot have a line after the Procedure tag, so we have to change that for 2.17. I would scan each file in 2.17 for that.

swopebe: I would not just remove this one, it should be merged with the description, which is lacking anyway.

secure_clusters/fine_grain_rbac_enable.adoc
```diff
@@ -25,11 +28,12 @@ See the following prerequisites that are needed to enable fine-grained role-base
 
 See the following steps to enable the `fine-grained-rbac` component in the `MultiClusterHub` resource:
```
swopebe: same here with line 26

secure_clusters/fine_grain_rbac_enable.adoc
```diff
@@ -25,11 +28,12 @@ See the following prerequisites that are needed to enable fine-grained role-base
...
+. Run the following command to patch the resource. The default name of the custom resource is `multiclusterhub`, and `open-cluster-management` is the default namespace. If you have a different name or namespace, update the command:
 
++
 [source,bash]
 ----
 oc patch mch multiclusterhub -n open-cluster-management --type=json -p='[{"op": "add", "path": "/spec/overrides/components/-", "value": {"name": "fine-grained-rbac", "enabled": true}}]'
 ----
 
-Next, see xref:../secure_clusters/fine_grain_rbac_ui.adoc#fine-grain-ui[Enabling fine-grained role-based access control with the console] for instructions on how to assign roles to users and groups from the console.
+Next, see xref:../secure_clusters/fine_grain_ui.adoc#fine-grain-ui[Enabling fine-grained role-based access control with the console] for instructions on how to assign roles to users and groups from the console.
```
swopebe: This would be considered "Next steps"

secure_clusters/rbac_enable_app_lifecycle.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+. Create and administer applications in the `cluster-name` managed cluster, with the `username` user. You must create a cluster role binding to the `open-cluster-management:admin:<cluster-name>` cluster role and bind it to `username` by entering the following command:
+
++
+[source,bash]
+----
+oc create clusterrolebinding <role-binding-name> --clusterrole=open-cluster-management:admin:<cluster-name> --user=<username>
+----
++
+This role has read and write access to all `application` resources on the managed cluster, `cluster-name`. Repeat this if access for other managed clusters is required.
```
swopebe, line 37: Do we need a note or anything here?

secure_clusters/rbac_enable_app_lifecycle.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+. To view an application on a managed cluster named `cluster-name` with the user named `username`, create a cluster role binding to the `open-cluster-management:view:` cluster role and bind it to `username`. Enter the following command:
+
++
+[source,bash]
+----
+oc create clusterrolebinding <role-binding-name> --clusterrole=open-cluster-management:view:<cluster-name> --user=<username>
+----
++
+This role has read access to all `application` resources on the managed cluster, `cluster-name`. Repeat this if access for other managed clusters is required.
```
swopebe, line 65: same--just asking, it may be fine.

secure_clusters/rbac_enable_observability.adoc
```diff
@@ -0,0 +1,75 @@
...
+
+[#rbac-enable-observability]
+= Enabling RBAC for Observability
+
+[role="_abstract"]
+To view the observability metrics for a managed cluster, you must have `view` access to that managed cluster on the hub cluster. 
+
+*Required access:* Cluster administrator
+
+Complete the following steps for observability RBAC features:
```
swopebe: we can remove this line 11 since we no longer need a line like this to intro the procedure, but I think we still will pass as long as it is above the Procedure tag. Since this is a fairly important book, up to you if you want to clean it up a little.

swopebe: maybe just for 2.17.

swopebe (changes requested):

secure_clusters/rbac_enable_app_lifecycle.adoc
dockerymick: it's not required, but if you want to explicitly call it out we can add it. It is not labeled as a note in 2.17 @swopebe

swopebe: Ok, I would not worry about it. For ACM 5 and in dita we can write it better. Ideally we don't say "In this role...." and add this afterwards. It should be information the user needs before the command, especially since we are telling them to "repeat" (take action)

Further, we are likely moving forward with this function in the doc bc it was not removed before ACM 5 code cut-off but app is mostly deprecated and we don't need to clean up the doc.

dockerymick: Understood, I will leave as is and add it to the list of things we need to address.

swopebe (approved):

PR #9116, opened 2026-08-24
https://redhat.atlassian.net/browse/ACM-41774--removing ACM apis exce…

swopebe:
…pt for Search


dockerymick (approved):
couple of conflicts , but this looks good

PR #9121, opened 2026-08-25
https://redhat.atlassian.net/browse/ACM-41774 -- 2.17 new topic and removals from 2.16

swopebe:


clusters/api/discover_apis_cli.adoc
```diff
@@ -0,0 +1,41 @@
+:_mod-docs-content-type: PROCEDURE
```
swopebe, line 1: This is the same new file that was added to 2.16.

apis/discover_apis_cli.adoc
```diff
@@ -0,0 +1,43 @@
+:_mod-docs-content-type: PROCEDURE
```
swopebe, line 1: This is the exact file that was added to 2.16.

apis/docinfo.xml
```diff
@@ -1,8 +1,8 @@
 <productname>{acm}</productname>
 <productnumber>{product-version}</productnumber>
-<subtitle>View a list of APIs for {acm-short}.</subtitle>
+<subtitle>View the Search API for {acm-short}.</subtitle>
 <abstract>
-    <para>View a list of APIs that you can use to create and manage application resources, channels, subscriptions, and to query information.</para>
+    <para>View the Search API that you can use to query information.</para>
```
dockerymick, line 5: did you intend to update this file?

dockerymick (approved):

apis/docinfo.xml
swopebe: @dockerymick I am actually moving that file, so I intentionally left it as-is....it will go to the search folder. I am not sure we will have an API folder after this, I have to see how it all turns out.

